PR #2, opened 2016-09-29
Add c make to official tbb repository

hjmjohnson:


antalmal:
It would be better if you could generate CMakelist.txt using TBB Makefile system likewise it is done for MSVS projects because supporting two or more independent project files is burdensome. TBB team considered cmake in the past and found it unaligned with the goals set by the project, thus it cannot be the main build system for TBB.

hjmjohnson:
@antalmal, Thank you for your consideration.  I would like to express my disappointment, especially in the light of an enthusiasm from the community to support such a system. A convenient CMake build definition would greatly facilitate the adoption of TBB in many projects. 

Will you please enumerate the goals of the TBB project for which CMake is unaligned?  I personally believe that reducing the burden of incorporating TBB into external projects on many different platform would promote use and adoption of this excellent tool.

There is clearly a long-standing community desire to have a cmake build system for tbb as evidenced by the a quick search for others who have struggled to incorporate TBB into their projects.

- https://software.intel.com/en-us/forums/intel-threading-building-blocks/topic/290889
- https://github.com/justusc/FindTBB
- https://github.com/wjakob/tbb
- https://software.intel.com/en-us/forums/embree-photo-realistic-ray-tracing-kernels/topic/635255
- http://stackoverflow.com/questions/28458300/building-vtk-and-thread-building-blocks-using-cmake
- https://sourceforge.net/p/mia/mia2/ci/master/tree/cmake/FindTBB.cmake
- https://bitbucket.org/sergiu/tbb-cmake
- http://answers.opencv.org/question/10/how-to-build-opencv-with-tbb-support/

The following are a short list of open source tools in my narrow domain that each use the Cmake build environment that also can depend on TBB:

- OpenCV
- VTK (and soon ITK)
- MLPack
- Armadillo
- GDCM
- LAPACK
- https://github.com/BRAINSia/BRAINSTools

A list of over 70 other important "reference toolkits" that have successfully increased their usability and adoption with cmake inclusion can be found here: https://cmake.org/

antalmal:
@hjmjohnson Thank you for your involvement. Full disclosure: I'm not TBB developer anymore and the final decision is to be made by other people . I'm participating in this discussion as a regular community member who just happened to have 9 years of TBB development experience.

Please also note, that I didn't say that TBB should never have CMake as its part. Actually, I'd like to see it supported. We share the same goal here. What I said is that by the time we considered it, it was not aligned with our requirements, and suggested an alternative implementation. Today, as TBB goes through a lot of changes, like change of licensing model, migration to git, publication on github, team refresh, this question can be reconsidered again and it is potentially possible to adopt cmake as the primary build system.

However, I believe rather that TBB does not have that many free resources to redesign the whole build and test system including CI infrastructure.. Unless there is someone brave enough to spend many months (and unlikely being payed for it) in order to collect all the use cases, design, re-implement, validate, and provide full support to what's already works well enough and satisfy all the requirements.
What you did is good for integrating into other peoples projects, to build the library as a user. But it does not look flexible and elaborated enough to substitute make system for TBB developers needs and for our needs of TBB's CI, which builds and tests the library with commercial quality in so many configurations. If curious enough, just go through index.html files about the build system and check whether your cmake project can support what is publicly documented at this moment (while there are much more undocumented functionality). If it is possible - please create a backward compatible Makefile interface which calls cmake underneath - it might increase the chances to be adopted. 

Otherwise, or more realistic way is what I suggested - make cmake project derivative from TBB's makefiles. It is not hard, it is what works for MSVC already, and it'll satisfy everyone: cmake fans will get cmake project, TBB team will still have one place to control the build configurations, and our infrastructure will keep working as is. What you need is to populate cmake project with the flags and the files being compiled by Makefile. MSVC project is created by defining a fake compiler as a script which adds the file and its build flags to the project instead of actual compilation. TBB can likely commit to generating such cmake projects as part of its nightly builds and pack into releases.

antalmal:
BTW, which LAPACK uses TBB?

hjmjohnson:
LAPACK does not use TBB, it uses CMake in a way that makes LAPACK much easier to incorporate into many development environments and eases developer burden to use LAPACK.  Sorry for the typo.

 

Thank you for the feedback.  It helps set realistic priorities and identify what would need to be done to move forward as efficiently as possible.

 

Hans

 

 

--

PR #20, opened 2017-06-06
Changed handling of size == 0 in scalable_aligned_malloc

ZeroEightSix:
I modified scalable_aligned_malloc to be consistent with scalable_malloc and allocate sizeof(size_t) bytes when the requested size is 0. The same can obviously be achieved in the calling code, but I thought it might be a good idea to have both functions behave the same.

### DIFF
--- a/src/tbbmalloc/frontend.cpp
+++ b/src/tbbmalloc/frontend.cpp
@@ -3019,10 +3019,11 @@ extern "C" int scalable_posix_memalign(void **memptr, size_t alignment, size_t s
 
 extern "C" void * scalable_aligned_malloc(size_t size, size_t alignment)
 {
-    if (!isPowerOfTwo(alignment) || 0==size) {
+    if (!isPowerOfTwo(alignment)) {
         errno = EINVAL;
         return NULL;
     }
+    if (!size) size = sizeof(size_t);
     void *tmp = allocateAligned(defaultMemPool, size, alignment);
     if (!tmp) errno = ENOMEM;
     return tmp;


alexey-katranov:
The main idea why scalable_malloc and scalable_aligned_malloc have different behavior is the compatibility with malloc and _aligned_malloc behavior:
[MSDN malloc states](https://msdn.microsoft.com/en-us/library/6ewkz86d.aspx): _"... If size is 0, malloc allocates a zero-length item in the heap and returns a valid pointer to that item. ..."_
[MSDN _aligned_malloc states](https://msdn.microsoft.com/en-us/library/8z34s9c6.aspx): _"... If alignment is not a power of 2 or size is zero, ... ,  this function returns NULL and sets errno to EINVAL."_

ZeroEightSix:
OK, fair enough.

The whole thing started with a quick emergency change that I had to make in our code to always use `scalable_aligned_malloc` on Mac. We override the `new` operator and macOS started requiring `new` to return memory aligned to at least `alignof(std::max_align_t)` which is 16 on 64-bit macOS. `scalable_malloc` doesn't meet this requirement, so I switched to `scalable_aligned_malloc`.

And yes, I do realize that having future versions of tbbmalloc return memory aligned to 16 bytes from `scalable_malloc` would be the best and preferred, so maybe I should request that somewhere?

Anyway, if you think compatibility with Microsoft's mallocs is more important than consistency between these two functions, I understand. It was worth a try. Thanks for your time!

alexey-katranov:
You are welcome. Personally, I agree with you that the inconsistency between `malloc` and `scalable_malloc` looks strange. However, many existing codes rely on implementation behavior of particular functions. In some cases, it is easier to have implementation behavior compatibility than change the legacy codes. Another curious example is that [`safer_aligned_malloc` accepts zero alignment](https://github.com/01org/tbb/blob/tbb_2017/src/tbbmalloc/proxy.cpp#L291-L295).

As far as I know, since Intel TBB 2017, [the memory allocator on 64-bit platforms aligns memory to 16 bytes by default for all allocations bigger than 8 bytes](https://github.com/01org/tbb/blob/tbb_2017/CHANGES#L202-L203).

If you want to request functionality, report a bug or discuss a problem, feel free to use [the issues page](https://github.com/01org/tbb/issues) or [the forum](https://software.intel.com/en-us/forums/intel-threading-building-blocks). If your have a private/confidential question or problem, you can [e-mail us](mailto:inteltbbdevelopers@intel.com).

PR #24, opened 2017-08-09
add 4D blocked range to TBB

jeffhammond:
Quantum chemistry and deep learning both make heavy use of 4D arrays.  I think it would be nice to support 4D ranges explicitly.  Whether this provides value over nesting 2 2D ranges or a 1D and a 3D range is unclear.  I'll look into that later.


akukanov:
A new `blocked_rangeNd` class is a preview feature since TBB 2018 Update 3, it allows to specify the range dimensionality as a template parameter.

PR #37, opened 2017-10-18
Use CHANGES date instead of build date

bmwiedemann:
in order to make builds reproducible.
See https://reproducible-builds.org/ for why this is good


tbbdev:
Hello, bmwiedemann.
We see your point and this change looks reasonable. Thanks!
We will inform you when implement reproducible builds (not for Linux only).

hjmjohnson:
👍 This would be a nice addition.

hjmjohnson:
This is very old.    If this suggestion will not be considered, please close the issue.  @tbbdev Is more work needed?

alexey-katranov:
The approach was significantly reworked for oneTBB and build is expected to be reproducible.

PR #52, opened 2018-05-08
Add ARMv8 support

siddhesh:
This patch adds ARMv8 support for libtbb and provides an optimized
__TBB_Pause implementation that avoids sched_yield.

I've taken the minimal approach to leave the current armv7 support mostly untouched, with the exception of __TBB_machine pause.  The aarch64 support mostly draws from gcc_generic.h since it provides a reasonable starting point for the implementation.  I've build-tested the -march=armv7-a on a raspberry pi and a full test on aarch64.

### DIFF
--- a/include/tbb/machine/gcc_arm.h
+++ b/include/tbb/machine/gcc_arm.h
@@ -26,10 +26,7 @@
 #error Do not include this file directly; include tbb_machine.h instead
 #endif
 
-//TODO: is ARMv7 is the only version ever to support?
-#if !(__ARM_ARCH_7A__)
-#error compilation requires an ARMv7-a architecture.
-#endif
+#if __ARM_ARCH_7A__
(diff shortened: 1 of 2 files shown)


alexey-katranov:
Thank you for your effort. The patch will be merged in one of the future updates.

siddhesh:
That's great, thanks Alexey!  It sounds like the merge happens outside of github.  Should I close this PR or do you prefer to keep it open for your internal workflow?

alexey-katranov:
Yes, the merge will be outside of github. I'd prefer to remain the pull request open that we can notify you when released.

alexey-katranov:
The pull request was merged into [TBB 2018 Update 5](https://github.com/01org/tbb/blob/tbb_2018/CHANGES#L25).

PR #58, opened 2018-05-15
Use process-local futex operations

Lastique:
The FUTEX_WAIT and FUTEX_WAKE operations are supposed to be applied to futexes that can be shared between processes. Using process-local equivalents can be more efficient.

Also add linux/futex.h include, which should nowdays be provided by all relevant Linux distributions.


include/tbb/machine/linux_common.h
```diff
@@ -34,15 +34,19 @@
 #define __TBB_USE_FUTEX 1
 #include <limits.h>
 #include <errno.h>
-// Unfortunately, some versions of Linux do not have a header that defines FUTEX_WAIT and FUTEX_WAKE.
+#include <linux/futex.h>
```
alexey-katranov: By historical reason, `machine/linux_common.h` is included for *BSD systems in [`machine.h`](https://github.com/01org/tbb/blob/tbb_2018/include/tbb/tbb_machine.h#L230-L251). These systems may not have the `linux/futex.h` header file (e.g. they may provide `sys/futex.h`). I would recommend to include `linux/futex.h` only when `__linux__` is defined (in other words, to guard it with `if __linux__ ... endif`).

Lastique: I was under impression that futexes are Linux-specific and expected `SYS_futex` to be not defined on other systems. However, it looks like OpenBSD supports them, but it is not covered by the conditions in `tbb_machine.h`.

I can add conditions for Linux and OpenBSD, but I will also have to disable the futex-related functions below, which presumably should not be defined when futexes are not available.

Lastique:
I've updated the PR.

include/tbb/tbb_machine.h
```diff
@@ -227,7 +227,7 @@ template<> struct atomic_selector<8> {
         #include "machine/linux_intel64.h"
     #endif
 
-#elif __linux__ || __FreeBSD__ || __NetBSD__
+#elif __linux__ || __FreeBSD__ || __NetBSD__ || __OpenBSD__
```
alexey-katranov: Is `|| __OpenBSD__` enough to build TBB on OpenBSD (and without this improvement the build fails)?

include/tbb/machine/linux_common.h
```diff
@@ -31,18 +31,39 @@
...
+#include <sys/futex.h>
 #define __TBB_USE_FUTEX 1
+#elif defined(__has_include)
+#if __has_include(<sys/futex.h>)
+#include <sys/futex.h>
+#define __TBB_USE_FUTEX 1
+#endif
+#endif
+
+#if __TBB_USE_FUTEX
```
alexey-katranov: I suppose the logic with `__TBB_USE_FUTEX` is surplus because all this code is guarded with the [`SYS_futex`](https://github.com/01org/tbb/pull/58/files#diff-1ddd140ec2d4d0fad745786ee9125f29R32) macro (i.e. the futex syscall is supported). It is Ok if no one header file is included because the values of `__TBB_FUTEX_WAIT` and `__TBB_FUTEX_WAKE` are hardcoded to [`0`](https://github.com/01org/tbb/pull/58/files#diff-1ddd140ec2d4d0fad745786ee9125f29R61) and [`1`](https://github.com/01org/tbb/pull/58/files#diff-1ddd140ec2d4d0fad745786ee9125f29R69) respectively in that case.

include/tbb/tbb_machine.h
Lastique: I did not test OpenBSD, I don't have access to the system. The reason I added this check is that OpenBSD supports futexes, and the other code in the guarded headers doesn't look system-specific.

include/tbb/machine/linux_common.h
Lastique: I actually don't like that the macros are defined to some specific values in TBB because the values are obviously system-dependent. Including the header provides these system values.

alexey-katranov: I agree that using headers is the best approach and your patch is worth contributing. However, if TBB is used on the system where the introduced logic will not find the required header file, disabling hardcoded behavior will break backward compatibility.
My suggestion is to include header files where possible (where we know what to include) and not to disable the old behavior for the sake of backward compatibility.

Lastique: Ok, will do.

Lastique:
I've updated the patch to support systems with no futex.h.

include/tbb/machine/linux_common.h
```diff
@@ -31,18 +31,37 @@
...
 #if defined(SYS_futex)
 
+/*
+ * Note that this header is included not only on Linux. Some BSD systems may also support futexes,
+ * but provide the definitions in other headers. For backward compatibility we also support
+ * systems that don't provide any header but support futexes, as indicated by the SYS_futex macro.
+ */
 #define __TBB_USE_FUTEX 1
+
+#if defined(__linux__)
```
alexey-katranov: We will slightly refactor this part of the code, à la:
```
#if defined(__has_include)
#define __TBB_has_include __has_include
#else
#define __TBB_has_include(x) 0
#endif

#if defined(__linux__) || __TBB_has_include(<linux/futex.h>)
#include <linux/futex.h>
#elif defined(__OpenBSD__) || __TBB_has_include(<sys/futex.h>)
#include <sys/futex.h>
#endif
```

Lastique:
Any news about this?

PR #63, opened 2018-06-28
fixed HashCompare constructor in concurrent_hash_map

arewedancer:
Hi,
  This is a followup fix on the issue reported on the TBB forum. https://software.intel.com/en-us/forums/intel-threading-building-blocks/topic/780000.

Now the example in the gist can build and work. 
https://gist.github.com/arewedancer/a6ac95ddd800b4da27a943a406bf8226

Thanks!

### DIFF
--- a/include/tbb/concurrent_hash_map.h
+++ b/include/tbb/concurrent_hash_map.h
@@ -755,6 +755,9 @@ public:
     };
 
     //! Construct empty table.
+    explicit concurrent_hash_map( const allocator_type &a = allocator_type() )
+        : internal::hash_map_base(), my_allocator(a)
+    {}
     explicit concurrent_hash_map( const HashCompare &c = HashCompare(), const allocator_type &a = allocator_type() )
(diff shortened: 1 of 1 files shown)


akukanov:
Hi @arewedancer, unfortunately fixing the issue in this way breaks source compatibility (in case an allocator was specified in a ctot call) and therefore is undersirable. I think a bunch of new ctors needs to be added; it's also aligned with the C++ library practice for associative containers.

arewedancer:
Hi, Alexey(@akukanov ),  Thanks for your feedback! So rolled back original constructor interface and added additional ones. Please review.

hjmjohnson:
@arewedancer Is this change compatible with 2019_U2 release?

akukanov:
@hjmjohnson, @arewedancer, the changes in 2019_U2 are essentially the integration of this merge request. This is mentioned in the CHANGES file as well, though we forgot to mention the author of the contribution - which we will fix.

arewedancer:
👍👍

PR #68, opened 2018-07-13
Patch for verbose tbbmalloc_proxy functionality

e4lam:
- TBBMALLOC_PROXY_HOOK_FAILURE is set to the list of failed checked opcodes
- When TBBMALLOC_PROXY_VERBOSE is set, we output the list of
  successfully checked opcodes in addition to the failed ones to
  TBBMALLOC_PROXY_HOOK_FAILURE


ivankochin:
@e4lam , sorry for the late response.
Supporting a set of debug environment variables is rather complex and preferred to be avoided. So, after more internal discussion, we have come up with an idea of public API function, that returns the pointer to a string with all the necessary information about replaced/falied routines (similar to your patch). You can then output to the file, stdout or even environment variable if it is more convenient. Is this solution suits you?

e4lam:
I think something like that sounds fine. The important part is to be able to:

1) Know that the hooking failed

2) Provide user diagnostics that can be sent back to us so that it can be patched without need the user's OS. This is important because we get Windows beta users that run into these problems as well.

ivankochin:
@e4lam , `TBB_malloc_replacement_log()` function that provides this functionality is available in [TBB2019 U4](https://github.com/01org/tbb/blob/2019_U4/CHANGES#L21-L22)

tbbdev:
Further discussion is expected to be done in https://github.com/intel/tbb/issues/187

PR #69, opened 2018-07-13
Patch exit() calls to abort() for better debugging when things go wrong

e4lam:


nik-ponomarev:
exit() has various benefits over abort(), like calling functions registered using atexit() and calling objects destructors. That is why we cannot just change all exit() calls to abort().
So, an alternative solution would be the introduction of a special scalable_allocation_mode or environment variable for enabling/disabling such calls replacements. Is this an acceptable solution for you?

e4lam:
Perhaps then we should have a facility like a tbb::set_malloc_fail_handler() similar to tbb::set_assertion_handler() for this.

I'd like to re-iterate that having any exit() calls in a low-level library like TBB is extremely bad. One cannot always predict what happens with publicly released software in the field and exit() calls make it impossible to gather information regarding what went wrong.

akukanov:
We have discussed this issue, and decided to remove both exit() calls. Failure to create the TLS key is yet another way how tbbmalloc initialization may fail, so we will make the behavior consistent (which is to return NULL from the allocation call that caused initialization). Failure to destroy the key should be even less of a problem - it likely happens during process termination, so exiting/aborting makes little sense.

@e4lam, if the default behavior is as I described above, would you need a configuration option or a failure handler to change it?

e4lam:
That is even better!

nik-ponomarev:
@e4lam, notifying you that `exit()` calls were removed from `ThreadId `class in the latest [TBB 2019 U1](https://github.com/01org/tbb/releases/tag/2019_U1)

akukanov:
@e4lam, can we close this PR?

PR #71, opened 2018-07-13
Patch to force tbbmalloc.dll to fast exit on process termination 

e4lam:
This fixes deadlocks when exiting a process that still has running threads.


nik-ponomarev:
@e4lam, notifying you that deadlock fix is available in the latest [TBB 2019 U1](https://github.com/01org/tbb/releases/tag/2019_U1)

e4lam:
Thanks! I still have not found time to update and try it but hopefully in the next month or so.

PR #72, opened 2018-07-13
Patch to workaround MSVC compiler bug when building with /permissive-

e4lam:
Patch to workaround MSVC compiler bug when building with /permissive- on latest Visual Studio 2017 version.


alexey-katranov:
Thank you for the patch. We want to move the helper class out of the function scope to the private part of the `task_group_base` class.

akukanov:
The issue is fixed in TBB 2019 Update 1.

PR #82, opened 2018-08-24
Do not capture build kernel version

bmwiedemann:
Do not capture build kernel version
so that it does not influence build results.
See https://reproducible-builds.org/ for why this is good.

Also os_version was unused and thus dropped.

### DIFF
--- a/build/linux.inc
+++ b/build/linux.inc
@@ -60,11 +60,9 @@ endif
 
 ifndef runtime
         export gcc_version:=$(shell gcc -dumpversion)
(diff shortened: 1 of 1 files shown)


bmwiedemann:
ping. Can I do something to help get this reviewed/tested/merged?

anton-potapov:
Closing as this part is rework in oneTBB
( as #37,  https://github.com/oneapi-src/oneTBB/pull/37#issuecomment-870809904 )

PR #95, opened 2018-10-11
Entend target visibility

Ignition:
Normally IMPORTED targets has scope in the directory in which it was created and below, but the GLOBAL option extends visibility.


hjmjohnson:
This is stale, @Ignition I'd recommend closing without incorporation if there is not a better rational for why this extended behavior is needed.

PR #106, opened 2018-12-01
Use nanosleep on Linux in atomic_backoff when spinning for a long time.

Lastique:
This allows to significantly reduce CPU system time while spinning, as
apparently nanosleep deschedules the thread for a longer time.

See https://github.com/01org/tbb/issues/105. On the test presented there,
this patch offers to reduce CPU system time by 25-30%. On a real world
application that test is based on this patch offers ~10% improvement.

However, the change causes test_task_arena failures, as described in
https://github.com/01org/tbb/issues/104.

### DIFF
--- a/include/tbb/machine/linux_common.h
+++ b/include/tbb/machine/linux_common.h
@@ -25,6 +25,9 @@
 #include <sched.h>
 #define __TBB_Yield()  sched_yield()
 
+#include <time.h>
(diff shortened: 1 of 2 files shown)


Lastique:
I've created an updated PR https://github.com/oneapi-src/oneTBB/pull/327 that is rebased against oneTBB code base. I'm not closing this PR for reference and in case if anyone needs this functionality for older TBB releases. However, you can close it if you like.

PR #122, opened 2019-02-09
Allow upgrading to writer without knowing if reader/writer

Niadb:
I found this very useful in some complex code that had to occasionally upgrade in a few spots

Sorry the spacing/tabs seems off, it keeps massacring the formatting when I try to submit this--

### DIFF
--- a/include/tbb/spin_rw_mutex.h
+++ b/include/tbb/spin_rw_mutex.h
@@ -123,7 +123,14 @@ public:
             is_writer = true;
             return mutex->internal_upgrade();
         }
-
+        //! Upgrade writer if currently a reader
+        /** Returns whether the upgrade happened without releasing and re-acquiring the lock */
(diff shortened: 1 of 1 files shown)


rfschtkt:
The odds of this being integrated are close to zero with only "a few spots" in "some complex code".

BTW, don't get me started about tabs, but your clue should be that TBB does not use them!

akukanov:
Actually, since TBB 2019 U5 the read-write mutexes in TBB allow upgrades and downgrades to "succeed" if the mutex is already in the requested state. In other words, it's now safe to call `upgrade` unconditionally, without first checking if the current thread is a reader or writer. The change was inspired by this PR, which can now be closed.

rfschtkt:
I guess it must have been more than "a few spots" in "some complex code" then...

PR #125, opened 2019-02-26
Work around CUDA on Mac

### LINKED ISSUE #44: tbb config picks up CUDA on macOS
subj. AFAIK `tbb` does not use CUDA anyhow, there is a suggestion here on how to workaround this issue https://github.com/spack/spack/issues/6000#issuecomment-358817701 which might be a hint to a proper solution.

ibaned:
See spack/spack#6000
Fixes #44

### DIFF
--- a/build/macos.inc
+++ b/build/macos.inc
@@ -49,7 +49,7 @@ ifndef arch
 endif
 
 ifndef runtime
-  clang_version:=$(shell clang -v 2>&1 >/dev/null | grep version | sed -e "s/.*version \(.*[0-9]\) .*/\1/")
+  clang_version:=$(shell clang --version 2>&1 >/dev/null | grep version | sed -e "s/.*version \(.*[0-9]\) .*/\1/")
   ifndef os_version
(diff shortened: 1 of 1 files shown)


ibaned:
Ah, sorry to not have carried over the context. As I understand it, when Clang detects CUDA, its -v flag prints extra text about CUDA. This gets misinterpreted by the build system as part of the computer system description and becomes part of the build directory name, including spaces. This leads to broken relative paths later on. —version seems not to print this extra CUDA text

hjmjohnson:
@ibaned Can that comment be part of the commit message so that it persist in the history outside of GitHub?

ibaned:
@hjmjohnson sure! I just force pushed to the PR branch, the commit message has the full description and credits @mikcieslak who actually came up with the solution.

hjmjohnson:
@ibaned Thanks.  Hopefully this will increase the likelihood that the PR is addressed quickly.

tbbdev:
Implemented.

PR #129, opened 2019-03-07
Remove superfluous semicolons to satisfy -pedantic.

chissg:
Compiling user code with `-pedantic` causes warnings which are turned into errors if the user also has `-Werror` activated.


hjmjohnson:
Is this patch set still open?

alexey-katranov:
It is [merged](https://github.com/oneapi-src/oneTBB/blob/tbb_2020/include/tbb/machine/gcc_generic.h#L215-L217). Thanks to everyone for the contribution and reminder.

Issue #146, opened 2019-05-20
Glibc version detection is broken for Clang


Timmmm:
TBB tries to detect the glibc version by looking at `__GNUC__`, `__GNUC_MINOR__` etc. which is the version of GCC. However Clang always seems to set this to 4. The code notes this.

```
// note that when ICC or Clang is in use, __TBB_GCC_VERSION might not fully match
// the actual GCC version on the system.
#define __TBB_GCC_VERSION (__GNUC__ * 10000 + __GNUC_MINOR__ * 100 + __GNUC_PATCHLEVEL__)
```

This is pretty annoying because GCC 4 is pretty ancient at this point, and you get warnings about deprecated functions if you try compiling the code with Clang.

Also there *are* proper macros for the glibc version now (easily since 6). See https://sourceforge.net/p/predef/wiki/Libraries/

Timmmm:
Note there has been a rather hacky attempt to fix this in [`linux.inc`](https://github.com/intel/tbb/blob/tbb_2019/build/linux.inc#L58) and [`linux.clang.inc`](https://github.com/intel/tbb/blob/tbb_2019/build/linux.clang.inc#L51):

        export gcc_version:=$(shell gcc -dumpversion)
...
       # TODO: do not assume that GCC minor and patchlevel versions are always single-digit.
       CPLUS_FLAGS += -DTBB_USE_GLIBCXX_VERSION=$(subst .,0,$(gcc_version))

This is rubbish because:

a) It can't handle non-single digit GCC versions (as noted in the comment). Somewhat surprisingly that hasn't actually happened since GCC 2.95.
b) `-dumpversion` only prints the major version since some time ago. Now you need to do `-dumpfullversion -dumpversion` (the second `-dumpversion` is for backwards compatibility) otherwise it will think GCC 7.0.0 is actually GCC 0.0.7.
c) This is only in the TBB build system - when you come to actually `#include` TBB headers in your project that depend on this value you don't get this `#define`! It doesn't get written to a `config.h` or anything like that.

Timmmm:
I had a further investigation, and it seems like there are no available macros to check the `libstdc++` version.:

* `__GNUC__` checks the GCC version but is always 4 for Clang.
* `__GLIBC__` checks the glibc version (i.e. not libstdc++).
* `__GLIBCXX__` is the *date* of the libstdc++ release, but it's useless because minor versions released after a newer major version have a higher number. I.e. 4.9 might have a higher number than 5.0.

[...]

PR #147, opened 2019-05-20
Fix gcc version dumping on recent GCC versions

Timmmm:
See #146 (though it doesn't really fix that).


tbbdev:
@Timmmm thank you for the proposal! It is implemented in [TBB 2019 U7](https://github.com/intel/tbb/blob/2019_U7/CHANGES).

PR #149, opened 2019-06-04
Allow to disable tbbmalloc, use posix_memalign

Lastique:
The first commit enables `posix_memalign` on UNIX-like systems. This API is universally supported nowdays, so there's no need to use the hand-rolled implementation on top of `malloc`.

The second one adds `TBB_USE_MALLOC` environment variable to force using system allocator.

By setting `TBB_USE_MALLOC` environment variable to 1 the user can disable scalable allocator from tbbmalloc and use system allocator through `malloc`/`posix_memalign`/`free` API. There are other, more performant memory allocators, like tcmalloc for example, which the user might want to use instead of tbbmalloc. Also, using the system allocator can be useful with memory debugging and profiling tools, like valgrind. Such tools usually provide their own implementations of memory allocation functions, and we want those functions to be used by TBB.


src/tbb/cache_aligned_allocator.cpp
```diff
@@ -19,15 +19,21 @@
 #include "tbb/tbb_allocator.h"
 #include "tbb/tbb_exception.h"
 #include "tbb_misc.h"
+#include "tbb_environment.h"
 #include "dynamic_link.h"
 #include <cstdlib>
+#include <stdlib.h> // posix_memalign
```
alexey-katranov: Does it make sense to include stdlib.h conditionaly (only when we really want to use `posix_memalign`)? I mean moving this inclusion to line 34.

src/tbb/cache_aligned_allocator.cpp
```diff
@@ -19,15 +19,21 @@
...
 
 #if _WIN32||_WIN64
 #include "tbb/machine/windows_api.h"
 #else
 #include <dlfcn.h>
 #endif /* _WIN32||_WIN64 */
 
+#if !_WIN32 && !_WIN64
+#define __TBB_USE_POSIX_MEMALIGN
+#endif
```
alexey-katranov: What about [`_aligned_malloc`](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/aligned-malloc?view=vs-2019) on Windows?

src/tbb/cache_aligned_allocator.cpp
```diff
@@ -112,7 +120,9 @@ static const dynamic_link_descriptor MallocLinkTable[] = {
     If that allocator is not found, it links to malloc and free. */
 void initialize_handler_pointers() {
     __TBB_ASSERT( MallocHandler==&DummyMalloc, NULL );
-    bool success = dynamic_link( MALLOCLIB_NAME, MallocLinkTable, 4 );
+    bool success = false;
+    if (!tbb::internal::GetBoolEnvironmentVariable("TBB_USE_MALLOC"))
+        success = dynamic_link( MALLOCLIB_NAME, MallocLinkTable, 4 );
```
alexey-katranov: Overall, we try to minimize the number of environment variables that can be used to specify the library behavior. It is an non-composable approach because different parts of application might want different settings.
What can be the reason to request TBB not to use tbbmalloc? Why not just remove tbbmalloc binary if it is unnecessary?

src/tbb/cache_aligned_allocator.cpp
```diff
@@ -121,7 +131,11 @@ void initialize_handler_pointers() {
         FreeHandler = &std::free;
         MallocHandler = &std::malloc;
         padded_allocate_handler = &padded_allocate;
+#if !defined(__TBB_USE_POSIX_MEMALIGN)
```
alexey-katranov: Usually if we have `else` branch in conditions we put the positive branch first.

src/tbb/cache_aligned_allocator.cpp
Lastique: I can do that.

src/tbb/cache_aligned_allocator.cpp
Lastique: I can add it, although I don't use Windows and can't test.

src/tbb/cache_aligned_allocator.cpp
Lastique: > It is an non-composable approach because different parts of application might want different settings.

With tbbmalloc and TBB, I don't think the user has such fine grained choice. He either uses tbbmalloc (through TBB) or not. I just want a way to disable it for my application, regardless of the TBB build options.

> Why not just remove tbbmalloc binary if it is unnecessary?

This is not possible in practice since tbbmalloc may be installed as part of system packages. It may be used by other applications and cannot be uninstalled.

> For the described reason, we can consider extending of the global_control functionality.

I considered adding a programming interface for disabling tbbmalloc, and I think this would be a useful addition. However, I chose environment variable approach for several reasons:

1. Environment variables can be set before starting the process. This means before any global constructors are run in any of the libraries. Potentionally, one of said constructors could invoke TBB and thus initialize tbbmalloc before the `main` attempts to disable it (thus disabling it will have no effect).
2. Environment variables can be used similarly to programming API. I.e. the application that is sure to be not affected by point 1 can set `TBB_USE_MALLOC` from its `main` before using TBB. In other words, environment variables are not worse than API.
3. Environment variables get propagated to child processes, which is especially useful in my particular case. I can set `TBB_USE_MALLOC` in my parent process and be sure that tbbmalloc is disabled in any child processes.

So for these reasons, even if API is added, I would still prefer an environment variable approach.

src/tbb/cache_aligned_allocator.cpp
Lastique: That's ok. I usually list the "more preferred" or "more usual" cases first to ease code reading. But I can swap the branches.

Lastique:
I've updated the PR according to your comments while keeping the environment variable.

src/tbb/cache_aligned_allocator.cpp
alexey-katranov: Your arguments are reasonable. However, we are really concerned about environment variables and we try to avoid using them if possible.
Can you consider another solution for the described problem? You can create a stub library, e.g.
```
> cat tbbmalloc_stub.c 
#include <stdlib.h>

void* scalable_malloc( size_t s ) {
    return malloc(s);
}

void scalable_free( void* p ) {
    free(p);
}

void* scalable_aligned_malloc( size_t s, size_t a) {
    void* p = NULL;
    int res = posix_memalign(&p, a, s);
    if (res != 0)
        p = NULL;
    return p;
}

void scalable_aligned_free( void* p) {
    free(p);
}
```
If you compile it to the library (e.g. `gcc -o libtbbmalloc_stub.so -shared -fPIC tbbmalloc_stub.c`), you can use it with LD_PRELOAD (e.g. `>LD_PRELOAD=libtbbmalloc_stub.so ./your_app`) to redirect TBB allocations to the stub library. Does this solution suit your needs?

Lastique: I'm already using `LD_PRELOAD` for other purposes, and given that I have to start child processes, managing `LD_PRELOAD` for the processes become much more complicated. Overall, this approach seems too complicated for me. I prefer the `TBB_USE_MALLOC` way.

Actually, I would rather prefer TBB *not* use tbbmalloc by default. tbbmalloc should be an optional replacement for the standard allocator, much like tcmalloc is. I believe, it already works that way, if the user `LD_PRELOAD`s libtbbmalloc_proxy.so. In this case there would be no need for `TBB_USE_MALLOC`. Would this be an acceptable change?

Lastique: Or, if you want to keep the dependency on tbbmalloc, maybe you would consider building a separate version of TBB without this dependency? E.g. libtbb_minimal.so. By this I mean a separate target in the Makefile that would produce this new library when building TBB.

Lastique: Also, to comment on your suggestion about libtbbmalloc_stub.so, I believe, it won't work. TBB uses `dlopen`/`dlsym` to bind `scalable_*` functions explicitly from tbbmalloc.

alexey-katranov: >I'm already using LD_PRELOAD for other purposes, and given that I have to start child processes, managing LD_PRELOAD for the processes become much more complicated. Overall, this approach seems too complicated for me. I prefer the TBB_USE_MALLOC way.

LD_PRELOAD and TBB_USE_MALLOC are both environment variables. How one environment variable can become much more complicated than another? You can define LD_PRELOAD though other variables in your application launcher, e.g. `LD_PRELOAD=$(LD_PRELOAD_MODULE1) $(TBB_USE_MALLOC)` (where `TBB_USE_MALLOC` is either `libtbbmalloc_stub.lib` or nothing).

> Actually, I would rather prefer TBB not use tbbmalloc by default. tbbmalloc should be an optional replacement for the standard allocator, much like tcmalloc is. I believe, it already works that way, if the user LD_PRELOADs libtbbmalloc_proxy.so. In this case there would be no need for TBB_USE_MALLOC. Would this be an acceptable change?

In the time when TBB was introduced, its performance was affected by non-scalable nature of the system malloc. I do not know if tcmalloc was invented that time (but at least it was not well know), hence tbbmalloc was invented as a scalable allocator. Nowdays, the system mallocs are much better than before; however, in general case tbbmalloc still has some benefits. Try to run [`tree_sum`](https://github.com/intel/tbb/tree/tbb_2019/examples/task/tree_sum) example on your machine.

>Or, if you want to keep the dependency on tbbmalloc, maybe you would consider building a separate version of TBB without this dependency? E.g. libtbb_minimal.so. By this I mean a separate target in the Makefile that would produce this new library when building TBB.

There is almost no value to create an additional binary because:
* it can be done manually
* tbbmalloc can be remove and system allocator will be used
* the stub library can used

Why all three approaches are not suitable for you?

> Also, to comment on your suggestion about libtbbmalloc_stub.so, I believe, it won't work. TBB uses `dlopen`/`dlsym` to bind `scalable_*` functions explicitly from tbbmalloc.

Consider [the comment](https://github.com/intel/tbb/blob/tbb_2019/src/tbb/dynamic_link.cpp#L81-L82) and [the implementation](https://github.com/intel/tbb/blob/tbb_2019/src/tbb/dynamic_link.cpp#L475). Moreover, it works in my tests.

Lastique: > How one environment variable can become much more complicated than another?

I'm composing `LD_PRELOAD` dynamically, depending on the program settings, the process to launch and what `LD_PRELOAD` contains on the main process startup. Adding `libtbbmalloc_stub.so` to the mix complicates things because I have one more library to worry about. E.g. I wouldn't want to add `libtbbmalloc_stub.so` if there is already `libtbbmalloc_proxy.so` or some version of it in `LD_PRELOAD`. And that implies I have to hardcode the version in my code. In general, I don't want to fiddle with `LD_PRELOAD` more than I absolutely have to.

> There is almost no value to create an additional binary because:
> - it can be done manually

I'm assuming, you mean by defining `__TBB_DYNAMIC_LOAD_ENABLED` to 0? The problem with this approach is that the resulting library will be `libtbb.so.2`, which is the same as the normally built one. Such specially built library cannot be packaged alongside with the normally built one.

I explicitly suggested a separate target in the Makefile so that the library is differently named (including SONAME) and can be packaged and installed side by side with the current `libtbb.so` without affecting the latter.

> - tbbmalloc can be remove and system allocator will be used

I already described why this cannot be done.

> - the stub library can used

Ok, thanks for pointing me to the comment, I missed it. However, as I said earlier, `LD_PRELOAD` approach seems too complicated for me.

Lastique: > Try to run tree_sum example on your machine.

I tried running the example on the patched TBB 2019 Update 8, here are the results:

```
Default run:
./tree_sum 
Tree creation using TBB scalable allocator
   half created serially: time = 112.0 msec
   half done in parallel: time = 27.9 msec
Calculations:
           SerialSumTree: time = 28.9 msec, sum=7.01275e+08
   SimpleParallelSumTree: time = 14.6 msec, sum=7.01275e+08
OptimizedParallelSumTree: time = 13.7 msec, sum=7.01275e+08
elapsed time : 0.229687 seconds
./tree_sum stdmalloc 
Tree creation using standard operator new
   half created serially: time = 171.6 msec
   half done in parallel: time = 199.3 msec
Calculations:
           SerialSumTree: time = 33.7 msec, sum=7.01275e+08
   SimpleParallelSumTree: time = 20.4 msec, sum=7.01275e+08
OptimizedParallelSumTree: time = 20.4 msec, sum=7.01275e+08
elapsed time : 0.482214 seconds

TBB_USE_MALLOC=1 run:
./tree_sum 
Tree creation using TBB scalable allocator
   half created serially: time = 109.7 msec
   half done in parallel: time = 26.6 msec
Calculations:
           SerialSumTree: time = 28.3 msec, sum=7.01275e+08
   SimpleParallelSumTree: time = 13.8 msec, sum=7.01275e+08
OptimizedParallelSumTree: time = 13.7 msec, sum=7.01275e+08
elapsed time : 0.22146 seconds
./tree_sum stdmalloc 
Tree creation using standard operator new
   half created serially: time = 180.4 msec
   half done in parallel: time = 216.8 msec
Calculations:
           SerialSumTree: time = 33.5 msec, sum=7.01275e+08
   SimpleParallelSumTree: time = 20.8 msec, sum=7.01275e+08
OptimizedParallelSumTree: time = 20.1 msec, sum=7.01275e+08
elapsed time : 0.508148 seconds

tcmalloc run:
./tree_sum 
Tree creation using TBB scalable allocator
   half created serially: time = 110.4 msec
   half done in parallel: time = 27.2 msec
Calculations:
           SerialSumTree: time = 27.4 msec, sum=7.01275e+08
   SimpleParallelSumTree: time = 13.8 msec, sum=7.01275e+08
OptimizedParallelSumTree: time = 13.7 msec, sum=7.01275e+08
elapsed time : 0.221178 seconds
./tree_sum stdmalloc 
Tree creation using standard operator new
   half created serially: time = 87.8 msec
   half done in parallel: time = 56.2 msec
Calculations:
           SerialSumTree: time = 35.3 msec, sum=7.01275e+08
   SimpleParallelSumTree: time = 18.9 msec, sum=7.01275e+08
OptimizedParallelSumTree: time = 19.4 msec, sum=7.01275e+08
elapsed time : 0.253918 seconds
```

Default run is `make test` with no further adjustments, "TBB_USE_MALLOC=1 run" is `make test` with `TBB_USE_MALLOC` set to 1 and "tcmalloc run" adds `LD_PRELOAD=libtcmalloc_minimal.so` to that. Note that in the last case there's little difference between `./tree_sum` and `./tree_sum stdmalloc` runs, which means tcmalloc performs pretty similar to tbbmalloc. Also note that `TBB_USE_MALLOC=1` does not meaningfully affect performance (in other words, using tbbmalloc in TBB itself doesn't provide a performance benefit). I did multiple runs, the results are mostly the same. This was run on a Core i7 2600K.

akukanov: @Lastique, would it work for you if in the environment variable you would need to specify a process name, as returned by `/proc/self/comm`, `GetModuleFileName`, `proc_selfname` etc.? This would limit the impact of the envirable to the process which name matches its value.

Lastique: I guess that could work, although it would break propagation to the child processes. One would have to set that variable for each child process that executes a different executable. This is possible in my case, but it could pose a problem in other cases.

Why are you concerned about the process name? Propagation to child processes seems like a natural thing to have.

akukanov: A big issue with envirables, especially on Windows, is that those are often set for the whole system. If such a variable controls behavior of some popular library, its impact might be much bigger than someone planned. This is why we believe using envirables in libraries is bad idea; it's an instrument for tuning either the system or a particular application, but not libraries.

Lastique:
Well, environment vars *are* the way to customize behavior of an 
application. If a library happens to be part of that application, that is 
fine - let the user control that part of the application. As long as it 
keeps functioning correctly, there's no problem. If the user wants to 
disable tbbmalloc system-wide, that's his right. As to whether that would 
be a reasonable thing to do is a different question, and I don't think TBB 
should make that decision. That's my humble opinion.

Lastique:
I've created https://github.com/oneapi-src/oneTBB/pull/326 as an updated version of this PR for oneTBB. I'm not closing this PR as a reference and in case if someone needs this functionality for older TBB releases. However, you can close it if you want.

PR #150, opened 2019-06-07
Fix detection of clang version when CUDA toolkit is installed

amadio:
When CUDA toolkit is installed, the version of `clang` is not detected properly.
See https://sft.its.cern.ch/jira/browse/ROOT-9678 for the original report.

Before this change:
```sh
$ clang -v 2>&1 >/dev/null | grep version | sed -e "s/.*version \(.*[0-9]\) .*/\1/"
10.0.0
Found CUDA installation: /usr/local/cuda, version unknown
```
After this change:
```sh
clang --version | sed -n "1s/.*version \(.*[0-9]\) .*/\1/p"
10.0.0
```

### DIFF
--- a/build/BSD.inc
+++ b/build/BSD.inc
@@ -25,7 +25,7 @@ ifndef arch
 endif
 
 ifndef runtime
-        clang_version:=$(shell clang -v 2>&1 | grep version | sed "s/.*version \([0-9]*\.[0-9]*\).*/\1/")
+        clang_version:=$(shell clang --version | sed -n "1s/.*version \(.*[0-9]\) .*/\1/p")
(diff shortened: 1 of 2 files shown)


amadio:
I see that this is also addressed by #125. Please let me know if I should close this.

akukanov:
Thanks for reminding about the issue. No, please do not close the PR yet; we will use one or the other patch and then report back.

amadio:
I've fixed this also on BSD, where the same problem appears.

tbbdev:
Included into [TBB 2019 U9](https://github.com/intel/tbb/blob/2019_U9/CHANGES).

PR #157, opened 2019-06-25
Fix gcc warning -Wclass-memaccess

zmajeed:
Fixes https://github.com/intel/tbb/issues/134

The warning was introduced in gcc 8.0: https://gcc.gnu.org/gcc-8/changes.html

Warning is documented at https://gcc.gnu.org/onlinedocs/gcc/C_002b_002b-Dialect-Options.html#index-Wclass-memaccess

Relevant gcc bug reports:

https://gcc.gnu.org/bugzilla/show_bug.cgi?id=81327

Original patch discussion:

https://gcc.gnu.org/ml/gcc-patches/2017-04/msg01571.html

### DIFF
--- a/include/tbb/concurrent_hash_map.h
+++ b/include/tbb/concurrent_hash_map.h
@@ -122,7 +122,7 @@ namespace interface5 {
 #endif
         //! Constructor
         hash_map_base() {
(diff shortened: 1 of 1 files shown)


tbbdev:
Related behavior was changed. The PR is not relevant anymore.

PR #159, opened 2019-06-28
Added support for move-only types to tbb::parallel_pipeline

rfschtkt:
I had a need to pass something like `std::vector<std::unique_ptr<something>>`, which was not supported. These are the necessary changes (or at least I assume it now supports the above type as well as simply `std::unique_ptr<int>`), plus some pedantic editing, including frivolous reformatting as well as a proposal to be more precise, e.g., about what really is meant by `tbb_trivially_copyable` and large objects, for the latter substituting `use_allocator` instead.

### DIFF
--- a/src/test/test_parallel_pipeline.cpp
+++ b/src/test/test_parallel_pipeline.cpp
@@ -27,6 +27,10 @@ int filter_node_count = 0;
 #include "tbb/tbb_allocator.h"
 #include "tbb/spin_mutex.h"
 
+#if __TBB_CPP11_RVALUE_REF_PRESENT
+#include <memory> // std::unique_ptr
+#endif
+
 const unsigned n_tokens = 8;
 // we can conceivably have two buffers used in the middle filter for every token in flight, so
 // we must allocate two buffers for every token.  Unlikely, but possible.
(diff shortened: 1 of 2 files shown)


src/test/test_parallel_pipeline.cpp
```diff
@@ -647,17 +651,22 @@ int TestMain() {
...
-        run_function<size_t,double>("size_t", "double");
-        run_function<size_t,bool>("size_t", "bool");
-        run_function<int,int>("int","int");
-        run_function<check_type<unsigned int>,size_t>("check_type<unsigned int>", "size_t");
-        run_function<check_type<unsigned short>,size_t>("check_type<unsigned short>", "size_t");
-        run_function<check_type<unsigned int>, check_type<unsigned int> >("check_type<unsigned int>", "check_type<unsigned int>");
-        run_function<check_type<unsigned int>, check_type<unsigned short> >("check_type<unsigned int>", "check_type<unsigned short>");
-        run_function<check_type<unsigned short>, check_type<unsigned short> >("check_type<unsigned short>", "check_type<unsigned short>");
-        run_function<double, check_type<unsigned short> >("double", "check_type<unsigned short>");
+        #define M(type1, type2) run_function<type1, type2>(#type1, #type2);
```
akukanov: I suggest to rename the macro to `TEST_WITH`.

akukanov:
@rfschtkt, a good implementation, thanks! We will gladly take it.

src/test/test_parallel_pipeline.cpp
rfschtkt: Then I would suggest RUN_FUNCTION. Less is more. More less is better than less less. :-)

rfschtkt:
Attention points:
- I only repeated the conditional use of wchar_t, but probably "long long" should also be tentative or with an environment-specific name (as in atomic.h), let alone "long double" (I don't know if that's relevant, it would have to be a 64-bit environment with a 64-bit long double and an old compiler, but irrelevance is the worst case here).
- I did not consider potential version compatibility issues, but at first sight it doesn't seem to matter, except for the test that explicitly looks at is_large_object->use_allocator?

rfschtkt:
I had made a simple __TBB_IF_ABLE_MOVE() macro, but then discovered the existing tbb::internal::move() function instead. I didn't immediately see something for a prvalue parameter, so I left that as-is (no use to provide a facility that is only used once). Pull request code freeze!

BTW, building on MacOS does not seem to use C++11 out of the box?

tbbdev:
Included into [TBB 2019 U9](https://github.com/intel/tbb/blob/2019_U9/CHANGES).

PR #162, opened 2019-07-02
fix gcc 9.1 -Wdeprecated-copy warnings by declaring default copy constructor and assignment operators

zmajeed:
Part 1 fixes for https://github.com/intel/tbb/issues/161


alexey-katranov:
Thank you for the patch. Unfortunately, we cannot accept the patch "as is" because it relies on C++11 and will not work with pre-C++11. In any case, we have already been working on the issue and the issue will be fixed the issue in one of the future releases.

zmajeed:
Sorry - did not realize C++03 is supported

PR #168, opened 2019-07-22
Support Huge Pages and per-NUMA node allocations on Windows

mirounga:
NUMA allocations are available only for Windows 7 or later OS. VS2019 projects set _WIN32_WINNT = 0x0601 to enable this advanced functionality.

### DIFF
--- a/build/vs2019/makefile.sln
+++ b/build/vs2019/makefile.sln
@@ -1,0 +1,80 @@
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio 2013
+VisualStudioVersion = 12.0.40629.0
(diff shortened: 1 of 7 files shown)


anton-potapov:
@mirounga please re-target the patch to the master, as tbb2019 no longer updated

anton-potapov:
@mirounga , unfortunately simple changing for the target branch is not enough, as the source base changed significantly after the [revamp](https://www.intel.com/content/www/us/en/developer/articles/technical/tbb-revamp.html)

anton-potapov:
@alexey-katranov that is your opinion about adding this functionality ?

kboyarinov:
Change the target branch back to tbb_2019 to view correct changes.

PR #171, opened 2019-08-07
Tentative rvalue reference support for parallel_reduce

rfschtkt:
My own take on rvalue reference support for parallel_reduce (see comments in pull request #169), extended with support for semigroups, i.e., without using an identity element.

For a while, shameless history rewriting was going on here, then on 2019-09-09 I squashed everything, and now I intend to only do straight commits... or that was the intention before I rebased into thematic commits for easier review.


rfschtkt:
Hmm, I would hope that the code and the comments there speak for themselves, without depending on any rationale hidden in a history that may not even remain accessible... Does it help if I characterise the current 9 commits as, respectively: trivial, frivolous (x4), preparation, material (x2), cleanup? Otherwise, please give an example of what you would like to see instead.

rfschtkt:
TODO: instrumented Body-based test?

BenFrantzDale:
So... what's next to make this happen?

rfschtkt:
Wait, wait, I've got another idea: a back-ported std::optional for the semigroup support! :-)

I'm open for suggestions to take this from experimental to releasable even though I'm not currently using any of this myself: should USING_LEFT_PARAMETER_FOR_SUM==0-specific code be eliminated, things that are not clear, etc.

rfschtkt:
Hmm, doesn't seem to work very well. I have the latest logical-commits chain along the first-parent ancestry as intended, but the previous versions seems to still cause merge conflicts rather than being ignored, even throwing GitHub into a tizzy, and I'm surprised that nobody has even complained about the experiment yet, so I might have to excise the old commits again?

rfschtkt:
<[https://www.youtube.com/watch?v=K8E_zMLCRNg](https://www.youtube.com/watch?v=K8E_zMLCRNg)>

rfschtkt:
Some tinkering with "Ancillary: Foo->Minimal" (only): MinimalBody member variables now always initialised, instead of assigned either inside or outside the constructor.

Still wondering where std::optional, backported or not, causes additional move-construction to happen... but not enough to try to find out.

BenFrantzDale:
What's needed to push this over the line?

rfschtkt:
I am resigned to the distinct possibility that that might never happen, yet still available to assist in any serious attempt to integrate some of the ideas I had fun exploring.

rfschtkt:
Oops! I did it again... Moved a comment from the "Experimental: std::optional" commit to the "Material: semigroup support" where it belongs, then squashed the previous commits about "Experimental: std::optional" with an improved comment and some asserts added.

BTW, I just noticed that my history rewriting seems to have discarded @hjmjohnson's review comments (sorry!)... maybe try again?

rfschtkt:
Moved some inline comments from "Experimental: std::optional" to "Material: rvalue support" where they belong, and minimally changed some others. But the outcome was negative anyway, so this commit can be disregarded.

BenFrantzDale:
It seems like this adds a lot, going beyond the scope of adding rvalue support. Is there a reason not to do the simple change of just having `lambda_reduce_body`'s `result` be `Value& result()` and have its `join` be `my_value = my_reduction(tbb ::internal::move(my_value), tbb::internal::move(rhs.my_value));`?

I appreciate the desire for a more-elegant solution and for avoiding reconstructing default values, but it feels like that's something that callers could do if they want (using `std::unique_ptr<T>(nullptr)` or `std::optional<T>(std::nullopt)` as the identity, for example.

BenFrantzDale:
PS: There's no Issue to go with this. Should I create one?

alexey-katranov:
The main issue that we need to rebase the patch on onetbb_2021 because we are not planning to update TBB 2020. Unfortunately, the code base has changed over parallel_reduce and I expect a lot of merge conflicts. The good news is that we can rely on C++11 and do not need to provide compatibility with C++03.

Sorry that I request additional work, can you reconsider (or at least rebase) the approach on the new code base?

rfschtkt:
You certainly don't need all the ideas I had fun exploring here (although you also shouldn't forget to `move()` out of `result()`, as my changes to the tests would have demonstrated). The semigroup stuff (as I've called it here) is clearly differentiated in some other languages, e.g., with `foldl1` instead of `foldl` (Haskell) or with `reduceLeft()` instead of `foldLeft()` (Scala), and it's there for a reason (it's always an imposition on the user to have to emulate one in terms of the other, in both directions), although the reason why it now seems better to me not to attempt to use overloading for that here is because of the potential confusion with the use of default values in C++17 `reduce()`. BTW, git has no idea how to even begin to rebase this...

rfschtkt:
I'm not entirely convinced yet that git best practices were top of mind when branch onetbb_2021 was created... There are also lots of differences between the relevant corresponding files, so perhaps it would be best to first make firm decisions on what will be used, and to then reapply those changes manually to the new branch.

BTW, I'm curious how std::reduce() fares against tbb::parallel_reduce(), considering that the former is entirely dependent on compiler smarts to optimise range operations. On the other hand, it might be nice to also have a default range operation in TBB...

alexey-katranov:
The onetbb_2021 branch is not real branch. It is a completely new code base that partly was copied from TBB 2020 but the most of the code was reconsidered.

With ranges, do you mean C++20 ranges support?

rfschtkt:
What I meant was that TBB uses separate and redundant operations, one for reducing over a range of elements and the other for joining intermediate results: the former could most easily be implemented in terms of the latter, computationally speaking, but taking it as a parameter also offers an opportunity for explicit vectorisation or any other intervention for optimisation, so I was just wondering how often this is significant and to what extent.

rfschtkt:
Semigroup support now with "1" suffix, and not_partitioner_t no longer needed for SFINAE-based disambiguation. Maybe next I'll try to make the range operation parameter optional... :-)

rfschtkt:
I don't think I'll be playing around with making range operations optional after all. The later operation-based functions entirely replace the original Body-based ones, but to provide the same functionality without range operations (in those cases where _only_ performance-related things like explicit vectorisation can be omitted), there would also have to be `tbb::parallel_transform_reduce()` functions. That's a lot of dead code to write just for fun amid other options.

I'll look into what's required to rebase this, but if it takes significant effort I would prefer the editorialising to occur prior to the effort with a firm commitment to integrate the result...

kboyarinov:
Since #1307 was merged to master, this PR can be closed. Thanks to everyone for the productive discussion.
Feel free to reopen the PR of open a separate PR/issue in case of any questions on the current implementation.

PR #179, opened 2019-09-11
Build warnings

rfschtkt:
This points out a few build warnings as seen with Clang on macOS, solving one by initialising a variable before use, and blindly suppressing two others by eliminating an unused variable or a lot of dead code. I have not (yet?) analysed what would be the correct solution for the latter two, and the last one seems particularly problematic (how do you test that your test code is indeed executed...). Moral of the story: warnings are people too, don't ignore them!

### DIFF
--- a/include/tbb/concurrent_hash_map.h
+++ b/include/tbb/concurrent_hash_map.h
@@ -663,21 +663,10 @@ protected:
         // Emplace construct an empty T object inside the pair
         return create_node(allocator, std::piecewise_construct,
                            std::forward_as_tuple(key), std::forward_as_tuple());
-#error // this is not what's under investigation right now
 #else
-        // TODO: A compiler warning has been observed if obj is not initialized,
-        //       but can we also demonstrate failure without the initialization,
-        //       and then add that as a regression test? Unfortunately, it seems that
-        //       the value accidentally starts out as if it were zero-initialized anyway,
-        //       unless the memory is spoiled inside this implementation file,
-        //       so we only have a demonstration but not a regression test.
-        {
-            volatile unsigned int i = 0xDEADBEEF; // this will prevent having obj already be zero by accident
-            ++i;
-        }
         // Use of a temporary object is impossible, because create_node takes a non-const reference.
-        // We can initialize obj as follows because concurrent_hash_map already requires T to be CopyConstructible:
-        T obj;// = T(); // without this initialization and with spoiled memory underneath, the contract is broken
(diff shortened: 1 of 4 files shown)


rfschtkt:
For the first issue I added some quick-and-dirty code to demonstrate what can go wrong (although unfortunately the issue seems unreachable to a regression test), and for the third issue I re-enabled the code and instead added some assertions that should fail but don't because the code is not being executed.

rfschtkt:
Is anything still holding you back from integrating this?

The first commit points out the problems, and both commits together constitute a proposed solution. The first issue is pretty clear, even though I failed to tease out failure from the outside without a little help from the inside. For the second and third issues I just guessed, leaving it to you to validate whether those changes are appropriate.

hjmjohnson:
Thanks @rfschtkt, hopefully, the TBB team will review and comment on your proposed bug fixes.

include/tbb/concurrent_hash_map.h
```diff
@@ -664,7 +664,9 @@ class concurrent_hash_map : protected internal::hash_map_base {
         return create_node(allocator, std::piecewise_construct,
                            std::forward_as_tuple(key), std::forward_as_tuple());
 #else
-        T obj; // Use of temporary object in impossible, because create_node takes non-const reference
+        // Use of a temporary object is impossible, because create_node takes a non-const reference.
+        // copy-initialization is possible because T is already required to be CopyConstructible.
+        T obj = T();
```
alexey-katranov: Why not `T obj()`?

rfschtkt: That would declare `obj` as a function taking `void` and returning `T`, because the compiler is obliged to prioritise this interpretation of the ambiguity. With C++11 you can avoid it by using `T obj{};`, which would not depend on the presence of a copy constructor for class object types (most/all compilers perform the copy elision and so only insist to see the declaration but do not require a definition), until the problem goes away in C++17 (copy constructor no longer needed, compiler is obliged to elide the copy, and you're not even supposed to use that term anymore which seems quite silly to me).

I should probably have differentiated based on language version, and then some code might work from C++11 instead of only from C++17, so I would recommend that you amend this.

alexey-katranov: Surely, you are right about [most vexing parse](https://en.wikipedia.org/wiki/Most_vexing_parse). Ignore my irrelevant comment. (Writing `()` I thought about `{}` that is not applicable for pre-C++11.)

tbbdev:
Implemented.

Issue #60, opened 2018-05-24
[Build] Failed in building a iOS library


hanton:
hi,

I run `make target=ios arch=arm64 compiler=clang` in the source code directory for building a library for iOS, and it fails with
```
...
lib -stdlib=libc++ -arch arm64 -L/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS11.3.sdk/usr/lib/system -L/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS11.3.sdk/usr/lib/ -miphoneos-version-min=8.0 
ld: warning: directory not found for option '-L/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS11.3.sdk/usr/lib/system'
```

Any ideas for fixing this issue?

The whole log is 
```
➜  tbb-2018_U3 make target=ios arch=arm64 compiler=clang
Created ./build/macos_arm64_clang_cc9.1.0_ios11.3_release and ..._debug directories
/Applications/Xcode.app/Contents/Developer/usr/bin/make -C "./build/macos_arm64_clang_cc9.1.0_ios11.3_debug"  -r -f ../../build/Makefile.tbb cfg=debug
../../build/Makefile.tbb:32: CONFIG: cfg=debug arch=arm64 compiler=clang target=ios runtime=cc_ios11.3
clang++ -o concurrent_hash_map.o -c -MMD -g -O0 -DTBB_USE_DEBUG -DUSE_PTHREAD  -stdlib=libc++ -arch arm64 -isysroot /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS11.3.sdk -miphoneos-version-min=8.0  -fPIC  -D__TBB_BUILD=1 -Wall -Wno-non-virtual-dtor -Wno-dangling-else    -I../../src -I../../src/rml/include -I../../include ../../src/tbb/concurrent_hash_map.cpp
[...]

PR #192, opened 2019-10-30
fix for #60 linker warning on iOS/MacOS using clang

willadsen:
solves #60  because $(SDKROOT)/usr/lib/system doesn't exist anymore in apple SDK's


anton-potapov:
@willadsen thanks for your contribution!  

It seems that this patch is no longer applicable , as build system migrated to CMake.

Issue #195, opened 2019-11-04
TBBConfig.cmake should not add '-isystem /usr/local' to gcc's CXX_INCLUDES flag


ldorau:
TBBConfig.cmake:
https://github.com/intel/tbb/blob/tbb_2019/cmake/templates/TBBConfig.cmake.in#L57
sets the property `INTERFACE_INCLUDE_DIRECTORIES` to `${CMAKE_CURRENT_LIST_DIR}/../../../include` (which equals `/usr/include`).
It causes that
` -isystem /usr/lib64/cmake/TBB/../../../include `
(equal `-isystem /usr/include`)
is being added to gcc's `CXX_INCLUDES` flag, what is wrong and causes a compilation error - see https://github.com/pmem/libpmemobj-cpp/issues/508

ldorau:
Seen the fix: https://github.com/pmem/libpmemobj-cpp/pull/514

ldorau:
It is a workaround, not a fix!

Aldrog:
From discussion of opencv/opencv#15795 I understood it that cmake will filter out directories like this if they'll be normalized. I think you can get a normalized path if you construct it with `file` cmake command.

Aldrog:
This one works:
```cmake
get_filename_component(COMPONENT_INCLUDE_DIRECTORY "${CMAKE_CURRENT_LIST_DIR}/@TBB_INC_REL_PATH@" ABSOLUTE)
set_target_properties(TBB::${_tbb_component} PROPERTIES
                    INTERFACE_INCLUDE_DIRECTORIES "${COMPONENT_INCLUDE_DIRECTORY}")
```

tbbdev:
@Aldrog thank you for the proposal!
It was applied in https://github.com/intel/tbb/blob/v2020.0/cmake/templates/TBBConfig.cmake.in#L50

RuRo:
I am having some issues with building `opencv-cuda` with `tbb v2020.1` on Arch Linux. I am not too sure, how `${CMAKE_CURRENT_LIST_DIR}/@TBB_INC_REL_PATH@` works, but it seems, that in my case `INTERFACE_INCLUDE_DIRECTORIES` got resolved to just `/include`.

The error is
```markdown
CMake Error in modules/core/CMakeLists.txt:
  Imported target "tbb" includes non-existent path

    "/include"

  in its INTERFACE_INCLUDE_DIRECTORIES.  Possible reasons include:

[...]

PR #196, opened 2019-11-04
Do not add '-isystem /usr/local' to gcc's CXX_INCLUDES flag

ldorau:
Fixes: #195
Refs: pmem/libpmemobj-cpp#508

---
This change is [Reviewable](https://reviewable.io/reviews/intel/tbb/196)

### DIFF
--- a/cmake/templates/TBBConfig.cmake.in
+++ b/cmake/templates/TBBConfig.cmake.in
@@ -53,8 +53,6 @@ foreach (_tbb_component ${TBB_FIND_COMPONENTS})
     if (EXISTS "${_tbb_release_lib}" OR EXISTS "${_tbb_debug_lib}")
         if (NOT TARGET TBB::${_tbb_component})
             add_library(TBB::${_tbb_component} SHARED IMPORTED)
-            set_target_properties(TBB::${_tbb_component} PROPERTIES
-                                  INTERFACE_INCLUDE_DIRECTORIES "${CMAKE_CURRENT_LIST_DIR}/@TBB_INC_REL_PATH@")
 
             if (EXISTS "${_tbb_release_lib}")
                 set_target_properties(TBB::${_tbb_component} PROPERTIES


ldorau:
And maybe TBB should enable NO_SYSTEM_FROM_IMPORTED? See the fix: pmem/libpmemobj-cpp#514

AlexVeprev:
@ldorau removal of `INTERFACE_INCLUDE_DIRECTORIES` property from the TBB targets will likely lead to compilation errors due to missed TBB headers.

> And maybe TBB should enable NO_SYSTEM_FROM_IMPORTED?

TBB doesn't touch NO_SYSTEM_FROM_IMPORTED property, it relies on CMake default behavior. Moreover, looks like there is no way to change the behavior from TBB side, because it works for **_consuming_** targets. From CMake documentation for [NO_SYSTEM_FROM_IMPORTED](https://cmake.org/cmake/help/latest/prop_tgt/NO_SYSTEM_FROM_IMPORTED.html#prop_tgt:NO_SYSTEM_FROM_IMPORTED):

> Do not treat include directories from the interfaces of **_consumed_** Imported Targets as SYSTEM.

Here is a discussion of this topic: https://gitlab.kitware.com/cmake/cmake/issues/17348

Since

> This property is initialized by the value of the CMAKE_NO_SYSTEM_FROM_IMPORTED variable if it is set when a target is created.

https://github.com/pmem/libpmemobj-cpp/pull/514 seems to be a proper solution for the issue.

ldorau:
I wonder why only TBB (no any other component used by libpmemobj-cpp) needs CMAKE_NO_SYSTEM_FROM_IMPORTED to be enabled? What does TBB do that it requires that?

AlexVeprev:
I can assume that TBB uses some system headers which lead to the issue, but they are not used by other components. BTW, from the error log in https://github.com/pmem/libpmemobj-cpp/issues/508 I see inclusion of test headers and system headers. Have you checked that it is not an issue in test headers?

Aldrog:
If you normalize path before setting it to `INTERFACE_INCLUDE_DIRECTORIES` then cmake will be able to handle it properly in all cases.
```cmake
get_filename_component(COMPONENT_INCLUDE_DIRECTORY "${CMAKE_CURRENT_LIST_DIR}/@TBB_INC_REL_PATH@" ABSOLUTE)
set_target_properties(TBB::${_tbb_component} PROPERTIES
                    INTERFACE_INCLUDE_DIRECTORIES "${COMPONENT_INCLUDE_DIRECTORY}")
```

AlexVeprev:
@Aldrog thank you very much! I was able to reproduce and fix it with your approach. I'll prepare a patch.

ldorau:
This is not a good fix for #195. Please continue discussion in #195.

PR #197, opened 2019-11-07
Remove non-virtual methods from receiver class template

Jauler:
In some cases it is convenient to build a single node type (sender/receiver) which is composed of multiple other nodes (e.g. function node with bounded queueing policy, which requires limiter and function nodes to be combined). In this case it is beneficial to forward calls to public receiver class methods to some other node, but in order to do that, those methods must be marked as virtual. This PR marks all receiver class template methods as such.

### DIFF
--- a/include/tbb/flow_graph.h
+++ b/include/tbb/flow_graph.h
@@ -461,11 +461,11 @@ public:
     __TBB_DEPRECATED typedef typename internal::async_helpers<T>::filtered_type filtered_type;
 
     //! Put an item to the receiver
-    bool try_put( const typename internal::async_helpers<T>::filtered_type& t ) {
+    virtual bool try_put( const typename internal::async_helpers<T>::filtered_type& t ) {
         return internal::untyped_receiver::try_put(t);
     }
 
-    bool try_put( const typename internal::async_helpers<T>::async_type& t ) {
+    virtual bool try_put( const typename internal::async_helpers<T>::async_type& t ) {
         return internal::untyped_receiver::try_put(t);
     }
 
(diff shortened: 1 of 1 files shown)


aleksei-fedotov:
Hi @Jauler , thanks for the PR!

Usually, [composite_node](https://www.threadingbuildingblocks.org/docs/help/hh_goto.htm?index.htm#reference/flow_graph/composite_node_cls.html) is used for combining multiple nodes together to form a subgraph.
Could you please describe your preference against composite_node?

Jauler:
Well I have a few cases:
 - input
 - some internal processing nodes
 - output

Internal processing nodes are the simplest ones, therefore I described them in my original message, and yes, these are covered well by `composite_node` I should have read documentation more rigorously and should have described all of the cases.

We are trying to build stream processing graph, therefore sometimes it will have to sit idle waiting for input and sometimes input might come faster than graph can process it, so we used out-of-tbb thread for reading input (and sending output), but we need some flow-control. TBB message passing protocol seems perfect for this, basically pulling from input thread when input is coming faster than we can process it (but thread must be blocked if too many messages are accumulating). And pushing if input is scarce. This way we would be able to avoid some busy-waiting, unecessary sleeping, or calling `try_put()/try_get()` just for it to return `false` too much. To implement this deriving from sender<T> seemed the way to go. But I might be missing something here...

Output is basically the same, just a mirror image of the input therefore we required both, sender<T> and receiver<T> to be virtual.

kboyarinov:
@Jauler, could you please retarget the PR to the master branch, if it is still applicable for oneTBB latest version.

Jauler:
Hi,

Unfortunately, I have no access to the original case. But I remember I was unable to extend the receiver class due to it having non-virtual methods. 

So the current update is best-effort to remember what needed to be done, and then apply that to master. It seems like two `try_put()` specializations are gone, therefore the only change left is the method declaration itself.

PR #198, opened 2019-11-11
Fix CPP11 detection for clang-cl (clang on Windows)

e4lam:
When building with clang-cl (https://clang.llvm.org/docs/MSVCCompatibility.html) on Windows, it defines both `_MSC_VER` and `__clang__` but behaves more like MSVC. This patch fixes the CPP11 detection so that one can build using clang-cl on a project that uses TBB built with MSVC.

### DIFF
--- a/include/tbb/tbb_config.h
+++ b/include/tbb/tbb_config.h
@@ -206,7 +206,7 @@
     #define __TBB_CPP14_VARIABLE_TEMPLATES_PRESENT          (__cplusplus >= 201402L)
     #define __TBB_CPP17_DEDUCTION_GUIDES_PRESENT            (__INTEL_COMPILER > 1910) // a future version
     #define __TBB_CPP17_INVOKE_RESULT_PRESENT               (__cplusplus >= 201703L)
-#elif __clang__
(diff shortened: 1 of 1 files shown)


anton-potapov:
@e4lam is this still needed for oneTBB ? 
I believe oneTBB should support clang-cl out of the box (@AlexVeprev , @isaevil  please correct me if I'm wrong)

If this patch is still needed, please retarget the patch against master, as tbb2020 no longer updated.  (Please note that `tbb_config.h` has moved to [_config.h](https://github.com/oneapi-src/oneTBB/blob/master/include/oneapi/tbb/detail/_config.h))

e4lam:
@anton-potapov I don't know as I'm unable to upgrade up to oneTBB because all of my third party dependencies would first need to upgrade to oneTBB, which as you might know, is non-trivial. All I can say was that I'm still currently using this patch for TBB 2020. In oneTBB, this code (even in `_config`) has changed substantially, it seems like support for older MSVC versions have been dropped. So for the sake of cleanliness, I'll just close this and re-open it again if needed (which would probably need to be significantly different anyways).

PR #199, opened 2019-11-13
Fix clang-cl support by avoiding TSX

jdonald:
We see linker errors with clang-cl as Transactional Synchronization Extensions instructions are not supported, so set TSX_AVAILABLE to off with clang-cl.

We (@fnabulsi, @bmcdb, myself) have only tested with clang-cl 8.0.0, but at the same time I haven't seen any indication that this would change in clang-cl 9.x

This seems to be different from the clang-cl C++11 issue tackled in #198.

### DIFF
--- a/include/tbb/tbb_config.h
+++ b/include/tbb/tbb_config.h
@@ -107,7 +107,7 @@
 #define __TBB_DEFINE_MIC 1
 #endif
 
-#define __TBB_TSX_AVAILABLE  ((__TBB_x86_32 || __TBB_x86_64) && !__TBB_DEFINE_MIC)
+#define __TBB_TSX_AVAILABLE  ((__TBB_x86_32 || __TBB_x86_64) && !__TBB_DEFINE_MIC && !(defined(__clang__) && defined(_WIN32)))
 
 /** Presence of compiler features **/
 


akukanov:
Is `defined(__clang__) && defined(_WIN32))` the right way to detect clang-cl? Or does the upstream Clang compiler also not support TSX on Windows, so the problem is not specific to clang-cl?

jdonald:
Good point. I have only recently been made aware that there's a difference between clang-cl.exe and clang.exe.

Whether using `_WIN32` or `_MSC_VER`, this check would detect both clang-cl.exe and clang.exe. As for whether the missing TSX problem is specific to clang-cl, that's unlikely but we haven't confirmed yet. We also haven't done any testing under clang[-cl] 9, and will provide an update once we do. We don't have an ETA, but it isn't blocking us at the moment.

anton-potapov:
@jdonald is this still needed for oneTBB ? 
I believe oneTBB should support clang-cl out of the box (@AlexVeprev , @isaevil  please correct me if I'm wrong)

If this patch is still needed, please retarget the patch against master, as tbb2020 no longer updated.  (Please note that `tbb_config.h` has moved to [_config.h](https://github.com/oneapi-src/oneTBB/blob/master/include/oneapi/tbb/detail/_config.h))

jdonald:
Confirming from commit history, as of some time in 2020 our code started working in clang/clang-cl without this line change. Closing.

PR #229, opened 2020-03-19
Recover ability to build debug libraries using main Makefile

### LINKED ISSUE #207: How to make a debug build?
Take a look at Makefile. How are we supposed to build a debug build of TBB if all paths and cfg are hard coded to release?

amadio:
Between 2019.9 and 2020.0 the debug builds that were always created have been removed from the main Makefile. However, the lines left wrongly use hard-coded values for release instead of the value for the current configuration. This fixes #207 by replacing the hard-coded values with values for the current configuration.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,29 +26,29 @@ default: tbb tbbmalloc $(if $(use_proxy),tbbproxy)
 all: tbb tbbmalloc tbbproxy test examples
 
 tbb: mkdir
-	$(MAKE) -C "$(work_dir)_release"  -r -f $(tbb_root)/build/Makefile.tbb cfg=release
+	$(MAKE) -C "$(work_dir)_$(cfg)"  -r -f $(tbb_root)/build/Makefile.tbb cfg=$(cfg)
 
 tbbmalloc: mkdir
-	$(MAKE) -C "$(work_dir)_release"  -r -f $(tbb_root)/build/Makefile.tbbmalloc cfg=release malloc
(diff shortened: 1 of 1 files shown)


TheUnixRoot:
I'd recommend to use a global variable instead of repeating the same two $(work_dir)_$(cfg) something like $(output_dir) for keeping it easier

alexey-katranov:
Is this fix important to be integrated in TBB 2020? We switched to cmake in [next version of TBB](https://github.com/oneapi-src/oneTBB/tree/onetbb_2021)

amadio:
If it's not painful to include this patch, I'd appreciate it. We still use the Makefile in Gentoo (see ebuilds in https://packages.gentoo.org/packages/dev-cpp/tbb). Is CMake going to be the only supported build system for TBB 2021?

amadio:
@alexey-katranov Hi, has this patch been included in any newer version of TBB? Also, will the Makefile build system still be supported alongside CMake?

alexey-katranov:
> Is CMake going to be the only supported build system for TBB 2021?

We do not plan to support Makefiles in oneTBB (the next version of TBB after TBB 2020). oneTBB has only CMake support.

>  has this patch been included in any newer version of TBB? 

Unfortunately, no. Currently, I would not expect any updates for TBB 2020.

amadio:
Closing, since with TBB 2021.1.1, the Makefile is no longer there to be patched.

PR #239, opened 2020-04-08
Ignore cmake/{TBBConfig,TBBConfigVersion}.cmake

e-kwsm:


tbbdev:
Thank you! Applied in [TBB 2020 U3](https://github.com/oneapi-src/oneTBB/blob/v2020.3/CHANGES#L26).

PR #240, opened 2020-04-08
Switch to Python3

e-kwsm:
Also, `HAS_NO_PRIVILEGE` variable is added to install in user site-package.


ivankochin:
Thank you! Addressed in [oneTBB Beta08](https://github.com/oneapi-src/oneTBB/tree/05b1dd0789a02dfa77fa802b7172f52e02e2d788).

PR #245, opened 2020-04-30
bug: compile error - add missing std namespace prefix

marksisson:
Signed-off-by: Mark Sisson <5761292+marksisson@users.noreply.github.com>

This pull request fixes compile error on macOS when defining TBB_USE_EXCEPTIONS=0


anton-potapov:
@marksisson thanks for your contribution, but it seems that this patch is no longer needed for oneTBB ,(https://github.com/oneapi-src/oneTBB/blob/478de5b1887c928e52f029d706af6ea640a877be/src/tbb/exception.cpp#L79)

anton-potapov:
@marksisson, please reopen if the patch is still applicable

PR #251, opened 2020-05-20
Fixing iterator comparison for C++20.

BRevzin:
With C++20, the comparison operators as declared are ambiguous. This fix makes it work on C++20 while also maintaining correct functionality for previous C++ versions. 

A short reproduction is:

```cpp
template <typename Derived>
struct B {
    bool operator==(Derived const&) const;
};
struct D : B<D> { };
bool b = D{} == D{}; // error on C++20
```

### DIFF
--- a/src/tbb/intrusive_list.h
+++ b/src/tbb/intrusive_list.h
@@ -76,11 +76,11 @@ class intrusive_list_base {
             return my_pos = &node(val);
         }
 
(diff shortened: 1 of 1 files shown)


tbbdev:
Thank you! Addressed in [TBB  2020 U3](https://github.com/oneapi-src/oneTBB/blob/v2020.3/CHANGES#L18).

PR #258, opened 2020-07-03
Add detection of arm64 architecture for macos builds

betajippity:
This commit adds detection of arm64 for macos builds on arm64 hardware.
The only change necessary is an additional case in the code where
macos.inc queries and detects the system architecture.

### DIFF
--- a/build/macos.inc
+++ b/build/macos.inc
@@ -36,11 +36,15 @@ ifndef arch
      export arch:=ppc32
    endif
  else
+  ifeq ($(shell /usr/sbin/sysctl -n hw.machine),arm64)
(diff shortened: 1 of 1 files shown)


claui:
Confirming that the `hw.machine` branch works just fine. It even returns `x86_64` when run under Rosetta.

anton-potapov:
@betajippity oneTBB has moved to CMake. please update the patch (if still applicable)

anton-potapov:
@betajippity , FIY oneTBB should work on ARM64 (at least on Windows), for more details see https://github.com/oneapi-src/oneTBB/pull/507

PR #263, opened 2020-07-24
LRU cache: Fix crash when my_number_of_lru_history_items == 0

HenryHeffanFormlabs:
it crashes because it tries to remove 1 item from an empty list.


kboyarinov:
Close due to duplication with PR #939.

PR #268, opened 2020-08-25
Fix assertion condition in unordered_map

trebconnell:
This condition seems incorrect, and causing unexpected assertions in our code.

Have worked around by disabling asserts.

### DIFF
--- a/include/tbb/internal/_concurrent_unordered_impl.h
+++ b/include/tbb/internal/_concurrent_unordered_impl.h
@@ -463,14 +463,14 @@ public:
     // Returns a public iterator version of the internal iterator. Public iterator must not
     // be a dummy private iterator.
     iterator get_iterator(raw_iterator it) {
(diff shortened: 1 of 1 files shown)


kboyarinov:
Hi, @trebconnell,

Thank you for reporting the problem. It is really seems like an issue in TBB 2020. Consider using the updated version of concurrent_unordered_map from the [new version of TBB](https://github.com/oneapi-src/oneTBB/tree/onetbb_2021) in your project.

alexey-katranov:
@trebconnell , currently, we are not going to fix the issue in TBB 2020. Can you use oneTBB or do you have migration issues?

trebconnell:
Makes sense.

I'm using TBB via another open source project (USD)...  I'll try switching over in a fork and hopefully there's not too many issues.
Otherwise I can continue using my current workaround of just disabling the assertions.

PR #294, opened 2020-10-20
Use arm specific definitions when compiling for Apple Silicon/iOS

jsbache:
Use gcc_arm (intrinsics) for Apple Silicon and iOS devices.
This prevents a fallback to the faulty generic implementation (see issue https://github.com/oneapi-src/oneTBB/issues/293)


jsbache:
We are seeing some issues that may be related to this change. I am retracting this PR.

PR #306, opened 2020-11-15
Allow explicitly specifying the make tool name for tbb_build.

siretty:
Problem:
While trying the source package integration using cmake TBBBuild.cmake/tbb_build automatically chooses the make_tool_name based on CMAKE_SYSTEM_NAME. When GNU Make is e.g. installed from [Chocolatey](https://chocolatey.org/packages/make) or [gnuwin32](http://gnuwin32.sourceforge.net/packages/make.htm) the binary is called 'make' and the build fails.

Solution:
This patch adds the option to explicitly set the make_tool_name when calling tbb_build through an additional one-value-argument called MAKE_TOOL_NAME.

### DIFF
--- a/cmake/README.rst
+++ b/cmake/README.rst
@@ -333,7 +333,7 @@ TBBBuild
 Module for building TBB library from the source code.
 
 Provides the following functions:
- ``tbb_build(TBB_ROOT <tbb_root> CONFIG_DIR <variable> [MAKE_ARGS <custom_make_arguments>])``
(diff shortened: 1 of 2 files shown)


AlexVeprev:
@siretty, thank you for the proposal! Currently there are no plans to update TBB 2020, have you considered using oneTBB 2021? You can find it on [onetbb_2021](https://github.com/oneapi-src/oneTBB/tree/onetbb_2021) and [master](https://github.com/oneapi-src/oneTBB/tree/master) branches. There is CMake-based build system instead of Makefiles.

PR #323, opened 2020-12-24
Add mingw support + other bug fixes

longnguyen2004:
Enabling `TBB_STRICT` currently errors out due to one [unused variable in `misc_ex.cpp`](https://github.com/oneapi-src/oneTBB/blob/master/src/tbb/misc_ex.cpp#L237), but for now everything else is working.
The white box `tbbmalloc` test fails to compile though, since the compiler defines `__STRICT_ANSI__`, which disables the macros in `float.h`. I tried `#undef`ing the macro, and while it works when building `tbb` and `tbbmalloc`, it failed when building the tests, which is weird. I'm trying to fix it.


src/tbb/def/win32-mingw-tbb.def
```diff
@@ -0,0 +1,143 @@
+; Copyright (c) 2005-2020 Intel Corporation
```
alexey-katranov, line 1: @AlexVeprev, @longnguyen2004 , the def files look very similar with Linux *def files. Can we unify them somehow?

src/tbb/tools_api/ittnotify_static.c
```diff
@@ -17,6 +17,7 @@
 #include "ittnotify_config.h"
 
 #if ITT_PLATFORM==ITT_PLATFORM_WIN
+#undef PATH_MAX
```
alexey-katranov, line 20:
```suggestion
#ifdef PATH_MAX
#undef PATH_MAX
#endif
```
Or
```suggestion
#ifndef PATH_MAX
#define PATH_MAX 512
#endif
```

alexey-katranov (commented):
The main question is about unification of def files, otherwise, it will burden future development of new entry points even more.
Notify: @AlexVeprev

src/tbb/def/win32-mingw-tbb.def
longnguyen2004: The `win32-mingw` file is similar to `lin32`, with `char` replaced with `wchar_t`. Likewise, `win64-mingw` is similar to `lin64`, with `char` replaced with `wchar_t` and `(unsigned) long` replaced with `(unsigned) long long`

longnguyen2004: By the way, how are you generating the MSVC and Linux def files? We can mix a bit between the two.

alexey-katranov: > By the way, how are you generating the MSVC and Linux def files? We can mix a bit between the two.

Manually, that is the main burden issue.

alexey-katranov: > The `win32-mingw` file is similar to `lin32`, with `char` replaced with `wchar_t`. Likewise, `win64-mingw` is similar to `lin64`, with `char` replaced with `wchar_t` and `(unsigned) long` replaced with `(unsigned) long long`

Interestingly, can we build mingw files from linux files on fly?

alexey-katranov:
As for ITT, disable it in src/tbb/CMakeLists.txt near lines 54-58 (perhaps, you do not need related changes). The macro `__TBB_USE_ITT_NOTIFY` should not be defined for the mingw build.

src/tbb/def/win32-mingw-tbb.def
longnguyen2004: It's certainly possible, but can it be done with pure CMake?
Here's what needs to be changed from the Linux def:
`lin32-tbb.def` (`char --> wchar_t`):
```
_ZN3tbb6detail2r115create_itt_syncEPvPKcS4_ --> _ZN3tbb6detail2r115create_itt_syncEPvPKwS4_
_ZN3tbb6detail2r117itt_set_sync_nameEPvPKc  --> _ZN3tbb6detail2r117itt_set_sync_nameEPvPKw
```

`lin64-tbb.def` (`char --> wchar_t`, `(unsigned) long --> (unsigned) long long`):
```
_ZN3tbb6detail2r115create_itt_syncEPvPKcS4_ --> _ZN3tbb6detail2r115create_itt_syncEPvPKwS4_
_ZN3tbb6detail2r117itt_set_sync_nameEPvPKc --> _ZN3tbb6detail2r117itt_set_sync_nameEPvPKw
_ZN3tbb6detail2r115allocate_memoryEm --> _ZN3tbb6detail2r115allocate_memoryEy
_ZN3tbb6detail2r122cache_aligned_allocateEm --> _ZN3tbb6detail2r122cache_aligned_allocateEy
_ZN3tbb6detail2r18allocateERPNS0_2d117small_object_poolEm --> _ZN3tbb6detail2r18allocateERPNS0_2d117small_object_poolEy
_ZN3tbb6detail2r18allocateERPNS0_2d117small_object_poolEmRKNS2_14execution_dataE --> _ZN3tbb6detail2r18allocateERPNS0_2d117small_object_poolEyRKNS2_14execution_dataE
_ZN3tbb6detail2r110deallocateERNS0_2d117small_object_poolEPvm --> _ZN3tbb6detail2r110deallocateERNS0_2d117small_object_poolEPvy
_ZN3tbb6detail2r110deallocateERNS0_2d117small_object_poolEPvmRKNS2_14execution_dataE --> _ZN3tbb6detail2r110deallocateERNS0_2d117small_object_poolEPvyRKNS2_14execution_dataE
_ZN3tbb6detail2r114notify_waitersEm --> _ZN3tbb6detail2r114notify_waitersEy
_ZN3tbb6detail2r16submitERNS0_2d14taskERNS2_18task_group_contextEPNS1_5arenaEm --> _ZN3tbb6detail2r16submitERNS0_2d14taskERNS2_18task_group_contextEPNS1_5arenaEy
_ZN3tbb6detail2r120isolate_within_arenaERNS0_2d113delegate_baseEl --> _ZN3tbb6detail2r120isolate_within_arenaERNS0_2d113delegate_baseEx
_ZN3tbb6detail2r18finalizeERNS0_2d121task_scheduler_handleEl --> _ZN3tbb6detail2r18finalizeERNS0_2d121task_scheduler_handleEx
_ZN3tbb6detail2r117parallel_pipelineERNS0_2d118task_group_contextEmRKNS2_11filter_nodeE --> _ZN3tbb6detail2r117parallel_pipelineERNS0_2d118task_group_contextEyRKNS2_11filter_nodeE
_ZN3tbb6detail2r126allocate_bounded_queue_repEm --> _ZN3tbb6detail2r126allocate_bounded_queue_repEy
_ZN3tbb6detail2r126wait_bounded_queue_monitorEPNS1_18concurrent_monitorEmlRNS0_2d113delegate_baseE --> _ZN3tbb6detail2r126wait_bounded_queue_monitorEPNS1_18concurrent_monitorEyxRNS0_2d113delegate_baseE
_ZN3tbb6detail2r128deallocate_bounded_queue_repEPhm --> _ZN3tbb6detail2r128deallocate_bounded_queue_repEPhy
_ZN3tbb6detail2r128notify_bounded_queue_monitorEPNS1_18concurrent_monitorEmm --> _ZN3tbb6detail2r128notify_bounded_queue_monitorEPNS1_18concurrent_monitorEyy
```
There's also `tbbmalloc`, and I'm working on it.
Basically everything using `tchar`, `std::size_t`, `std::(u)intptr_t` and `std::ptrdiff_t` have to be replaced for 64 bit.

longnguyen2004: Also this is the [Python script](https://pastebin.com/tbrFGUWB) I use to create the def file from the Linux def file. I haven't implemented the symbol replacement yet, but it can be done with minimal effort.

longnguyen2004:
> The macro `__TBB_USE_ITT_NOTIFY` should not be defined for the mingw build.

I don't remove it and it seems to build fine, so I think we should keep it there.

cmake/compilers/Clang.cmake
```diff
@@ -39,6 +47,9 @@ if (NOT APPLE)
     set(TBB_WARNING_SUPPRESS -Wno-parentheses)
 endif()
 
+if (NOT CMAKE_CXX_COMPILER_VERSION VERSION_LESS 12)
+    set(TBB_LIB_COMPILE_FLAGS -mwaitpkg)
```
AlexVeprev, line 51: `-mwaitpkg` is [x86](https://clang.llvm.org/docs/ClangCommandLineReference.html#x86) option, so it should be applied only for this architecture, consider moving it under `if (CMAKE_SYSTEM_PROCESSOR STREQUAL x86_64)` on line 33.

Consider adding the same key to GNU.cmake.

src/tbb/def/win32-mingw-tbb.def
AlexVeprev: If we have specific rules for replacements, I'd agree on using script on fly, but I'd prefer to make implementation in CMake to avoid additional dependency on Python. It can be done on our side later.

longnguyen2004:
After my `__STRICT_ANSI__` fix, the tests now build successfully, and 126/128 tests passed
```
mingw32-make test
Running tests...
Test project D:/github/onetbb/build
        Start   1: test_tick_count
  1/128 Test   #1: test_tick_count ..........................   Passed    1.48 sec
        Start   2: test_allocators
  2/128 Test   #2: test_allocators ..........................   Passed    0.30 sec
        Start   3: test_arena_constraints
  3/128 Test   #3: test_arena_constraints ...................   Passed    0.28 sec
        Start   4: test_arena_priorities
  4/128 Test   #4: test_arena_priorities ....................   Passed    0.16 sec
        Start   5: test_dynamic_link
  5/128 Test   #5: test_dynamic_link ........................   Passed    0.15 sec
        Start   6: test_concurrent_lru_cache
  6/128 Test   #6: test_concurrent_lru_cache ................   Passed    0.21 sec
        Start   7: test_concurrent_unordered_map
  7/128 Test   #7: test_concurrent_unordered_map ............   Passed    4.84 sec
        Start   8: test_concurrent_unordered_set
  8/128 Test   #8: test_concurrent_unordered_set ............   Passed    7.33 sec
        Start   9: test_concurrent_map
  9/128 Test   #9: test_concurrent_map ......................   Passed    1.24 sec
        Start  10: test_concurrent_set
 10/128 Test  #10: test_concurrent_set ......................   Passed    8.37 sec
        Start  11: test_concurrent_priority_queue
 11/128 Test  #11: test_concurrent_priority_queue ...........   Passed    0.52 sec
        Start  12: test_parallel_for
 12/128 Test  #12: test_parallel_for ........................   Passed   60.55 sec
        Start  13: test_parallel_for_each
 13/128 Test  #13: test_parallel_for_each ...................   Passed    0.59 sec
        Start  14: test_parallel_reduce
 14/128 Test  #14: test_parallel_reduce .....................   Passed  207.44 sec
        Start  15: test_parallel_sort
 15/128 Test  #15: test_parallel_sort .......................   Passed    1.47 sec
        Start  16: test_parallel_invoke
 16/128 Test  #16: test_parallel_invoke .....................   Passed    0.47 sec
        Start  17: test_parallel_scan
 17/128 Test  #17: test_parallel_scan .......................   Passed    3.61 sec
        Start  18: test_parallel_pipeline
 18/128 Test  #18: test_parallel_pipeline ...................   Passed    4.88 sec
        Start  19: test_eh_algorithms
 19/128 Test  #19: test_eh_algorithms .......................   Passed    3.02 sec
        Start  20: test_blocked_range
 20/128 Test  #20: test_blocked_range .......................   Passed    0.17 sec
        Start  21: test_concurrent_vector
 21/128 Test  #21: test_concurrent_vector ...................   Passed    0.71 sec
        Start  22: test_task_group
 22/128 Test  #22: test_task_group ..........................   Passed    8.72 sec
        Start  23: test_concurrent_hash_map
 23/128 Test  #23: test_concurrent_hash_map .................   Passed    0.33 sec
        Start  24: test_task_arena
 24/128 Test  #24: test_task_arena ..........................   Passed   32.67 sec
        Start  25: test_enumerable_thread_specific
 25/128 Test  #25: test_enumerable_thread_specific ..........   Passed    0.76 sec
        Start  26: test_concurrent_queue
 26/128 Test  #26: test_concurrent_queue ....................   Passed    0.35 sec
        Start  27: test_resumable_tasks
 27/128 Test  #27: test_resumable_tasks .....................   Passed    1.21 sec
        Start  28: test_mutex
 28/128 Test  #28: test_mutex ...............................   Passed    2.39 sec
        Start  29: test_function_node
 29/128 Test  #29: test_function_node .......................   Passed    3.24 sec
        Start  30: test_multifunction_node
 30/128 Test  #30: test_multifunction_node ..................   Passed    3.16 sec
        Start  31: test_broadcast_node
 31/128 Test  #31: test_broadcast_node ......................   Passed    0.18 sec
        Start  32: test_buffer_node
 32/128 Test  #32: test_buffer_node .........................   Passed    0.34 sec
        Start  33: test_composite_node
 33/128 Test  #33: test_composite_node ......................   Passed    2.50 sec
        Start  34: test_continue_node
 34/128 Test  #34: test_continue_node .......................   Passed    2.31 sec
        Start  35: test_eh_flow_graph
 35/128 Test  #35: test_eh_flow_graph .......................   Passed    1.38 sec
        Start  36: test_flow_graph
 36/128 Test  #36: test_flow_graph ..........................   Passed    0.86 sec
        Start  37: test_flow_graph_priorities
 37/128 Test  #37: test_flow_graph_priorities ...............   Passed    0.36 sec
        Start  38: test_flow_graph_whitebox
 38/128 Test  #38: test_flow_graph_whitebox .................   Passed    0.22 sec
        Start  39: test_indexer_node
 39/128 Test  #39: test_indexer_node ........................   Passed    0.59 sec
        Start  40: test_join_node
 40/128 Test  #40: test_join_node ...........................   Passed    0.71 sec
        Start  41: test_join_node_key_matching
 41/128 Test  #41: test_join_node_key_matching ..............   Passed    0.77 sec
        Start  42: test_join_node_msg_key_matching
 42/128 Test  #42: test_join_node_msg_key_matching ..........   Passed    0.74 sec
        Start  43: test_limiter_node
 43/128 Test  #43: test_limiter_node ........................   Passed    0.44 sec
        Start  44: test_priority_queue_node
 44/128 Test  #44: test_priority_queue_node .................   Passed    0.23 sec
        Start  45: test_queue_node
 45/128 Test  #45: test_queue_node ..........................   Passed    0.72 sec
        Start  46: test_sequencer_node
 46/128 Test  #46: test_sequencer_node ......................   Passed    0.20 sec
        Start  47: test_split_node
 47/128 Test  #47: test_split_node ..........................   Passed    0.21 sec
        Start  48: test_tagged_msg
 48/128 Test  #48: test_tagged_msg ..........................   Passed    0.16 sec
        Start  49: test_overwrite_node
 49/128 Test  #49: test_overwrite_node ......................   Passed    2.98 sec
        Start  50: test_write_once_node
 50/128 Test  #50: test_write_once_node .....................   Passed    3.02 sec
        Start  51: test_async_node
 51/128 Test  #51: test_async_node ..........................   Passed    8.03 sec
        Start  52: test_input_node
 52/128 Test  #52: test_input_node ..........................   Passed    0.15 sec
        Start  53: test_profiling
 53/128 Test  #53: test_profiling ...........................   Passed    0.17 sec
        Start  54: test_concurrent_queue_whitebox
 54/128 Test  #54: test_concurrent_queue_whitebox ...........   Passed    0.13 sec
        Start  55: test_intrusive_list
 55/128 Test  #55: test_intrusive_list ......................   Passed    1.22 sec
        Start  56: test_semaphore
 56/128 Test  #56: test_semaphore ...........................   Passed    7.46 sec
        Start  57: test_environment_whitebox
 57/128 Test  #57: test_environment_whitebox ................   Passed    0.20 sec
        Start  58: test_handle_perror
 58/128 Test  #58: test_handle_perror .......................   Passed    0.12 sec
        Start  59: test_hw_concurrency
 59/128 Test  #59: test_hw_concurrency ......................   Passed    0.15 sec
        Start  60: test_eh_thread
 60/128 Test  #60: test_eh_thread ...........................   Passed    0.11 sec
        Start  61: test_global_control
 61/128 Test  #61: test_global_control ......................   Passed    0.71 sec
        Start  62: test_task
 62/128 Test  #62: test_task ................................   Passed    1.58 sec
        Start  63: test_concurrent_monitor
 63/128 Test  #63: test_concurrent_monitor ..................   Passed    0.20 sec
        Start  64: test_tbb_fork
 64/128 Test  #64: test_tbb_fork ............................   Passed    6.42 sec
        Start  65: test_tbb_header
 65/128 Test  #65: test_tbb_header ..........................***Exception: SegFault  0.24 sec
        Start  66: conformance_tick_count
 66/128 Test  #66: conformance_tick_count ...................   Passed    0.10 sec
        Start  67: conformance_allocators
 67/128 Test  #67: conformance_allocators ...................   Passed    6.70 sec
        Start  68: conformance_arena_constraints
 68/128 Test  #68: conformance_arena_constraints ............   Passed    0.13 sec
        Start  69: conformance_mutex
 69/128 Test  #69: conformance_mutex ........................   Passed    1.71 sec
        Start  70: conformance_task_group_context
 70/128 Test  #70: conformance_task_group_context ...........   Passed    0.33 sec
        Start  71: conformance_task_arena
 71/128 Test  #71: conformance_task_arena ...................   Passed    0.14 sec
        Start  72: conformance_concurrent_lru_cache
 72/128 Test  #72: conformance_concurrent_lru_cache .........   Passed    0.12 sec
        Start  73: conformance_concurrent_unordered_map
 73/128 Test  #73: conformance_concurrent_unordered_map .....   Passed    0.29 sec
        Start  74: conformance_concurrent_unordered_set
 74/128 Test  #74: conformance_concurrent_unordered_set .....   Passed    0.98 sec
        Start  75: conformance_concurrent_map
 75/128 Test  #75: conformance_concurrent_map ...............   Passed    0.51 sec
        Start  76: conformance_concurrent_set
 76/128 Test  #76: conformance_concurrent_set ...............   Passed    0.37 sec
        Start  77: conformance_concurrent_priority_queue
 77/128 Test  #77: conformance_concurrent_priority_queue ....   Passed    0.33 sec
        Start  78: conformance_parallel_for
 78/128 Test  #78: conformance_parallel_for .................   Passed    4.02 sec
        Start  79: conformance_parallel_for_each
 79/128 Test  #79: conformance_parallel_for_each ............   Passed    0.19 sec
        Start  80: conformance_parallel_reduce
 80/128 Test  #80: conformance_parallel_reduce ..............   Passed    0.36 sec
        Start  81: conformance_parallel_scan
 81/128 Test  #81: conformance_parallel_scan ................   Passed    0.12 sec
        Start  82: conformance_parallel_sort
 82/128 Test  #82: conformance_parallel_sort ................   Passed    0.11 sec
        Start  83: conformance_parallel_pipeline
 83/128 Test  #83: conformance_parallel_pipeline ............   Passed    1.08 sec
        Start  84: conformance_parallel_invoke
 84/128 Test  #84: conformance_parallel_invoke ..............   Passed    0.53 sec
        Start  85: conformance_blocked_range
 85/128 Test  #85: conformance_blocked_range ................   Passed    0.43 sec
        Start  86: conformance_blocked_range2d
 86/128 Test  #86: conformance_blocked_range2d ..............   Passed    1.41 sec
        Start  87: conformance_blocked_range3d
 87/128 Test  #87: conformance_blocked_range3d ..............   Passed    0.44 sec
        Start  88: conformance_blocked_rangeNd
 88/128 Test  #88: conformance_blocked_rangeNd ..............   Passed    0.22 sec
        Start  89: conformance_concurrent_vector
 89/128 Test  #89: conformance_concurrent_vector ............   Passed   13.36 sec
        Start  90: conformance_global_control
 90/128 Test  #90: conformance_global_control ...............   Passed  431.34 sec
        Start  91: conformance_concurrent_hash_map
 91/128 Test  #91: conformance_concurrent_hash_map ..........   Passed    6.75 sec
        Start  92: conformance_enumerable_thread_specific
 92/128 Test  #92: conformance_enumerable_thread_specific ...   Passed   13.72 sec
        Start  93: conformance_combinable
 93/128 Test  #93: conformance_combinable ...................   Passed    1.85 sec
        Start  94: conformance_concurrent_queue
 94/128 Test  #94: conformance_concurrent_queue .............   Passed   22.21 sec
        Start  95: conformance_resumable_tasks
 95/128 Test  #95: conformance_resumable_tasks ..............   Passed    0.09 sec
        Start  96: conformance_version
 96/128 Test  #96: conformance_version ......................   Passed    0.12 sec
        Start  97: conformance_function_node
 97/128 Test  #97: conformance_function_node ................   Passed   10.28 sec
        Start  98: conformance_multifunction_node
 98/128 Test  #98: conformance_multifunction_node ...........   Passed   10.30 sec
        Start  99: conformance_input_node
 99/128 Test  #99: conformance_input_node ...................   Passed    0.10 sec
        Start 100: conformance_continue_node
100/128 Test #100: conformance_continue_node ................   Passed    0.11 sec
        Start 101: conformance_async_node
101/128 Test #101: conformance_async_node ...................   Passed    0.09 sec
        Start 102: conformance_overwrite_node
102/128 Test #102: conformance_overwrite_node ...............   Passed    0.07 sec
        Start 103: conformance_write_once_node
103/128 Test #103: conformance_write_once_node ..............   Passed    0.14 sec
        Start 104: conformance_buffer_node
104/128 Test #104: conformance_buffer_node ..................   Passed    0.12 sec
        Start 105: conformance_queue_node
105/128 Test #105: conformance_queue_node ...................   Passed    0.09 sec
        Start 106: conformance_priority_queue_node
106/128 Test #106: conformance_priority_queue_node ..........   Passed    0.12 sec
        Start 107: conformance_sequencer_node
107/128 Test #107: conformance_sequencer_node ...............   Passed    0.08 sec
        Start 108: conformance_limiter_node
108/128 Test #108: conformance_limiter_node .................   Passed    0.06 sec
        Start 109: conformance_broadcast_node
109/128 Test #109: conformance_broadcast_node ...............   Passed    0.07 sec
        Start 110: conformance_composite_node
110/128 Test #110: conformance_composite_node ...............   Passed    0.08 sec
        Start 111: conformance_indexer_node
111/128 Test #111: conformance_indexer_node .................   Passed    0.15 sec
        Start 112: conformance_split_node
112/128 Test #112: conformance_split_node ...................   Passed    0.12 sec
        Start 113: conformance_join_node
113/128 Test #113: conformance_join_node ....................   Passed    0.08 sec
        Start 114: conformance_graph
114/128 Test #114: conformance_graph ........................   Passed    0.10 sec
        Start 115: test_scalable_allocator
115/128 Test #115: test_scalable_allocator ..................   Passed    3.46 sec
        Start 116: test_malloc_pools
116/128 Test #116: test_malloc_pools ........................   Passed    2.92 sec
        Start 117: test_malloc_init_shutdown
117/128 Test #117: test_malloc_init_shutdown ................   Passed    2.12 sec
        Start 118: test_malloc_regression
118/128 Test #118: test_malloc_regression ...................   Passed    0.41 sec
        Start 119: test_malloc_shutdown_hang
119/128 Test #119: test_malloc_shutdown_hang ................   Passed    3.79 sec
        Start 120: test_malloc_compliance
120/128 Test #120: test_malloc_compliance ...................   Passed   25.41 sec
        Start 121: test_malloc_used_by_lib
121/128 Test #121: test_malloc_used_by_lib ..................Exit code 0xc0000409
***Exception:   2.09 sec
        Start 122: test_malloc_lib_unload
122/128 Test #122: test_malloc_lib_unload ...................   Passed    0.17 sec
        Start 123: test_malloc_pure_c
123/128 Test #123: test_malloc_pure_c .......................   Passed    0.16 sec
        Start 124: test_malloc_whitebox
124/128 Test #124: test_malloc_whitebox .....................   Passed   11.35 sec
        Start 125: test_malloc_atexit
125/128 Test #125: test_malloc_atexit .......................   Passed    0.38 sec
        Start 126: test_malloc_overload
126/128 Test #126: test_malloc_overload .....................   Passed    0.13 sec
        Start 127: test_malloc_overload_disable
127/128 Test #127: test_malloc_overload_disable .............   Passed    0.17 sec
        Start 128: test_malloc_new_handler
128/128 Test #128: test_malloc_new_handler ..................   Passed    0.11 sec

98% tests passed, 2 tests failed out of 128

Total Test time (real) = 996.06 sec

The following tests FAILED:
         65 - test_tbb_header (SEGFAULT)
        121 - test_malloc_used_by_lib (Exit code 0xc0000409
)
Errors while running CTest
mingw32-make: *** [Makefile:113: test] Error 8
```
One SEGFAULT, and one buffer overflow, I'll see what's wrong with it.

cmake/compilers/GNU.cmake
```diff
@@ -12,17 +12,31 @@
...
+
 set(TBB_WARNING_LEVEL -Wall -Wextra $<$<BOOL:${TBB_STRICT}>:-Werror> -Wfatal-errors)
 set(TBB_TEST_WARNING_FLAGS -Wshadow -Wcast-qual -Woverloaded-virtual -Wnon-virtual-dtor)
 
 set(TBB_MMD_FLAG -MMD)
 if (CMAKE_SYSTEM_PROCESSOR STREQUAL x86_64)
     set(TBB_COMMON_COMPILE_FLAGS -mrtm)
+    check_cxx_compiler_flag(-mwaitpkg WAITPKG_SUPPORTED)
+    if (WAITPKG_SUPPORTED)
+        set(TBB_LIB_COMPILE_FLAGS -mwaitpkg)
```
AlexVeprev, line 33: @alexey-katranov do you think this flag is applicable only for library build or for tests also?

cmake/compilers/GNU.cmake
```diff
@@ -48,6 +62,10 @@ if ("${CMAKE_SYSTEM_PROCESSOR}" MATCHES "mips")
     set(TBB_TEST_COMPILE_FLAGS ${TBB_TEST_COMPILE_FLAGS} -DTBB_TEST_LOW_WORKLOAD)
 endif()
 
+if (MINGW)
+    add_compile_options(-U __STRICT_ANSI__)
```
AlexVeprev, line 66: Consider appending the options to `TBB_COMMON_COMPILE_FLAGS` variable

longnguyen2004:
@AlexVeprev I think we can maintain a list of functions that use `tchar`, `std::size_t`, `std::(u)intptr_t` and `std::ptrdiff_t`, then create a replacement table based on that. What do you think?

longnguyen2004:
Only one test fails now, which is `test_tbb_header`. It failed while trying to catch a `tbb::bad_last_alloc()` exception, but ended up getting a "NULL" (`0xffffffffffffffff`) exception instead, which caused an error when trying to get the `what()` string. What is causing this issue?

longnguyen2004:
@AlexVeprev Regarding the CMake script, should we execute it at configure time (`execute_command`) or build time (`add_custom_command`)?

cmake/compilers/Clang.cmake
```diff
@@ -24,11 +30,17 @@ if (NOT TBB_STRICT AND COMMAND tbb_remove_compile_flag)
     tbb_remove_compile_flag(-Werror)
 endif()
 
-if (CMAKE_SYSTEM_PROCESSOR STREQUAL x86_64)
+if (CMAKE_SYSTEM_PROCESSOR MATCHES "(amd64)|(AMD64)|(x86_64)")
```
AlexVeprev, line 33: Seems the expression can be a bit simplified:
`"(amd64|AMD64|x86_64)"`

cmake/compilers/GNU.cmake
```diff
@@ -48,6 +60,11 @@ if ("${CMAKE_SYSTEM_PROCESSOR}" MATCHES "mips")
     set(TBB_TEST_COMPILE_FLAGS ${TBB_TEST_COMPILE_FLAGS} -DTBB_TEST_LOW_WORKLOAD)
 endif()
 
+if (MINGW)
+    set(TBB_COMMON_COMPILE_FLAGS ${TBB_COMMON_COMPILE_FLAGS} -U __STRICT_ANSI__ -D_WIN32_WINNT=0x0601)
+    set(TBB_TEST_COMPILE_FLAGS ${TBB_TEST_COMPILE_FLAGS} -Wa,-mbig-obj)
```
AlexVeprev, line 65: Shouldn't it be added into `Clang.cmake` as well?

cmake/scripts/gen_mingw_def.cmake
```diff
@@ -0,0 +1,114 @@
+set(tbb_32_FROM
```
AlexVeprev, line 1: Such hardcoded symbols are fine to start with, but such approach will require manual adding new symbols here if they require similar replacements. I wonder if some universal rules can be applied, like `Kc` -> `Kw`, `Em` -> `Ey`, but seems they can't be reliable enough to work in 100% cases and do not touch unrelated parts.

Considering hardcoded symbols: I think there is not much value in comparison with just separate def files for MinGW, such automatic conversion may work worse in future development: developer has a set of def files in one place where he should add a new symbol, so he may hardly remember to add the symbol to the script in some other location.

cmake/scripts/gen_mingw_def.cmake
```diff
@@ -0,0 +1,114 @@
...
+    foreach(bitness 32 64)
+        message("Generating win${bitness}-mingw-${component}.def")
+        file(WRITE ${CMAKE_CURRENT_SOURCE_DIR}/../../src/${component}/def/win${bitness}-mingw-${component}.def ${DEF_START})
+        file(STRINGS ${CMAKE_CURRENT_SOURCE_DIR}/../../src/${component}/def/lin${bitness}-${component}.def INPUT)
+        foreach(line IN LISTS INPUT)
+            if(line MATCHES "(local:)|(}\;)")
+                break()
+            elseif(line MATCHES "[A-Za-z0-9_]+\;")
+                string(REPLACE ";" "" symbol_no_semicolon "${line}")
+                foreach(from to IN ZIP_LISTS ${component}_${bitness}_FROM ${component}_${bitness}_TO)
```
AlexVeprev, line 103: `ZIP_LISTS` looks nice, but [it's new in CMake 3.17](https://cmake.org/cmake/help/latest/release/3.17.html#commands) while oneTBB claims support starting from [CMake 3.1](https://github.com/oneapi-src/oneTBB/blob/master/CMakeLists.txt#L15). So `ZIP_LISTS` should either be replaced or wrapped with CMake version check.

AlexVeprev:
> @AlexVeprev Regarding the CMake script, should we execute it at configure time (execute_process) or build time (add_custom_command)?

Phase doesn't matter much; the main point is that an output file should be recreated if source file is changed, seems it may be done easier with `add_custom_command`.

cmake/compilers/Clang.cmake
longnguyen2004: Clang+LLD on Windows defaults to outputting big object files by default, so we don't need the flag.
Speaking of object files, currently gcc fails to compile `test_eh_flow_graph` on Windows, due to the linker complaining about string table overflow. We might have to turn on optimizations for the test, to reduce the symbol length.

cmake/compilers/GNU.cmake
longnguyen2004: Clang+LLD on Windows defaults to outputting big object files by default, so we don't need the flag.
Speaking of object files, currently gcc fails to compile `test_eh_flow_graph` on Windows, due to the linker complaining about string table overflow. We might have to turn on optimizations for the test, to reduce the symbol length.

cmake/scripts/gen_mingw_def.cmake
longnguyen2004: Like I said in my issue, the proper way to do this is to add the proper export attributes (`__declspec(dllexport)`, `__attribute__((visibility("default")))`) to the symbols in the source files itself. However, doing this can be time-consuming for such a large codebase.
Also, is there a specific need for the def files that the export attributes can't do?

alexey-katranov: > the proper way to do this is to add export attributes (__declspec(dllexport), __attribute__((visibility("default")))) to the symbols in the source files itself

The Microsoft [documentation](https://docs.microsoft.com/en-us/cpp/build/determining-which-exporting-method-to-use?view=msvc-160) recommends using .def files because "_This is very convenient if you are designing a DLL for use by many applications because you can add new functionality and also ensure that it continues to work correctly with the applications that already rely on it_".

In addition, it warns about `__declspec(dllexport)`: "_If you rebuild the DLL with new exports, you also have to rebuild the applications because the decorated names for exported C++ functions might change if you use a different version of the compiler to rebuild it_".

longnguyen2004: > they can't be reliable enough to work in 100% cases and do not touch unrelated parts.

I agree. We need to change `tchar`, `std::size_t`, `std::(u)intptr_t`, `std::ptrdiff_t` and somehow preserve actual `(unsigned) long`. With only the def files, this is not possible. Also there are many different combinations of parameters, and hardcoding all of them would be a pain.

longnguyen2004: > In addition, it warns about `__declspec(dllexport)`: "_If you rebuild the DLL with new exports, you also have to rebuild the applications because the decorated names for exported C++ functions might change if you use a different version of the compiler to rebuild it_".

But the same situation would also happen with def files. If a new compiler version changes the name decorating rules, then the def files won't be correct anymore, and we'll have to maintain 2 files for the old and the new version, which in my opinion is worse than using `__declspec(dllexport)`. Thankfully, both the Itanium and Microsoft C++ ABI are now fairly stable, so we don't have to worry too much about symbol names over e.g. CRT mismatch.

> "This is very convenient if you are designing a DLL for use by many applications because you can add new functionality and also ensure that it continues to work correctly with the applications that already rely on it"

This is really only useful when we're exporting using ordinals, in which case it might be helpful to maintain compatibility with old DLLs. But considering that the library would be tied to the finished product, and also the fact that we don't provide prebuilt binaries, the chance of a DLL swap in a finished product is slim. Besides, we're exporting by names anyway, so it doesn't really matter. I haven't seen a case when `__declspec(dllexport)` actually fails.

longnguyen2004:
Regarding export attributes: Should we use it or not? It has been well tested by many other libraries at this point, and it would save us from having to maintain multiple def files, which will reduce development costs in the future.

alexey-katranov:
> Regarding export attributes: Should we use it or not? It has been well tested by many other libraries at this point, and it would save us from having to maintain multiple def files, which will reduce development costs in the future.

We cannot made the decision right now because of public holidays in Russia (till January 10). If you want to move forward, the manually created `*.def` files is an acceptable approach.

longnguyen2004:
Then it's better to leave this PR here and wait until the decision is made, than to implement a temporary solution only to replace it with a better one. It's not a very high priority right now, but it's nice to have as much compatibility as possible.

longnguyen2004:
How's the status on this? Will the team switch to using export directives or not?

alexey-katranov:
> How's the status on this? Will the team switch to using export directives or not?

We are ready to try. 

The main concern is about backward compatibility: we do not want to export more than minimally required. The def file approach forces us to carefully add exported functions while the new approach is not so strict. Do we have guarantee that only marked with export directives functions will be exported? Can we compare the current list of exported symbols with the new approach?

longnguyen2004:
> Can we compare the current list of exported symbols with the new approach?

You can inspect the compiled object file using `dumpbin` for MSVC, or `nm` for GCC

> The def file approach forces us to carefully add exported functions while the new approach is not so strict

Using export directives, you can achieve the same granularity with def files, by only adding the export directive to the required functions, or you can export the entire class, in which case all public and private symbols will be exported, which might be needed if you have inline functions calling private methods. Either way, it's better than having to maintain multiple def files.

cmake/compilers/GNU.cmake
alexey-katranov: It is required only for the library build.

cmake/compilers/GNU.cmake
alexey-katranov: `bigobj` is Windows specific. We already use it for the MSVC compiler, so I do not see any issues to use it for gcc as well.

longnguyen2004:
Here's a minimal export header example:
```
#if TBB_BUILDING
    #if _WIN32
        #define TBB_EXPORT __declspec(dllexport)
    #elif __APPLE__ || __linux__
        #define TBB_EXPORT __attribute__(visibility("default"))
    #endif
#else
    #define TBB_EXPORT
#endif
```

Include this file for each public header, then add `TBB_EXPORT` to the functions and classes to be exported:
`TBB_EXPORT int Foo();` and `class TBB_EXPORT Foo;`

We also have to set `DEFINE_SYMBOL` to `TBB_BUILDING` for each CMake target, so the compiler will define it when we're building TBB.

alexey-katranov:
Let's decide on the next steps. I do not think we proceed with the replacement of def files quickly; however, I would like to move forward with mingw support. Does it make to move forward with the current def files (without automatic generation) or will you able to switch to the exports approach at least for mingw?

longnguyen2004:
Either one is fine. If we want to prepare for the change to export directives, we can create an export header that only defines the macro on mingw:
```
#if TBB_BUILDING
    #if __MINGW32__
        #define TBB_EXPORT __declspec(dllexport)
    #else
        #define TBB_EXPORT // nothing, def files will still work
    #endif
#else
    #define TBB_EXPORT
#endif
```

By doing so, we can prepare the codebase for other platforms, then we can add support for other platforms by modifying the export header.

longnguyen2004:
If we're good with that approach, I'll open a new PR with the changes and close this one.

longnguyen2004:
Closing in favor of #351

PR #326, opened 2020-12-28
Allow to disable tbbmalloc, use posix_memalign (oneTBB)

Lastique:
This is an updated version of https://github.com/oneapi-src/oneTBB/pull/149 which targets the reworked code base of oneTBB 2021.1.1.

The first commit enables `posix_memalign` and `memalign` on UNIX-like systems. This API is universally supported nowdays, so there's no need to use the hand-rolled implementation on top of `malloc`.

The second one adds `TBB_USE_MALLOC` environment variable to force using system allocator.

By setting `TBB_USE_MALLOC` environment variable to 1 the user can disable scalable allocator from tbbmalloc and use system allocator through `malloc`/`posix_memalign`/`free` API. There are other, more performant memory allocators, like tcmalloc for example, which the user might want to use instead of tbbmalloc. Also, using the system allocator can be useful with memory debugging and profiling tools, like valgrind. Such tools usually provide their own implementations of memory allocation functions, and we want those functions to be used by TBB.

The original PR https://github.com/oneapi-src/oneTBB/pull/149 contains some benchmarks and rationale as to why we prefer an environment variable as a way to disable tbbmalloc.

### DIFF
--- a/src/tbb/allocator.cpp
+++ b/src/tbb/allocator.cpp
@@ -31,6 +31,20 @@
 #include <dlfcn.h>
 #endif /* _WIN32||_WIN64 */
 
+#if (!_WIN32 && !_WIN64) || defined(__CYGWIN__)
+#include <stdlib.h> // posix_memalign, free
+// With glibc it is safe to use memalign(), as the allocated memory can be freed with free()
+#if defined(__GLIBC__)
+#include <malloc.h> // memalign
+#define __TBB_USE_MEMALIGN
+#else
+#define __TBB_USE_POSIX_MEMALIGN
+#endif
+#elif defined(_MSC_VER) || defined(__MINGW32__)
+#include <malloc.h> // _aligned_malloc, _aligned_free
+#define __TBB_USE_MSVC_ALIGNED_MALLOC
+#endif
+
 #if __TBB_WEAK_SYMBOLS_PRESENT
 
 #pragma weak scalable_malloc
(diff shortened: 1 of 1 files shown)


Lastique:
The patch to enable support for `memalign`, `posix_memalign`, etc. has been extracted into a separate PR https://github.com/oneapi-src/oneTBB/pull/671.

src/tbb/allocator.cpp
```diff
@@ -115,7 +116,9 @@ static const dynamic_link_descriptor MallocLinkTable[] = {
     If that allocator is not found, it links to malloc and free. */
 void initialize_handler_pointers() {
     __TBB_ASSERT(allocate_handler == &initialize_allocate_handler, NULL);
-    bool success = dynamic_link(MALLOCLIB_NAME, MallocLinkTable, 4);
+    bool success = false;
+    if (!GetBoolEnvironmentVariable("TBB_USE_MALLOC"))
+        success = dynamic_link(MALLOCLIB_NAME, MallocLinkTable, 4);
```
alexey-katranov, line 121: Have you considered an approach to extend `global_control` functionality? Usually, we try to avoid configuration via environment variables.

Lastique: Yes, the rationale and discussion about `global_control` is in the previous version of this PR: https://github.com/oneapi-src/oneTBB/pull/149. The short version is that we want to ensure that tbbmalloc is not enabled if TBB is used somewhere during global initialization of the program, and we want to propagate that setting to child processes.

PR #327, opened 2020-12-28
Use nanosleep on Linux in atomic_backoff when spinning for a long time (oneTBB)

Lastique:
This is an updated version of https://github.com/oneapi-src/oneTBB/pull/106, which is rebased on top of oneTBB 2021.1.1 code base.

This allows to significantly reduce CPU system time while spinning, as `nanosleep` deschedules the thread and allows other threads to progress. Unlike `sched_yield`, `nanosleep` allows the thread to migrate to other CPUs, as well as other threads to migrate to the current CPU, which is what allows for more total progress.

See https://github.com/oneapi-src/oneTBB/issues/105. On the test presented
there, this patch offers to reduce CPU system time by 25-30%. On a real world
application that test is based on this patch offers ~10% improvement.

### DIFF
--- a/include/oneapi/tbb/detail/_machine.h
+++ b/include/oneapi/tbb/detail/_machine.h
@@ -20,6 +20,7 @@
 #include "_config.h"
 #include "_assert.h"
 
+#include <time.h>
 #include <atomic>
 #include <climits>
 #include <cstdint>
@@ -58,6 +59,15 @@ static inline void yield() {
 #else
 using std::this_thread::yield;
 #endif
+
+static inline void long_yield() {
+#if defined(__linux__)
+    struct ::timespec ts = {};
+    ::nanosleep(&ts, 0);
+#else
+    yield();
+#endif
+}
 
 //--------------------------------------------------------------------------------------------------
 // atomic_fence implementation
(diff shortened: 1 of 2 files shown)


pavelkumbrasev:
Now we are trying to use "[mutex](https://spec.oneapi.io/versions/latest/elements/oneTBB/source/mutual_exclusion/mutex_cls.html)" and "[rw_mutex](https://spec.oneapi.io/versions/latest/elements/oneTBB/source/mutual_exclusion/rw_mutex_cls.html)" in cases where high contention might cause slowdowns. They are implemented using an adaptive approach, it guarantees that the thread that cannot acquire the lock spins before blocking(for example on cond var).  You can check current [wait ](https://github.com/oneapi-src/oneTBB/blob/09346731f1279feb23c81f569eabe6010c620872/include/oneapi/tbb/detail/_waitable_atomic.h#L67) implementation in waitable_atomic. If you still observe slowdowns on atomic_backoff waits please let as know.

Lastique:
We're still using it in our production builds of TBB. Though we haven't tried the latest developments in the TBB code base as there haven't been a release yet.

PR #328, opened 2020-12-28
Optimize x86 atomic_fence

Lastique:
The first commit provides an optimized `atomic_fence` implementation for x86 on gcc-compatible compilers.

On x86 (32 and 64-bit) any lock-prefixed instruction provides sequential consistency guarantees for atomic operations and is more efficient than mfence. You can see some tests in [this](https://shipilev.net/blog/2014/on-the-fence-with-dependencies/) article.
    
We are choosing a "lock not" on a dummy byte on the stack for the following reasons:

- The "not" instruction does not affect flags or clobber any registers. The memory operand is presumably accessible through esp/rsp.
- The dummy byte variable is at the top of the stack, which is likely hot in cache.
- The dummy variable does not alias any other data on the stack, which means the "lock not" instruction won't introduce any false data dependencies with prior or following instructions.

In order to avoid various sanitizers and valgrind complaining, we have to initialize the dummy variable to zero prior to the operation.
    
Additionally, for memory orders weaker than seq_cst there is no need for any special instructions, and we only need a compiler fence. For the relaxed memory order we don't need even that.

The second commit removes explicit mfence on Windows. The existing `std::atomic_thread_fence` already provides the necessary instructions to maintain memory order according to its argument.

### DIFF
--- a/include/oneapi/tbb/detail/_machine.h
+++ b/include/oneapi/tbb/detail/_machine.h
@@ -76,22 +76,15 @@ using std::this_thread::yield;
 #endif
 
 //--------------------------------------------------------------------------------------------------
-// atomic_fence implementation
+// atomic_fence_seq_cst implementation
 //--------------------------------------------------------------------------------------------------
 
-static inline void atomic_fence(std::memory_order order) {
+static inline void atomic_fence_seq_cst() {
 #if (__TBB_x86_64 || __TBB_x86_32) && defined(__GNUC__) && __GNUC__ < 11
-    if (order == std::memory_order_seq_cst)
-    {
-        unsigned char dummy = 0u;
-        __asm__ __volatile__ ("lock; notb %0" : "+m" (dummy) :: "memory");
-    }
-    else if (order != std::memory_order_relaxed)
-    {
-        __asm__ __volatile__ ("" ::: "memory");
-    }
+    unsigned char dummy = 0u;
+    __asm__ __volatile__ ("lock; notb %0" : "+m" (dummy) :: "memory");
 #else
-    std::atomic_thread_fence(order);
+    std::atomic_thread_fence(std::memory_order_seq_cst);
 #endif
 }
 
--- a/include/oneapi/tbb/detail/_machine.h
+++ b/include/oneapi/tbb/detail/_machine.h
@@ -84,6 +84,17 @@ using std::this_thread::yield;
 #endif
 
 static inline void atomic_fence(std::memory_order order) {
+#if (__TBB_x86_64 || __TBB_x86_32) && defined(__GNUC__) && __GNUC__ < 11
+    if (order == std::memory_order_seq_cst)
+    {
+        unsigned char dummy = 0u;
+        __asm__ __volatile__ ("lock; notb %0" : "+m" (dummy) :: "memory");
+    }
+    else if (order != std::memory_order_relaxed)
+    {
+        __asm__ __volatile__ ("" ::: "memory");
+    }
+#else
 #if _MSC_VER && (__TBB_x86_64 || __TBB_x86_32)
     if (order == std::memory_order_seq_cst ||
         order == std::memory_order_acq_rel ||
@@ -95,6 +106,7 @@ static inline void atomic_fence(std::memory_order order) {
     }
 #endif /*_MSC_VER && (__TBB_x86_64 || __TBB_x86_32)*/
     std::atomic_thread_fence(order);
+#endif
 }
 
 //--------------------------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


include/oneapi/tbb/detail/_machine.h
```diff
@@ -79,22 +79,20 @@ using std::this_thread::yield;
 // atomic_fence implementation
 //--------------------------------------------------------------------------------------------------
 
-#if _MSC_VER && (__TBB_x86_64 || __TBB_x86_32)
-#pragma intrinsic(_mm_mfence)
-#endif
-
 static inline void atomic_fence(std::memory_order order) {
-#if _MSC_VER && (__TBB_x86_64 || __TBB_x86_32)
```
alexey-katranov, line 87: Microsoft Compiler used to generate [strange code](https://godbolt.org/z/q9zqn3MjG) that allocates two registers and slowdowns nearby code. For our use cases `mfence` is not big deal, rather than broken optimization due to inefficient register allocation. We still use old Microsoft Compiler for the releases and I would remain the workaround as is. 

GCC 11 generates really good code: https://godbolt.org/z/1eobjq5PT, so the proposed patch is going to be outdated.  So, it seems unnecessary for modern GCC and for official releases (we use Intel Compiler on Linux that does not bother itself and generates `mfence`).

If you think that the patch is useful for your use case, I would refactor out `order` argument because we use only `memory_order_seq_cst` and limit the workaround to GCC version prior to 11.

Lastique: > Microsoft Compiler used to generate strange code that allocates two registers and slowdowns nearby code.

The godbolt output you linked shows that MSVC does basically the same - replaces `mfence` with `lock cmpxchg`. Yes, `cmpxchg` is not the best replacement, but I would still expect it to be far better than `mfence`, even with two clobbered registers. See [here](https://uops.info/table.html?search=lock%20cmpxchg&cb_lat=on&cb_tp=on&cb_uops=on&cb_ports=on&cb_SKL=on&cb_ICL=on&cb_ZEN3=on&cb_measurements=on&cb_doc=on&cb_base=on&cb_sse=on) and [here](https://uops.info/table.html?search=mfence&cb_lat=on&cb_tp=on&cb_uops=on&cb_ports=on&cb_SKL=on&cb_ICL=on&cb_ZEN3=on&cb_measurements=on&cb_doc=on&cb_base=on&cb_sse=on) for performance numbers. The bottom line is `mfence` is very slow (about half the throughput), after all security mitigations, so replacing it with a `lock`-prefixed operation is a win.

I can leave the `mfence` in place if you insist, I just don't see the point.

> GCC 11 generates really good code, so the proposed patch is going to be outdated.

Yes, gcc 11 have optimized the thread fence intrinsics. It uses `lock or [rsp], 0` as a replacement though, which might be less efficient than the asm statement I proposed because:

- `or` instruction affects flags,
- `[rsp]` is not necessarily the location of the latest accessed data on the stack, as the compiler tends to allocate the stack frame beforehand. This means that `[rsp]` is less likely to be in the cache than a variable that is inside the stack frame, near the latest allocated variable in the preceding code.

I cannot say how much these nuances affect performance, probably not by much. I can limit the asm code to gcc versions prior to 11 if you prefer.

> If you think that the patch is useful for your use case, I would refactor out `order` argument

Sure, I can do that, though I'd prefer to remove `order` in a separate commit.

alexey-katranov: I see the point. I checked the code, we already eliminated `atomic_fence` from the hottest paths, so even inefficient register allocation does not seem to be the issue. Hence, let's go with your solution because it is indifferent for Intel-based CPU.

> I can limit the asm code to gcc versions prior to 11 if you prefer, though I see no harm in leaving it enabled unconditionally. 

The main harm is two versions of the code that can become different over the time and require multiple configurations to test them. Moreover, the assembler limits compiler capabilities to optimize nearby code and generate better code for future CPUs (what if someone invents one more good instruction?). In addition, memory model is really complex topic and it is better to have implementation as simple as possible to avoid unforeseen issues and apply such tools like Thread Sanitizer to check the correctness but Thread Sanitizer does not understand assembler. Unfortunately, it does not understand standalone barriers but there are some attempts (mentioned in https://github.com/google/sanitizers/issues/1415) to add them. Therefore, I prefer to keep workarounds limited and rely on generic coded as well tested and reasonably performant solutions.

> Sure, I can do that, though I'd prefer to remove order in a separate commit.

Up to you, so you can consider not to create branches for other than `memory_order_seq_cst` memory order.

Lastique: Done.

alexey-katranov (approved):

PR #330, opened 2021-01-05
Add support for Bazel

Vertexwahn:
This PR helps users of the [Bazel](https://bazel.build/) build system to integrate oneTBB in their builds.

Supporting two build systems (CMake and Bazel) is of course more effort to maintain but helps to spread oneTBB.
In the worst case the Bazel build does simply not work - in the best case someone who uses Bazel can benefit from it.
The Bazel build files will not influence the CMake build, and can coexist without any side effects. 

Also, other projects support CMake and Bazel side by side (e.g. [Catch2](https://github.com/catchorg/Catch2), [GoogleTest](https://github.com/google/googletest), [OpenEXR](https://github.com/AcademySoftwareFoundation/openexr) etc.)

I am willing to support the Bazel build of this lib for at least for a half year.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,9 @@ Here are [Release Notes]( https://software.intel.com/en-us/articles/intel-oneapi
 [System Requirements](https://software.intel.com/en-us/articles/intel-oneapi-threading-building-blocks-system-requirements).
 
 ## Documentation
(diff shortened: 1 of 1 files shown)


froody:
I need this!

AlexVeprev:
Discussion is here: https://github.com/oneapi-src/oneTBB/pull/442

PR #335, opened 2021-01-12
Fix TBB vars.sh for zsh in Linux

bieito98:
Solution for [this issue](https://community.intel.com/t5/Intel-oneAPI-HPC-Toolkit/setvars-sh-failing-on-macOS-Catalina-using-zsh-works-for-bash/m-p/1199001), marked as Solved, but the solution proposed in the post is just a workaround, and initiates other subshell.

Here, we replicate the Bash line for Zsh, but using "${(%):-%x}" instead of $BASH_SOURCE (more info: [https://stackoverflow.com/a/28336473](https://stackoverflow.com/a/28336473))

I think that the same fix could be applied to the Mac version, but I cannot test it.

### DIFF
--- a/integration/linux/env/vars.sh
+++ b/integration/linux/env/vars.sh
@@ -19,10 +19,7 @@
 #   intel64|ia32 - architecture, intel64 is default.
 
 if [ -n "$ZSH_EVAL_CONTEXT" ]; then
-    case "$0" in
-    (/*) TBBROOT="$(dirname $0)"/.. ;;
-    (*) TBBROOT="$(dirname "$(command pwd)/$0")"/.. ;;
-    esac
+    TBBROOT="$(cd "$(dirname "${(%):-%x}")" && pwd -P)"/..
 elif [ -n "$KSH_VERSION" ]; then
     TBBROOT="${.sh.file%/*}/.."
 else


AlexVeprev:
Hi @bieito98,
Thank you for the PR!
`master` branch contains significantly changed `vars.sh` with the updated Zsh handling: https://github.com/oneapi-src/oneTBB/blob/master/integration/linux/env/vars.sh#L108. Do you think it is ok now?

bieito98:
Hi!
Yes, I think the new `vars.sh` solves the issue. I have tested it in my computer (ZSH 5.8 on Linux), and it is working well now. Thanks!

With the old file:
```shell
bieito@pc ~ % zsh -l
:: initializing oneAPI environment ...
   ZSH version = 5.8
:: dpl -- latest
:: mpi -- latest
:: advisor -- latest
:: dev-utilities -- latest
:: itac -- latest
:: compiler -- latest
:: dpcpp-ct -- latest
:: inspector -- latest
:: tbb -- latest
ERROR: libtbb.so.12 library does not exist in /home/bieito/../lib/intel64/gcc4.8.
:: vtune -- latest
:: oneAPI environment initialized ::
```

With the file from `master`:
```shell
bieito@pc ~ % zsh -l
:: initializing oneAPI environment ...
   ZSH version = 5.8
:: dpl -- latest
:: mpi -- latest
:: advisor -- latest
:: dev-utilities -- latest
:: itac -- latest
:: compiler -- latest
:: dpcpp-ct -- latest
:: inspector -- latest
:: tbb -- latest
:: vtune -- latest
:: oneAPI environment initialized ::
```

AlexVeprev:
@bieito98 thank you for the confirmation! I'm closing the PR then.

PR #339, opened 2021-01-22
Update 12-b-HandsOn-oneDPL.ipynb

reble:
Reorder includes. Workaround to support GCC9, GCC10


anton-potapov:
@reble is this still relevant?

reble:
Yes, the code won't compile with current oneAPI version. Examples as-is are targeting beta API. Patch is recommended.

PR #350, opened 2021-02-22
Jsbache/asan fix

jsbache:
Modifications allowing TBB to be run with address sanitizer when compiling for Apple Silicon.
TBB's generic implementation relies on 4 byte compare&swap. For these methods we need to disable ASAN  because they are also called with 1 byte and 2 byte types. The four byte read is safe as TBB forces 4 byte alignment of related values.


anton-potapov:
@jsbache Could you please update the patch with latest master (there were several ASAN related fixes) ?

jsbache:
I think the original PR was against the 2020 branch. The current PR shows 10176 changed files and is not what I want to propose. I am closing this PR.

PR #354, opened 2021-02-27
Use gcc_generics on Apple mSilicon

jsbache:
clang supports gcc_generics on Apple Silicon and we should use this backend to get an efficient implementation of CAS and other low level atomic primitives


anton-potapov:
@jsbache, thanks for the contribution! However, oneTBB has moved to c++11 `std::atomic` so this patch no longer applies.

Issue #267, opened 2020-08-17
Feature Request: parallel_call_once/parallel_once_flag


BenFrantzDale:
From https://software.intel.com/en-us/node/841360 (`isolated_task_group`) there's an example `class parallel_once` which acts a bit like `std::call_once` https://en.cppreference.com/w/cpp/thread/call_once, but rather than blocking, follow-on threads can help get the work done. Prior to finding that example code, I was thinking about the same idea, but I think the better way to do it would be to follow the standard's API and provide a `tbb::once_flag` and `tbb::parallel_call_once` (or maybe just `tbb::call_once`?)

The example on https://software.intel.com/en-us/node/841360 seems very close; it just lacks proper exception handling, since `std::call_once` doesn't count exceptions as successful calls. In a correct parallel version, the initial calling thread would throw, but then a waiting (helping) thread would get to try calling its function while any other waiting threads would re-try helping until all calls either throw or one succeeds.

The other change I'd consider vs the example is to potentially have the `tbb::isolated_task_group` be an `std::optional< tbb::isolated_task_group>` (or have it heap-allocated) so the `tbb::isolated_task_group` only has to exist while in use. (I'm not clear on the cost in memory and/or other resources of keeping potentially thousands of `tbb::isolated_task_group` instances around. For reference, for at least one implementation, `sizeof(std::once_flag) == 4` https://godbolt.org/z/WhcesE)

If this is of interest, I could help implement it.

alexey-katranov:
The idea seems interesting.

As for isolated_task_group, the main problem how to delete it (the complete flag can be set earlier than all thread leaves the `task_group`). I have a complicated idea with 4 state for once_flag and additional reference count (barrier) for helping threads. I will share the details later if there is no better solution. 

Another issue is user threads, each user thread works in its own `arena`, so it is not so easy to join a user thread to help. Perhaps, attached `task_arena` can be used (but it is also complicated solution how to transfer this information to the helping threads).

So, if you are not afraid of possible complexities, feel free to open full request with your contribution (use onetbb_2021 branch for the baseline). And/or we can discuss in details how to overcome the issues.

BenFrantzDale:
Here's what I've got: https://godbolt.org/z/qTdET3

alexey-katranov:
[...]

PR #357, opened 2021-03-04
Add Collaborative call once class

HenryHeffanFormlabs:
Hi Alexei,
Here is a version of the collaborative call once class that @BenFrantzDale and I believe work. Im not sure if there is more I need to do besides add the files. Please let me know!
Best,
Henry


include/oneapi/tbb/collaborative_once_flag.h
```diff
@@ -61,7 +61,7 @@ class run_once {
                 cancellation_status = my_context.is_group_execution_cancelled();
             });
         };
-        my_arena.execute([&f]() { return tbb::this_task_arena::isolate(f); });
+        my_arena.execute(f);
```
HenryHeffanFormlabs, line 64: This change is needed to make the `collaborative_once_flag` actually run in parallel. I think it is correct because the first thread gets trapped inside the call once, so it cant deadlock, but then the other threads are free to do what they want. Does this seem right to you?

include/oneapi/tbb/collaborative_once_flag.h
```diff
@@ -0,0 +1,126 @@
...
+                    throw;
+                }
+                // Let go, mark done, and return:
+                std::atomic_store_explicit(&my_runner, std::shared_ptr<detail::run_once>(nullptr),
+                                           std::memory_order_release);
+                my_state.store(state::done, std::memory_order_release);
+                return;
+            } else {
+                // Get our own pointer so it can't go away while we aren't looking.
+                if (auto runner = std::atomic_load_explicit(&my_runner, std::memory_order_acquire)) {
```
alexey-katranov, line 111: it seems that we are busy waiting for runner publication or for `my_state` completion that requires some sort of backoff.

include/oneapi/tbb/collaborative_once_flag.h
```diff
@@ -0,0 +1,126 @@
...
+               state::done) { // Loop until someone else finishes or it's Done.
+            state expected = state::uninitialized;
+            if (my_state.compare_exchange_weak(expected, state::in_progress)) {
+                // We won the race to be the primary thread.
+                auto runner = std::make_shared<detail::run_once>();
+                std::atomic_store_explicit(&my_runner, runner, std::memory_order_release);
+                try {
+                    tbb::task_group_status wait_result = runner->run_and_wait(f);
+                    if (wait_result == tbb::task_group_status::canceled) {
+                        throw canceled_exception(); // Translate cancelation to an exception.
```
alexey-katranov, line 95: Usually, exceptions are not recommended way for expected program flow while cancellation can be a part of normal program flow (e.g. looking for a element in container). I would propose to pass the context as an argument to `collaborative_call_once`.

include/oneapi/tbb/collaborative_once_flag.h
```diff
@@ -0,0 +1,126 @@
...
+                    }
+                } catch (...) {
```
alexey-katranov, line 97: consider using `on_exception`

include/oneapi/tbb/collaborative_once_flag.h
```diff
@@ -0,0 +1,126 @@
...
+    /*! A call to wait() potentially participates in the work of run_and_wait().
+     * If run_and_wait() throws or is canceled, this returns canceled.
+     */
+    task_group_status wait() {
+        bool cancellation_status = false;
+        auto f = [&]() {
+            detail::try_call([&] {
+                detail::d1::wait(my_wait_context, my_context);
+            }).on_completion([&] {
+                cancellation_status = my_context.is_group_execution_cancelled();
```
alexey-katranov, line 61: It seems we do not need all cancellation logic in `wait`. Additionally, consider the comment about `exception` vs `cancellation`.

include/oneapi/tbb/collaborative_once_flag.h
```diff
@@ -0,0 +1,126 @@
...
+
+} // namespace detail
+
+class canceled_exception : public std::exception {};
+
+// The concurrent_once_flag idea may be added to tbb at some point: oneapi-src/oneTBB#267
+class collaborative_once_flag : detail::d0::no_copy {
+    enum class state { uninitialized, in_progress, done };
+    std::atomic<state> my_state = state::uninitialized;
+    std::shared_ptr<detail::run_once> my_runner = nullptr;
```
alexey-katranov, line 78: We will think about shrinking the state into one pointer as discussed in #267

alexey-katranov (commented):
@HenryHeffanFormlabs , we are Ok with the patch and we will proceed it internally for further integration into the master. The comments are to inform you what aspect we are going to change/improve. If any comment/concerns, feel free to share.

Thank you for the patch.

include/oneapi/tbb/collaborative_once_flag.h
HenryHeffanFormlabs: That seems reasonable! What would that look like in this case?

```
class LazyEvaluated {
    tbb::collaborative_once_flag;
    double cachedPropoerty;
public:
    double getProperty() {
        tbb::collaberative_call_once([]() {
             ... work ...
             cachedPropoerty = result;
        });
        return cachedProperty;
    }
};
```

`getProperty()` could conceivably could be called from different contexts at the same time? Would that be ok?

alexey-katranov: `collaberative_call_once` is synchronization point that should provide an _acquire_ semantics. So, `cachedProperty` can be read concurrently without introducing a data race. In addition, the primary thread (that proceed the function until the end) should provide _release_ semantics.

HenryHeffanFormlabs: Hi Alexey, I was just wondering if the proper usage would be to pass the context into the getProperty() call, for it to be created on the stack in `getProperty()`, or if the LazyEvaluation class would have a context as a member?

include/oneapi/tbb/collaborative_once_flag.h
alexey-katranov: Surely, it should not be a new isolated region as it was before. Should be it the same isolated region that created by the primary thread or no isolation is Ok here? There are pros and cons for the both options, so let's remain as is while we do not have strong reason for either of the options.

include/oneapi/tbb/collaborative_once_flag.h
alexey-katranov: I think it should be similar to other parallel algorithms when the context is passed as the last optional argument. If the user needs something from the context, they can specify it as the last argument to `tbb::collaberative_call_once`. If not specified, a default context will be created on the stack. In addition, user can capture the explicit context into the lambda (e.g. for cancellation purposes).

HenryHeffanFormlabs: sounds good!

include/oneapi/tbb/collaborative_once_flag.h
BenFrantzDale: Yes. I really like the idea of shared storage on the stack of the first caller to get there. I see that as a potential follow-on performance improvement here, where we'd potentially replace the `std::shared_ptr<detail::run_once>` with a `tbb::stack_shared_temporary<detail::run_once>` (or however we'd spell it).

(For reference, on clang, `sizeof(std::once_flag) == 4)` and `sizeof(std::mutex) == 40)`, but with `std::once_flag`, the threads don't need to "find each other", they just need to know if they should block.)

include/oneapi/tbb/collaborative_once_flag.h
BenFrantzDale: @aleksei-fedotov in the case that a follow-on collaborator is called with a context and the first-to-arrive doesn't bail out prematurely, the behavior would be that the follow-on collaborator ignores its context, correct?

BenFrantzDale: > Usually, exceptions are not recommended way for expected program flow while cancellation can be a part of normal program flow (e.g. looking for a element in container). I would propose to pass the context as an argument to `collaborative_call_once`.

This is a good question. I've been bitten hard by cancelation causing me to think I've succeeded leading to UB (https://github.com/oneapi-src/oneTBB/issues/253). That said, the Committee has decided that `std::stop_token` shouldn't use exceptions for cancelation (early versions of `std::stop_token` had a `throw_if_canceled` member function). OTOH, `std::call_once` either succeeds or throws; there's no other way out. We could have it be
```
template <typename Fn>
[[nodiscard]] task_group_status collaborative_call_once(collaborative_once_flag& flag, Fn&& fn);
```
Then if client code wants it to throw on cancelation, we can easily wrap it. But if we do that, please please make it `[[nodiscard]]`!

include/oneapi/tbb/collaborative_once_flag.h
BenFrantzDale: `tbb::internal::atomic_backoff`? That seems too useful to hide in `internal`.

include/oneapi/tbb/collaborative_once_flag.h
alexey-katranov: > @aleksei-fedotov in the case that a follow-on collaborator is called with a context and the first-to-arrive doesn't bail out prematurely, the behavior would be that the follow-on collaborator ignores its context, correct?

The context of first-to-arrive call is expected to be used for all collaborators. Overall, it seems strange if different contexts are used (or some threads do not specify the context). (P.S. I believe you wanted to notify @alexey-katranov).

As for exception vs cancellation, consider the following serial example:
```
try {
    bool zero_element_found = false;
    for (auto v : container) {
        if (!v.ptr) {
            throw bad_element;                         // exception on error
        }
        if (*v.ptr == 0) {
            zero_element_found = true;
            break;                                     // cancelation on normal program flow condition
        }
    }
    if (zero_element_found) { ... }
} catch ( ... ) {
    // report error
}
```
The main issue that neither exception nor cancelation will not work as expected in a threaded application because the both cancelation and exception are local to the executing thread. While the `break` condition is semi working because threads that met the condition will save some cycles, the exception condition can be dangerous because the exception will be thrown on some other thread that might miss sending a signal about the completion to the main thread.

oneTBB provides mechanisms to handle the both situations. The exception condition is handled automatically that any exception is captured and rethrown on the waiting thread, so it works like the exception occurs on the waiting thread. To improve the performance, the exception propagation mechanism triggers the cancelation mechanism that sets the flag into `tbb::task_group_context` that any not currently running task associated with this context should be skipped. However, we do not need to work with context (only to request the cancelation), i.e. the example can be transformed to parallel execution in the following way:
```
try {
    std::atomic<bool> zero_element_found = false;
    parallel_for (auto v : container) {
        if (!v.ptr) {
            throw bad_element;                         // exception on error
        }
        if (*v.ptr == 0) {
            zero_element_found = true;
            tbb::task::current_context()->cancel_group_execution();  // cancelation on normal program flow condition
         }
    }
    if (zero_element_found) { ... }
} catch ( ... ) {
    // report error
}
```
So, in some cases, you do not need to pass `task_group_context` explicitly to a functor or parallel algorithm.

The same idea can be applied to `collaborative_call_once` that any exception is processed like in a serial code and cancelation can be explicitly requested if necessary.

As `std::call_once` it has serial semantics and cancelation is not really required (user can use their own flags if necessary).

BenFrantzDale:
@akukanov have you folks had a chance to experiment with this? Let me know if you have any questions.

alexey-katranov:
We are in process of integration. Everything is on track.

alexey-katranov:
#397 is merged

PR #276, opened 2020-09-10
Uses __copy__ in proxy.hpp for aliases when available

SteveBronder:
I was compiling oneTBB 2020 v2 with gcc > 9 and getting warnings like the following

```bash
../../src/tbbmalloc/proxy.cpp:215:7: warning: ‘void* __libc_calloc(size_t, size_t)’ specifies less restrictive attributes than its target ‘void* calloc(size_t, size_t)’: ‘alloc_size’, ‘leaf’, ‘malloc’, ‘nothrow’ [-Wmissing-attributes]
  215 | void *__libc_calloc(size_t num, size_t size) __attribute__ ((alias ("calloc")));
      |       ^~~~~~~~~~~~~
../../src/tbbmalloc/proxy.cpp:123:14: note: ‘void* __libc_calloc(size_t, size_t)’ target declared here
  123 | void *PREFIX(calloc)(ZONE_ARG size_t num, size_t size) __THROW
      |              ^~~~~~
../../src/tbbmalloc/proxy.cpp:84:22: note: in definition of macro ‘PREFIX’
   84 | #define PREFIX(name) name
```

I saw you had a pragma to ignore those, but I think you can just use `__copy__` instead

This does the same thing as [#1637](https://github.com/jemalloc/jemalloc/pull/1637/files#diff-d2be2656d37670b0c7f5f4320c0670ce) in jemalloc with a little check to make sure the copy attribute exists.

PR #359, opened 2021-03-05
[fix] Uses __copy__ in proxy.hpp for aliases when complier has it

SteveBronder:
@alexey-katranov This is just #276 but updated to master. I let that PR linger for too long and the rebase was v difficult for such a small change so I just opened up a new PR.

I was compiling oneTBB 2020 v2 with gcc > 9 and getting warnings like the following

```bash
../../src/tbbmalloc/proxy.cpp:215:7: warning: ‘void* __libc_calloc(size_t, size_t)’ specifies less restrictive attributes than its target ‘void* calloc(size_t, size_t)’: ‘alloc_size’, ‘leaf’, ‘malloc’, ‘nothrow’ [-Wmissing-attributes]
  215 | void *__libc_calloc(size_t num, size_t size) __attribute__ ((alias ("calloc")));
      |       ^~~~~~~~~~~~~
../../src/tbbmalloc/proxy.cpp:123:14: note: ‘void* __libc_calloc(size_t, size_t)’ target declared here
  123 | void *PREFIX(calloc)(ZONE_ARG size_t num, size_t size) __THROW
      |              ^~~~~~
../../src/tbbmalloc/proxy.cpp:84:22: note: in definition of macro ‘PREFIX’
   84 | #define PREFIX(name) name
```

I saw you had a pragma to ignore those, but I think you can just use `__copy__` instead

This does the same thing as [#1637](https://github.com/jemalloc/jemalloc/pull/1637/files#diff-d2be2656d37670b0c7f5f4320c0670ce) in jemalloc with a little check to make sure the copy attribute exists.

### DIFF
--- a/src/tbbmalloc_proxy/proxy.cpp
+++ b/src/tbbmalloc_proxy/proxy.cpp
@@ -59,6 +59,13 @@
 // Use MallocMutex implementation
 typedef MallocMutex ProxyMutex;
 
+// Adds aliasing and copy attributes to function if available
+#if __has_attribute(__copy__)
+  #define __TBB_ALIAS_ATTR_COPY(name) __attribute__((alias (#name), __copy__(name)))
+#else
+  #define __TBB_ALIAS_ATTR_COPY(name) __attribute__((alias (#name)))
+#endif
+
 // In case there is no std::get_new_handler function
 // which provides synchronized access to std::new_handler
 #if !__TBB_CPP11_GET_NEW_HANDLER_PRESENT
@@ -123,10 +130,7 @@ static inline void initPageSize()
     #pragma GCC diagnostic push
     #pragma GCC diagnostic ignored "-Wmissing-attributes"
 #endif
-extern "C" void *__TBB_malloc_proxy(size_t) __attribute__ ((alias ("malloc")));
-#if __GNUC__ == 9
-    #pragma GCC diagnostic pop
-#endif
+extern "C" void *__TBB_malloc_proxy(size_t) __TBB_ALIAS_ATTR_COPY(malloc);
 
 static void *orig_msize;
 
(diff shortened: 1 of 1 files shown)


SteveBronder:
Ping

alexey-katranov:
Thank you for the patch. I will let you know if any issues with integration.

src/tbbmalloc_proxy/proxy.cpp
```diff
@@ -59,6 +59,13 @@
 // Use MallocMutex implementation
 typedef MallocMutex ProxyMutex;
 
+// Adds aliasing and copy attributes to function if available
+#if __has_attribute(__copy__)
```
alexey-katranov, line 63: Not all compilers recognize `__has_attribute`, i will wrap it into `#if defined __has_attribute/#endif` (as recommend in https://gcc.gnu.org/onlinedocs/cpp/_005f_005fhas_005fattribute.html)

src/tbbmalloc_proxy/proxy.cpp
```diff
@@ -265,28 +269,23 @@ struct mallinfo mallinfo() __THROW
...
-    #pragma GCC diagnostic pop
-#endif
+void *__libc_malloc(size_t size) __TBB_ALIAS_ATTR_COPY(malloc);
+void *__libc_calloc(size_t num, size_t size) __TBB_ALIAS_ATTR_COPY(calloc);
+void *__libc_memalign(size_t alignment, size_t size) __TBB_ALIAS_ATTR_COPY(memalign);
+void *__libc_pvalloc(size_t size) __TBB_ALIAS_ATTR_COPY(pvalloc);
+void *__libc_valloc(size_t size) __TBB_ALIAS_ATTR_COPY(valloc);
+
+// Last place tbb alias attr copy is used
+#undef __TBB_ALIAS_ATTR_COPY
```
alexey-katranov, line 288: Usually, we do not undef macros (moreover, it is a cpp file). Do you have any objections against removing it?

alexey-katranov (commented):
No changes are required (we apply the comments during integration if no objections).

src/tbbmalloc_proxy/proxy.cpp
SteveBronder: I'm fine with leaving the macro. I was following the google style guide for [macros](https://google.github.io/styleguide/cppguide.html#Preprocessor_Macros) but whatever works for tbb is fine for me

> #define macros right before you use them, and #undef them right after.

src/tbbmalloc_proxy/proxy.cpp
SteveBronder: Yes that's a v good idea, thank you!

SteveBronder:
Hi Alexey, would you like me to apply the above or has it already gone through integration? I'm happy to make the changes above if it's easier for yinz

alexey-katranov:
The patch is in process of integration. No any additional changes are required. Thank you!

SteveBronder:
Np!

alexey-katranov:
The patch seems integrated. Feel free to reopen if any questions/concerns.

PR #360, opened 2021-03-10
Fixes compliation failure on Mac under LLVM and GCC.

sighingnow:
On Mac, the clang and gcc is recognized as `Clang` and `GNU` rather than `AppleClang`, thus requires such apple specific settings.

Have tested with brew install llvm (llvm-11) and gcc (gcc-10).

Signed-off-by: Tao He <sighingnow@gmail.com>

### DIFF
--- a/cmake/compilers/Clang.cmake
+++ b/cmake/compilers/Clang.cmake
@@ -12,8 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-set(TBB_LINK_DEF_FILE_FLAG -Wl,--version-script=)
-set(TBB_DEF_FILE_PREFIX lin${TBB_ARCH})
+# --version-script= is not supported on MacOS
+if (APPLE)
+    set(TBB_LINK_DEF_FILE_FLAG -Wl,-exported_symbols_list,)
+    set(TBB_DEF_FILE_PREFIX mac${TBB_ARCH})
+else()
+    set(TBB_LINK_DEF_FILE_FLAG -Wl,--version-script=)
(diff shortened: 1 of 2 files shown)


AlexVeprev:
Looks related to https://github.com/oneapi-src/oneTBB/issues/344.

@sighingnow, have you considered something like
```cmake
if (APPLE)
    include(${CMAKE_CURRENT_LIST_DIR}/AppleClang.cmake)
    return()
endif()
```

in the beginning of GNU.cmake and Clang.cmake?

I've applied such quick fix [here](https://github.com/conda-forge/tbb-feedstock/pull/78/files#diff-68aa83f395d12af7ac6d50800d2dc7fe448e8d292ac7c545674bd7f3e5dcfff3), but haven't investigated it deeply.

sighingnow:
> have you considered something like ...

That could work, but raises further concerns.

+ The arguments passed to appleclang and clang is not exactly same, I'm not sure if there is any noticeable impact.
+ For further development, there's indeed chances to have totally different settings for appleclang and clang (as well as for appleclang and gnu), at that time, we need to go back here and drop the `include`.

AlexVeprev:
@sighingnow, thank you for the PR! It has been applied in master with small updates:
https://github.com/oneapi-src/oneTBB/blob/9e15720bc7744f85dff611d34d65e9099e077da4/cmake/compilers/Clang.cmake#L15
https://github.com/oneapi-src/oneTBB/blob/9e15720bc7744f85dff611d34d65e9099e077da4/cmake/compilers/GNU.cmake#L15

Could you please check if current master works on your side?

sighingnow:
It works well with llvm-11 on Mac, thanks for applying the patch!

PR #394, opened 2021-05-11
Fix build with gcc11+glibc-2.34

kraj:
Pass a valid pointer to pthread_setspecific to avoid GCC 11 warning.
Fixes

src/tbb/tls.h:44:46: error: 'int pthread_setspecific(pthread_key_t, const void*)' expecting 1 byte in a region of size 0 [-Werror=stringop-overread]
|    44 |     void set( T value ) { pthread_setspecific(my_key, (void*)value); }
|       |                           ~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~
| compilation terminated due to -Wfatal-errors.

Signed-off-by: Khem Raj <raj.khem@gmail.com>

### DIFF
--- a/src/tbb/parallel_pipeline.cpp
+++ b/src/tbb/parallel_pipeline.cpp
@@ -243,7 +243,8 @@ public:
         return end_of_input_tls.get() != 0;
     }
     void set_my_tls_end_of_input() {
-        end_of_input_tls.set(1);
+	// Use an arbitrary but valid pointer as value.
+        end_of_input_tls.set(array_size);
     }
 };
 


src/tbb/parallel_pipeline.cpp
```diff
@@ -243,7 +243,8 @@ class input_buffer {
         return end_of_input_tls.get() != 0;
     }
     void set_my_tls_end_of_input() {
-        end_of_input_tls.set(1);
+	// Use an arbitrary but valid pointer as value.
+        end_of_input_tls.set(array_size);
```
alexey-katranov, line 247: I am not sure that I understand the waring and approach how it is fixed (`array_size` is also integral type similar to `1`).

alexey-katranov: Thank you for the explanation.

It seems to be specific to `pthread_setspecific`. Can we try to encapsulate the workaround into `basic_tls`? Will it work if we rework implementation near [`tbb/tls.h::44`](https://github.com/oneapi-src/oneTBB/blob/master/src/tbb/tls.h#L44)?:
```diff 
-void set( T value ) { pthread_setspecific(my_key, (void*)value); }
+// Promote `value` into pointer-size integral to prevent GCC 11+ warning.
+void set( T value ) { pthread_setspecific(my_key, (void*)(std::uintptr_t)value; }
```

alexey-katranov: Can you please rebase on the master branch?

alexey-katranov (changes requested):
Can you please rebase on the master branch?

src/tbb/parallel_pipeline.cpp
kraj: I don't this it will work. problem is the value being passed not the type.

alexey-katranov: For some unclear reason I cannot reproduce the issue on godbolt: https://godbolt.org/z/rYseqo367

I am slightly confused that we need some changes on user side of `tls`. It is implementation detail that `tls` uses `pthread_setspecific` (on other systems it uses something else). Moreover, the `my_tls_end_of_input` function (line 243) expects some non-zero value, maybe `array_size` is not zero, but it seems strange to pass something unknown that affect program flow. Do not we have some other workarounds inside the `tls::set` function to resolve the warning, e.g. suppress with compiler specific pragmas if there is no better solution?

kraj: you need glibc/master ( soon to be 2.34 ) release

kraj:
> Can you please rebase on the master branch?

done

kraj:
@alexey-katranov can you review it again now?

src/tbb/parallel_pipeline.cpp
alexey-katranov: The concern [is addressed](https://github.com/oneapi-src/oneTBB/pull/519/files#diff-2595b33d006f9bcc276fdb214d03b3c8f7cd7ccde63358b534ad69bccb20d2b8R242-R247) in #519

anton-potapov:
@kraj , does this still needed after #519 landed ?

kraj:
> @kraj , does this still needed after #519 landed ?

I am not sure, have not tested master yet in long time

anton-potapov:
@kraj, friendly reminder

PR #395, opened 2021-05-12
build: Fix paths in .pc file with absolute GNUInstallDirs

jtojnar:
It is not generally true that `CMAKE_INSTALL_<dir>` variables are relative paths:

https://github.com/jtojnar/cmake-snips#concatenating-paths-when-building-pkg-config-files


src/tbb/CMakeLists.txt
```diff
@@ -137,6 +137,8 @@ else()
     if (BUILD_TYPE STREQUAL "debug")
         set(TBB_LIB_NAME ${TBB_LIB_NAME}${CMAKE_DEBUG_POSTFIX})
     endif()
+    join_paths(libdir_for_pc_file "\${prefix}" "${CMAKE_INSTALL_LIBDIR}")
```
AlexVeprev, line 140: While `join_paths` seems to work properly we have a simpler case here, so I'd suggest to use couple of `if`'s instead of additional module:
```cmake
if (IS_ABSOLUTE "${CMAKE_INSTALL_LIBDIR}")
    set(_libdir_for_pc_file "${CMAKE_INSTALL_LIBDIR}")
else()
    set(_libdir_for_pc_file "\${prefix}/${CMAKE_INSTALL_LIBDIR}")
endif()

if (IS_ABSOLUTE "${CMAKE_INSTALL_INCLUDEDIR}")
    set(_includedir_for_pc_file "${CMAKE_INSTALL_INCLUDEDIR}")
else()
    set(_includedir_for_pc_file "\${prefix}/${CMAKE_INSTALL_INCLUDEDIR}")
endif()

AlexVeprev (commented):
@jtojnar, thank you for finding this weak place. Do you have examples where this leads to failures? I've read your explanations in cmake-snips, but I wonder if there are real examples of such failures with TBB.

Changes in `target_include_directories` looks good to me (good catch!), but I'd propose to simplify pkg-config part.

src/tbb/CMakeLists.txt
jtojnar: Yeah, that should be enough in this case.

jtojnar:
We set `CMAKE_INSTALL_LIBDIR` and `CMAKE_INSTALL_INCLUDEDIR` to absolute paths in Nixpkgs (in bigger packages, they can even have different prefix) so tbb master had non-existent paths in the pkg-config file, preventing packages using pkg-config resolution from finding tbb.

AlexVeprev:
@jtojnar, thank you! Now it looks good to me!

@Iliamish, please review/check it on your side.

Iliamish (approved):

PR #397, opened 2021-05-17
Add collaborative_call_once algorithm

isaevil:
`collaborative_call_once` is a function similar to [`std::call_once`](https://en.cppreference.com/w/cpp/thread/call_once) but allows waiting threads to join oneTBB parallel construction called inside the initializing functor. The semantics of exceptions is also similar to `std::call_once`: whenever the exception is thrown, it is rethrown on the initializing thread and this or other thread can try to call initializing functor again until it is completed without an exception.

This pull request integrates #357. It is reworked to avoid dynamic memory allocation and fit the state into one word (see state machine in [#267](https://github.com/oneapi-src/oneTBB/issues/267#issuecomment-676074464)). Notify: @BenFrantzDale @HenryHeffanFormlabs


BenFrantzDale (commented):
Thanks for considering our idea. This looks great!

I made a few little suggestion in the comments.

include/oneapi/tbb/collaborative_call_once.h
```diff
@@ -0,0 +1,185 @@
...
+            task_group_context::default_traits | task_group_context::concurrent_wait)
+        {}
+
+    ~run_once() {
+        spin_wait_while(ref_count, [&](std::int64_t value) { return value > 0; }, std::memory_order_acquire);
+    }
+
+    void increase_ref() { ref_count++; }
+
+    void decrease_ref() { ref_count--; }
```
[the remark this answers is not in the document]
isaevil: As far as I know, `std::shared_lock` is available since C++14 standard. Minimal supported version of standard for oneTBB is C++11.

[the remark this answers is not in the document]
BenFrantzDale: Gotcha. Could use `lock()`/`unlock()` the same way; I'm not sure how others would feel about if that abuses convention.

include/oneapi/tbb/collaborative_call_once.h
```diff
@@ -0,0 +1,185 @@
...
+    std::atomic<std::uintptr_t> my_state{ state::uninitialized };
+
+    template <typename Fn, typename... Args>
+    friend void collaborative_call_once(collaborative_once_flag& flag, Fn&& f, Args&&... args);
+
+    template <typename Fn>
+    void do_collaborative_call_once(Fn&& f) {
+        std::uintptr_t expected = my_state.load(std::memory_order_acquire);
+        run_once local_runner;
+        while (expected != state::done) {
```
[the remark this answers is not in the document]
isaevil: @alexey-katranov does it makes sense to make an early exit before initialization of runner?
As for the second point: it is necessary to create a runner on stack of each thread, they are competing to publish address of their `local_runner`. Thus, shared runner will be published already on the `compare_exchange_strong`.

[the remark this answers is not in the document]
BenFrantzDale: Right. I see several ways we could go:
1. Leave it as is. (Seems like it leaves an easy improvement on the table.)
2. Add my `if`, allowing subsequent calls to return ASAP when the flag is "used".
3. We could additionally have the definitely-moonlighting threads not construct theirs. That is, if on the first pass through, don't construct `local_runner` if we immediately see that `expected` is already pointing at something. That's a bit more work, that maybe could worth if there wind up being lots of moonlighers. But probably not since worst case it's only the number of worker threads.
4. Try to avoid unnecessary construction in the race case: That would basically require another bit and have the "winner" set the bit to "I'm busy constructing the `local_runner`" which others would have to spin-wait on. That actually might be simpler code? Not sure if it's worth it, though.

My preference is for 2 unless you guys think 3 or 4 is worth the complexity.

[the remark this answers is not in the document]
alexey-katranov: @BenFrantzDale, good catch. I think 4 can be easily implemented with the following wrapper:
```c++
template <T>
class thread_safe_optional {
    std::atomic<bool> m_initialized{false};
    union {
        T m_value;
    }

    thread_safe_optional( const atomic_optional & other ) = delete;
public:
    thread_safe_optional() = default;
    ~thread_safe_optional() {
        if (m_initialized.load(std::memory_order_relaxed)) {
            m_value.~T();
        }
    }
    template <class... Args>
    T& unsafe_emplace( Args&&... args ) {
        new (&m_value) T(std::forward<Args>(args)...);
        return m_value;
    }
    T& get() {
        spin_wait_while_eq(m_initialized, false);
        return m_value;
    }
};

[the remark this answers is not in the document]
Comment (BenFrantzDale) [reply include/oneapi/tbb/collaborative_call_once.h:126]: I like that. That lets us avoid duplicate constructions and also streamline the already-called case.

Are you missing an `m_initialized.store(std::memory_order_release)` after `new`?

I don't love the name `unsafe_emplace` since it's not obvious what's unsafe about it, especially since it's a `thread_safe_optional`. Maybe `forcibly_emplace` (to emphasize that it's potential UB if it's already emplaced)?

I'd put a comment on `get()` that it's `get()` in the `std::future` sense (blocking). I with no context, I expect `get()` to return a pointer (ala the shared pointers). Maybe clearer from the calling side would be to split it apart and have
```
//! wait() like std::future:
void wait() const { spin_wait_while_eq(m_initialized, false); }
//! value() like std::optional:
T& value() { return m_value; }
```
Then the winning thread would `opt.force_emplace();` and the loosing thread would `opt.wait(); auto& foo = opt.value();`?


Is the `std::memory_order_relaxed` (rather than `std::memory_order_acquire`) in the d'tor because the assumption is that the destroying thread that matters is the one that emplaced?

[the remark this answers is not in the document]
Comment (alexey-katranov) [reply include/oneapi/tbb/collaborative_call_once.h:126]: > Are you missing an m_initialized.store(std::memory_order_release) after new?

Yes, thank you!

> I don't love the name unsafe_emplace since it's not obvious what's unsafe about it,

Typically, we use `unsafe_` prefix for not thread safe methods in concurrent containers. In any case, it is just a draft and definitely requires better naming.

> Then the winning thread would opt.force_emplace(); and the loosing thread would opt.wait(); auto& foo = opt.value();?

It seems reasonable

> Is the std::memory_order_relaxed (rather than std::memory_order_acquire) in the d'tor because the assumption is that the destroying thread that matters is the one that emplaced?

Yes, moreover, we expect that the pointer to `thread_safe_optional` is distributed across the threads with happens-before semantics, i.e. even dtor on another thread should be Ok with relaxed.

Overall, it can be better not to introduce an isolated wrapper because it will require careful semantics polishing. The similar methods can be introduced for `run_once` with narrow logic required for the particular case.

File: include/oneapi/tbb/collaborative_call_once.h (hunk 1 above)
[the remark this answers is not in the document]
Comment (alexey-katranov) [reply include/oneapi/tbb/collaborative_call_once.h:88]: I like the idea but we limited with C++11. I am not sure that it makes sense to implement [Mutex named requirement](https://en.cppreference.com/w/cpp/named_req/Mutex) because the following code will be misleading that it is executed by multiple threads:
```
std::lock_guard lock(runner);
runner->wait();
```

Maybe it makes sense to use `try_call/on_copmletion` guard but see below. 

> runner->increaese_ref();
> ... // Be sure this doesn't throw!
> runner->decrease_ref();

We protected ourselves from exceptions with the following expression
```
 [runner] () noexcept { runner->wait(); }();
```
Any exception will cause termination due to `noexcept` specifier.

However, we still need to think about better names.

File: include/oneapi/tbb/collaborative_call_once.h (hunk 2 above)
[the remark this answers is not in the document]
Comment (isaevil) [reply include/oneapi/tbb/collaborative_call_once.h]: I tried to introduce initialization logic for runner class directly. Now `task_arena` and contexts can only be created by winner thread by calling `make_runner()` method.
Moonlighters must wait in method `wait_for_init()` for initialization of storage with `task_arena`.

File: test/tbb/test_collaborative_call_once.cpp [hunk 3]
```diff
@@ -0,0 +1,372 @@
...
+//! Correctness test for Fibonacci example
+//! \brief \ref interface \ref requirement
+TEST_CASE("fibonacci example") {
+    constexpr int N = 25;
+
+    auto naive = naive_recursive_fib(N);
+
+    auto collaborative = collaborative_recursive_fib(N);
+
+    REQUIRE(naive == collaborative);
```
alexey-katranov, line 367: Collaborative Fib has O(N) complexity, so does it make sense to use big N? E.g., 93 and compare the result with `12200160415121876738`?

test/tbb/test_collaborative_call_once.cpp
```diff
@@ -0,0 +1,372 @@
...
+    //     }
+    // };
+    auto work = [&]() {
+        call_ct++;
+        // busy_wait(time_initial_delay);
+        utils::doDummyWork(100000);
+        tbb::parallel_for(tbb::blocked_range<size_t>(0, num_work_pieces, 1), [&](const tbb::blocked_range<size_t>& range) {
+            for (size_t i = range.begin(); i != range.end(); ++i) {
+                // busy_wait(time_per_work_piece);
+                utils::doDummyWork(100000);
```
alexey-katranov, line 298: It seems as synchronization based on sleeps. Consider using barrier to guarantee that all threads joined the parallel loop.

test/tbb/test_collaborative_call_once.cpp
```diff
@@ -0,0 +1,372 @@
...
+            thread.join();
+        }
+        REQUIRE(f.ct == 1);
+    }
+    {
+        increment_functor f;
+
+        utils::SpinBarrier barrier(utils::get_platform_max_threads()*5);
+        tbb::collaborative_once_flag flag;
+        utils::NativeParallelFor(utils::get_platform_max_threads()*5, [&](std::size_t) {
```
alexey-katranov, line 193: We want to check cases when there are more threads than `max_nfs_size`. Does it make sense to rely on `TBB_TEST_LOW_WORKLOAD` for reduced workload but in general case create many threads, e.g. `4*max_nfs_size`?

test/tbb/test_collaborative_call_once.cpp
```diff
@@ -0,0 +1,372 @@
...
+#include "tbb/task_arena.h"
+
+#include <atomic>
+#include <mutex>
+
+//! \file test_collaborative_call_once.cpp
+//! \brief Tests for [algorithms.collaborative_call_once] functionality
+
+struct increment_functor {
+    std::atomic<int> ct{0};
```
alexey-katranov, line 41:
```suggestion
    int ct{0};
```

test/tbb/test_collaborative_call_once.cpp
```diff
@@ -0,0 +1,372 @@
...
+    }
+}
+
+#if TBB_USE_EXCEPTIONS
+
+//! Test for collaborative_call_once exception handling
+//! \brief \ref error_guessing
+TEST_CASE("handles exceptions - state reset") {
+
+    std::atomic<bool> b{ false };
```
alexey-katranov, line 212:
```suggestion
    bool b{ false };
```

test/tbb/test_collaborative_call_once.cpp
```diff
@@ -0,0 +1,372 @@
...
+
+#endif
+
+//! Test for multiple help from moonlighting threads
+//! \brief \ref interface \ref requirement
+TEST_CASE("multiple help") {
+    // Test multiple threads do the work:
+    std::mutex mutex;
+    std::map<int, int> id_ct_map;
+    std::atomic<int> call_ct{0};
```
alexey-katranov, line 278:
```suggestion
    int call_ct{0};
```

test/tbb/test_collaborative_call_once.cpp
```diff
@@ -0,0 +1,372 @@
...
+#include "common/exception_handling.h"
+#include "common/utils_concurrency_limit.h"
+
+
+#include "tbb/collaborative_call_once.h"
+#include "tbb/parallel_invoke.h"
+#include "tbb/parallel_reduce.h"
+#include "tbb/task_arena.h"
+
+#include <atomic>
```
alexey-katranov, line 34:
```suggestion
```

test/tbb/test_collaborative_call_once.cpp
```diff
@@ -0,0 +1,372 @@
...
+#pragma warning( push )
+#pragma warning( disable: 4702 )
+#endif
+
+#include "common/test.h"
+#include "common/exception_handling.h"
+#include "common/utils_concurrency_limit.h"
+
+
+#include "tbb/collaborative_call_once.h"
```
alexey-katranov, line 29:
```suggestion
#include "common/config.h"

// Include first to check missed header dependencies
#include "tbb/collaborative_call_once.h"

#include "common/test.h"
#include "common/exception_handling.h"
#include "common/utils_concurrency_limit.h"

```

alexey-katranov (changes requested):

include/oneapi/tbb/collaborative_call_once.h
```diff
@@ -0,0 +1,211 @@
...
+                do {
+                    auto max_value = expected | (max_nfs_size-1);
+                    expected = spin_wait_while_eq(my_state, max_value);
+                // "expected > state::done" prevents storing values, when state is uninitialized
+                } while (expected > state::done && !my_state.compare_exchange_strong(expected, expected + 1));
+
+                if (auto runner = reinterpret_cast<once_runner*>(maskoff_pointer(expected))) {
+                    runner->wait_for_init();
+
+                    runner->increase_ref();
```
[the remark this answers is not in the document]
BenFrantzDale: Not saying we should but if they were named `lock()`/`unlock()`, then we could use `std::lock_guard<once_runner>` which is in C++11. Even if it's not a mutex, it feels nicer (to me, anyway) to do
```
std::lock_guard<once_runner> hold_ref_count{*runner};
my_state.fetch_sub(1);
runner->assist();
```
If you don't like the name `lock_guard` there, you could put `using reference_count_holder = std::lock_guard<once_runner>;` in `once_runner`, so this code reads as
```
once_runner::reference_count_holder hold_ref_count{*runner};
```
The `noexcept` lambda is scary to me. Should it instead be that `once_runner::assist()` is `noexcept`?

test/tbb/test_collaborative_call_once.cpp
```diff
@@ -0,0 +1,372 @@
...
+    };
+
+    tbb::collaborative_once_flag flag;
+
+    utils::NativeParallelFor(utils::get_platform_max_threads()*5, [&](std::size_t) {
+        while(run_again) {
+            try {
+                tbb::collaborative_call_once(flag, throwing_func);
+            } catch(const call_once_exception&) {
+            }
```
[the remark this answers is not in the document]
isaevil: Actually, we are expecting the exception here. In this test case we are trying to stress test algorithm's exception handling, because on exception `collaborative_once_flag` sets its state to `uninitialized`.

test/tbb/test_collaborative_call_once.cpp
```diff
@@ -0,0 +1,372 @@
...
+        }
+    });
+}
+
+template<typename Fn, typename... Args>
+void call_once_threads(Fn& body, Args&&... args) {
+    tbb::collaborative_once_flag flag;
+    std::vector<std::thread> threads;
+
+#if __GNUC__ && !__TBB_GCC_PARAMETER_PACK_IN_LAMBDAS_PRESENT
```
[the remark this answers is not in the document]
isaevil: I set check for `__GNUC__`, because I think, that the general way to test here- it is using forwarding to `collaborative_call_once` directly. `stored_pack` here - is a workaround for GCC 4.8.5 bug: https://godbolt.org/z/YbTM1qfPs

include/oneapi/tbb/collaborative_call_once.h
```diff
@@ -0,0 +1,211 @@
...
+                local_runner.make_runner();
+
+                auto local_expected = reinterpret_cast<std::uintptr_t>(&local_runner);
+
+                try_call([&] {
+                    local_runner.run_once(std::forward<Fn>(f));
+                }).on_exception([&] {
+                    while (!my_state.compare_exchange_strong(local_expected, state::uninitialized)) {
+                        local_expected = reinterpret_cast<std::uintptr_t>(&local_runner);
+                    }
```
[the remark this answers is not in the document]
isaevil: I think, we'll get a logical race, because since moonlighters don't throw any exceptions, they can easily overwrite `my_state` after it has been set to `unitialized` state.

include/oneapi/tbb/collaborative_call_once.h
```diff
@@ -0,0 +1,211 @@
...
+                });
+
+                while (!my_state.compare_exchange_strong(local_expected, state::done)) {
+                    local_expected = reinterpret_cast<std::uintptr_t>(&local_runner);
+                }
```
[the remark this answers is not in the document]
isaevil: I guess it relates to conversation above too.

include/oneapi/tbb/collaborative_call_once.h
```diff
@@ -0,0 +1,222 @@
...
+                break;
+            } else {
+                // moonlighting thread
+                do {
+                    auto max_value = expected | (max_nfs_size-1);
+                    expected = spin_wait_while_eq(my_state, max_value);
+                // "expected > state::done" prevents storing values, when state is uninitialized
+                } while (expected > state::done && !my_state.compare_exchange_strong(expected, expected + 1));
+
+                if (auto runner = reinterpret_cast<once_runner*>(maskoff_pointer(expected))) {
```
alexey-katranov, line 184:
```suggestion
                if (auto runner = once_runner::from_bits(expected)) {
```

include/oneapi/tbb/collaborative_call_once.h
```diff
@@ -0,0 +1,222 @@
...
+    template <typename Fn, typename... Args>
+    friend void collaborative_call_once(collaborative_once_flag& flag, Fn&& f, Args&&... args);
+
+    template <typename Fn>
+    void do_collaborative_call_once(Fn&& f) {
+        std::uintptr_t expected = my_state.load(std::memory_order_acquire);
+        once_runner local_runner;
+
+        while (expected != state::done) {
+            if (expected == state::uninitialized && my_state.compare_exchange_strong(expected, reinterpret_cast<std::uintptr_t>(&local_runner))) {
```
alexey-katranov, line 157:
```suggestion
            if (expected == state::uninitialized && my_state.compare_exchange_strong(expected, local_runner.to_bits())) {
```
Apply in other places, as well

include/oneapi/tbb/collaborative_call_once.h
```diff
@@ -0,0 +1,222 @@
...
+    enum state {uninitialized, done};
+    std::atomic<std::uintptr_t> my_state{ state::uninitialized };
+
+    template <typename Fn, typename... Args>
+    friend void collaborative_call_once(collaborative_once_flag& flag, Fn&& f, Args&&... args);
+
+    template <typename Fn>
+    void do_collaborative_call_once(Fn&& f) {
+        std::uintptr_t expected = my_state.load(std::memory_order_acquire);
+        once_runner local_runner;
```
alexey-katranov, line 154:
```suggestion
        once_runner runner;
```

include/oneapi/tbb/collaborative_call_once.h
```diff
@@ -0,0 +1,222 @@
...
+
+        while (expected != state::done) {
+            if (expected == state::uninitialized && my_state.compare_exchange_strong(expected, reinterpret_cast<std::uintptr_t>(&local_runner))) {
+                // winner thread
+                local_runner.make_runner();
```
alexey-katranov, line 159:
```suggestion
                local_runner.init();
```

include/oneapi/tbb/collaborative_call_once.h
```diff
@@ -0,0 +1,222 @@
...
+                // The exception is handled with the winner thread
+                task_group_context stub_context;
+                wait(m_storage.my_wait_context, stub_context);
+            });
+        });
+    }
+
+};
+
+class runner_lifetime_checker : no_copy {
```
alexey-katranov, line 133: Make it a nested class in `once_runner` and make private `ref` manipulation functions. 

`runner_lifetime_checker`->`once_runner::lifetime_tracker`

include/oneapi/tbb/collaborative_call_once.h
```diff
@@ -0,0 +1,222 @@
...
+        bool operator()() const override {
+            my_func();
+            return true;
+        }
+
+    private:
+        Func& my_func;
+    };
+
+    std::atomic<std::int64_t> ref_count{0};
```
alexey-katranov, line 69:
```suggestion
    std::atomic<std::int64_t> m_ref_count{0};
```

include/oneapi/tbb/collaborative_call_once.h
```diff
@@ -0,0 +1,222 @@
...
+inline constexpr std::uintptr_t maskoff_pointer(std::uintptr_t ptr) {
+    return ptr & (std::size_t(-1) << bit_count);
+}
+
+class alignas(max_nfs_size) once_runner {
+
+    struct storage_t {
+        task_arena my_arena{ task_arena::attach{} };
+        wait_context my_wait_context{0};
+        task_group_context my_context{ task_group_context::bound,
```
alexey-katranov, line 51: Do you really need the context inside the storage?

test/tbb/test_collaborative_call_once.cpp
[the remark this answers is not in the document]
alexey-katranov: Maybe we will rename `__TBB_GCC_PARAMETER_PACK_IN_LAMBDAS_PRESENT ` to `__TBB_PARAMETER_PACK_IN_LAMBDAS_BROKEN`?

include/oneapi/tbb/collaborative_call_once.h
```diff
@@ -0,0 +1,222 @@
...
+    }
+};
+
+
+template <typename Fn, typename... Args>
+void collaborative_call_once(collaborative_once_flag& flag, Fn&& fn, Args&&... args) {
+    // Using stored_pack to suppress bug in GCC 4.8
+    // with parameter pack expansion in lambda
+    auto stored_pack = save_pack(std::forward<Args>(args)...);
+    auto func = [&] { call(std::forward<Fn>(fn), stored_pack); };
```
alexey-katranov, line 206: I do not think that all compilers should suffer from `save_pack` because of GCC 4.8. Let's implement two versions: one for GCC 4.8 and another one is generic with variadic in lambdas.

include/oneapi/tbb/collaborative_call_once.h
```diff
@@ -0,0 +1,222 @@
...
+        m_runner.increase_ref();
+    }
+    ~runner_lifetime_checker() {
+        m_runner.decrease_ref();
+    }
+};
+
+class collaborative_once_flag : no_copy {
+    enum state {uninitialized, done};
+    std::atomic<std::uintptr_t> my_state{ state::uninitialized };
```
alexey-katranov, line 146: Consider consistent naming: use either `m_` or `my_` across of new code.

alexey-katranov (changes requested):
Consider additional tests:
- multiple arenas
- move only arguments

include/oneapi/tbb/collaborative_call_once.h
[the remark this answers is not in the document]
isaevil: We decided not to use `lock_guard`  as `reference_count_holder`, but we introduced helper class `once_runner::lifetime_tracker`, that increase reference count on construction and decrease it on destruction. Also removed scary lambda and declared `once_runner::assist()` as `noexcept`.

include/oneapi/tbb/collaborative_call_once.h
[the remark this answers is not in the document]
isaevil: See conversation below.

include/oneapi/tbb/collaborative_call_once.h
```diff
@@ -0,0 +1,229 @@
...
+    void do_collaborative_call_once(Fn&& f) {
+        std::uintptr_t expected = m_state.load(std::memory_order_acquire);
+        once_runner runner;
+
+        while (expected != state::done) {
+            if (expected == state::uninitialized && m_state.compare_exchange_strong(expected, runner.to_bits())) {
+                // winner thread
+                runner.init();
+
+                expected = runner.to_bits();
```
alexey-katranov, line 166:
```suggestion
                auto set_state = [runner_bits = runner.to_bits(), &m_state] (std::uintptr_t desired) {
                    do {
                        auto expected = runner_bits;
                        spin_wait_until_eq(m_state, expected);
                    } while (!m_state.compare_exchange_strong(expected, desired));
                });
```

include/oneapi/tbb/collaborative_call_once.h
```diff
@@ -0,0 +1,229 @@
...
+                try_call([&] {
+                    runner.run_once(std::forward<Fn>(f));
+                }).on_exception([&] {
+                    while (!m_state.compare_exchange_strong(expected, state::uninitialized)) {
+                        expected = runner.to_bits();
+                    }
```
alexey-katranov, line 172:
```suggestion
                    set_state(state::uninitialized);
```

include/oneapi/tbb/collaborative_call_once.h
```diff
@@ -0,0 +1,229 @@
...
+                });
+
+                while (!m_state.compare_exchange_strong(expected, state::done)) {
+                    expected = runner.to_bits();
+                }
```
alexey-katranov, line 177:
```suggestion
                set_state(state::done);
```

test/tbb/test_collaborative_call_once.cpp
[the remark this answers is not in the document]
alexey-katranov: I suppose we expect `const call_once_exception&` not an arbitrary type (`...`). The idea to enforce that we receive only `const call_once_exception&` (the comment is put on `}` not on the catch - it is supposed to remain as is)

alexey-katranov (changes requested):
It is in a good shape. Just minor polishing.

alexey-katranov (dismissed):

alexey-katranov:
@BenFrantzDale , @HenryHeffanFormlabs, we finished with productization and the patch is ready for the commit. Do you have any last minute concerns or remarks before we merge?

alexey-katranov (approved):

BenFrantzDale (approved):
Looks good to me. I like what you folks have done with it!

PR #400, opened 2021-05-17
ASAN fixes for concurrent_skip_list

anton-potapov:
Address SANitiser spotted use after scope issue in the `concurrent_skip_list` (`concurrent_map`/`concurrent_set`). 

Changes are: 
 - removed reference to container allocator from `skip_list_node` to avoid have it dangling due to container move/assignment 
 - added special `skip_list_node::create` `skip_list_node::destroy` static  methods
 - moved `concurrent_skip_list`, `*_map`/`*multimap`, `*set`/`*multiset` into `d2` namespace due layout change of `skip_list_node`

@alexey-katranov . @kboyarinov please take a look

### DIFF
--- a/include/oneapi/tbb/detail/_concurrent_skip_list.h
+++ b/include/oneapi/tbb/detail/_concurrent_skip_list.h
@@ -49,7 +49,7 @@
 
 namespace tbb {
 namespace detail {
-namespace d1 {
+namespace d2 {
 
 template <typename Value, typename Allocator>
 class skip_list_node {
@@ -72,24 +72,51 @@ public:
     using pointer = typename value_allocator_traits::pointer;
     using const_pointer = typename value_allocator_traits::const_pointer;
 
-    skip_list_node( size_type levels, container_allocator_type& alloc )
-        : my_container_allocator(alloc), my_height(levels), my_index_number(0)
-    {
-        for (size_type l = 0; l < my_height; ++l) {
-            allocator_traits::construct(my_container_allocator, &get_atomic_next(l), nullptr);
-        }
-    }
+    //In perfect world these constructor and destructor would have been private, 
+    //however this seems technically impractical due to use of allocator_traits.
 
-    ~skip_list_node() {
-        for (size_type l = 0; l < my_height; ++l) {
-            allocator_traits::destroy(my_container_allocator, &get_atomic_next(l));
-        }
-    }
+    //Should not be called directly, instead use create method
+    skip_list_node( size_type levels )
+        : my_height(levels), my_index_number(0)
+    {}
+
+    //Should not be called directly, instead use destroy method
+    ~skip_list_node() {}
 
     skip_list_node( const skip_list_node& ) = delete;
     skip_list_node( skip_list_node&& ) = delete;
     skip_list_node& operator=( const skip_list_node& ) = delete;
     skip_list_node& operator=( skip_list_node&& ) = delete;
+
+    static skip_list_node* create( container_allocator_type& alloc, size_type height ) {
+        size_type sz = calc_node_size(height);
+        static_assert(std::is_same<typename allocator_traits::value_type, std::uint8_t>::value, "skip_list_node assumes that passed in allocator operates on bytes");
+        auto* node = reinterpret_cast<skip_list_node*>(allocator_traits::allocate(alloc, sz));
+
+        //Construct the node itself
+        allocator_traits::construct(alloc, node, height);
+
+        //Construct the level pointers
+        for (size_type l = 0; l < height; ++l) {
+            allocator_traits::construct(alloc, &node->get_atomic_next(l), nullptr);
+        }
+
+        return node;
+    }
+
+    static void destroy( container_allocator_type& alloc, skip_list_node* node ) {
+        //Destroy the level pointers
+        for (size_type l = 0; l < node->height(); ++l) {
+            allocator_traits::destroy(alloc, &node->atomic_next(l));
+        }
+        size_type sz = calc_node_size(node->height());
+        // Destroy the node itself
+        allocator_traits::destroy(alloc, node);
+
+        // Deallocate the node
+        allocator_traits::deallocate(alloc, reinterpret_cast<std::uint8_t*>(node), sz);
+    }
+
 
     pointer storage() {
         return &my_value;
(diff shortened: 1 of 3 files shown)


include/oneapi/tbb/detail/_concurrent_skip_list.h
```diff
@@ -105,11 +97,13 @@ class skip_list_node {
         return res;
     }
 
-    atomic_node_ptr& atomic_next( size_type level ) {
```
kboyarinov, line 108: May be we can add a new member function: `atomic_next_address`, return `&get_atomic_next` and use it directly in `construct`?
In this case we can avoid using boolean flags just to work around the assertion problem.

include/oneapi/tbb/detail/_concurrent_skip_list.h
```diff
@@ -1089,11 +1094,15 @@ class concurrent_skip_list {
...
-        // Destroy the node
+        //Destroy the level pointers
+        for (size_type l = 0; l < node->height(); ++l) {
+            node_allocator_traits::destroy(my_node_allocator, &node->atomic_next(l));
+        }
+        // Destroy the node itself
         node_allocator_traits::destroy(my_node_allocator, node);
+
         // Deallocate the node
+        size_type sz = calc_node_size(node->height());
```
kboyarinov, line 1105: Seems like UB takes place here. `sz` calculation executes `node->height()`, but the destructor of `node` was called on the previous line.

include/oneapi/tbb/detail/_concurrent_skip_list.h
```diff
@@ -72,19 +72,11 @@ class skip_list_node {
...
+    skip_list_node( size_type levels )
+        : my_height(levels), my_index_number(0)
+    {}
 
-    ~skip_list_node() {
-        for (size_type l = 0; l < my_height; ++l) {
-            allocator_traits::destroy(my_container_allocator, &get_atomic_next(l));
-        }
-    }
+    ~skip_list_node() {}
```
kboyarinov, line 79: May be remove the declaration at all?

kboyarinov (changes requested):

include/oneapi/tbb/detail/_concurrent_skip_list.h
anton-potapov: good catch

include/oneapi/tbb/detail/_concurrent_skip_list.h
anton-potapov: IMHO both solutions are not that straightforward. 
However extending existing method with a parameter seems to be a lesser evil than introducing new method (with functionality overlapping with existing one, which by itself is misleading).

include/oneapi/tbb/detail/_concurrent_skip_list.h
anton-potapov: It is not possible. Due to union member defaulted/compiler generated constructor is deleted.  (https://eel.is/c++draft/class.dtor#7)

kboyarinov (dismissed):

include/oneapi/tbb/detail/_concurrent_skip_list.h
```diff
@@ -72,25 +72,51 @@ class skip_list_node {
...
 
     skip_list_node( const skip_list_node& ) = delete;
     skip_list_node( skip_list_node&& ) = delete;
     skip_list_node& operator=( const skip_list_node& ) = delete;
     skip_list_node& operator=( skip_list_node&& ) = delete;
 
+    static skip_list_node* create( container_allocator_type& alloc, size_type height ) {
+        size_type sz = calc_node_size(height);
+        auto* node = reinterpret_cast<skip_list_node*>(allocator_traits::allocate(alloc, sz));
```
kboyarinov, line 93: One more comment regarding this place: I would add `static_assert` that either `std::is_same_v<allocator_traits::value_type, std::uint8_t>` or `sizeof(allocator_traits::value_type) == sizeof(std::uint8_t)`.

anton-potapov: added

include/oneapi/tbb/detail/_concurrent_skip_list.h
```diff
@@ -700,9 +731,13 @@ class concurrent_skip_list {
     }
 
     void internal_move(concurrent_skip_list&& other) {
-        my_head_ptr.store(other.my_head_ptr.load(std::memory_order_relaxed), std::memory_order_relaxed);
+        node_ptr head = my_head_ptr.exchange(other.my_head_ptr.load(std::memory_order_relaxed), std::memory_order_relaxed);
```
alexey-katranov, line 734: `exchange` is a relatively expensive operation on x86 (even with the relaxed semantics - it is not supported with memory operands). 
```suggestion
        node_ptr head = my_head_ptr.load(std::memory_order_relaxed);
        my_head_ptr.store(other.my_head_ptr.load(std::memory_order_relaxed), std::memory_order_relaxed);
```

alexey-katranov (changes requested):

include/oneapi/tbb/detail/_concurrent_skip_list.h
```diff
@@ -700,9 +731,13 @@ class concurrent_skip_list {
...
         other.my_head_ptr.store(nullptr, std::memory_order_relaxed);
 
+        if (head != nullptr) {
+            delete_node(head);
+        }
```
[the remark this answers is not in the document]
kboyarinov: If I understand correct, the problem takes place when `internal_move` is called during move assignment. I have checked the code - we call `clear()` method before the `internal_move`. The problem happends because `clear()` does not delete the head node for the optimization.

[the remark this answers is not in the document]
anton-potapov: actually `clear()` is called in the `operator=` of the `concurrent_skip_list`. 

However this part of the patch is moved to separate PR

include/oneapi/tbb/detail/_concurrent_skip_list.h
anton-potapov: rolled back

alexey-katranov (approved):

include/oneapi/tbb/detail/_concurrent_skip_list.h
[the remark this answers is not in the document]
anton-potapov: moved to #409

PR #401, opened 2021-05-18
ASAN fix for test_eh_algorithm

anton-potapov:
- fixed read out of bounds

### DIFF
--- a/test/tbb/test_eh_algorithms.cpp
+++ b/test/tbb/test_eh_algorithms.cpp
@@ -1539,18 +1539,21 @@ void TestWithDifferentFiltersAndConcurrency() {
         g_NumThreads = static_cast<int>(concurrency_level);
         g_Master = std::this_thread::get_id();
         if (g_NumThreads > 1) {
+
+            const tbb::filter_mode modes[] = {
+                tbb::filter_mode::parallel,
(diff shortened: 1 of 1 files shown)


Iliamish (approved):

test/tbb/test_eh_algorithms.cpp
```diff
@@ -1539,18 +1539,21 @@ void TestWithDifferentFiltersAndConcurrency() {
...
+            };
+
+            const int NumFilterTypes = sizeof(modes)/sizeof(modes[0]);
+
             // Execute in all the possible modes
-            for ( size_t j = 0; j < 4; ++j ) {
+            for ( size_t j = 0; j < NumFilterTypes; ++j ) {
                 tbb::global_control control(tbb::global_control::max_allowed_parallelism, g_NumThreads);
                 g_ExceptionInMaster = (j & 1) != 0;
                 g_SolitaryException = (j & 2) != 0;
```
aleksei-fedotov, line 1555: I think the number of loop iterations equal to 4 here is primarily to cover all possible values of pair <`g_ExceptionInMaster`, `g_SolitaryException`>. With `NumFilterTypes` equal to 3, the test does not cover <true, true> case. Was that idea considered during this change?

PR #402, opened 2021-05-18
Fix the logic that depends on optional copy elision

ivankochin:
Disabling the [copy elision](https://en.cppreference.com/w/cpp/language/copy_elision) optimization causes unexpected behavior in several tests and interfaces. The patch gets rid of the dependency on this optional optimization.

#### raii_guard
Without explicitly defined move semantic and with disabled optional copy elision the interface `make_raii_guard()` causes the additional temporary object that will just be moved out and then destroyed. But during the temporary object destruction, this object will contain `true` in the `is_active` field which will cause the calling of the `on_exception()` without any exception.
#### attached arena construction
The execution of the code
```c++
oneapi::tbb::task_arena attached_arena = oneapi::tbb::task_arena(oneapi::tbb::task_arena::attach());
```
with disabled optional copy elision optimization causes the construction of additional temporary object which will be attached, but after it, this temporary object will be copy-constructed to the `attached_arena` variable using copy constructor which is not expected at corresponding test cases.
#### doctest internal fix
The code
```c++
auto DOCTEST_ANONYMOUS(_DOCTEST_CAPTURE_) = doctest::detail::MakeContextScope(lambda_name)
```
with disabled optional copy elision optimization causes the construction of additional temporary which causes the segmentation fault during `WARN_MESSSAGE()` message printing. The change that was applied in this patch fixes the problem.

### DIFF
--- a/cmake/compilers/Clang.cmake
+++ b/cmake/compilers/Clang.cmake
@@ -48,5 +48,7 @@ if (ANDROID_PLATFORM)
     set(TBB_COMMON_COMPILE_FLAGS ${TBB_COMMON_COMPILE_FLAGS} $<$<NOT:$<CONFIG:Debug>>:-D_FORTIFY_SOURCE=2>)
 endif()
 
+set(TBB_COMMON_COMPILE_FLAGS ${TBB_COMMON_COMPILE_FLAGS} -fno-elide-constructors)
+
 # TBB malloc settings
 set(TBBMALLOC_LIB_COMPILE_FLAGS -fno-rtti -fno-exceptions)
--- a/cmake/compilers/Clang.cmake
+++ b/cmake/compilers/Clang.cmake
@@ -48,7 +48,5 @@ if (ANDROID_PLATFORM)
     set(TBB_COMMON_COMPILE_FLAGS ${TBB_COMMON_COMPILE_FLAGS} $<$<NOT:$<CONFIG:Debug>>:-D_FORTIFY_SOURCE=2>)
 endif()
 
-set(TBB_COMMON_COMPILE_FLAGS ${TBB_COMMON_COMPILE_FLAGS} -fno-elide-constructors)
-
 # TBB malloc settings
 set(TBBMALLOC_LIB_COMPILE_FLAGS -fno-rtti -fno-exceptions)
--- a/include/oneapi/tbb/detail/_template_helpers.h
+++ b/include/oneapi/tbb/detail/_template_helpers.h
@@ -184,12 +184,7 @@ class raii_guard {
 public:
     raii_guard( Func f ) : my_func(f), is_active(true) {}
 
-    raii_guard (raii_guard&& g) : my_func(g.my_func), is_active(g.is_active) {
-        g.is_active = false;
-    }
-    raii_guard& operator=(raii_guard&& g) {
-        my_func = std::move(g.my_func);
-        is_active = g.is_active;
+    raii_guard (raii_guard&& g) : my_func(std::move(g.my_func)), is_active(g.is_active) {
         g.is_active = false;
     }
 
--- a/include/oneapi/tbb/detail/_template_helpers.h
+++ b/include/oneapi/tbb/detail/_template_helpers.h
@@ -22,7 +22,7 @@
 
 #include <cstddef>
 #include <cstdint>
-
+#include <utility>
 #include <type_traits>
 #include <memory>
 #include <iterator>
@@ -182,9 +182,15 @@ using pack_element_t = typename pack_element<N, Args...>::type;
 template <typename Func>
 class raii_guard {
 public:
-    raii_guard( Func f ) : my_func(f), is_active(true) {}
+    static_assert(
+        std::is_nothrow_copy_constructible<Func>::value &&
+        std::is_nothrow_move_constructible<Func>::value,
+        "Throwing an exception during the Func copy or move construction cause an unexpected behavior."
+    );
 
-    raii_guard (raii_guard&& g) : my_func(std::move(g.my_func)), is_active(g.is_active) {
+    raii_guard( Func f ) noexcept : my_func(f), is_active(true) {}
+
+    raii_guard( raii_guard&& g ) noexcept : my_func(std::move(g.my_func)), is_active(g.is_active) {
         g.is_active = false;
     }
 
--- a/include/oneapi/tbb/detail/_template_helpers.h
+++ b/include/oneapi/tbb/detail/_template_helpers.h
@@ -184,6 +184,15 @@ class raii_guard {
 public:
     raii_guard( Func f ) : my_func(f), is_active(true) {}
 
+    raii_guard (raii_guard&& g) : my_func(g.my_func), is_active(g.is_active) {
+        g.is_active = false;
+    }
+    raii_guard& operator=(raii_guard&& g) {
+        my_func = std::move(g.my_func);
+        is_active = g.is_active;
+        g.is_active = false;
+    }
+
     ~raii_guard() {
         if (is_active) {
             my_func();
--- a/test/common/doctest.h
+++ b/test/common/doctest.h
@@ -1522,7 +1522,7 @@ namespace detail {
     public:
         explicit ContextScope(const L &lambda) : ContextScopeBase(), lambda_(lambda) {}
 
-        ContextScope(ContextScope &&other) : ContextScopeBase(), lambda_(std::move(other.lambda_)) {}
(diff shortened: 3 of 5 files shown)


test/tbb/test_task_arena.cpp
```diff
@@ -1218,9 +1218,14 @@ namespace TestReturnValueNS {
...
             std::size_t moved = cnts[StateTrackableBase::MoveInitialized];
             REQUIRE(cnts[StateTrackableBase::Destroyed] == copied + moved);
-            // The number of copies/moves should not exceed 3: function return, store to an internal storage,
-            // acquire internal storage.
-            REQUIRE((copied == 0 && moved <=3));
+            // The number of copies/moves should not exceed 3 if copy elision takes a place:
+            // function return, store to an internal storage, acquire internal storage.
+            // For compilation, without copy elision, this number may be grown up to 7.
+            REQUIRE((copied == 0 && moved <=7));
+            WARN_MESSAGE(moved <= 3,
```
ivankochin, line 1225:
```suggestion
            REQUIRE((copied == 0 && moved <= 7));
            WARN_MESSAGE(moved <= 3,
```

include/oneapi/tbb/detail/_template_helpers.h
```diff
@@ -184,6 +184,15 @@ class raii_guard {
 public:
     raii_guard( Func f ) : my_func(f), is_active(true) {}
 
+    raii_guard (raii_guard&& g) : my_func(g.my_func), is_active(g.is_active) {
```
anton-potapov, line 187: my_func should be moved as well (but not copied )

include/oneapi/tbb/detail/_template_helpers.h
```diff
@@ -184,6 +184,15 @@ class raii_guard {
...
+        g.is_active = false;
```
anton-potapov, line 188: BTW what if copy/move of the function throws an exception ?

include/oneapi/tbb/detail/_template_helpers.h
```diff
@@ -184,6 +184,15 @@ class raii_guard {
...
+    }
+    raii_guard& operator=(raii_guard&& g) {
```
anton-potapov, line 190: Do we really need this function ? 
IMHO both assignment operators for `raii_guard` should be banned. 


And if we do __realy__ need them - first question what if the object being assigned is already armed (i.e. has function assigned and `is_active == true`) ?

cmake/compilers/Clang.cmake
```diff
@@ -48,5 +48,7 @@ if (ANDROID_PLATFORM)
     set(TBB_COMMON_COMPILE_FLAGS ${TBB_COMMON_COMPILE_FLAGS} $<$<NOT:$<CONFIG:Debug>>:-D_FORTIFY_SOURCE=2>)
 endif()
 
+set(TBB_COMMON_COMPILE_FLAGS ${TBB_COMMON_COMPILE_FLAGS} -fno-elide-constructors)
```
anton-potapov, line 51: does it really needed with proper move constructor in place ?

ivankochin: This option required here only for testing and should be removed before the merge

include/oneapi/tbb/detail/_template_helpers.h
ivankochin: In such a case, the logic would be broken and the `on_exception()` call may be performed without any exception in `try_call()` inside the temporary object destructor. But I don't know how to handle it, maybe we can use `std::move_if_noexcept`, but it still doesn't fix the case when the copy construction of the functor throws an exception.

Also, I think that this is a corner case, since it is an internal interface, and most of the use cases just use lambdas with reference capture or simple callables which do not produce exceptions.

include/oneapi/tbb/detail/_template_helpers.h
ivankochin: I agree that the `raii_guard` object assignment seems strange. This member was removed from the interface.

include/oneapi/tbb/detail/_template_helpers.h
ivankochin: It was decided to check it using `static_assert`

test/common/doctest.h
```diff
@@ -1511,22 +1511,31 @@ namespace detail {
...
     template <typename L> class DOCTEST_INTERFACE ContextScope : public ContextScopeBase
     {
         const L &lambda_;
 
     public:
-        explicit ContextScope(const L &lambda) : lambda_(lambda) {}
+        explicit ContextScope(const L &lambda) : ContextScopeBase(), lambda_(lambda) {}
 
-        ContextScope(ContextScope &&other) : lambda_(other.lambda_) {}
+        ContextScope(ContextScope &&other) : ContextScopeBase(static_cast<ContextScopeBase&&>(other)), lambda_(other.lambda_) {}
```
aleksei-fedotov, line 1527: Let's write a comment for what reason`std::move` cannot be used here.

ivankochin, line 1527:
```suggestion
        // std::move cannot be applied here since there is no utility header included in this place (by some reason all headers are included later)
        ContextScope(ContextScope &&other) : ContextScopeBase(static_cast<ContextScopeBase&&>(other)), lambda_(other.lambda_) {}
```

aleksei-fedotov (approved):

include/oneapi/tbb/detail/_template_helpers.h
```diff
@@ -182,7 +182,17 @@ using pack_element_t = typename pack_element<N, Args...>::type;
...
 class raii_guard {
 public:
-    raii_guard( Func f ) : my_func(f), is_active(true) {}
+    static_assert(
+        std::is_nothrow_copy_constructible<Func>::value &&
+        std::is_nothrow_move_constructible<Func>::value,
+        "Throwing an exception during the Func copy or move construction cause an unexpected behavior."
+    );
+
+    raii_guard( Func f ) noexcept : my_func(f), is_active(true) {}
```
anton-potapov, line 191: Why not add an overload accepting functor via rvalue ref ?

include/oneapi/tbb/detail/_template_helpers.h
```diff
@@ -182,7 +182,17 @@ using pack_element_t = typename pack_element<N, Args...>::type;
...
+
+    raii_guard( raii_guard&& g ) noexcept : my_func(std::move(g.my_func)), is_active(g.is_active) {
```
anton-potapov, line 193: please be so kind to add a test for it

anton-potapov (commented):
sorry for post-review comments , but anuway

include/oneapi/tbb/detail/_template_helpers.h
ivankochin: In this part of the functionality (under this I mean `raii_guard`, `try_call`, `on_exception`, `on_completion`...), all functors objects are passed by copy to let the compiler better optimize the code. We already looked at this, check the assembler and it looks like compilers inline ALL `try_call()` cases and just call the required functors.

So it was decided not to apply passing by reference here to keep the inlining.

include/oneapi/tbb/detail/_template_helpers.h
ivankochin: What cases do you want to test? 

Getting `std::terminate` in the case when the functor copy/move constructor is noexcept but throws an exception? Isn't the termination, in this case, guaranteed by the C++ standard?

include/oneapi/tbb/detail/_template_helpers.h
anton-potapov: I asking not to change existing one method, but instead add a _new_ one

include/oneapi/tbb/detail/_template_helpers.h
anton-potapov: simple case that given (`noexcept`) function is called once (in appropriate time), if the one `raii_guard` instance is move constricted from the other one

include/oneapi/tbb/detail/_template_helpers.h
ivankochin: But what is the purpose of this overload and why do you think that we need to add such one?

include/oneapi/tbb/detail/_template_helpers.h
ivankochin: As I know the common approach here is do not to create a separate test for internal functionality if it already covered by public interfaces testing. If I understand your case correctly, then it already well covered by concurrent container testing.

PR #407, opened 2021-05-21
Fix atomic usage in conformance_parallel_pipeline

alexey-katranov:
Fix two issues in conformance_parallel_pipleine:
1. The serial filters should not use atomics
2. The parallel filters should use atomics

Signed-off-by: Alexei Katranov <alexei.katranov@intel.com>

### DIFF
--- a/test/conformance/conformance_parallel_pipeline.cpp
+++ b/test/conformance/conformance_parallel_pipeline.cpp
@@ -279,7 +279,7 @@ TEST_CASE("Testing filter modes")
     {
         oneapi::tbb::global_control control(oneapi::tbb::global_control::max_allowed_parallelism, concurrency_level);
 
-        std::atomic<short> serial_checker{0};
+        short serial_checker{0};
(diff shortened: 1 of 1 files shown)


test/conformance/conformance_parallel_pipeline.cpp
```diff
@@ -331,18 +331,17 @@ TEST_CASE("Testing max token number")
...
-                                [&active_tokens,&counter](oneapi::tbb::flow_control&fc)
+                                [&active_tokens](oneapi::tbb::flow_control&fc)
                                 {
                                     ++active_tokens;
                                     doWork();
                                     CHECK_MESSAGE(active_tokens <= n_tokens, "max number of tokens is exceed");
                                     --active_tokens;
-                                    if(++counter>=max_counter)
-                                    {
+                                    if (--input_counter < 0) {
```
anton-potapov, line 342: why not make existing variable `counter` an `atomic` and avoid using global variables ?

alexey-katranov: it seems to be the common approach across all test cases.

anton-potapov (approved):

PR #409, opened 2021-05-24
ASAN: fix memory leak in concurrent_skip_list move assign operator

anton-potapov:
ASAN spotted a memory leak in `concurrent_skip_list::operator=(concurrent_skip_list&&)`

### The problem 
`internal_move(concurrent_skip_list&&)` unconditionally overwrites content of `my_head_ptr`, thus leaking it previous value.

### What is  done:
 - `concurrent_skip_list::operator=(concurrent_skip_list&&)` now deletes  the `my_head_ptr`  before calling the `internal_move(concurrent_skip_list&&)`
 - introduced internal helper member function `delete_head`

### DIFF
--- a/include/oneapi/tbb/detail/_concurrent_skip_list.h
+++ b/include/oneapi/tbb/detail/_concurrent_skip_list.h
@@ -346,10 +346,7 @@ public:
 
     ~concurrent_skip_list() {
         clear();
-        node_ptr head = my_head_ptr.load(std::memory_order_relaxed);
-        if (head != nullptr) {
-            delete_node(head);
-        }
+        delete_head();
     }
 
     concurrent_skip_list& operator=( const concurrent_skip_list& other ) {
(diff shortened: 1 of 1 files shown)


include/oneapi/tbb/detail/_concurrent_skip_list.h
```diff
@@ -1111,6 +1107,14 @@ class concurrent_skip_list {
         return create_node(max_level);
     }
 
+    void delete_head() {
+        node_ptr head = my_head_ptr.load(std::memory_order_relaxed);
+        if (head != nullptr) {
+            delete_node(head);
+        }
+        my_head_ptr.store(nullptr, std::memory_order_relaxed);
```
kboyarinov, line 1115: May be move it under the `if` statement to avoid restoring `nullptr` into the head?

kboyarinov (changes requested):

include/oneapi/tbb/detail/_concurrent_skip_list.h
anton-potapov: moved

kboyarinov (approved):

PR #411, opened 2021-05-24
Fix shellcheck scan errors and warnings

isaevil:
`SETVARS_ARGS` variable is not parsed properly, if it contains more than one argument. There are also `shellcheck` scan error and warnings fixes in this PR.

### DIFF
--- a/integration/linux/env/vars.sh
+++ b/integration/linux/env/vars.sh
@@ -141,7 +141,7 @@ TBBROOT=$(get_script_path "${vars_script_name:-}")/..
 TBB_TARGET_ARCH="intel64"
 
 if [ -n "${SETVARS_ARGS:-}" ]; then
-  tbb_arg_ia32="$(expr "${SETVARS_ARGS:-}" : '^.*[:space:]*\(ia32\)[:space:]*')"
+  tbb_arg_ia32="$(expr "${SETVARS_ARGS:-}" : '^.*\(ia32\)')"
   if [ -n "${tbb_arg_ia32:-}" ]; then
     TBB_TARGET_ARCH="ia32"
   fi
--- a/integration/linux/env/vars.sh
+++ b/integration/linux/env/vars.sh
@@ -140,32 +140,11 @@ TBBROOT=$(get_script_path "${vars_script_name:-}")/..
 
 TBB_TARGET_ARCH="intel64"
 
-if [ -n "${ZSH_VERSION:-}" ]; then
-    setopt sh_word_split
-fi
-
-# shellcheck disable=2154
-if [ -n "$SETVARS_ARGS" ] && [ "$SETVARS_CALL" = "1" ]; then
-    for arg in $SETVARS_ARGS; do
-        case "$arg" in
-        (intel64|ia32)
-            TBB_TARGET_ARCH="${arg}"
-            ;;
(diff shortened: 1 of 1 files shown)


integration/linux/env/vars.sh
```diff
@@ -140,8 +140,10 @@ TBBROOT=$(get_script_path "${vars_script_name:-}")/..
 
 TBB_TARGET_ARCH="intel64"
 
-if [ ! -z "$SETVARS_ARGS" ] && [ "$SETVARS_CALL" "==" "1" ]; then
-    for arg in "$SETVARS_ARGS"; do
+# shellcheck disable=2236
+# shellcheck disable=2154
+if [ ! -z "$SETVARS_ARGS" ] && [ "$SETVARS_CALL" "=" "1" ]; then
```
AlexVeprev, line 145: Do you think it makes sense to put `=` into quotes?

integration/linux/env/vars.sh
```diff
@@ -140,8 +140,10 @@ TBBROOT=$(get_script_path "${vars_script_name:-}")/..
 
 TBB_TARGET_ARCH="intel64"
 
-if [ ! -z "$SETVARS_ARGS" ] && [ "$SETVARS_CALL" "==" "1" ]; then
-    for arg in "$SETVARS_ARGS"; do
+# shellcheck disable=2236
```
AlexVeprev, line 143: Why do you prefer to suppress it but not fix?
Explanations in https://github.com/koalaman/shellcheck/wiki/SC2236 look reasonable.

AlexVeprev (approved):

integration/linux/env/vars.sh
```diff
@@ -140,49 +140,33 @@ TBBROOT=$(get_script_path "${vars_script_name:-}")/..
...
-            TBB_TARGET_ARCH="${arg}"
-            ;;
-            (*)
-            >&2 echo "ERROR: Unknown argument $arg"
-            return 255 2>/dev/null || exit 255
-            ;;
-        esac
-    done
+if [ -n "${SETVARS_ARGS:-}" ]; then
+  tbb_arg_ia32="$(expr "${SETVARS_ARGS:-}" : '^.*[:space:]*\(ia32\)[:space:]*')"
```
xmnboy, line 144: Now that I look at this a second time, there's no value, in this case, to include the `[:space:]*` expressions, it's not providing more precision. So it would be simpler to just use the following regex:
```
'^.*\(ia32\)'
```
until we can define a more precise way of doing this in a POSIX conformant way.

xmnboy (approved):
See my other comment about simplifying the regex. Completely up to you if want to take that action, either way the results should be identical.

integration/linux/env/vars.sh
AlexVeprev: Is there any need in `()`? What if just `'^.*ia32'`?

AlexVeprev: Never mind, `()` are needed. From `man expr`:
> Pattern matches return the string matched between \\( and \\) or null; if \\( and \\) are not used, they return the number of characters matched or 0.

isaevil: Applied.

AlexVeprev (approved):

PR #412, opened 2021-05-25
Remove the same test cases to fix MIPS compilation

ivankochin:
The `relocation truncated to fit: R_MIPS_GOT_PAGE against` issue occurred during the MIPS debug compilation when GOT taking more than 64Kb. Now, this issue appeared during the `test_join_node_msg_key_matching` compilation. There are two same test cases in this test, so removing one of them reduces the size of GOT fixes the issue.

The possible more general solution is to split too complex tests into several exe files on MIPS to reduce the size of GOT and get rid of the problem.

### DIFF
--- a/test/tbb/test_join_node_msg_key_matching.cpp
+++ b/test/tbb/test_join_node_msg_key_matching.cpp
@@ -87,7 +87,7 @@ using T10 = make_tuple < T9, MyMessageKeyWithBrokenKey<std::string, size_t>>;
 // the compiler might generate huge object file in debug (>64M)
 #define TEST_CASE_TEMPLATE_N_ARGS(dec) TEST_CASE_TEMPLATE(dec, T, T2, T3, T4, T5, T6, T7, T8, T9, T10)
 #else
-#define TEST_CASE_TEMPLATE_N_ARGS(dec) TEST_CASE_TEMPLATE(dec, T, T2, T5)
+#define TEST_CASE_TEMPLATE_N_ARGS(dec) TEST_CASE_TEMPLATE(dec, T, T2, T10)
 #endif
 
 //! Serial test with different tuple sizes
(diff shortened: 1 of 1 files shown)


test/tbb/test_join_node_msg_key_matching.cpp
```diff
@@ -94,7 +87,7 @@ using T10 = make_tuple < T9, MyMessageKeyWithBrokenKey<std::string, size_t>>;
 // the compiler might generate huge object file in debug (>64M)
 #define TEST_CASE_TEMPLATE_N_ARGS(dec) TEST_CASE_TEMPLATE(dec, T, T2, T3, T4, T5, T6, T7, T8, T9, T10)
 #else
-#define TEST_CASE_TEMPLATE_N_ARGS(dec) TEST_CASE_TEMPLATE(dec, T, T2, T5, T8, T10)
+#define TEST_CASE_TEMPLATE_N_ARGS(dec) TEST_CASE_TEMPLATE(dec, T, T2, T5)
```
alexey-katranov, line 90: Is it possible to test with T10 instead of T5 to cover the big number of arguments?

ivankochin: What do you propose? Just change the T5 to T10? If yes, then it's possible of course. 

Also, I can fully revert this change and place back both T8 and T10, but I did this change to prevent this error in the near future.

alexey-katranov: > What do you propose? Just change the T5 to T10? If yes, then it's possible of course.

Yes

ivankochin:
```suggestion
#define TEST_CASE_TEMPLATE_N_ARGS(dec) TEST_CASE_TEMPLATE(dec, T, T2, T10)
```

alexey-katranov (approved):

PR #413, opened 2021-05-25
Update doctest version to 2.4.6

ivankochin:
Now our testing system includes the 2.3.5 [doctest framework](https://github.com/onqtam/doctest) version. This patch updates the doctest framework to the latest version (2.4.6 at this moment).

Also, during the development process, several fixes and extensions were applied to the doctest code. All the changes that apply the fixes or add required configurations support should be included in this patch.  Current changes in origin doctest code:
- https://github.com/oneapi-src/oneTBB/pull/402 - fix for copy elision absence case
- Intel Compiler support
- TestSuite default initialization warning solution
- Typo fixing
- Fix several warnings inside doctest
- Fix strange lib unloading issue with MSVC static library

### DIFF
--- a/cmake/compilers/MSVC.cmake
+++ b/cmake/compilers/MSVC.cmake
@@ -30,6 +30,8 @@ if (MSVC_VERSION LESS_EQUAL 1900)
     # https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4503
     set(TBB_TEST_COMPILE_FLAGS ${TBB_TEST_COMPILE_FLAGS} /wd4503)
 endif()
+# Warning suppression C4127: conditional expression is constant
+set(TBB_TEST_COMPILE_FLAGS ${TBB_TEST_COMPILE_FLAGS} /wd4127)
 set(TBB_LIB_COMPILE_FLAGS -D_CRT_SECURE_NO_WARNINGS /GS)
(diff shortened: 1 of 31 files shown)


cmake/compilers/MSVC.cmake
```diff
@@ -30,6 +30,10 @@ if (MSVC_VERSION LESS_EQUAL 1900)
     # https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4503
     set(TBB_TEST_COMPILE_FLAGS ${TBB_TEST_COMPILE_FLAGS} /wd4503)
 endif()
+# Warning suppression C4127: conditional expression is constant
+set(TBB_TEST_COMPILE_FLAGS ${TBB_TEST_COMPILE_FLAGS} /wd4127)
+# Warning suppression C4800: forcing value to bool 'true' or 'false' (performance warning)
+set(TBB_TEST_COMPILE_FLAGS ${TBB_TEST_COMPILE_FLAGS} /wd4800)
```
alexey-katranov, line 36: Do we really want to suppress this warnings unconditionally? We will not test that public header files do not produce these warnings.

ivankochin: Yes, you're right, the local suppression seems like a better solution for this.

Iliamish (dismissed):

alexey-katranov (approved):

PR #415, opened 2021-05-26
ASAN fix for test_malloc_regression

vlserov:
Malloc regression test checks memory leaks by comparing allocated memory before and after the test.
ASAN allocator never releases memory to OS, instead of it just marks those memory as "freed".
Generic OS functions for getting currently allocated memory aren't working correctly under ASAN.
Have to use __sanitizer_get_current_allocated_bytes function in this case.

### DIFF
--- a/test/common/config.h
+++ b/test/common/config.h
@@ -74,4 +74,16 @@ const unsigned MByte = 1024*1024;
 #define __TBB_TEST_SKIP_AFFINITY 1
 #endif
 
+#ifdef __has_feature
+	#define __TBB_HAS_FEATURE(x) __has_feature(x)
+#else
+	#define __TBB_HAS_FEATURE(x) 0
+#endif
+
+#ifdef __SANITIZE_ADDRESS__
+	#define __TBB_TEST_USE_ADDRESS_SANITIZER 1
+#else
+	#define __TBB_TEST_USE_ADDRESS_SANITIZER __TBB_HAS_FEATURE(address_sanitizer)
+#endif
+
 #endif /* __TBB_test_common_config_H */
--- a/test/common/config.h
+++ b/test/common/config.h
@@ -74,16 +74,12 @@ const unsigned MByte = 1024*1024;
 #define __TBB_TEST_SKIP_AFFINITY 1
 #endif
 
-#ifdef __has_feature
-	#define __TBB_HAS_FEATURE(x) __has_feature(x)
-#else
-	#define __TBB_HAS_FEATURE(x) 0
-#endif
-
 #ifdef __SANITIZE_ADDRESS__
 	#define __TBB_TEST_USE_ADDRESS_SANITIZER 1
-#else
-	#define __TBB_TEST_USE_ADDRESS_SANITIZER __TBB_HAS_FEATURE(address_sanitizer)
+#elif defined(__has_feature)
+#if __has_feature(address_sanitizer)
+      #define __TBB_TEST_USE_ADDRESS_SANITIZER 1
+#endif
 #endif
 
 #endif /* __TBB_test_common_config_H */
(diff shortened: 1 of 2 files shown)


test/common/memory_usage.h
```diff
@@ -86,7 +90,9 @@ namespace utils {
     /* Returns 0 if not implemented on platform. */
     std::size_t GetMemoryUsage(MemoryStatType stat = currentUsage) {
         utils::suppress_unused_warning(stat);
-#if __TBB_WIN8UI_SUPPORT || defined(WINAPI_FAMILY)
+#if __TBB_TEST_USE_ADDRESS_SANITIZER
+        return __sanitizer_get_current_allocated_bytes();
```
alexey-katranov, line 94: Can you please add a comment why we need to use special functions with ASAN?

anton-potapov:
> 
> 
> Using __sanitizer_get_current_allocated_bytes function for getting currently allocated memory under Address SANitizer

Could you please update the description what the problem is  - and how the patch is intended to fix it ?

test/common/memory_usage.h
vlserov: Updated with comment below:
Under ASAN generic memory usage functions provide information with ASAN overhead.

test/common/config.h
```diff
@@ -74,4 +74,10 @@ const unsigned MByte = 1024*1024;
 #define __TBB_TEST_SKIP_AFFINITY 1
 #endif
 
+#if defined(__has_feature)
+#if __has_feature(address_sanitizer)
+    #define __TBB_TEST_USE_ADDRESS_SANITIZER 1
```
anton-potapov, line 79: It seems `__has_feature` is the Clang-only extension and some alternative way is needed here is needed to support GCC as well

alexey-katranov: `__SANITIZE_ADDRESS__`
https://gcc.gnu.org/onlinedocs/gcc-4.8.0/cpp/Common-Predefined-Macros.html

alexey-katranov (dismissed):

test/common/config.h
```diff
@@ -74,4 +74,16 @@ const unsigned MByte = 1024*1024;
...
+	#define __TBB_HAS_FEATURE(x) __has_feature(x)
+#else
+	#define __TBB_HAS_FEATURE(x) 0
+#endif
+
+#ifdef __SANITIZE_ADDRESS__
+	#define __TBB_TEST_USE_ADDRESS_SANITIZER 1
+#else
+	#define __TBB_TEST_USE_ADDRESS_SANITIZER __TBB_HAS_FEATURE(address_sanitizer)
+#endif
```
anton-potapov, line 87:
```suggestion
#ifdef __SANITIZE_ADDRESS__
	#define __TBB_TEST_USE_ADDRESS_SANITIZER 1
#elif defined(__has_feature)
#if __has_feature(address_sanitizer)
      #define __TBB_TEST_USE_ADDRESS_SANITIZER 1
#endif
#endif
```

vlserov: Possible "undef" warning for __TBB_TEST_USE_ADDRESS_SANITIZER define in case using clang without address_sanitizer.

anton-potapov: could you please check it ? 
As similar check here https://github.com/oneapi-src/oneTBB/blob/master/include/oneapi/tbb/detail/_config.h#L438-L444 seems to not produce warnings

vlserov: It's not a real issue, since warring "undef" is switched off by default.

alexey-katranov (approved):

anton-potapov:
**LGTM**

anton-potapov (approved):

PR #417, opened 2021-05-27
Reworked missed work avoidance mechanism

pavelkumbrasev:
When threads execute `get_task` or `steal_task` they can omit unsuitable tasks (different isolation or affinity). The omitted tasks are virtually extracted from task pool, that makes them temporary invisible for other threads. It might lead to situation when another thread marks the arena `EMPTY` (as a result of `is_out_of_work()`), while these tasks are still in the arena. If tasks were omitted we need to notify the arena like new work was added (calling `advertise_new_work()`).

The proposed idea is that omit tasks in such way that other threads can not observe it. It will prevent unnecessary arena state transition and premature threads leaving. 

Therefore, the patch introduce shadow tail and head which move after real task extracted(do not participate in tasks omitting mechanism). 

### DIFF
--- a/src/tbb/arena_slot.cpp
+++ b/src/tbb/arena_slot.cpp
@@ -60,6 +60,7 @@ d1::task* arena_slot::get_task_impl(size_t T, execution_data_ext& ed, bool& task
 
 d1::task* arena_slot::get_task(execution_data_ext& ed, isolation_type isolation) {
     __TBB_ASSERT(is_task_pool_published(), nullptr);
+    __TBB_ASSERT(shadow_tail.load(std::memory_order_relaxed) == tail.load(std::memory_order_relaxed), nullptr);
     // The current task position in the task pool.
     std::size_t T0 = tail.load(std::memory_order_relaxed);
     // The bounds of available tasks in the task pool. H0 is only used when the head bound is reached.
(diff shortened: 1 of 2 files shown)


pavelkumbrasev:
@aleksei-fedotov @anton-potapov @alexey-katranov, could you please review this patch?

aleksei-fedotov:
> 
> 
> @aleksei-fedotov @anton-potapov @alexey-katranov, could you please review this patch?

Yep. Working on it!

akukanov:
Rebased on top of the current code base. However, I think the patch needs to be revisited first. Two tests failing on ARM likely mean that "shadowing" the head and the tail may require some extra synchronization.
> The following tests FAILED:
	  5 - test_collaborative_call_once (Subprocess aborted)
	 66 - test_task (Subprocess aborted)

Update: apparently, the patch does not even achieve its goal, because `reset_task_pool_and_leave`, which also resets the shadow state, can be used in the middle of `get_task` while omitted tasks still remain in the pool (yes, this sounds strange - but the actual bounds are stored in `get_task` as `H0` and `T0`, and restored later).

akukanov:
Reimplemented in a different way in #1779; closing this PR.

Issue #408, opened 2021-05-23
Build fails on FreeBSD (13.0)


emaste:
```
FAILED: test/CMakeFiles/conformance_blocked_range2d.dir/conformance/conformance_blocked_range2d.cpp.o 
/usr/bin/c++  -I../test/.. -I../test -I../src/tbb/../../include -O2 -g -DNDEBUG -Wall -Wextra -Werror -Wshadow -Wcast-qual -Woverloaded-virtual -Wnon-virtual-dtor -pthread -std=c++11 -MD -MT test/CMakeFiles/conformance_blocked_range2d.dir/conformance/conformance_blocked_range2d.cpp.o -MF test/CMakeFiles/conformance_blocked_range2d.dir/conformance/conformance_blocked_range2d.cpp.o.d -o test/CMakeFiles/conformance_blocked_range2d.dir/conformance/conformance_blocked_range2d.cpp.o -c ../test/conformance/conformance_blocked_range2d.cpp
In file included from ../test/conformance/conformance_blocked_range2d.cpp:20:
../test/common/utils_concurrency_limit.h:173:5: error: unknown type name 'cpu_set_t'; did you mean 'cpuset_t'?
    cpu_set_t* mask = nullptr;
    ^~~~~~~~~
    cpuset_t
/usr/include/sys/_cpuset.h:50:24: note: 'cpuset_t' declared here
typedef struct _cpuset cpuset_t;
                       ^
In file included from ../test/conformance/conformance_blocked_range2d.cpp:20:
../test/common/utils_concurrency_limit.h:174:20: error: unknown type name 'cpu_set_t'; did you mean 'cpuset_t'?
    ncpus = sizeof(cpu_set_t) * CHAR_BIT;
                   ^~~~~~~~~
                   cpuset_t
```

Appears to be introduced in 4523a7615eaed49c4ed75654b349ca41b92c6381

ivankochin:
Hi @emaste,
Could you please try out the following solution: https://github.com/oneapi-src/oneTBB/tree/fix-freebsd-testing ?

emaste:
Thanks, it resolves that error. Now getting this error:
```
In file included from ../test/conformance/conformance_global_control.cpp:20:
../test/common/utils_concurrency_limit.h:182:16: error: use of undeclared identifier 'CPU_ALLOC'
        mask = CPU_ALLOC(ncpus);
               ^
[...]

PR #425, opened 2021-06-01
Fix tests compilation on FreeBSD

ivankochin:
There are several issues during the compilation and linking oneTBB library on the FreeBSD OS. This patch fixes it.

One of the issues is related to the TBBbind compilation (and linking with the HWLOC library), so this patch also actualizes the TBBbind related CMake documentation and adds a new option for disabling the TBBBind library build.

### DIFF
--- a/test/common/utils_concurrency_limit.h
+++ b/test/common/utils_concurrency_limit.h
@@ -171,9 +171,15 @@ int limit_number_of_threads( int max_threads ) {
 #define OS_AFFINITY_SYSCALL_PRESENT ((__linux__ && !__ANDROID__) || (__FreeBSD_version >= 701000))
 
 #if OS_AFFINITY_SYSCALL_PRESENT
+
+#if __FreeBSD__
+using mask_t = cpuset_t;
+#else
+using mask_t = cpu_set_t;
+#endif
 void get_thread_affinity_mask(std::size_t& ncpus, std::vector<int>& free_indexes) {
-    cpu_set_t* mask = nullptr;
-    ncpus = sizeof(cpu_set_t) * CHAR_BIT;
+    mask_t* mask = nullptr;
+    ncpus = sizeof(mask_t) * CHAR_BIT;
     do {
         mask = CPU_ALLOC(ncpus);
         if (!mask) break;
@@ -206,7 +212,7 @@ void pin_thread_imp(std::size_t ncpus, std::vector<int>& free_indexes, std::atom
     ASSERT(free_indexes.size() > 0, nullptr);
     int mapped_idx = free_indexes[curr_idx++ % free_indexes.size()];
 
-    cpu_set_t *target_mask = CPU_ALLOC(ncpus);
+    mask_t *target_mask = CPU_ALLOC(ncpus);
     ASSERT(target_mask, nullptr);
     CPU_ZERO_S(size, target_mask);
     CPU_SET_S(mapped_idx, size, target_mask);
--- a/test/common/utils_concurrency_limit.h
+++ b/test/common/utils_concurrency_limit.h
@@ -168,18 +168,13 @@ int limit_number_of_threads( int max_threads ) {
 
 #endif // __TBB_TEST_SKIP_AFFINITY
 
-#define OS_AFFINITY_SYSCALL_PRESENT ((__linux__ && !__ANDROID__) || (__FreeBSD_version >= 701000))
+#define OS_AFFINITY_SYSCALL_PRESENT (__linux__ && !__ANDROID__)
 
 #if OS_AFFINITY_SYSCALL_PRESENT
 
-#if __FreeBSD__
-using mask_t = cpuset_t;
-#else
-using mask_t = cpu_set_t;
-#endif
 void get_thread_affinity_mask(std::size_t& ncpus, std::vector<int>& free_indexes) {
-    mask_t* mask = nullptr;
-    ncpus = sizeof(mask_t) * CHAR_BIT;
+    cpu_set_t* mask = nullptr;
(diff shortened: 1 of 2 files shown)


ivankochin:
Seems like the issues that were reported in #408 were fixed. But now we got the following error during the build:
```
In file included from /home/travis/build/actions-test/onetbb/test/common/test.h:32,
from /home/travis/build/actions-test/onetbb/test/tbbmalloc/test_malloc_overload.cpp:45:
/home/travis/build/actions-test/onetbb/test/common/doctest.h: In constructor 'doctest::detail::Subcase::Subcase(const doctest::String&, const char*, int)':
/home/travis/build/actions-test/onetbb/test/common/doctest.h:3824:5: error: incorrect sharing of tree nodes
3824 | Subcase::Subcase(const String& name, const char* file, int line)
```

But we cannot investigate it due to lack of environment so the request is blocked at this moment.

martymac:
Hello,

I am working on updating FreeBSD port. What is the status of those patches ? May I use the to update the port to 2021.3.0 ?

Thanks in advance,
Best regards,

Ganael.

alexey-katranov:
The patch is WIP; however, we really want to finish it and enable FreeBSD. The main issue that we do not have regular access to the FreeBSD system. It would be really helpful if you test the patch on FreeBSD and share the issues with us. Thank you in advance.

martymac:
Hi,

I'll test the patches ASAP and tell you how it goes, thanks!

martymac:
Hello,

I could build the port with success! See build logs here (on FreeBSD 13.0-RELEASE / amd64):

http://box.martymac.org/FreeBSD-Packages/data/FBSD130amd64-default/2021-07-05_18h17m06s/logs/onetbb-2021.3.0.log

with all patches but the following one:

https://github.com/oneapi-src/oneTBB/pull/425/commits/47ce878bb8949c0a91d41ea0e927087971be9098

which does not seem to be necessary since I could build the tbbBind library without problem.

I'll include them in next port update, thanks!

ivankochin:
@martymac, Thanks for testing!
I've removed redundant commit(47ce878) and rebased the branch. So could you please test the branch one more time?

martymac:
Hi @ivankochin

Thanks, seems good to me but edc39b8 and 5c5d500 could be merged into a single modification as the final result is just removing FreeBSD support from OS_AFFINITY_SYSCALL_PRESENT macro.

martymac:
As a side note,

47ce878 is not necessary to build on FreeBSD, *but* it may provide an interesting option (TBB_TBBBIND_BUILD) to have for the build process.

ivankochin:
> As a side note,
 47ce878 is not necessary to build on FreeBSD, but it may provide an interesting option (TBB_TBBBIND_BUILD) to have for the build process.


The option that provides a similar behavior has already merged to master as a separate patch. See `TBB_DISABLE_HWLOC_AUTOMATIC_SEARCH` [here](https://github.com/oneapi-src/oneTBB/blob/master/cmake/README.md#build-system-description)

martymac:
OK, thanks!

test/common/utils_concurrency_limit.h
```diff
@@ -168,9 +168,10 @@ int limit_number_of_threads( int max_threads ) {
 
 #endif // __TBB_TEST_SKIP_AFFINITY
 
-#define OS_AFFINITY_SYSCALL_PRESENT ((__linux__ && !__ANDROID__) || (__FreeBSD_version >= 701000))
+#define OS_AFFINITY_SYSCALL_PRESENT (__linux__ && !__ANDROID__)
 
 #if OS_AFFINITY_SYSCALL_PRESENT
+
```
ivankochin, line 174:
```suggestion
#if OS_AFFINITY_SYSCALL_PRESENT
```

test/common/utils_concurrency_limit.h
```diff
@@ -168,9 +168,10 @@ int limit_number_of_threads( int max_threads ) {
 
 #endif // __TBB_TEST_SKIP_AFFINITY
 
-#define OS_AFFINITY_SYSCALL_PRESENT ((__linux__ && !__ANDROID__) || (__FreeBSD_version >= 701000))
+#define OS_AFFINITY_SYSCALL_PRESENT (__linux__ && !__ANDROID__)
```
ivankochin, line 171:
```suggestion
// TODO: consider using cpuset_setaffinity/sched_getaffinity on FreeBSD to enable the functionality
#define OS_AFFINITY_SYSCALL_PRESENT (__linux__ && !__ANDROID__)
```

test/common/utils_concurrency_limit.h
```diff
@@ -168,9 +168,10 @@ int limit_number_of_threads( int max_threads ) {
...
 
 #if OS_AFFINITY_SYSCALL_PRESENT
+
 void get_thread_affinity_mask(std::size_t& ncpus, std::vector<int>& free_indexes) {
```
ivankochin, line 175:
```suggestion
#if OS_AFFINITY_SYSCALL_PRESENT
void get_thread_affinity_mask(std::size_t& ncpus, std::vector<int>& free_indexes) {
```

alexey-katranov (approved):

ivankochin:
@martymac, the patch was merged to master.
There is still an opportunity to enable this part of testing on FreeBSD (see [TODO](https://github.com/oneapi-src/oneTBB/pull/425/files#diff-9cb6b25d2378b3e543b66036a9fd57e1cd67bc0b2f9c9e0e5953d589df1faac4R171)), but we have no FreeBSD system for testing on. So we are open to contributions.

martymac:
@ivankochin Thanks!

PR #426, opened 2021-06-02
Solve the issue with different incompatible HWLOC library versions in one executable

ivankochin:
HWLOC library has two incompatible versions: 1.x and 2.x. But the part of C-style entry points in both of these versions have the same names which cause the following issue: If the application links with one HWLOC version and oneTBB loads the incompatible HWLOC library version using `dlopen()`, then the HWLOC shared object that was shared later will use some entry points from the already loaded version. That's because oneTBB loads the HWLOC shared object with the `RTLD_GLOBAL`  flag.

This patch fixes this issue by loading the HWLOC with the right flags (`RTLD_LOCAL` and `RTLD_DEEPBIND`) which places the symbols into the local scope and use only loaded symbols.

### DIFF
--- a/src/tbb/dynamic_link.cpp
+++ b/src/tbb/dynamic_link.cpp
@@ -409,9 +409,9 @@ namespace r1 {
     }
 
 #if !_WIN32
-    int define_flags(bool is_local) {
+    int loading_flags(bool local_binding) {
         int flags = RTLD_NOW;
-        if (is_local) {
+        if (local_binding) {
             flags = flags | RTLD_LOCAL;
 #if __linux__ && !__ANDROID__
             flags = flags | RTLD_DEEPBIND;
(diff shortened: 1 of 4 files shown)


src/tbb/dynamic_link.cpp
```diff
@@ -448,8 +464,8 @@ namespace r1 {
         // TODO: May global_symbols_link find weak symbols?
         dynamic_link_handle library_handle = ( flags & DYNAMIC_LINK_GLOBAL ) ? global_symbols_link( library, descriptors, required ) : 0;
 
-        if ( !library_handle && ( flags & DYNAMIC_LINK_LOAD ) )
-            library_handle = dynamic_load( library, descriptors, required );
+        if ( !library_handle && ( ( flags & DYNAMIC_LINK_LOAD ) || ( flags & DYNAMIC_LINK_LOCAL_BINDING ) ) )
```
alexey-katranov, line 467:
```suggestion
        if ( !library_handle && ( flags & DYNAMIC_LINK_LOAD ) )
```

src/tbb/governor.cpp
```diff
@@ -390,7 +390,7 @@ const char* load_tbbbind_shared_object() {
     if (si.dwNumberOfProcessors > 32) return nullptr;
 #endif /* _WIN32 && !_WIN64 */
     for (const auto& tbbbind_version : {TBBBIND_2_4_NAME, TBBBIND_2_0_NAME, TBBBIND_NAME}) {
-        if (dynamic_link(tbbbind_version, TbbBindLinkTable, LinkTableSize)) {
+        if (dynamic_link(tbbbind_version, TbbBindLinkTable, LinkTableSize, nullptr, DYNAMIC_LINK_LOCAL_BINDING)) {
```
alexey-katranov, line 393:
```suggestion
        if (dynamic_link(tbbbind_version, TbbBindLinkTable, LinkTableSize, nullptr, DYNAMIC_LINK_DEFAULT | DYNAMIC_LINK_LOCAL_BINDING)) {
```

ivankochin: The approach was discussed offline and we agreed on different option.

alexey-katranov (approved):

PR #427, opened 2021-06-03
Improve wakeup signal propagation in private server

pavelkumbrasev:
Decreased size of critical section on hot path in private server.  Factor out  slack allocation from the critical section to wake up threads faster.

Signed-off-by: pkumbras <pavel.kumbrasev@intel.com>

### DIFF
--- a/src/tbb/private_server.cpp
+++ b/src/tbb/private_server.cpp
@@ -155,7 +155,7 @@ private:
         which in turn each wake up two threads, etc. */
     void propagate_chain_reaction() {
         // First test of a double-check idiom.  Second test is inside wake_some(0).
-        if( my_asleep_list_root.load(std::memory_order_acquire) )
+        if( my_asleep_list_root.load(std::memory_order_relaxed) )
             wake_some(0);
     }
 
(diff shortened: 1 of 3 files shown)


src/tbb/private_server.cpp
```diff
@@ -351,49 +352,58 @@ inline bool private_server::try_insert_in_asleep_list( private_worker& t ) {
...
-                do {
-                    if( old<=0 ) goto done;
-                } while( !my_slack.compare_exchange_strong(old,old-1) );
-            }
+
+        while( my_asleep_list_root.load(std::memory_order_relaxed) && w<wakee+2 && allotted_slack) {
+            --allotted_slack;
             // Pop sleeping worker to combine with claimed unit of slack
             auto old = my_asleep_list_root.load(std::memory_order_relaxed);
             my_asleep_list_root.store(old->my_next, std::memory_order_relaxed);
```
aleksei-fedotov, line 396: I believe we can be even better here by saving one more store to `my_asleep_list_root`, that is update it only once, despite the value of `allotted_slack`.

src/tbb/private_server.cpp
```diff
@@ -351,49 +352,58 @@ inline bool private_server::try_insert_in_asleep_list( private_worker& t ) {
...
+
+    int allotted_slack = 0;
+    while (allotted_slack < 2) {
+        // Chain reaction; Try to claim unit of slack
+        int old = my_slack.load(std::memory_order_relaxed);
+        do {
+            if (old <= 0) goto done;
+        } while (!my_slack.compare_exchange_strong(old, old - 1));
+        ++allotted_slack;
+    }
```
aleksei-fedotov, line 385: Since `my_slack` is an atomic, can we save some updates of it by draining `additional_slack` out first?

aleksei-fedotov (approved):
A couple of more ideas how to improve the waking signal even more. Feel free to experiment with in this or the other patch.

src/tbb/private_server.cpp
pavelkumbrasev: This optimization will lead to logical race, between workers recall from market `adjust_job_count_estimate` with negative delta and `wake_some` with positive delta. Negative and positive delta can be combined  into zero or negative slack. However with the proposed optimization we still wake one or two threads while no slack.

src/tbb/private_server.cpp
alexey-katranov: I think you will still need a store to a local variable (maybe optimized into a register) to go through `old->my_next`. So, the difference is only between the store in the register + store to memory or two stores into the same memory. I would not expect notable performance difference because CPU is quite smart to optimize the stores into the same memory location.

PR #431, opened 2021-06-07
Fix leaking proxy task objects

alexey-katranov:
It seems during revamp we forget to add clean up logic for proxy task when destroying arena.

Signed-off-by: Alexei Katranov <alexei.katranov@intel.com>

### DIFF
--- a/test/tbb/test_task_arena.cpp
+++ b/test/tbb/test_task_arena.cpp
@@ -1644,7 +1644,7 @@ void StressTestMixFunctionality() {
 
                     break;
                 }
-                case detach_observer :
+                case detach_observer:
                 {
                     auto arena_number = get_random_arena() % arenas_pool.size();
                     auto curr_arena = arenas_pool.begin();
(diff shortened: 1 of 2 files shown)


test/tbb/test_task_arena.cpp
```diff
@@ -1668,15 +1668,18 @@ void StressTestMixFunctionality() {
...
-                        });
+                            for (; sum < 10; ++sum);
+                        };
+                        tbb::parallel_for(tbb::blocked_range<std::size_t>(0, 10000), body, tbb::auto_partitioner{});
+                        tbb::parallel_for(tbb::blocked_range<std::size_t>(0, 10000), body, aff);
                     });
 
                     break;
-                }
+                }
```
alexey-katranov, line 1682:
```suggestion
                }
```

test/tbb/test_task_arena.cpp
```diff
@@ -1668,11 +1668,14 @@ void StressTestMixFunctionality() {
...
 
                     curr_arena->arena->execute([] () {
-                        tbb::parallel_for(tbb::blocked_range<std::size_t>(0, 10000), [] (tbb::blocked_range<std::size_t>&) {
+                        static tbb::affinity_partitioner aff;
+                        auto body = [](tbb::blocked_range<std::size_t>&) {
                             std::atomic<int> sum{};
                             // Make some work
-                            for (; sum < 10; ++sum) ;
-                        });
+                            for (; sum < 10; ++sum);
```
aleksei-fedotov, line 1675: Replace with DummyWork from utils?

test/tbb/test_task_arena.cpp
```diff
@@ -1668,11 +1668,14 @@ void StressTestMixFunctionality() {
...
+                        };
+                        tbb::parallel_for(tbb::blocked_range<std::size_t>(0, 10000), body, tbb::auto_partitioner{});
+                        tbb::parallel_for(tbb::blocked_range<std::size_t>(0, 10000), body, aff);
```
aleksei-fedotov, line 1678: Add the use case with `static_partitioner` as well?

aleksei-fedotov (approved):

test/tbb/test_task_arena.cpp
alexey-katranov: Can you please describe the supposed reason? The affinity and static partitioners use the same mechanism of affinity tasks that we want to test. So, it seems there is no value to increase testing because it will test the same stuff.

aleksei-fedotov: > The affinity and static partitioners use the same mechanism of affinity tasks that we want to test

Well, not exactly the same - it is a bit different, otherwise there won't be different partitioners. Moreover, consider if one day the mechanism diverges more in these partitioners while the tests remain intact. We won't notice there is a problem with management of the tasks in the new mechanism, because I don't believe someone will recall to change this test accordingly.

pavelkumbrasev (approved):
In this patch we test the leak in proxy tasks not partitioners.
Consider write better test for partitioners(about leaks).
LGTM

PR #432, opened 2021-06-08
Enable intermodular optimization during the compilation

ivankochin:
Intermodular (or Link Time) optimizations may be used to reduce binaries' size and improve the performance of the libraries. It achieves by removing some parts of unused code from the binary during the linkage stage.  This patch enables such optimizations on different compilers.

Additional materials:
- [LLVM LTO documentation](https://llvm.org/docs/LinkTimeOptimization.html)
- [GCC LTO documentation](https://gcc.gnu.org/wiki/LinkTimeOptimization)
- [Intel IPO documentation](https://software.intel.com/content/www/us/en/develop/documentation/cpp-compiler-developer-guide-and-reference/top/optimization-and-programming-guide/interprocedural-optimization-ipo.html)


cmake/compilers/Clang.cmake
```diff
@@ -48,5 +48,7 @@ if (ANDROID_PLATFORM)
     set(TBB_COMMON_COMPILE_FLAGS ${TBB_COMMON_COMPILE_FLAGS} $<$<NOT:$<CONFIG:Debug>>:-D_FORTIFY_SOURCE=2>)
 endif()
 
+set(TBB_IPO_LINK_FLAG -flto)
```
isaevil, line 51: Doesn't `Clang` require `-flto` flag during the compile stage too?

src/tbb/CMakeLists.txt
```diff
@@ -88,6 +89,11 @@ set_target_properties(tbb PROPERTIES
     LINK_DEPENDS ${CMAKE_CURRENT_SOURCE_DIR}/def/${TBB_DEF_FILE_PREFIX}-tbb.def
 )
 
+set_property(TARGET tbb APPEND PROPERTY LINK_FLAGS_RELWITHDEBINFO ${TBB_IPO_LINK_FLAG})
+if (WIN32)
+    set_property(TARGET tbb APPEND PROPERTY LINK_FLAGS_RELWITHDEBINFO /INCREMENTAL:NO)
+endif()
+
```
isaevil, line 96: Could this property (`LINK_FLAGS_RELWITHDEBINFO`) be added inside `set_target_properties`? Is there a possibility to use [cmake-generator-expressions](https://cmake.org/cmake/help/latest/manual/cmake-generator-expressions.7.html) here to shrink this into one line?

cmake/compilers/Clang.cmake
ivankochin: Yes, it does, thanks!

src/tbb/CMakeLists.txt
ivankochin: I tried several approaches but cannot find a way how to do it using CMake generator expressions. And also I think that the code without generator expressions looks more simple and straightforward.

isaevil: Ok, I see. Maybe it makes sense to move general part (92 line) inside `set_target_properties`.
Is it OK that we use the `-flto` flag whether we are compiling in `debug` or `relwithdebinfo`, but adding to link flags only while compiling in `relwithdebinfo`?

pavelkumbrasev (dismissed):
The changes seem ok to me.

alexey-katranov:
Did you consider [CMAKE INTERPROCEDURAL_OPTIMIZATION](https://cmake.org/cmake/help/v3.20/prop_tgt/INTERPROCEDURAL_OPTIMIZATION.html)?

ivankochin:
> Did you consider [CMAKE INTERPROCEDURAL_OPTIMIZATION](https://cmake.org/cmake/help/v3.20/prop_tgt/INTERPROCEDURAL_OPTIMIZATION.html)?

Yes, I did, but our CMake configs require CMake 3.2 minimal version and this CMake version does not contain an interface for checking the IPO availability ([CheckIPOSupported](https://cmake.org/cmake/help/latest/module/CheckIPOSupported.html#module:CheckIPOSupported)), so just specifying the `INTERPROCEDURAL_OPTIMIZATION` causes warnings that should be suppressed. So at old CMake versions, we cannot know was optimization applied or not.

Also, the `INTERPROCEDURAL_OPTIMIZATION` adds corresponding compilation flags not to all configurations. For example, the compilers which are used for our official packages compile without LTO flags even if `INTERPROCEDURAL_OPTIMIZATION` is specified.

src/tbb/CMakeLists.txt
ivankochin: @isaevil, I reworked the whole CMake part to enable such optimization for all configurations. Please review one more time.

src/tbbmalloc/CMakeLists.txt
```diff
@@ -73,19 +74,26 @@ set_target_properties(tbbmalloc PROPERTIES
 
 set(CMAKE_CXX_IMPLICIT_LINK_LIBRARIES "")
 
+set_property(TARGET tbbmalloc APPEND PROPERTY LINK_FLAGS_RELWITHDEBINFO )
```
isaevil, line 77: Is this line still relevant?

ivankochin: No, I forgot to remove this code, thanks for the comment!

isaevil (dismissed):
Changes LGTM!

pavelkumbrasev (approved):

PR #438, opened 2021-06-08
Use mallinfo2 API where possible

kraj:
glibc 2.33+ has introduced this API and actively deprecated the old
mallinfo API, therefore it will be prudent to enquire the platform for
what it can support and conditionalize the code accordingly with
preference to use mallinfo2 if available.

Signed-off-by: Khem Raj <raj.khem@gmail.com>

### DIFF
--- a/src/tbbmalloc_proxy/def/lin64-proxy.def
+++ b/src/tbbmalloc_proxy/def/lin64-proxy.def
@@ -26,6 +26,7 @@ aligned_alloc;
 valloc;
 pvalloc;
 mallinfo;
+mallinfo2;
 mallopt;
 malloc_usable_size;
 __libc_malloc;
--- a/src/tbbmalloc_proxy/proxy.cpp
+++ b/src/tbbmalloc_proxy/proxy.cpp
@@ -256,10 +256,16 @@ int mallopt(int /*param*/, int /*value*/) __THROW
     return 1;
 }
 
-struct mallinfo mallinfo() __THROW
+#ifdef HAVE_MALLINFO2
+#define MALLINFO mallinfo2
+#else
+#define MALLINFO mallinfo
+#endif
+
+struct MALLINFO MALLINFO() __THROW
 {
-    struct mallinfo m;
-    memset(&m, 0, sizeof(struct mallinfo));
+    struct MALLINFO m;
+    memset(&m, 0, sizeof(struct MALLINFO));
 
     return m;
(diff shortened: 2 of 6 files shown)


src/tbbmalloc_proxy/proxy.cpp
```diff
@@ -256,10 +256,16 @@ int mallopt(int /*param*/, int /*value*/) __THROW
...
+
+struct MALLINFO MALLINFO() __THROW
 {
-    struct mallinfo m;
-    memset(&m, 0, sizeof(struct mallinfo));
+    struct MALLINFO m;
+    memset(&m, 0, sizeof(struct MALLINFO));
 
     return m;
 }
```
alexey-katranov, line 271: I believe tbbmalloc_proxy should export the both `mallinfo` and `mallinfo2` to support the old and new code.

src/tbbmalloc_proxy/def/lin64-proxy.def
```diff
@@ -26,6 +26,7 @@ aligned_alloc;
 valloc;
 pvalloc;
 mallinfo;
+mallinfo2;
```
alexey-katranov, line 29: Will not it break the code if tbbmalloc_proxy is built with old libc but used on the system with new libc? What the `mallinfo2` symbol will be defined in with old libc?

test/tbbmalloc/test_malloc_overload.cpp
```diff
@@ -448,7 +448,11 @@ TEST_CASE("Main set of tests") {
     CheckMemalignFuncOverload(aligned_alloc, free);
 #endif
 
+#ifdef HAVE_MALLINFO2
+    struct mallinfo2 info = mallinfo2();
+#else
     struct mallinfo info = mallinfo();
+#endif
```
alexey-katranov, line 455: The both function should be replaced and tested.

anton-potapov:
@kraj, friendly reminder

test/tbbmalloc/test_malloc_overload.cpp
alexey-katranov: I think we need to test `mallinfo` always and `mallinfo2` when availble, e.g.
```suggestion
#ifdef HAVE_MALLINFO2
    struct mallinfo2 info = mallinfo2();
#endif
    struct mallinfo info = mallinfo();
```

PR #440, opened 2021-06-08
Add support for QNX Neutrino

### LINKED ISSUE #439: Add support for QNX Neutrino

pablorcum:
Fixes #439 .
Extends support for other Unix non-Linux systems.


pablorcum:
I have fixed the issue causing the Linux workflows to fail. Please rerun them.

src/tbb/allocator.cpp
```diff
@@ -97,7 +97,7 @@ static const dynamic_link_descriptor MallocLinkTable[] = {
 #define MALLOCLIB_NAME "libtbbmalloc" DEBUG_SUFFIX ".dylib"
 #elif __FreeBSD__ || __NetBSD__ || __OpenBSD__ || __sun || _AIX || __ANDROID__
 #define MALLOCLIB_NAME "libtbbmalloc" DEBUG_SUFFIX ".so"
-#elif __linux__  // Note that order of these #elif's is important!
+#elif __unix__  // Note that order of these #elif's is important!
 #define MALLOCLIB_NAME "libtbbmalloc" DEBUG_SUFFIX ".so.2"
```
alexey-katranov, line 101: I think it should be similar to FreeBSD because we do not manage soname on non-Linux platforms, e.g.
```suggestion
#elif __linux__  // Note that order of these #elif's is important!
#define MALLOCLIB_NAME "libtbbmalloc" DEBUG_SUFFIX ".so.2"
#elif __unix__
#define MALLOCLIB_NAME "libtbbmalloc" DEBUG_SUFFIX ".so"
```
Pay attention to other places as well.

src/tbb/misc.h
```diff
@@ -25,7 +25,7 @@
 #include "oneapi/tbb/info.h"
 #endif /*__TBB_ARENA_BINDING*/
 
-#if __linux__ || __FreeBSD__
+#if __unix__ || __FreeBSD__
```
alexey-katranov, line 28: I think FreeBSD defines unix, i.e.
```suggestion
#if __unix__
```
Check other places as well

src/tbb/misc_ex.cpp
```diff
@@ -76,7 +78,7 @@ static void set_thread_affinity_mask( std::size_t maskSize, const basic_mask_t*
 }
 
 static void get_thread_affinity_mask( std::size_t maskSize, basic_mask_t* threadMask ) {
-#if __linux__
+#if __unix__
     if( sched_getaffinity( 0, maskSize, threadMask ) )
 #else /* FreeBSD */
     if( cpuset_getaffinity( CPU_LEVEL_WHICH, CPU_WHICH_TID, -1, maskSize, threadMask ) )
```
alexey-katranov, line 84: AFAIK, FreeBSD does not support `sched_getaffinity` but defines `__unix__`. Does it really works on arbitrary Unix? The affinity API is not part of POSIX.

src/tbb/misc.h
pablorcum: FreeBSD does not define unix as far as I've just checked. On the other hand, unix defines linux.

src/tbb/allocator.cpp
pablorcum: Soname is available in unix platforms, but not in FreeBSD.

src/tbb/misc_ex.cpp
pablorcum: As said in the previous comment, FreeBSD does not define unix.

src/tbb/misc.h
alexey-katranov: It seems I am outdated, on FreeBSD 11 with clang 3.8.0, I have the following output
```shell
$clang++ -dM -E -x c++ /dev/null | grep unix
#define __unix 1
#define __unix__ 1
#define unix 1
```
Was something changed in newer versions?

pablorcum: In other words with the compiler macro __unix__, you'll capture any unix system including FreeBSD and QNX. However, only using the compiler macro __FreeBSD__ leaves out other unix systems like QNX.

src/tbb/misc_ex.cpp
ivankochin: @pablorcum 
According to [this page](https://forums.freebsd.org/threads/gcc-general-macro-for-bsd.47769/), the `__unix__` macro is defined on FreeBSD systems. Unfortunately, I have no available FreeBSD machine to test this statement.

Maybe the `__linux__ && __QNXNTO__` condition can be used instead of `__unix__`?

pablorcum: Yes, that's also possible. I don't have anything against that. My personal preference is to keep it as general and reusable as possible, therefore I tend to use __unix__ instead of __linux__ in my other projects, but we can keep __linux__ || __QNXNTO__ for the sake of backwards compatibility.

alexey-katranov: @ivankochin , are you sure that it is a good idea to limit the patch for QNX only? I suppose enabling Unix in generic way and implementing some exclusions (e.g. for FreeBSD) is the good approach for other Unix-like platforms.

ivankochin: @alexey-katranov, Ok, then we may use a common approach with `__unix__` everywhere except the affinity interfaces case. Do you know more differences between FreeBSD and common POSIX interfaces?

alexey-katranov: As I understand, there is no affinity interfaces in POSIX. So, for each platform (or group of platforms) it should be implemented in its own way (e.g. disabled if possible).

alexey-katranov:
@pablorcum , can you please rebase the patch on the latest master that we can run internal CI?

pablorcum:
@alexey-katranov Rebased and tested that builds for QNX.

src/tbb/misc_ex.cpp
alexey-katranov: Swap Unix and *BSD blocks:
```suggestion
#if __FreeBSD__ || __NetBSD__ || __OpenBSD__
    if( cpuset_getaffinity( CPU_LEVEL_WHICH, CPU_WHICH_TID, -1, maskSize, threadMask ) )
#else /* __unix__ */
    if( sched_getaffinity( 0, maskSize, threadMask ) )
```

src/tbb/semaphore.h
```diff
@@ -35,24 +35,26 @@
 
 #include <atomic>
 
-#if __linux__ || __FreeBSD__ || __NetBSD__ || __OpenBSD__
+#if __unix__ || __FreeBSD__ || __NetBSD__ || __OpenBSD__
+
```
alexey-katranov, line 39:
```suggestion
#if __unix__
```

src/tbb/semaphore.h
```diff
@@ -87,7 +89,7 @@ the actual parameter values to match Linux: 0 for wait, 1 for wake.
 #endif
 
 #endif // SYS_futex
-#endif // __linux__ || __FreeBSD__ || __NetBSD__ || __OpenBSD__
+#endif // __unix__ || __FreeBSD__ || __NetBSD__ || __OpenBSD__
```
alexey-katranov, line 92:
```suggestion
#endif // __unix__
```

src/tbbmalloc/frontend.cpp
```diff
@@ -775,7 +775,7 @@ static inline unsigned int highestBitPos(unsigned int n)
     unsigned int pos;
 #if __ARCH_x86_32||__ARCH_x86_64
 
-# if __linux__||__APPLE__||__FreeBSD__||__NetBSD__||__OpenBSD__||__sun||__MINGW32__
+# if __unix__||__APPLE__||__FreeBSD__||__NetBSD__||__OpenBSD__||__sun||__MINGW32__
```
alexey-katranov, line 778:
```suggestion
# if __unix__||__APPLE__||__MINGW32__
```

src/tbbmalloc_proxy/proxy.h
```diff
@@ -22,7 +22,7 @@
 
 // MALLOC_UNIXLIKE_OVERLOAD_ENABLED depends on MALLOC_CHECK_RECURSION stuff
 // TODO: limit MALLOC_CHECK_RECURSION to *_OVERLOAD_ENABLED only
-#if __linux__ || __APPLE__ || __sun || __FreeBSD__ || MALLOC_UNIXLIKE_OVERLOAD_ENABLED
+#if __unix__ || __APPLE__ || __sun || __FreeBSD__ || MALLOC_UNIXLIKE_OVERLOAD_ENABLED
```
alexey-katranov, line 25:
```suggestion
#if __unix__ || __APPLE__ || MALLOC_UNIXLIKE_OVERLOAD_ENABLED
```

test/common/memory_usage.h
```diff
@@ -30,7 +30,7 @@
 #include <sanitizer/allocator_interface.h>
 #endif
 
-#if __linux__ || __sun
+#if __unix__ || __sun
```
alexey-katranov, line 33:
```suggestion
#if __unix__
```

src/tbb/rml_tbb.cpp
```diff
@@ -50,7 +50,7 @@ namespace rml {
 #define RML_SERVER_NAME "irml" DEBUG_SUFFIX ".dll"
 #elif __APPLE__
 #define RML_SERVER_NAME "libirml" DEBUG_SUFFIX ".dylib"
-#elif __linux__
+#elif __unix__
 #define RML_SERVER_NAME "libirml" DEBUG_SUFFIX ".so.1"
 #elif __FreeBSD__ || __NetBSD__ || __OpenBSD__ || __sun || _AIX
 #define RML_SERVER_NAME "libirml" DEBUG_SUFFIX ".so"
```
alexey-katranov, line 56:
```suggestion
#elif __FreeBSD__ || __NetBSD__ || __OpenBSD__ || __sun || _AIX
#define RML_SERVER_NAME "libirml" DEBUG_SUFFIX ".so"
#elif __unix__
#define RML_SERVER_NAME "libirml" DEBUG_SUFFIX ".so.1"
```

alexey-katranov (commented):
Everything seems good. Can you please apply minor polishing (in some places FreeBSD should be check before Unix)?

pablorcum:
Suggested changes committed and validated build again on QNX.

src/tbb/misc_ex.cpp
```diff
@@ -63,7 +65,7 @@ static const dynamic_link_descriptor iompLinkTable[] = {
 #endif
 
 static void set_thread_affinity_mask( std::size_t maskSize, const basic_mask_t* threadMask ) {
-#if __linux__
+#if __unix__
     if( sched_setaffinity( 0, maskSize, threadMask ) )
 #else /* FreeBSD */
     if( cpuset_setaffinity( CPU_LEVEL_WHICH, CPU_WHICH_TID, -1, maskSize, threadMask ) )
```
alexey-katranov, line 71:
```suggestion
#if __FreeBSD__ || __NetBSD__ || __OpenBSD__
    if( sched_setaffinity( 0, maskSize, threadMask ) )
#else /* __unix__ */
    if( cpuset_setaffinity( CPU_LEVEL_WHICH, CPU_WHICH_TID, -1, maskSize, threadMask ) )
```

src/tbb/misc_ex.cpp
```diff
@@ -135,7 +137,7 @@ static void initialize_hardware_concurrency_info () {
     int err;
     int availableProcs = 0;
     int numMasks = 1;
-#if __linux__
+#if __unix__
```
alexey-katranov, line 140: For some reason GitHub does not allow multiline comment with untouched code. Swap the Unix and *BSD blocks here as well.

src/tbb/misc_ex.cpp
```diff
@@ -147,7 +149,7 @@ static void initialize_hardware_concurrency_info () {
         const int curMaskSize = BasicMaskSize * numMasks;
         processMask = new basic_mask_t[numMasks];
         std::memset( processMask, 0, curMaskSize );
-#if __linux__
+#if __unix__
```
alexey-katranov, line 152: For some reason GitHub does not allow multiline comment with untouched code. Swap the Unix and *BSD blocks here as well.

alexey-katranov (changes requested):
Sorry I missed several places of Unix/*BSD macro blocks ordering

src/tbb/misc_ex.cpp
pablorcum: If I swap this, then unix would be the default case. Is that ok?

src/tbb/misc_ex.cpp
pablorcum: If I swap this, then unix would be the default case. Is that ok?

src/tbb/misc_ex.cpp
alexey-katranov: It is acceptable, currently, the main issue that *BSD system will use __unix__ branch that will not work

src/tbb/misc_ex.cpp
alexey-katranov: It is acceptable, currently, the main issue that *BSD system will use __unix__ branch that will not work

src/tbb/misc_ex.cpp
pablorcum: I have applied the suggestion including a minor simplification.

src/tbb/misc_ex.cpp
pablorcum: I have applied the suggestion including a minor simplification.

src/tbb/misc_ex.cpp
alexey-katranov: The internal CI found that my suggestion was incorrect. Lines 70 and line 72 are also should be swapped...

pablorcum: Your suggestion is ok. I forgot to change one thing in the first CI try. I fixed it later and the last CI run is succesful.

alexey-katranov:
Just minor issues from internal CI. You just need to copy past copyright from other file and add a new line at the end in QCC.cmake. All other stuff is ok.

1. In file onetbb_source_code/cmake/compilers/QCC.cmake wrong copyright years or missing copyrights
2. QCC.cmake:
```
@@ -1,4 +1,3 @@
 include(${CMAKE_CURRENT_SOURCE_DIR}/cmake/compilers/GNU.cmake)
 
 # Remove dl library not present in QNX systems
-unset(TBB_COMMON_LINK_LIBS)
\ No newline at end of file
```

pablorcum:
I have updated QCC.cmake according to the comments.

alexey-katranov (approved):
Thank you a lot for your contribution!

PR #443, opened 2021-06-15
Fix failed mips debug build

Iliamish:


test/tbb/test_join_node.h
```diff
@@ -26,6 +26,12 @@
 #endif
 #endif
 
+#ifdef __MIPS
+    #define MAX_TUPLE_TEST_SIZE_LOCAL 3
+#else
+    #define MAX_TUPLE_TEST_SIZE_LOCAL MAX_TUPLE_TEST_SIZE
+#endif
+
```
alexey-katranov, line 34: We usually use `TBB_TEST_LOW_WORKLOAD` macro for similar things.

ivankochin: Agree, look here please, this macro allows to reduce the amount of testing on MIPS and it already set: https://github.com/oneapi-src/oneTBB/blob/master/cmake/compilers/GNU.cmake#L54

test/tbb/test_join_node.h
```diff
@@ -26,6 +26,12 @@
 #endif
 #endif
 
+#ifdef __MIPS
+    #define MAX_TUPLE_TEST_SIZE_LOCAL 3
+#else
+    #define MAX_TUPLE_TEST_SIZE_LOCAL MAX_TUPLE_TEST_SIZE
+#endif
```
ivankochin, line 33: I would propose to redefine the `MAX_TUPLE_TEST_SIZE` instead of introducing new macro?
```suggestion
#if TBB_TEST_LOW_WORKLOAD && defined(MAX_TUPLE_TEST_SIZE)
    #undef MAX_TUPLE_TEST_SIZE
    #define MAX_TUPLE_TEST_SIZE 3
#endif
```

ivankochin (approved):

PR #447, opened 2021-06-18
Fix GCC 11 warning

anton-potapov:
gcc 11 issue a warning, about possibly incorrect destructor being called for `dummy` node: 
`onetbb/src/tbb/../../include/tbb/../oneapi/tbb/detail/_concurrent_unordered_base.h:172:20: error: array subscript ‘tbb::detail::d1::value_node<AllocatorAwareData<std::scoped_allocator_adaptor<std::allocator<int> > >, long unsigned int>[0]’ is partly outside array bounds of ‘unsigned char [16]’ [-Werror=array-bounds]
  172 |     ~value_node() {}
      |                    ^
compilation terminated due to -Wfatal-errors.`

### DIFF
--- a/include/oneapi/tbb/detail/_concurrent_unordered_base.h
+++ b/include/oneapi/tbb/detail/_concurrent_unordered_base.h
@@ -910,6 +910,10 @@ private:
             // Deallocate the memory
             node_allocator_traits::deallocate(dummy_node_allocator, node, 1);
         } else {
+            // GCC 11.1 issues a warning here that incorrect destructor might be called for dummy_nodes
+            #if (__TBB_GCC_VERSION >= 110100 && __TBB_GCC_VERSION < 120000 ) && !__clang__ && !__INTEL_COMPILER
+            volatile
+            #endif
             value_node_ptr val_node = static_cast<value_node_ptr>(node);
             value_node_allocator_type value_node_allocator(my_segments.get_allocator());
             // Destroy the value


alexey-katranov:
Does it make sense to report the issue to GCC? Or similar issue is already opened?

include/oneapi/tbb/detail/_concurrent_unordered_base.h
```diff
@@ -910,6 +910,10 @@ class concurrent_unordered_base {
             // Deallocate the memory
             node_allocator_traits::deallocate(dummy_node_allocator, node, 1);
         } else {
+            // GCC 11.1 issues a warning here that incorrect destructor might be called for dummy_nodes
+            #if (__TBB_GCC_VERSION >= 110100 && __TBB_GCC_VERSION < 110200 ) && !__clang__ && !__INTEL_COMPILER
```
pavelkumbrasev, line 914: Did GCC disable this option by default from 110200? Or are you going to update this file each GCC update?

anton-potapov: Updated to 12.0 to recheck it automatically when gcc 12.0 arrives

anton-potapov:
> 
> 
> Does it make sense to report the issue to GCC? Or similar issue is already opened?

it does - At the moment trying to create minimal reproducer

alexey-katranov (approved):

anton-potapov:
rechecked with gcc 11.1, the patch is still makes sense

PR #448, opened 2021-06-21
Upgrade from HWLOC 2.4 to HWLOC 2.5

ivankochin:
The HWLOC 2.5 release was published. This release contains the topology flag that allows does not cross the process binding borders. Also, this is the first major HWLOC release with a fix for Hybrid CPUs parsing on Linux. Due to these facts, it was decided to remove the `tbbbind_2_4` library and introduce a new `tbbbind_2_5` library version. Please ask if you need additional information regards the motivation for this change.

Changes applied by this patch:
* Introduce tbbbind_2_5 library
* Remove tbbbind_2_4 library
* Use the HWLOC_TOPOLOGY_FLAG_RESTRICT_TO_CPUBINDING topology flag (introduces as a part of HWLOC 2.5) to restrict topology discovery to process binding


CMakeLists.txt
```diff
@@ -100,6 +100,7 @@ option(TBB_NO_APPCONTAINER "Apply /APPCONTAINER:NO (for testing binaries for Win
 option(TBB4PY_BUILD "Enable tbb4py build" OFF)
 option(TBB_CPF "Enable preview features of the library" OFF)
 option(TBB_FIND_PACKAGE "Enable search for external oneTBB using find_package instead of build from sources" OFF)
+option(TBB_TBBBIND_BUILD "Enable building TBBbind libraries" ON)
```
aleksei-fedotov, line 103: Add this option conditionall? For example,
```cmake
if (APPLE)
    message(STATUS "TBBBind build targets are disabled due to unsupported environment")
    set(TBB_TBBBIND_BUILD FALSE)
else()
    option(TBB_TBBBIND_BUILD "Enable building TBBbind libraries" ON)
endif()
```
This will allow not having any messages below, but have simple thing instead:
```cmake
if (TBB_TBBBIND_BUILD)
    add_subdirectory(src/tbbbind)
endif()
```
What do you think about this approach?

src/tbbbind/tbb_bind.cpp
```diff
@@ -85,6 +86,16 @@ class platform_topology {
...
+            #if __HWLOC_RESTRICT_TO_CPUBINDING_TOPOLOGY_FLAG_PRESENT
+            if ( groups_num == 1 &&
+                 hwloc_topology_set_flags(topology,
+                     HWLOC_TOPOLOGY_FLAG_IS_THISSYSTEM |
+                     HWLOC_TOPOLOGY_FLAG_RESTRICT_TO_CPUBINDING
+                 ) != 0
+            ) {
+                return;
+            }
+            #endif
```
aleksei-fedotov, line 98: It seems that we can avoid introducing new macro by having something like:
```suggestion
            if ( groups_num == 1 &&
                 hwloc_topology_set_flags(topology,
                     HWLOC_TOPOLOGY_FLAG_IS_THISSYSTEM 
#if HWLOC_TOPOLOGY_FLAG_RESTRICT_TO_CPUBINDING
                     | HWLOC_TOPOLOGY_FLAG_RESTRICT_TO_CPUBINDING
#endif
                 ) != 0
            ) {
                return;
            }
```
What do you think?

test/common/common_arena_constraints.h
```diff
@@ -69,6 +86,7 @@ int get_processors_groups_count() { return 1; }
 #endif
 
 #define __HWLOC_HYBRID_CPUS_INTERFACES_PRESENT (HWLOC_API_VERSION >= 0x20400)
+#define __HWLOC_RESTRICT_TO_CPUBINDING_TOPOLOGY_FLAG_PRESENT (HWLOC_API_VERSION >= 0x20500)
```
aleksei-fedotov, line 89: Not sure I understand why we need it defined here in the test. I can't find where it is used.

test/common/common_arena_constraints.h
```diff
@@ -153,6 +171,15 @@ class system_info {
 
     system_info() {
         hwloc_require_ex(hwloc_topology_init, &topology);
+#if __HWLOC_WINDOWS_AFFINITY_RESPECTING_INTERFACES_PRESENT
```
aleksei-fedotov, line 174: Looks confusing. Why WINDOWS specific? Can you please add a comment in the code what this macros is about and why it needs to be used instead of `__HWLOC_RESTRICT_TO_CPUBINDING_TOPOLOGY_FLAG_PRESENT`?

test/common/common_arena_constraints.h
```diff
@@ -38,7 +38,24 @@
 
 #if (_WIN32 || _WIN64) && __TBB_HWLOC_VALID_ENVIRONMENT
 #include <windows.h>
-int get_processors_groups_count() { return GetActiveProcessorGroupCount(); }
+int get_processors_groups_count() {
```
aleksei-fedotov, line 41: To be more consistent with the Windows API:
```suggestion
int get_processor_group_count() {
```

cmake/README.md
```diff
@@ -42,17 +43,19 @@ Some useful options:
 
 #### TBBBind library configuration
 
-The TBBbind library has two versions: `tbbbind` and `tbbbind_2_0`. Each of these versions is linked with corresponding HWLOC library version: `tbbbind` links with HWLOC 1.11.x , `tbbbind_2_0` links with HWLOC 2.x.
+The TBBbind library has three versions: `tbbbind`, `tbbbind_2_0` and `tbbbind_2_5`. Each of these versions is linked with corresponding HWLOC library version: `tbbbind` links with HWLOC 1.11.x , `tbbbind_2_0` links with HWLOC 2.1-2.4, `tbbbind_2_5` links with HWLOC 2.5 and later.
 
-To enable the TBBBind library build or special tests that performed inside the HWLOC environment, please specify the following CMake variables:
+The suitable HWLOC library version search is enabled by default, but if you want to specify the path to the library manually, or build several TBBBind versions at one time you may specify the following variables CMake variables:
```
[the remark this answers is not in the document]
ivankochin: Why do you remove the following information:
```
build several TBBBind versions at one time
```

I think it is important.

src/tbbbind/tbb_bind.cpp
ivankochin: What is the purpose of this change?

CMakeLists.txt
ivankochin: I think it is a good idea! Please take a look at the patch and review the new logic.

CMakeLists.txt
```diff
@@ -100,6 +100,15 @@ option(TBB_NO_APPCONTAINER "Apply /APPCONTAINER:NO (for testing binaries for Win
...
+
+if (APPLE)
+    set(TBB_TBBBIND_BUILD OFF)
+endif()
```
alexey-katranov, line 107: It look strange, we are adding option to enable tbbbind build but turn off on some platforms. What is the option usage model? It seems that it does not respect the user setting.

ivankochin: It was decided to rework option semantic and rename it to `TBB_DISABLE_HWLOC_AUTOMATIC_SEARCH `

src/tbbbind/tbb_bind.cpp
ivankochin: I cannot do it since it is an enum, not a macro.

cmake/README.md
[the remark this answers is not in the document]
ivankochin: Discussed offline

cmake/README.md
```diff
@@ -42,17 +43,21 @@ Some useful options:
...
  - `CMAKE_HWLOC_<HWLOC_VER>_LIBRARY_PATH` - path to the corresponding HWLOC version shared library on Linux or path to `.lib` file on Windows.
  - `CMAKE_HWLOC_<HWLOC_VER>_INCLUDE_PATH` - path to the corresponding HWLOC version includes directory.
 
 Windows requires one additional variable for correct TBBBind library building:
  - `CMAKE_HWLOC_<HWLOC_VER>_DLL_PATH` - path to the corresponding HWLOC version `.dll` file.
 
-`HWLOC_VER` variable used above can be replaced with one of the two values: `1_11` for the `tbbbind` library configuration and `2` for the `tbbbind_2_0` library configuration.
+`HWLOC_VER` substring used above can be replaced with one of the three values: `1_11` for the `tbbbind` library configuration, `2` for the `tbbbind_2_0`, and `2_5` for the `tbbbind_2_5` library configuration.
+
+If you specify variables for several TBBBind versions, then all these versions may be built during one build session.
```
aleksei-fedotov, line 58: Why may be? Maybe:
```suggestion
If you specify variables for several TBBBind versions, then building process for all of these versions will be performed during single build session.
```
?

src/tbbbind/tbb_bind.cpp
aleksei-fedotov: Should not we include it in the macro definition then? Something like:
```cpp
#define __TBBBIND_HWLOC_TOPOLOGY_FLAG_RESTRICT_TO_CPUBINDING_PRESENT \
    (HWLOC_API_VERSION >= 0x20500 && HWLOC_TOPOLOGY_FLAG_RESTRICT_TO_CPUBINDING !=0)
```

ivankochin:
> Should not we include it in the macro definition then? Something like:
> 
> #define __TBBBIND_HWLOC_TOPOLOGY_FLAG_RESTRICT_TO_CPUBINDING_PRESENT \
    (HWLOC_API_VERSION >= 0x20500 && HWLOC_TOPOLOGY_FLAG_RESTRICT_TO_CPUBINDING !=0)

As I understand, the preprocessor doesn't know anything about the enums, that's why we cannot use them in the preprocessor's macros definition.

aleksei-fedotov (approved):

aleksei-fedotov (approved):

PR #451, opened 2021-06-21
task_group extensions : task_handle comparison operators unit tests

anton-potapov:
- unit tests for empty check / comparison operators

### DIFF
--- a/test/tbb/test_task_group.cpp
+++ b/test/tbb/test_task_group.cpp
@@ -1175,6 +1175,25 @@ TEST_CASE("Task handle empty check"){
     CHECK_MESSAGE(h != nullptr, "delayed task returned by task_group::delayed should not be empty");
 }
 
+//! Test for comparison operations
+//! \brief \ref interface
+TEST_CASE("Task handle comparison/empty checks"){
+    tbb::task_group tg;
+
+    tbb::task_handle h;
+
+    bool empty =  ! static_cast<bool>(h);
+    CHECK_MESSAGE(empty, "default constructed task_handle should be empty");
+    CHECK_MESSAGE(h == nullptr, "default constructed task_handle should be empty");
+    CHECK_MESSAGE(nullptr == h, "default constructed task_handle should be empty");
+
+    h = tg.defer([]{});
+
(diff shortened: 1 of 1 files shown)


test/tbb/test_task_group.cpp
```diff
@@ -1175,6 +1175,25 @@ TEST_CASE("Task handle empty check"){
...
 }
 
+//! Test for comparison operations
+//! \brief \ref interface
+TEST_CASE("Task handle comparison/empty checks"){
+    tbb::task_group tg;
+
+    tbb::task_handle h;
+
+    bool empty =  ! static_cast<bool>(h);
```
aleksei-fedotov, line 1185: Why do we need an explicit cast here? Just to test it or the code is not compiling without it? I expect it to compile without an explicit cast.

test/tbb/test_task_group.cpp
```diff
@@ -1175,6 +1175,25 @@ TEST_CASE("Task handle empty check"){
...
+    CHECK_MESSAGE(empty, "default constructed task_handle should be empty");
+    CHECK_MESSAGE(h == nullptr, "default constructed task_handle should be empty");
+    CHECK_MESSAGE(nullptr == h, "default constructed task_handle should be empty");
+
+    h = tg.defer([]{});
+
+    CHECK_MESSAGE(h != nullptr, "delayed task returned by task_group::delayed should not be empty");
+    CHECK_MESSAGE(nullptr != h, "delayed task returned by task_group::delayed should not be empty");
```
aleksei-fedotov, line 1193:
```suggestion
    CHECK_MESSAGE(h != nullptr, "deferred task returned by task_group::defer() should not be empty");
    CHECK_MESSAGE(nullptr != h, "deferred task returned by task_group::defer() should not be empty");
```
?

test/tbb/test_task_group.cpp
aleksei-fedotov: Got it! Thanks!

aleksei-fedotov (approved):
Looks good to me!

PR #457, opened 2021-06-24
Fix dlopen flags for loading tbbmalloc and rml

ivankochin:
Now after [this commit](https://github.com/oneapi-src/oneTBB/commit/0354fc109fe53a5d18c7fe2faae58fe224043862), `tbbmalloc` and `rml` libraries load via `dlopen` with `RTLD_LOCAL` and `RTLD_DEEPBIND` flags, but the correct behavior for such operations is loading with `RTLD_GLOBAL`.

Also this patch removes `RTLD_DEEPBIND` flag during loading the `tbbbind` libraries when the library was compiled with address sanitizer statistic to get rid of the following runtime error:
```
You are trying to dlopen a shared library <library_name> with RTLD_DEEPBIND flag which is incompatible with sanitizer runtime (see https://github.com/google/sanitizers/issues/611 for details). If you want to run <library_name> library under sanitizers please remove RTLD_DEEPBIND from dlopen flags.
```

### DIFF
--- a/include/oneapi/tbb/detail/_config.h
+++ b/include/oneapi/tbb/detail/_config.h
@@ -28,6 +28,10 @@
 /* Check which standard library we use. */
 #include <cstddef>
 
+#if !defined(__has_feature)
+#define __has_feature(feature) 0
+#endif
+
 #if _MSC_VER
     #define __TBB_EXPORTED_FUNC   __cdecl
     #define __TBB_EXPORTED_METHOD __thiscall
@@ -204,6 +208,11 @@
     #define __TBB_USE_OPTIONAL_RTTI 1
 #else
     #define __TBB_USE_OPTIONAL_RTTI (__GXX_RTTI || __RTTI || __INTEL_RTTI__)
+#endif
+
+/** Address sanitizer detection **/
+#if __SANITIZE_ADDRESS__ || (__clang__ && __has_feature(address_sanitizer))
+#define __TBB_SANITIZE_ADDRESS 1
 #endif
 
 /** Library features presence macros **/
--- a/include/oneapi/tbb/detail/_config.h
+++ b/include/oneapi/tbb/detail/_config.h
@@ -28,10 +28,6 @@
 /* Check which standard library we use. */
 #include <cstddef>
 
-#if !defined(__has_feature)
-#define __has_feature(feature) 0
-#endif
-
 #if _MSC_VER
     #define __TBB_EXPORTED_FUNC   __cdecl
     #define __TBB_EXPORTED_METHOD __thiscall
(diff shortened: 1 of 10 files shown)


alexey-katranov (approved):

include/oneapi/tbb/detail/_config.h
```diff
@@ -206,6 +210,11 @@
     #define __TBB_USE_OPTIONAL_RTTI (__GXX_RTTI || __RTTI || __INTEL_RTTI__)
 #endif
 
+/** Address sanitizer detection **/
+#if __SANITIZE_ADDRESS__ || (__clang__ && __has_feature(address_sanitizer))
+#define __TBB_SANITIZE_ADDRESS 1
+#endif
+
```
anton-potapov, line 217: Please move the definition from tests config : https://github.com/oneapi-src/oneTBB/blob/3a6580652d2b697823f38669fb4685da812df13c/test/common/config.h#L88-L94

include/oneapi/tbb/detail/_config.h
```diff
@@ -28,6 +28,10 @@
 /* Check which standard library we use. */
 #include <cstddef>
 
+#if !defined(__has_feature)
+#define __has_feature(feature) 0
+#endif
+
```
anton-potapov, line 34: It seems to me that this is not needed

anton-potapov (changes requested):

src/tbb/dynamic_link.h
```diff
@@ -71,8 +71,9 @@ using dynamic_link_handle = void*;
 const int DYNAMIC_LINK_GLOBAL        = 0x01;
 const int DYNAMIC_LINK_LOAD          = 0x02;
 const int DYNAMIC_LINK_WEAK          = 0x04;
+const int DYNAMIC_LINK_LOCAL         = 0x08;
 
-const int DYNAMIC_LINK_LOCAL_BINDING = 0x08 | DYNAMIC_LINK_LOAD;
+const int DYNAMIC_LINK_LOCAL_BINDING = DYNAMIC_LINK_LOCAL | DYNAMIC_LINK_LOAD;
```
anton-potapov, line 76: It seems that this is not used

ivankochin: It used here: https://github.com/oneapi-src/oneTBB/blob/master/src/tbb/governor.cpp#L393

anton-potapov (approved):

anton-potapov:
Please update to the latest master as it seems that not all usages of `__TBB_TEST_USE_ADDRESS_SANITIZER` has been replaced with new one

alexey-katranov (approved):

PR #465, opened 2021-06-29
Multiple fixes for constraints

kboyarinov:
Add fixes for multiple problems with C++20 constraints on Windows and Linux. Added workaround for a problem on Windows: "failed to select the destructor for the class" in case of constrained destructor (in tests)

### DIFF
--- a/examples/test_all/fibonacci/fibonacci.cpp
+++ b/examples/test_all/fibonacci/fibonacci.cpp
@@ -211,8 +211,8 @@ struct IntHashCompare {
     bool equal(const int j, const int k) const {
         return j == k;
     }
-    unsigned long hash(const int k) const {
-        return (unsigned long)k;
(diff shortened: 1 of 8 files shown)


aleksei-fedotov (dismissed):
Looks good to me!

kboyarinov:
I have deeply investigated the destructors problems and it seems like MSVC is the only compiler which behaves correct!
See the link below for more details on C++20 prospective destructors.
https://en.cppreference.com/w/cpp/language/destructor#:~:text=member%20access%20operator.-,Prospective%20destructor,implicitly%20declared%20prospective%20destructor%20is%20also%20the%20destructor%20for%20T.,-(since%20C%2B%2B20

I have rewriten the test code to respect the modern C++ rule.

alexey-katranov (approved):

PR #466, opened 2021-06-29
Fix LTO flags on Windows for old CMake versions

ivankochin:
In the CMake version older than 13 we use the `target_link_libraries` command to pass the linker flags. This option detects the link flags if they prefixed by dash(-). So we should pass only flags prefixed by dash(-) for building with CMake < 13.


cmake/README.md
```diff
@@ -274,6 +274,7 @@ Variable | Description
 `TBB_LIB_LINK_LIBS `       | Applied to libraries as link libraries
 `TBB_TEST_COMPILE_FLAGS`   | Applied to tests as compile options
 
+Please specify the `LINK` options prefixed with dash(-) for Visual Sudio compiler with CMake < 13 to avoid issues caused by `target_link_libraries` CMake command usage.
```
zheltovs, line 277: `for Visual Sudio compiler with CMake < 13` -> `for MSVC (Visual Studio) compiler with CMake < 3.13`?

ivankochin:
```suggestion
Please specify the `LINK` options prefixed with dash(-) for MSVC(Visual Studio) compiler with CMake < 3.13 to avoid issues caused by `target_link_libraries` CMake command usage.
```
Right?

zheltovs: Yes

zheltovs (approved):

PR #467, opened 2021-06-29
ASAN: test_malloc_whitebox

vlserov:
Reverting ASAN fix for test_malloc_regression #415
Turn off memory leaks test in test_malloc_regression test case

### DIFF
--- a/test/common/memory_usage.h
+++ b/test/common/memory_usage.h
@@ -26,10 +26,6 @@
 #include "utils.h"
 #include "utils_assert.h"
 
-#if __TBB_USE_ADDRESS_SANITIZER
-#include <sanitizer/allocator_interface.h>
-#endif
-
 #if __linux__ || __sun
 #include <sys/resource.h>
 #include <unistd.h>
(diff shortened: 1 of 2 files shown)


test/common/memory_usage.h
```diff
@@ -133,6 +130,16 @@ namespace utils {
 #endif
     }
 
+    std::size_t GetApplicationMemoryUsage(MemoryStatType stat = currentUsage) {
```
alexey-katranov, line 133: What is the difference between `GetApplicationMemoryUsage` and `GetMemoryUsage`? Why it cannot be one function?

test/tbbmalloc/test_malloc_regression.cpp
```diff
@@ -158,13 +158,16 @@ void TestAlignedMsize()
         }
 }
 
-//! \brief \ref error_guessing
+#if __TBB_USE_ADDRESS_SANITIZER
+TEST_CASE("Memory leaks test is not applicable under ASAN\n" * doctest::skip(true)) {}
+#else
```
alexey-katranov, line 163: Do not remove the briefs. In addition, each TEST_CASE (incl. skipped) should have a brief. Maybe it make sense just to exclude the TEST_CASE, e.g.
```suggestion
#if !__TBB_USE_ADDRESS_SANITIZER
//! \brief \ref error_guessing
```

alexey-katranov (approved):

PR #480, opened 2021-07-05
Fix internal allocator related issues

alexey-katranov:
* For some reason, GCC does not instrument `tbbmalloc` with TSan when `lto` is enabled. 
* Fix race in internal allocator on handler pointer initialization. 
* Simplify checks in `conformance_parallel_for`

### DIFF
--- a/src/tbb/allocator.cpp
+++ b/src/tbb/allocator.cpp
@@ -225,16 +225,16 @@ void __TBB_EXPORTED_FUNC deallocate_memory(void* p) {
 }
 
 bool __TBB_EXPORTED_FUNC is_tbbmalloc_used() {
-    auto alloc_handler = allocate_handler.load(std::memory_order_acquire);
-    if (alloc_handler == &initialize_allocate_handler) {
+    auto handler_snapshot = allocate_handler.load(std::memory_order_acquire);
+    if (handler_snapshot == &initialize_allocate_handler) {
         initialize_cache_aligned_allocator();
     }
-    alloc_handler = allocate_handler.load(std::memory_order_relaxed);
-    __TBB_ASSERT(alloc_handler != &initialize_allocate_handler && deallocate_handler != nullptr, NULL);
+    handler_snapshot = allocate_handler.load(std::memory_order_relaxed);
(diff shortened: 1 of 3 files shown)


src/tbb/allocator.cpp
```diff
@@ -218,15 +225,16 @@ void __TBB_EXPORTED_FUNC deallocate_memory(void* p) {
 }
 
 bool __TBB_EXPORTED_FUNC is_tbbmalloc_used() {
-    if (allocate_handler == &initialize_allocate_handler) {
-        void* void_ptr = allocate_handler(1);
-        deallocate_handler(void_ptr);
+    auto alloc_handler = allocate_handler.load(std::memory_order_acquire);
```
ivankochin, line 228:
```suggestion
    auto handler_snapshot = allocate_handler.load(std::memory_order_acquire);
```
The `alloc_handler` and `allocate_handler` seem so similar, so I want to propose to improve readability by changing the local copy name.

alexey-katranov: Applied

test/conformance/conformance_parallel_for.cpp
```diff
@@ -184,11 +184,8 @@ void Flog() {
             }
                 break;
             }
-            for( int j=0; j<i; ++j )
-                CHECK( Array[j]==1);
-            for( int j=i; j<N; ++j )
-                CHECK( Array[j]==0);
-            CHECK( FooBodyCount==1);
+            CHECK(std::find_if_not(Array + i, Array + N, [](const std::atomic<int>& v) { return v.load(std::memory_order_relaxed) == 0; }) == Array + N);
```
alexey-katranov, line 187:
```suggestion
            CHECK(std::find_if_not(Array, Array + i, [](const std::atomic<int>& v) { return v.load(std::memory_order_relaxed) == 1; }) == Array + i);
            CHECK(std::find_if_not(Array + i, Array + N, [](const std::atomic<int>& v) { return v.load(std::memory_order_relaxed) == 0; }) == Array + N);
```

ivankochin (approved):

PR #484, opened 2021-07-08
ASAN: memory leak in test_arena_constraints/conformance_arena_constra…

vlserov:
…ints

Memory leak has been caused by ThirdParty library.
Adding LSAN suppression list for fitler known issues.


cmake/sanitize.cmake
```diff
@@ -34,5 +34,8 @@ if (NOT ${FLAG_DISPLAY_NAME})
     "please try another compiler or omit TBB_SANITIZE variable")
 endif()
 
+set(TBB_LSAN_OPTIONS "suppressions=${CMAKE_CURRENT_SOURCE_DIR}/cmake/lsan.suppressions")
+set(TBB_TESTS_ENVIRONMENT LSAN_OPTIONS=${TBB_LSAN_OPTIONS})
```
ivankochin, line 38:
```suggestion
set(TBB_LSAN_ENVIRONMENT LSAN_OPTIONS=suppressions=${CMAKE_CURRENT_SOURCE_DIR}/cmake/lsan.suppressions)
```

`TBB_LSAN_OPTIONS` is used only in one place, so maybe just merge two options into one.

ivankochin (commented):
It seems like you should also define the LSAN suppression variables [here](https://github.com/oneapi-src/oneTBB/pull/484/files#diff-33394812ba204689144fd2f80832db83853ba1cb32403edb4e15fe4893e675fdR220).

cmake/sanitize.cmake
vlserov: it's good point merge two variable in one for now, but I would like to use neutral name for this variable.
```suggestion
set(TBB_TESTS_ENVIRONMENT ${TBB_TESTS_ENVIRONMENT} LSAN_OPTIONS="suppressions=${CMAKE_CURRENT_SOURCE_DIR}/cmake/lsan.suppressions")
```

cmake/lsan.suppressions
```diff
@@ -0,0 +1,2 @@
+# LSAN suppression for ltdl library known issue.
```
ivankochin, line 1: I think that this file is not related to CMake. Can we consider different locations, e.g. `<root>/sanitizers/`? @AlexVeprev what do you think about this?

ivankochin (commented):
If CI is passed, then LGTM, except the previous comment.

PR #498, opened 2021-07-18
Fix HWLOC_LIBRARY_DIRS when using pkg-config to find hwloc

Logarithmus:
Without this change `cmake` reports this:
```
WARNING: Target "conformance_arena_constraints" requests linking to directory "/usr/lib64".  Targets may link only to libraries.  CMake is dropping the item.
WARNING: Target "test_arena_constraints" requests linking to directory "/usr/lib64".  Targets may link only to libraries.  CMake is dropping the item.
```

And then generated `build.ninja` contains this:

```
...
build gnu_10.2_cxx11_64_relwithdebinfo/libtbbbind_2_5.so.3.3: CXX_SHARED_LIBRARY_LINKER__tbbbind_2_5_RelWithDebInfo src/tbbbind/CMakeFiles/tbbbind_2_5.dir/tbb_bind.cpp.o | ../src/tbbbind/def/lin64-tbbbind.def
  LANGUAGE_COMPILE_FLAGS = -O2 -g -DNDEBUG
  LINK_FLAGS = -Wl,--version-script=/home/artur/Programming/C++/oneTBB/src/tbbbind/def/lin64-tbbbind.def  -flto /usr/lib64
  LINK_LIBRARIES = -lhwloc  -ldl
...
```

This happens because `HWLOC_LIBRARY_DIRS` variable, which is set by `pkg-config`, contains library directories without `-L` prefix (see https://cmake.org/cmake/help/latest/module/FindPkgConfig.html#:~:text=their%20absolute%20paths-,%3CXXX%3E_LIBRARY_DIRS,-the%20paths%20of), so the linker tries to link `/usr/lib64` directory as library and obviously fails. So instead of passing `HWLOC_LIBRARY_DIRS` to `target_link_libraries`, one should pass it to `target_link_directories` instead.

I'm using Void Linux, I tried to reproduce this bug in Alpine docker container with `pkgconf` installed, but failed for unknown reason. Anyway, this bug obviously exists and prevents from updating Void Linux package.

### DIFF
--- a/src/tbbbind/CMakeLists.txt
+++ b/src/tbbbind/CMakeLists.txt
@@ -37,6 +37,12 @@ function(tbbbind_build TBBBIND_NAME REQUIRED_HWLOC_TARGET)
         $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
         ${HWLOC_INCLUDE_DIRS} # pkg-config defined
     )
+    if (COMMAND target_link_directories)
+	target_link_directories(${TBBBIND_NAME}
+		PRIVATE
+		${HWLOC_LIBRARY_DIRS} # pkg-config defined
+	)
+    endif()
 
     target_compile_options(${TBBBIND_NAME}
         PRIVATE
@@ -66,7 +72,6 @@ function(tbbbind_build TBBBIND_NAME REQUIRED_HWLOC_TARGET)
             ${TBB_LIB_LINK_FLAGS}
             ${TBB_COMMON_LINK_FLAGS}
             ${TBB_IPO_LINK_FLAGS}
-            ${HWLOC_LIBRARY_DIRS} # pkg-config defined
         )
     else()
         target_link_libraries(${TBBBIND_NAME}
@@ -74,7 +79,6 @@ function(tbbbind_build TBBBIND_NAME REQUIRED_HWLOC_TARGET)
             ${TBB_LIB_LINK_FLAGS}
             ${TBB_COMMON_LINK_FLAGS}
             ${TBB_IPO_LINK_FLAGS}
-            ${HWLOC_LIBRARY_DIRS} # pkg-config defined
         )
     endif()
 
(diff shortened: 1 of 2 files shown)


src/tbbbind/CMakeLists.txt
```diff
@@ -37,6 +37,10 @@ function(tbbbind_build TBBBIND_NAME REQUIRED_HWLOC_TARGET)
         $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
         ${HWLOC_INCLUDE_DIRS} # pkg-config defined
     )
+	target_link_directories(${TBBBIND_NAME}
+		PRIVATE
+		${HWLOC_LIBRARY_DIRS} # pkg-config defined
+	)
```
ivankochin, line 43: The `target_link_directories` command was introduced in CMake 3.13, but the oneTBB CMake minimal supported version is 3.1. So I would propose you check the command availability before the usage:
```suggestion
    if (COMMAND target_link_directories)
	target_link_directories(${TBBBIND_NAME}
		PRIVATE
		${HWLOC_LIBRARY_DIRS} # pkg-config defined
	)
    endif()
```
@AlexVeprev what do you think, should we have any `else` branch here to handle link directories when using CMake < 3.13?

Logarithmus:
@ivankochin what's next?

ivankochin:
@Logarithmus I've created the new solution for this problem: #508

My main root cause for the new solution is that your solution doesn't take into attention the `HWLOC_LIBRARY_DIRS` hint if CMake does not contain the `target_link_directories` command. But the solution described in #508 is designed to be cross-version. Please review the proposed solution and leave your comment if needed.

Logarithmus:
> @Logarithmus I've created the new solution for this problem: #508
> 
> My main root cause for the new solution is that your solution doesn't take into attention the `HWLOC_LIBRARY_DIRS` hint if CMake does not contain the `target_link_directories` command. But the solution described in #508 is designed to be cross-version. Please review the proposed solution and leave your comment if needed.

I see that you are clearly more familiar with CMake than me, so I think your solution would be better than mine.

PR #501, opened 2021-07-20
Prevent linking HWLOC 2.4 with tbbbind_2_5

ivankochin:
The `tbbbind_2_5` library uses only one enum flag that was introduced as a part of HWLOC 2.5. So it requires to be compiled with the HWLOC 2.5 headers but may be linked with HWLOC 2.4 during the runtime. Such a situation caused an undefined behavior since the HWLOC library does not support the forward binary compatibility. That's why this patch adds an explicit call of the HWLOC 2.5 interface to prevent linking the `tbbbind_2_5` library with HWLOC 2.4


src/tbbbind/tbb_bind.cpp
```diff
@@ -223,6 +223,15 @@ class platform_topology {
         }
     }
 
+    // Explicitly call the interface that was introduced in the HWLOC 2.5 release to prevent linking
+    // HWLOC 2.4 during the tbbbind_2_5 library loding.
```
ivankochin, line 227:
```suggestion
    // HWLOC 2.4 during the tbbbind_2_5 library loading.
```

src/tbbbind/tbb_bind.cpp
```diff
@@ -223,6 +223,15 @@ class platform_topology {
         }
     }
 
+    // Explicitly call the interface that was introduced in the HWLOC 2.5 release to prevent linking
+    // HWLOC 2.4 during the tbbbind_2_5 library loading.
+    void backward_compatibility_guarantee() {
```
ivankochin, line 228:
```suggestion
    void enforce_hwloc_2_5_runtime_linkage() {
        // Explicitly call the interface that was introduced in the HWLOC 2.5 release to prevent linking
        // HWLOC 2.4 during the tbbbind_2_5 library loading.
```

src/tbbbind/tbb_bind.cpp
```diff
@@ -242,6 +251,8 @@ class platform_topology {
         numa_topology_parsing();
         core_types_topology_parsing();
 
+        backward_compatibility_guarantee();
```
ivankochin, line 254:
```suggestion
        enforce_hwloc_2_5_runtime_linkage();
```

src/tbbbind/tbb_bind.cpp
ivankochin:
```suggestion
    void enforce_hwloc_2_5_runtime_linkage() {
        // Without the call of this function HWLOC 2.4 can be successfully loaded during the tbbbind_2_5 loading.
        // It is possible since tbbbind_2_5 don't use any new entry points that were introduced in HWLOC 2.5
        // But tbbbind_2_5 compiles with HWLOC 2.5 header, therefore such situation requires binary forward compatibility
        // which are not guaranteed by the HWLOC library. To enforce linkage tbbbind_2_5 only with HWLOC >= 2.5 version
        // this function calls the interface that is available in the HWLOC 2.5 only.
```

alexey-katranov (approved):

alexey-katranov (approved):

PR #502, opened 2021-07-21
ASAN: test_task_group timeout

vlserov:
Prevent infinite loop during memory leak check

### DIFF
--- a/test/tbb/test_task_group.cpp
+++ b/test/tbb/test_task_group.cpp
@@ -763,6 +763,11 @@ namespace TestIsolationNS {
 }
 #endif
 
+#if __TBB_USE_ADDRESS_SANITIZER
+//! Test for thread safety for the task_group
+//! \brief \ref error_guessing \ref resource_usage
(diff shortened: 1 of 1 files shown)


test/tbb/test_task_group.cpp
```diff
@@ -258,16 +258,21 @@ void TestThreadSafety() {
     // Test and warm up allocator.
     tests();
 
-    // Ensure that cosumption is stabilized.
-    std::size_t initial = utils::GetMemoryUsage();
-    for (;;) {
+#if !__TBB_USE_ADDRESS_SANITIZER
+    constexpr size_t num_trials = 1000;
+    bool no_memory_leak = false;
```
alexey-katranov, line 263: I am Ok to disable the test case for ASAN; however, the limited number of trails might lead to test instability in some corner cases. Can we remain it endless while disabling with ASAN?

anton-potapov (approved):
LGTM

PR #505, opened 2021-07-22
Add statistics collection to thread data

kononovk:
Co-Authored-By: Zakhar Zakharov <zakhar.zakharov@intel.com>

### DIFF
--- a/include/oneapi/tbb/detail/_config.h
+++ b/include/oneapi/tbb/detail/_config.h
@@ -127,6 +127,14 @@
     #define TBB_USE_ASSERT TBB_USE_DEBUG
 #endif // TBB_USE_ASSERT
 
+#ifndef __TBB_STATISTICS
+#if __TBB_STATISTICS_SUM
+    #define __TBB_STATISTICS 1
+#else
+    #define __TBB_STATISTICS 0
+#endif // __TBB_STATISTICS_SUM
+#endif // __TBB_STATISTICS
+
 #ifndef TBB_USE_PROFILING_TOOLS
 #if TBB_USE_DEBUG
     #define TBB_USE_PROFILING_TOOLS 2
@@ -479,6 +487,12 @@
 #else // __TBB_CPP20_CONCEPTS_PRESENT
     #define __TBB_requires(...)
 #endif // __TBB_CPP20_CONCEPTS_PRESENT
+
+#if __TBB_STATISTICS
+    #define __TBB_add_statistics(TD, STAT) TD->add_statistics(governor::statistics::STAT)
+#else
+    #define __TBB_add_statistics(TD, STAT)
+#endif // __TBB_STATISTICS
 
 /** Macros of the form __TBB_XXX_BROKEN denote known issues that are caused by
     the bugs in compilers, standard or OS specific libraries. They should be
--- a/include/oneapi/tbb/detail/_config.h
+++ b/include/oneapi/tbb/detail/_config.h
@@ -489,7 +489,7 @@
 #endif // __TBB_CPP20_CONCEPTS_PRESENT
 
 #if __TBB_STATISTICS
-    #define __TBB_add_statistics(TD, STAT) TD->add_statistics(governor::statistics::STAT)
+    #define __TBB_add_statistics(TLS, STAT) TLS->add_statistics(governor::statistics::STAT)
 #else
     #define __TBB_add_statistics(TD, STAT)
 #endif // __TBB_STATISTICS
--- a/include/oneapi/tbb/detail/_config.h
+++ b/include/oneapi/tbb/detail/_config.h
@@ -491,7 +491,7 @@
 #if __TBB_STATISTICS
     #define __TBB_add_statistics(TLS, STAT) TLS->add_statistics(governor::statistics::STAT)
 #else
-    #define __TBB_add_statistics(TD, STAT)
+    #define __TBB_add_statistics(TLS, STAT)
 #endif // __TBB_STATISTICS
 
 /** Macros of the form __TBB_XXX_BROKEN denote known issues that are caused by
--- a/src/tbb/arena_slot.cpp
+++ b/src/tbb/arena_slot.cpp
@@ -59,6 +59,10 @@ d1::task* arena_slot::get_task_impl(size_t T, execution_data_ext& ed, bool& task
 }
 
 d1::task* arena_slot::get_task(execution_data_ext& ed, isolation_type isolation) {
+#if __TBB_STATISTICS
+    thread_data *tls = governor::get_thread_data();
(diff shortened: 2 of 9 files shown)


src/tbb/governor.h
```diff
@@ -51,7 +58,7 @@ class governor {
 
     // TODO: consider using thread_local (measure performance and side effects)
     //! TLS for scheduler instances associated with individual threads
-    static basic_tls<thread_data*> theTLS;
+    static basic_tls<thread_data*> theTLS;
```
kononovk, line 61: delete whitespace

include/oneapi/tbb/detail/_config.h
```diff
@@ -127,6 +127,14 @@
     #define TBB_USE_ASSERT TBB_USE_DEBUG
 #endif // TBB_USE_ASSERT
 
+#ifndef __TBB_STATISTICS
+#if __TBB_STATISTICS_SUM
+    #define __TBB_STATISTICS 1
+#else
+    #define __TBB_STATISTICS 0
+#endif // __TBB_STATISTICS_SUM
+#endif // __TBB_STATISTICS
```
ivankochin, line 136: Shouldn't it be like this:
```suggestion
#ifndef __TBB_STATISTICS
    #define __TBB_STATISTICS 0
#endif // __TBB_STATISTICS

#if __TBB_STATISTICS
    #ifndef __TBB_STATISTICS_SUM
        #define __TBB_STATISTICS_SUM 0
     #endif // __TBB_STATISTICS_SUM
#endif
```

include/oneapi/tbb/detail/_config.h
```diff
@@ -480,6 +488,12 @@
     #define __TBB_requires(...)
 #endif // __TBB_CPP20_CONCEPTS_PRESENT
 
+#if __TBB_STATISTICS
+    #define __TBB_add_statistics(TD, STAT) TD->add_statistics(governor::statistics::STAT)
```
ivankochin, line 492:
```suggestion
    #define __TBB_add_statistics(TLS, STAT) TLS->add_statistics(governor::statistics::STAT)
```

src/tbb/governor.h
ivankochin, line 61:
```suggestion
    static basic_tls<thread_data*> theTLS;
```

include/oneapi/tbb/detail/_config.h
kononovk: No, because if you set flag `__TBB_STATISTICS_SUM` we must set also flag `__TBB_STATISTICS`

src/tbb/governor.h
```diff
@@ -151,6 +158,68 @@ class governor {
...
+    enum class statistics : std::size_t {
+        exec_task = 0,
+        get_task,
+        task_res_not_null,
+        steal_task,
+        try_steal_task,
+        victim_pool_empty,
+        size,
+    };
+    using array_stat = std::array<std::uint32_t, static_cast<std::size_t>(statistics::size)>;
```
ivankochin, line 172:
```suggestion
    using statistic_array = std::array<std::uint32_t, static_cast<std::size_t>(statistics::size)>;
```

src/tbb/thread_data.h
```diff
@@ -116,6 +116,10 @@ class thread_data : public ::rml::job
     }
 
     ~thread_data() {
+#if __TBB_STATISTICS
+        governor::accumulator.emplace(statistics);
+#endif /* __TBB_STATISTICS */
+        context_list_cleanup();
```
ivankochin, line 122: Why should we call this method here?

kononovk: We shouldn't

src/tbb/governor.h
```diff
@@ -151,6 +158,68 @@ class governor {
...
     }
+
+#if __TBB_STATISTICS
+    enum class statistics : std::size_t {
+        exec_task = 0,
+        get_task,
+        task_res_not_null,
+        steal_task,
+        try_steal_task,
+        victim_pool_empty,
```
ivankochin, line 169:
```suggestion
        execute_task = 0,
        try_get_task,
        successfull_get_task,
        try_steal_task,
        successfull_steal_task,
        victim_pool_empty,
```
@alexey-katranov, @pavelkumbrasev what do you think about such renaming?

src/tbb/governor.h
```diff
@@ -151,6 +158,68 @@ class governor {
...
+        size,
+    };
+    using statistic_array = std::array<std::uint32_t, static_cast<std::size_t>(statistics::size)>;
+
+    class collect_statistics : no_copy {
+    public:
+        collect_statistics() = default;
+
+        ~collect_statistics() = default;
```
ivankochin, line 178: Seems redundant

src/tbb/governor.h
```diff
@@ -151,6 +158,68 @@ class governor {
...
+        }
+
+        // thread-unsafe method
+        void print_statistics() {
+                int offset = 2;
+                /* std::string not so bad because of Small String Optimization */
+                std::string column_names[] = {
+                    "executed", "getted",
+                    "result != null", "steal", "try steal",
+                    "vic pool empty",
```
ivankochin, line 192: Consider more verbose names and making this string static

src/tbb/arena_slot.cpp
```diff
@@ -59,6 +59,10 @@ d1::task* arena_slot::get_task_impl(size_t T, execution_data_ext& ed, bool& task
 }
 
 d1::task* arena_slot::get_task(execution_data_ext& ed, isolation_type isolation) {
+#if __TBB_STATISTICS
+    thread_data *tls = governor::get_thread_data();
```
ivankochin, line 63: Can we just take tls using the function arguments from the outer level?

src/tbb/market.cpp
```diff
@@ -203,7 +203,7 @@ bool market::release ( bool is_public, bool blocking_terminate ) {
             theMarket = NULL;
         }
     }
-    if( do_release ) {
+    if ( do_release ) {
```
ivankochin, line 206:
```suggestion
    if( do_release ) {
```

kononovk: No, check another codestyle in this file https://github.com/oneapi-src/oneTBB/blob/master/src/tbb/market.cpp#L222

src/tbb/governor.h
kononovk: We deleted copy/move constructors and copy/move assigned operators. Seems its ok

src/tbb/governor.h
kononovk: Names should be as short a possible, otherwise the table with be wider then the screen..

PR #508, opened 2021-07-26
Fix the HWLOC linking when it was found using pkg-config

ivankochin:
One more solution for the issue described in #498, but this solution uses the `HWLOC_LIBRARY_DIRS` hint not only for CMake versions that contain the `target_link_directories` interface.


ivankochin:
@AlexVeprev what do you think about such an approach?

src/tbbbind/CMakeLists.txt
```diff
@@ -70,21 +70,24 @@ function(tbbbind_build TBBBIND_NAME REQUIRED_HWLOC_TARGET)
...
             ${TBB_LIB_LINK_FLAGS}
             ${TBB_COMMON_LINK_FLAGS}
             ${TBB_IPO_LINK_FLAGS}
-            ${HWLOC_LIBRARY_DIRS} # pkg-config defined
         )
     endif()
 
+    # Get the full path to HWLOC if it was found by pkg-config
+    if (${HWLOC_TARGET_LINK_LIBRARIES} STREQUAL ${HWLOC_LIBRARIES})
+        find_library(HWLOC_TARGET_LINK_LIBRARIES ${HWLOC_LIBRARIES} PATHS ${HWLOC_LIBRARY_DIRS})
```
AlexVeprev, line 85: From #498 

Will you have to duplicate it here https://github.com/oneapi-src/oneTBB/blob/ad4c9bcc47382858c88a9ea58b7c3416072ea583/test/CMakeLists.txt#L266 ?

Maybe it makes sense to do such stuff somewhere near https://github.com/oneapi-src/oneTBB/blob/e96dbf4b48e689ab810b0428d0858a6abc2c83bd/cmake/hwloc_detection.cmake#L55 ?

I know you want to make the approach working for CMake 3.1, but just in case you can consider using pkg_check_modules with IMPORTED_TARGET (available since CMake 3.6). Or you can try creation of imported target on your own.

AlexVeprev (commented):
Interesting idea, but take into account update of test/CMakeLists.txt - it may lead to duplication

src/tbbbind/CMakeLists.txt
ivankochin: Great idea, thanks!

cmake/hwloc_detection.cmake
```diff
@@ -51,9 +51,9 @@ if (NOT HWLOC_TARGET_EXPLICITLY_DEFINED AND
     NOT TBB_DISABLE_HWLOC_AUTOMATIC_SEARCH
 )
     find_package(PkgConfig QUIET)
-    if (PKG_CONFIG_FOUND)
-        pkg_search_module(HWLOC hwloc)
-        if (HWLOC_FOUND)
+    if (PKG_CONFIG_FOUND AND CMAKE_VERSION VERSION_GREATER 3.6)
```
AlexVeprev, line 54: `VERSION_GREATER`  will lead to `>3.6`, but you need `>=3.6`. CMake provides `VERSION_GREATER_EQUAL` but it is not available in CMake 3.1. So, please consider this:
```suggestion
    if (PKG_CONFIG_FOUND AND NOT CMAKE_VERSION VERSION_LESS 3.6)
```

AlexVeprev (commented):
Generally looks good to me

AlexVeprev (approved):

PR #516, opened 2021-07-30
Code coverage fixes

isaevil:
In scope of this patch `TBB_BUILD` and `TBBMALLOC_BUILD` options were added. It was done to support scenarios when we are testing without tbbmalloc. It allows us to cover place where we fail to link with tbbmalloc and start using `std_cache_aligned_allocate/deallocate`.

Signed-off-by: Ilya Isaev <ilya.isaev@intel.com>


pavelkumbrasev (commented):
The code part looks fine.

cmake/README.md
```diff
@@ -10,6 +10,8 @@ TBB_SANITIZE:STRING - Sanitizer parameter, passed to compiler/linker
 TBB_SIGNTOOL:FILEPATH - Tool for digital signing, used in post install step for libraries if provided.
 TBB_SIGNTOOL_ARGS:STRING - Additional arguments for TBB_SIGNTOOL, used if TBB_SIGNTOOL is set.
 TBB4PY_BUILD:BOOL - Enable  Intel(R) oneAPI Threading Building Blocks (oneTBB) Python module build (OFF by default)
+TBB_BUILD:BOOL - Enable  Intel(R) oneAPI Threading Building Blocks (oneTBB) build (ON by default)
+TBBMALLOC_BUILD:BOOL - Enable  Intel(R) oneAPI Threading Building Blocks (oneTBB) memory allocator build (ON by default)
```
AlexVeprev, line 14:
```suggestion
TBB_BUILD:BOOL - Enable Intel(R) oneAPI Threading Building Blocks (oneTBB) build (ON by default)
TBBMALLOC_BUILD:BOOL - Enable Intel(R) oneAPI Threading Building Blocks (oneTBB) memory allocator build (ON by default)
TBB4PY_BUILD:BOOL - Enable Intel(R) oneAPI Threading Building Blocks (oneTBB) Python module build (OFF by default)
```

test/CMakeLists.txt
```diff
@@ -361,247 +361,253 @@ unset(_sde_find_name)
...
-if (NOT "${CMAKE_SYSTEM_PROCESSOR}" MATCHES "mips")
-    # TODO: Fix for MIPS
-    tbb_add_test(SUBDIR tbb NAME test_tbb_fork DEPENDENCIES TBB::tbb)
-endif()
-
-tbb_add_test(SUBDIR tbb NAME test_tbb_header DEPENDENCIES TBB::tbb)
-target_sources(test_tbb_header PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/tbb/test_tbb_header_secondary.cpp)
-if (TBB_OPENMP_FLAG AND NOT CMAKE_SYSTEM_PROCESSOR MATCHES "(mips)")
-    tbb_add_test(SUBDIR tbb NAME test_openmp DEPENDENCIES TBB::tbb)
+if (TBB_BUILD)
```
AlexVeprev, line 364: I have some doubts regarding using variable with such name here. They are pretty nice in scope of including/excluding `add_subdirectory`, but considering `test/CMakeLists.txt` such names don't describe the behavior precisely for cases of testing _external_ TBB. It may be more precisely to use something like `if (TARGET TBB::tbb)` here, what do you think?

cmake/README.md
```diff
@@ -10,6 +10,8 @@ TBB_SANITIZE:STRING - Sanitizer parameter, passed to compiler/linker
 TBB_SIGNTOOL:FILEPATH - Tool for digital signing, used in post install step for libraries if provided.
 TBB_SIGNTOOL_ARGS:STRING - Additional arguments for TBB_SIGNTOOL, used if TBB_SIGNTOOL is set.
 TBB4PY_BUILD:BOOL - Enable  Intel(R) oneAPI Threading Building Blocks (oneTBB) Python module build (OFF by default)
```
AlexVeprev, line 12: Have you considered dependencies? E.g. TBB4PY depends on TBB: https://github.com/oneapi-src/oneTBB/blob/9e15720bc7744f85dff611d34d65e9099e077da4/python/CMakeLists.txt#L41

AlexVeprev:
Please provide the motivation for new options in the description.

test/CMakeLists.txt
isaevil: Seems reasonable

cmake/README.md
isaevil: If we try to configure oneTBB like this `cmake -DTBB4PY_BUILD=ON -DTBB_BUILD=OFF`, configuration will fail with error `non-existent target "tbb"`. I think that is OK to fail here because such command does not have sense and we let user to know about it.

CMakeLists.txt
```diff
@@ -192,10 +194,14 @@ if (TBB_FIND_PACKAGE OR TBB_DIR)
...
     if (NOT "${CMAKE_SYSTEM_PROCESSOR}" MATCHES "mips")
-        add_subdirectory(src/tbbmalloc)
-        add_subdirectory(src/tbbmalloc_proxy)
+        if (TBBMALLOC_BUILD)
+            add_subdirectory(src/tbbmalloc)
+            add_subdirectory(src/tbbmalloc_proxy)
+        endif()
         if (APPLE)
             message(STATUS "TBBBind build targets are disabled due to unsupported environment")
         else()
```
AlexVeprev, line 207: There is `add_subdirectory(src/tbbbind)` one line below.

We've discussed that normally tbbbind is skipped due to lack of hwloc paths, but take into account auto-detection of hwloc using pkg-config: https://github.com/oneapi-src/oneTBB/blob/357c9b27fa84eaf748bc556585c26995c446eceb/cmake/hwloc_detection.cmake#L53

User still can force disabling hwloc detection through `TBB_DISABLE_HWLOC_AUTOMATIC_SEARCH`. It's just FYI.

AlexVeprev (approved):

PR #519, opened 2021-08-02
GCC 11: suppress stringop-overread warning

vlserov:
pthread_setspecific has attribute access none for second parameter.
GCC 11 causes "stringop-overread" warning in case providing non-pointer value to pthread_setspecific.

test_malloc_new_handler: using thread_local instead of tbb::detail::r1::tls
test_malloc_init_shutdown: Replace magic-value with valid pointer

### DIFF
--- a/test/tbbmalloc/test_malloc_new_handler.cpp
+++ b/test/tbbmalloc/test_malloc_new_handler.cpp
@@ -33,9 +33,7 @@
 #pragma warning (disable: 4800)
 #endif //#if _MSC_VER
 
-#include "../../src/tbb/tls.h"
-
-tbb::detail::r1::tls<bool> new_handler_called;
+thread_local bool new_handler_called = false;
 void customNewHandler() {
     new_handler_called = true;
     throw std::bad_alloc();
--- a/src/tbb/parallel_pipeline.cpp
+++ b/src/tbb/parallel_pipeline.cpp
@@ -144,7 +144,7 @@ class input_buffer {
     const bool is_ordered;
 
     //! for parallel filters that accepts NULLs, thread-local flag for reaching end_of_input
-    using end_of_input_tls_t = basic_tls<std::intptr_t>;
+    using end_of_input_tls_t = basic_tls<input_buffer*>;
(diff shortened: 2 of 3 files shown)


src/tbb/tls.h
```diff
@@ -41,7 +41,14 @@ class basic_tls {
         return pthread_key_create(&my_key, dtor);
     }
     int  destroy()      { return pthread_key_delete(my_key); }
-    void set( T value ) { pthread_setspecific(my_key, (void*)value); }
+    void set(T value) {
+        void *
+#if __TBB_GCC_VERSION >= 110100 && !__clang__ && !__INTEL_COMPILER
```
pavelkumbrasev, line 46: Do we want to add this workaround for all future versions of GCC?

vlserov: I guess the warning is valid since non-pointer value is used for pointer type.
So behavior won't changed in future GCC release, but updated for double check in future GCC release.

test/tbbmalloc/test_malloc_new_handler.cpp
```diff
@@ -33,9 +33,7 @@
 #pragma warning (disable: 4800)
 #endif //#if _MSC_VER
 
-#include "../../src/tbb/tls.h"
-
-tbb::detail::r1::tls<bool> new_handler_called;
```
alexey-katranov, line 38: Did it cause warning as well? If so, do we have a workaround for other places where `r1::tls` is used?

phprus: I not found any other usages of ```r1::tls```.

vlserov: This warning affect only `r1::basic_tls` and `r1::tls` with non-pointer type as template parameter.
There is no other cases of usage `r1::basic_tls` or `r1::tls` with non-pointer type.

alexey-katranov: Thank you for the explanation.

alexey-katranov (approved):

PR #523, opened 2021-08-03
add numa optimization

kononovk:


src/tbb/arena.cpp
```diff
@@ -23,6 +23,7 @@
 #include "oneapi/tbb/detail/_task.h"
 #include "oneapi/tbb/info.h"
 #include "oneapi/tbb/tbb_allocator.h"
+#include "misc.h"
```
pavelkumbrasev, line 26: Move it to library include

src/tbb/arena.cpp
```diff
@@ -90,6 +91,14 @@ template <bool as_worker>
 std::size_t arena::occupy_free_slot(thread_data& tls) {
     // Firstly, external threads try to occupy reserved slots
     std::size_t index = as_worker ? out_of_arena : occupy_free_slot_in_range( tls,  0, my_num_reserved_slots );
+    if (index == out_of_arena && has_numa_optimization()) {
+        std::pair<int, int> my_range = {my_num_reserved_slots, info::default_concurrency()};
```
pavelkumbrasev, line 95: auto + make_pair?

src/tbb/arena.cpp
```diff
@@ -90,6 +91,14 @@ template <bool as_worker>
...
+        auto it = numa_intervals.find(governor::get_my_current_numa_node());
+        if (it != numa_intervals.end())
+            my_range = it->second;
+        index = occupy_free_slot_in_range(tls, max(static_cast<int>(my_num_reserved_slots), my_range.first),
```
pavelkumbrasev, line 99: But my_num_reserved_slots always equal to my_range.first, because you init my_range.first with this value

src/tbb/arena.cpp
```diff
@@ -198,6 +207,22 @@ arena::arena ( market& m, unsigned num_slots, unsigned num_reserved_slots, unsig
     }
     my_fifo_task_stream.initialize(my_num_slots);
     my_resume_task_stream.initialize(my_num_slots);
+
+    if (has_numa_optimization()) {
+        auto first_numa_id = governor::get_my_current_numa_node();
+        auto left = governor::get_numa_cores_count(first_numa_id);
+        numa_intervals[first_numa_id] = {0, left};
```
pavelkumbrasev, line 214: If I understand idea correctly, shouldn't  it `{1, left}`?

src/tbb/arena.h
```diff
@@ -288,6 +289,8 @@ struct arena_base : padded<intrusive_list_node> {
     //! Used to trap accesses to the object after its destruction.
     std::uintptr_t my_guard;
 #endif /* TBB_USE_ASSERT */
+    //! Map from numa nodes to range of slots to this numa node
+    std::unordered_map<numa_node_id, std::pair<int, int>> numa_intervals;
```
pavelkumbrasev, line 293: I would suggest add __TBB_HIERARCHICAL_STEALING_OPTIMIZATION and use inside TBB.
To decrease debug cost. (because class layouts are changed and so on)

src/tbb/arena.h
```diff
@@ -353,6 +356,11 @@ class arena: public padded<arena_base>
         return num_workers_active() > my_num_workers_allotted.load(std::memory_order_relaxed);
     }
 
+    //! Check that numa optimization works for this arena
+    bool has_numa_optimization() const noexcept {
+        return info::numa_nodes().size() > 1 && static_cast<int>(my_num_slots) == info::default_concurrency() && my_num_reserved_slots == 1u;
```
pavelkumbrasev, line 361: Instead info::default_concurrency you should use governor::default_num_threads()

src/tbb/arena.h
```diff
@@ -555,7 +563,56 @@ inline d1::task* arena::steal_task(unsigned arena_index, FastRandom& frnd, execu
         // No slots to steal from
         return nullptr;
     }
-    // Try to steal a task from a random victim.
+
+    std::pair<int, int> range {0, 0};
```
pavelkumbrasev, line 567: Revert this method and write it with masks.

src/tbb/arena.h
```diff
@@ -549,26 +573,37 @@ void arena::advertise_new_work() {
...
 }
 
-inline d1::task* arena::steal_task(unsigned arena_index, FastRandom& frnd, execution_data_ext& ed, isolation_type isolation) {
-    auto slot_num_limit = my_limit.load(std::memory_order_relaxed);
-    if (slot_num_limit == 1) {
-        // No slots to steal from
-        return nullptr;
+/*  Trying to find index in the passed range using pool mask
+    Pre-condition: range.second <= my_limit */
+inline long long arena::get_index_for_steal_in_range(FastRandom& frnd, const std::pair<std::size_t, std::size_t>& range) {
```
ivankochin, line 578:
```suggestion
inline int arena::get_index_for_steal_in_range(FastRandom& frnd, const std::pair<std::size_t, std::size_t>& range) {
```
Why do you return `long long` if `try_steal_from_slot()` accepts the `slot_number` parameter as `int`?

src/tbb/arena.h
```diff
@@ -549,26 +573,37 @@ void arena::advertise_new_work() {
     }
 }
 
-inline d1::task* arena::steal_task(unsigned arena_index, FastRandom& frnd, execution_data_ext& ed, isolation_type isolation) {
-    auto slot_num_limit = my_limit.load(std::memory_order_relaxed);
-    if (slot_num_limit == 1) {
-        // No slots to steal from
-        return nullptr;
+/*  Trying to find index in the passed range using pool mask
+    Pre-condition: range.second <= my_limit */
```
ivankochin, line 577: Why don't check this pre-condition using `__TBB_ASSERT`?

PR #524, opened 2021-08-04
Add pool mask optimization

Ri7ay:


src/tbb/arena.h
```diff
@@ -545,22 +549,43 @@ void arena::advertise_new_work() {
...
+                break;
+            }
+            --k;
+        }
+    }
+
+    if (!find_steal_pool) {
+        return nullptr;
+    }
+
```
pavelkumbrasev, line 583:
```suggestion
    std::size_t target_slot = (frnd.get() % k) + 1;
    std::size_t i = 0; 
    for (; target_slot > 0 && i < slot_num_limit; ++i) {
        if (pool_mask[i].load(std::memory_order_relaxed)) {
            --target_slot;
        }
    }

    if (i < slot_num_limit) {
        k = i;
    }

```

pavelkumbrasev (commented):
I would suggest create `find_steal_slot` method, that should be method` arena`.
And you can call it from arena :: steal_task. So you can put there all find slot logic.

src/tbb/arena.h
```diff
@@ -356,8 +360,11 @@ class arena: public padded<arena_base>
     //! If necessary, raise a flag that there is new job in arena.
     template<arena::new_work_type work_type> void advertise_new_work();
 
+    //! Get gets the index of a non-empty task pool
+    long long get_index_for_steal(FastRandom& frnd, std::size_t slot_num_limit);
```
kononovk, line 364: Why public? Maybe move this function into `steal_task` as lambda?

pavelkumbrasev: 1) I would suggest not expand steal_task size(amount of lines) and create new method.
2) Because my_limit is atomic, so we get snapshot in steal_task and pass it to help method.

kononovk: @pavelkumbrasev why we cant get snapshot in get_index_for_steal ? we dont use it in steal task

src/tbb/arena_slot.cpp
```diff
@@ -25,6 +25,24 @@ namespace r1 {
...
+
+void arena_slot::publish_task_pool(thread_data* td) {
+        __TBB_ASSERT ( task_pool == EmptyTaskPool, "someone else grabbed my arena slot?" );
+        __TBB_ASSERT ( head.load(std::memory_order_relaxed) < tail.load(std::memory_order_relaxed),
+                "entering arena without tasks to share" );
+        // Release signal on behalf of previously spawned tasks (when this thread was not in arena yet)
+        task_pool.store(task_pool_ptr, std::memory_order_release );
+        td->my_arena->pool_mask[td->my_arena_index].store(1, std::memory_order_relaxed);
+    }
+
```
kononovk, line 44: code style

PR #532, opened 2021-08-10
CMake: suppress alignment warning for ARM64-MSVC

zheltovs:
Signed-off-by: Zheltov, Sergey1 <sergey1.zheltov@intel.com>


pavelkumbrasev (commented):
There is no possibility that someone will get an error because of the substitution of this flag in the compilation?

cmake/compilers/MSVC.cmake
```diff
@@ -21,6 +21,9 @@ string(REGEX REPLACE "/W[0-4]" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
 
 # Warning suppression C4324: structure was padded due to alignment specifier
 set(TBB_WARNING_LEVEL $<$<NOT:$<CXX_COMPILER_ID:Intel>>:/W4> $<$<BOOL:${TBB_STRICT}>:/WX>)
+if ("${MSVC_CXX_ARCHITECTURE_ID}" STREQUAL "ARM64")
+    set(TBB_ALIGNMENT_WARNING_SUPPRESS /wd4324)
+endif()
 set(TBB_WARNING_SUPPRESS /wd4324)
```
alexey-katranov, line 27: Consider the following approach:
```suggestion
set(TBB_ALIGNMENT_WARNING_SUPPRESS /wd4324)
if ("${MSVC_CXX_ARCHITECTURE_ID}" STREQUAL "ARM64")
    set(TBB_PUBLIC_WARNING_SUPPRESS ${TBB_ALIGNMENT_WARNING_SUPPRESS})
else()
    set(TBB_WARNING_SUPPRESS ${TBB_ALIGNMENT_WARNING_SUPPRESS})
endif()
```

cmake/compilers/MSVC.cmake
```diff
@@ -21,6 +21,9 @@ string(REGEX REPLACE "/W[0-4]" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
 
 # Warning suppression C4324: structure was padded due to alignment specifier
```
alexey-katranov, line 22: Move this line to be just before `set(TBB_WARNING_SUPPRESS /wd4324)`

alexey-katranov (approved):
Consider updating release notes.

test/CMakeLists.txt
```diff
@@ -33,6 +33,7 @@ function(tbb_add_test)
         PRIVATE
         ${TBB_CXX_STD_FLAG}
         ${TBB_WARNING_LEVEL}
+        ${TBB_ALIGNMENT_WARNING_SUPPRESS}
```
AlexVeprev, line 36: Just as idea how to get rid of this variable:
```suggestion
        # Warning suppression C4324 for ARM64: structure was padded due to alignment specifier
        $<$<STREQUAL:${MSVC_CXX_ARCHITECTURE_ID},ARM64>:/wd4325>
```

Is not it too complicated? :)

zheltovs: It really looks better, thanks for the suggestion!

PR #533, opened 2021-08-12
Add statistics collection

kononovk:
Add statistics collection in TBB

Configure cmake: `cmake .. -DCMAKE_CXX_FLAGS="-D__TBB_STATISTICS=1"`
Example output:
```
executed  try get  got  try steal  stolen  empty vic pool
16        16       15   0          0       0
0         0        0    0          0       0
26        26       22   6          4       0
18        18       16   2          2       0
SUMMARY:
60        60       53   8          6       0
```
Configure cmake: `cmake .. -DCMAKE_CXX_FLAGS="-D__TBB_STATISTICS_SUM=1"`
Example output:
```
executed  try get  got  try steal  stolen  empty vic pool
95        95       71   31         23      0
```


src/tbb/governor.h
```diff
@@ -148,6 +155,69 @@ class governor {
...
+        void print_statistics() {
+            int offset = 2;
+            /* std::string not so bad because of Small String Optimization */
+            const static std::string column_names[] = {
+                "executed", "try get",
+                "got", "try steal", "stolen",
+                "empty vic pool",
+            };
+            for (auto& el : column_names) {
+                std::cout << std::left << std::setw(el.size() + offset) << el;
```
kononovk, line 192: This is not very convenient, that we should remember order in this array, to solve this problem we can add a part of magic enum in utils, maybe : https://github.com/Neargye/magic_enum - its statically convert enum into string using `__PRETTY_FUNCTION__` compilers extension

src/tbb/arena_slot.cpp
```diff
@@ -142,12 +147,22 @@ d1::task* arena_slot::get_task(execution_data_ext& ed, isolation_type isolation)
 
     __TBB_ASSERT( (std::intptr_t)tail.load(std::memory_order_relaxed) >= 0, nullptr );
     __TBB_ASSERT( result || tasks_omitted || is_quiescent_local_task_pool_reset(), nullptr );
+
+    if (result) {
```
kononovk, line 151: To compiler optimization when `__TBB_STATISTICS==0`:
```suggestion
    if (__TBB_STATISTICS && result) {
```

src/tbb/governor.h
```diff
@@ -148,6 +155,69 @@ class governor {
...
+#if !__TBB_STATISTICS_SUM
+            std::cout << "SUMMARY:" << std::endl;
+#endif
+            for (std::size_t k = 0; k < summary.size(); ++k) {
+                std::cout << std::left << std::setw(column_names[k].size() + offset) << summary[k];
+            }
+            std::cout << std::endl;
+        }
+        std::mutex m_;
+        std::vector<array_stat> accumulate_statistics {};
```
kononovk, line 216: - naming
- use `private`

test/tbb/test_statistics.cpp
```diff
@@ -0,0 +1,41 @@
...
+    tbb::task_scheduler_handle handler = tbb::task_scheduler_handle::get();
+
+    tbb::parallel_for(tbb::blocked_range<int>(0, values.size()),
+                       [&](tbb::blocked_range<int> r) {
+        for (int i = r.begin(); i < r.end(); ++i) {
+            values[i] = i;
+        }
+    });
+    tbb::finalize(handler);
+    std::cout << std::flush;
```
kononovk, line 39: We don't need iostream here

test/tbb/test_statistics.cpp
```diff
@@ -0,0 +1,41 @@
...
+#include "common/test.h"
+
+
+//! \file test_statistics.cpp
+//! \brief Test for [internal] functionality
+TEST_CASE("Simple") {
+    auto values = std::vector<double>(1000000, 0);
+    tbb::task_scheduler_handle handler = tbb::task_scheduler_handle::get();
+
+    tbb::parallel_for(tbb::blocked_range<int>(0, values.size()),
```
kononovk, line 32: `int` with `size_t`

src/tbb/governor.h
```diff
@@ -148,6 +155,70 @@ class governor {
...
+    enum class statistics : std::size_t {
+        executed = 0,
+        try_get,
+        got,
+        try_steal,
+        stolen,
+        empty_vic_pool,
+        size,
+    };
+    using array_stat = std::array<std::uint32_t, static_cast<std::size_t>(statistics::size)>;
```
ivankochin, line 169: What about using `std::map<statistics, std::size_t>` instead of `std::array` to store results?

src/tbb/governor.h
```diff
@@ -148,6 +155,70 @@ class governor {
         return false;
 #endif
     }
+
+#if __TBB_STATISTICS
+    enum class statistics : std::size_t {
```
ivankochin, line 160:
```suggestion
    enum class statistic_type : std::size_t {
```

src/tbb/governor.h
ivankochin: I think that using a third-party solution is overcomplicated for this task, what do you think about storing the strings inside the map where the keys are enum values and the values are strings?

src/tbb/governor.h
```diff
@@ -148,6 +155,70 @@ class governor {
...
+        try_get,
+        got,
+        try_steal,
+        stolen,
+        empty_vic_pool,
+        size,
+    };
+    using array_stat = std::array<std::uint32_t, static_cast<std::size_t>(statistics::size)>;
+
+    class collect_statistics : no_copy {
```
ivankochin, line 171:
```suggestion
    class statistic_holder : no_copy {
```

src/tbb/governor.h
```diff
@@ -148,6 +155,70 @@ class governor {
...
+            std::cout << "SUMMARY:" << std::endl;
+#endif
+            for (std::size_t k = 0; k < summary.size(); ++k) {
+                std::cout << std::left << std::setw(column_names[k].size() + offset) << summary[k];
+            }
+            std::cout << std::endl;
+        }
+    private:
+        std::mutex m_;
+        std::vector<array_stat> accumulate_statistics_ {};
```
ivankochin, line 217: Usually, we don't use underscore suffix to mark the member as private. Please consider different naming. I would propose `statistics_collection` or `statistics_container`.

src/tbb/governor.h
ivankochin: I think that using a third-party component is an overcomplicated solution for this purpose. What do you think about storing the strings inside a map where the keys are the corresponding enum values?

src/tbb/governor.h
kononovk: By they are exactly the same by API, but array has no extra allocations, so statistics will not slow the program and will be more correct

PR #541, opened 2021-08-20
Fix using tbb::this_task_arena::current_thread_index() with uninitilized arena

pavelkumbrasev:
Replaced `int` on `slot_id` to avoid problem with uninitilized arena.

### DIFF
--- a/src/tbb/task_dispatcher.cpp
+++ b/src/tbb/task_dispatcher.cpp
@@ -133,7 +133,7 @@ d1::slot_id __TBB_EXPORTED_FUNC execution_slot(const d1::execution_data* ed) {
         return ed_ext->task_disp->m_thread_data->my_arena_index;
     } else {
         thread_data* td = governor::get_thread_data_if_initialized();
-        return td ? int(td->my_arena_index) : -1;
+        return td ? td->my_arena_index : -1;
     }
 }
 
--- a/src/tbb/task_dispatcher.cpp
+++ b/src/tbb/task_dispatcher.cpp
@@ -133,7 +133,7 @@ d1::slot_id __TBB_EXPORTED_FUNC execution_slot(const d1::execution_data* ed) {
         return ed_ext->task_disp->m_thread_data->my_arena_index;
     } else {
         thread_data* td = governor::get_thread_data_if_initialized();
-        return td ? td->my_arena_index : -1;
+        return td ? td->my_arena_index : d1::slot_id(-1);
     }
 }
 
--- a/test/tbb/test_task_arena.cpp
+++ b/test/tbb/test_task_arena.cpp
@@ -1752,7 +1752,7 @@ struct enqueue_test_helper {
 //! Test for uninitilized arena
 //! \brief \ref requirement \ref error_guessing
 TEST_CASE("Test current_thread_index") {
-    CHECK(tbb::this_task_arena::current_thread_index() < 0);
+    CHECK((tbb::this_task_arena::current_thread_index() == tbb::task_arena::not_initialized));
 }
 
 // This test requires TBB in an uninitialized state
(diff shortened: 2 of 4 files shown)


isaevil:
Could you add a description?

test/tbb/test_task_arena.cpp
```diff
@@ -1748,7 +1748,13 @@ struct enqueue_test_helper {
 
 //--------------------------------------------------//
 
-// This test should be first
+//! Test for uninitilized arena
+//! \brief \ref requirement \ref error_guessing
+TEST_CASE("Test current_thread_index") {
+    CHECK(tbb::this_task_arena::current_thread_index() < 0);
```
alexey-katranov, line 1754:
```suggestion
    CHECK(tbb::this_task_arena::current_thread_index() == task_arena::not_initialized);
```

test/tbb/test_task_arena.cpp
```diff
@@ -1748,7 +1748,13 @@ struct enqueue_test_helper {
 
 //--------------------------------------------------//
 
-// This test should be first
+//! Test for uninitilized arena
```
alexey-katranov, line 1751:
```suggestion
// This test requires TBB in an uninitialized state
//! Test for uninitilized arena
```

test/tbb/test_task_arena.cpp
pavelkumbrasev: I tried to do it but it ended with `/test/tbb/test_task_arena.cpp:1754: undefined reference to `tbb::detail::d1::task_arena_base::not_initialized'`

alexey-katranov (dismissed):

isaevil (approved):

test/tbb/test_task_arena.cpp
```diff
@@ -1748,7 +1748,13 @@ struct enqueue_test_helper {
...
+//! \brief \ref requirement \ref error_guessing
+TEST_CASE("Test current_thread_index") {
```
ivankochin, line 1754: As I understand this test should be performed first to work correctly. But I cannot find information about test cases order inside the DocTest documentation. What do you think about adding some protection mechanism which will define the cases in the right order or just showing the message and fails if the test cases order is incorrect?

As I understand, the order of test cases may be set using [subcases](https://github.com/onqtam/doctest/blob/master/doc/markdown/tutorial.md#test-cases-and-subcases).

ivankochin (commented):
https://github.com/oneapi-src/oneTBB/blob/master/src/tbb/task_dispatcher.cpp#L136

In this line, we assigning `-1` to the `unsigned short` variable. I know that it's a valid scenario, and the value just will be equal to `std::numeric_limits<unsigned short>::max()` but may it would be convenient to make the `slot_id` type signed? Does it breaks the backward compatibility?

Also I think that in this line we shouldn't convert the `td->my_arena_index` to `int`.

test/tbb/test_task_arena.cpp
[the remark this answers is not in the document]
pavelkumbrasev: It is error_guessing, I would prefer add it to tests.

test/tbb/test_task_arena.cpp
pavelkumbrasev: I added error message: "TBB was initialized".

ivankochin (dismissed):

src/tbb/task_dispatcher.cpp
```diff
@@ -133,7 +133,7 @@ d1::slot_id __TBB_EXPORTED_FUNC execution_slot(const d1::execution_data* ed) {
         return ed_ext->task_disp->m_thread_data->my_arena_index;
     } else {
         thread_data* td = governor::get_thread_data_if_initialized();
-        return td ? int(td->my_arena_index) : -1;
+        return td ? td->my_arena_index : -1;
```
ivankochin, line 136:
```suggestion
        return td ? td->my_arena_index : d1::slot_id(-1);
```

ivankochin (approved):

alexey-katranov (approved):

PR #544, opened 2021-08-24
Cover the constraints_threads_per_core() reserved interface by testing

ivankochin:
Now we have the uncovered entry-point inside the `tbb::info` namespace. This request test this function to improve code coverage.

### DIFF
--- a/test/tbb/test_arena_constraints.cpp
+++ b/test/tbb/test_arena_constraints.cpp
@@ -217,5 +217,5 @@ TEST_CASE("Test concurrency getters output for constraints with custom concurren
 //! \brief \ref interface \ref error_guessing
 TEST_CASE("Testing constraints_threads_per_core() reserved entry point") {
     tbb::task_arena::constraints c{};
-    REQUIRE(tbb::detail::r1::constraints_threads_per_core(c) == tbb::task_arena::automatic);
+    REQUIRE((tbb::detail::r1::constraints_threads_per_core(c) == tbb::task_arena::automatic));
 }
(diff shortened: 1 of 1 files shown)


test/tbb/test_arena_constraints.cpp
```diff
@@ -212,3 +212,10 @@ TEST_CASE("Test concurrency getters output for constraints with custom concurren
     c.set_max_threads_per_core(1);
     check_concurrency_level(c);
 }
+
+//! Testing constraints_threads_per_core() reserved entry point
+//! \brief \ref interface \ref error_guessing
+TEST_CASE("Testing constraints_threads_per_core() reserved entry point") {
+    tbb::task_arena::constraints c{};
+    REQUIRE(tbb::detail::r1::constraints_threads_per_core(c) == tbb::task_arena::automatic);
```
alexey-katranov, line 220: Do we expect that this behavior will be preserved in future that `constraints_threads_per_core` will return `automatic` for the default constructed constraints?

ivankochin: No, I think it will return something like a maximal number of threads per core for the given constraints.

alexey-katranov: It seems that the test will not be backward compatible in that case. Consider just calling this entry point.

test/tbb/test_arena_constraints.cpp
```diff
@@ -212,3 +212,10 @@ TEST_CASE("Test concurrency getters output for constraints with custom concurren
     c.set_max_threads_per_core(1);
     check_concurrency_level(c);
 }
+
+//! Testing constraints_threads_per_core() reserved entry point
+//! \brief \ref interface \ref error_guessing
```
alexey-katranov, line 217:
```suggestion
//! \brief \ref error_guessing
```
It is some internal stuff, not interface.

alexey-katranov (approved):

PR #553, opened 2021-08-30
Fix stack overflow protection mechanism

alexey-katranov:
The stack size is incorrectly supposed to be 4 MB on Windows. However, MSDN [states](https://docs.microsoft.com/en-us/windows/win32/procthread/thread-stack-size) that "_The default stack reservation size used by the linker is 1 MB_". It leads to two issues:

1. The stack overflow protection mechanism does not work on Windows properly, i.e. stealing can be allowed even if the stack is fully consumed.
2. The [stack anchor calculation](https://github.com/oneapi-src/oneTBB/blob/35147e0021727916c789287ba52a3499eb3ab348/src/tbb/scheduler_common.h#L490-L492) might overflow and block the stealing completely (that is reproduced in https://github.com/oneapi-src/oneTBB/issues/478#issuecomment-891669250) due to main thread stack mapping on low address).

The fix changes the assumption to 1 MB on Windows. If compiled with modern Windows, the system API is used to query stack size instead of the assumption.

### DIFF
--- a/src/tbb/global_control.cpp
+++ b/src/tbb/global_control.cpp
@@ -93,7 +93,7 @@ class alignas(max_nfs_size) stack_size_control : public control_storage {
         static auto ThreadStackSizeDefault = [] {
             ULONG_PTR hi, lo;
             GetCurrentThreadStackLimits(&lo, &hi);
-            return std::size_t(hi - lo);
+            return hi - lo;
         }();
         return ThreadStackSizeDefault;
 #else
--- a/src/tbb/global_control.cpp
+++ b/src/tbb/global_control.cpp
@@ -89,7 +89,16 @@ public:
 
 class alignas(max_nfs_size) stack_size_control : public control_storage {
     std::size_t default_value() const override {
+#if _WIN32_WINNT >= 0x0602 /* _WIN32_WINNT_WIN8 */
+        static auto ThreadStackSizeDefault = [] {
+            ULONG_PTR hi, lo;
+            GetCurrentThreadStackLimits(&lo, &hi);
+            return std::size_t(hi - lo);
+        }();
+        return ThreadStackSizeDefault;
+#else
         return ThreadStackSize;
+#endif
     }
     void apply_active(std::size_t new_active) override {
         control_storage::apply_active(new_active);
--- a/src/tbb/governor.cpp
+++ b/src/tbb/governor.cpp
@@ -132,11 +132,11 @@ void governor::one_time_init() {
 static std::uintptr_t get_stack_base(std::size_t stack_size) {
     // Stacks are growing top-down. Highest address is called "stack base",
     // and the lowest is "stack limit".
-#if USE_WINTHREAD
+#if __TBB_USE_WINAPI
     suppress_unused_warning(stack_size);
     NT_TIB* pteb = (NT_TIB*)NtCurrentTeb();
     __TBB_ASSERT(&pteb < pteb->StackBase && &pteb > pteb->StackLimit, "invalid stack info in TEB");
-    return pteb->StackBase;
+    return std::uintptr_t(pteb->StackBase);
 #else /* USE_PTHREAD */
     // There is no portable way to get stack base address in Posix, so we use
     // non-portable method (on all modern Linux) or the simplified approach
--- a/src/tbb/governor.cpp
+++ b/src/tbb/governor.cpp
@@ -136,7 +136,7 @@ static std::uintptr_t get_stack_base(std::size_t stack_size) {
     suppress_unused_warning(stack_size);
     NT_TIB* pteb = (NT_TIB*)NtCurrentTeb();
     __TBB_ASSERT(&pteb < pteb->StackBase && &pteb > pteb->StackLimit, "invalid stack info in TEB");
-    return std::uintptr_t(pteb->StackBase);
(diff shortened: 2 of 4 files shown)


phprus:
Is there any progress?

alexey-katranov:
Thank you for the reminder, unfortunately, there is no any progress.

src/tbb/global_control.cpp
```diff
@@ -89,7 +89,16 @@ class alignas(max_nfs_size) allowed_parallelism_control : public control_storage
 
 class alignas(max_nfs_size) stack_size_control : public control_storage {
     virtual std::size_t default_value() const override {
+#if _WIN32_WINNT >= 0x0602 /* _WIN32_WINNT_WIN8 */
+        static auto ThreadStackSizeDefault = [] {
+            ULONG_PTR hi, lo;
+            GetCurrentThreadStackLimits(&lo, &hi);
+            return std::size_t(hi - lo);
```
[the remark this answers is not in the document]
alexey-katranov: While it seems we can emulate `GetCurrentThreadStackLimits` via [`TIB`](https://en.wikipedia.org/wiki/Win32_Thread_Information_Block), we should use `DeallocationStack` (not 
`StackLimit`) for the real stack address. The implementation `GetCurrentThreadStackLimits` does the same (x64 implementation):
```
mov         r8,qword ptr gs:[30h]      // get TIB
mov         rax,qword ptr [r8+1478h]   // TIB + 1478h is DeallocationStack
mov         qword ptr [rcx],rax        // return LowLimit
mov         rax,qword ptr [r8+8]       // TIB + 8h is StackBase
mov         qword ptr [rdx],rax        // return HighLimit
ret
```
So, on x86_32 the `DeallocationStack` offset will be another and it seems the current approach is the lesser of two evils.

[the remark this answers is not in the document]
phprus: I had no crashes on tests on Windows 7 (x86 and x64) after applying this patch.
The simple 1 MB limit works in all my Windows 7 environments.

src/tbb/governor.cpp
```diff
@@ -132,11 +132,11 @@ void governor::one_time_init() {
 static std::uintptr_t get_stack_base(std::size_t stack_size) {
     // Stacks are growing top-down. Highest address is called "stack base",
     // and the lowest is "stack limit".
-#if USE_WINTHREAD
+#if __TBB_USE_WINAPI
     suppress_unused_warning(stack_size);
     NT_TIB* pteb = (NT_TIB*)NtCurrentTeb();
     __TBB_ASSERT(&pteb < pteb->StackBase && &pteb > pteb->StackLimit, "invalid stack info in TEB");
-    return pteb->StackBase;
+    return std::uintptr_t(pteb->StackBase);
```
anton-potapov, line 139:
```suggestion
    return reinterpret_cast<std::uintptr_t>(pteb->StackBase);
```

src/tbb/governor.cpp
```diff
@@ -132,11 +132,11 @@ void governor::one_time_init() {
 static std::uintptr_t get_stack_base(std::size_t stack_size) {
     // Stacks are growing top-down. Highest address is called "stack base",
     // and the lowest is "stack limit".
-#if USE_WINTHREAD
```
anton-potapov, line 135: It seems that this macro no longer defined by oneTBB build system (while classic TBB [does define it](https://github.com/oneapi-src/oneTBB/blob/eca91f16d7490a8abfdee652dadf457ec820cc37/build/windows.cl.inc#L119-L120))

src/tbb/governor.cpp
```diff
@@ -132,11 +132,11 @@ void governor::one_time_init() {
...
+#if __TBB_USE_WINAPI
     suppress_unused_warning(stack_size);
     NT_TIB* pteb = (NT_TIB*)NtCurrentTeb();
     __TBB_ASSERT(&pteb < pteb->StackBase && &pteb > pteb->StackLimit, "invalid stack info in TEB");
-    return pteb->StackBase;
+    return std::uintptr_t(pteb->StackBase);
 #else /* USE_PTHREAD */
```
anton-potapov, line 140:
```suggestion
#else /* __TBB_USE_POSIX */
```

anton-potapov: please change macro name at line 167 accordingly

src/tbb/global_control.cpp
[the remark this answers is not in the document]
anton-potapov:
```suggestion
            ULONG_PTR hi, lo;
            GetCurrentThreadStackLimits(&lo, &hi);
            return static_cast<std::size_t>(hi - lo);
```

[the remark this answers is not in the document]
anton-potapov: I believe type cast is not needed here at whole, but if you wish

src/tbb/governor.cpp
alexey-katranov: It is never used in oneTBB.

src/tbb/governor.cpp
alexey-katranov: I would remove this comment. Moreover the comment on line 167 is misleading because `endif` comment usually contains `if` condition.

anton-potapov (approved):
**LGTM**

PR #556, opened 2021-08-31
[DOC]: New guidelines

alexandraepan:
Update index.rst and README.md, new INSTALL.md, and fix typo in CONTRIBUTING.md. 

Signed-off-by: Alexandra Epanchinzeva alexandra.epanchinzeva@intel.com


INSTALL.md
```diff
@@ -0,0 +1,34 @@
+# Installation from Sources
+
+Required Software:
```
omalyshe, line 3: I think we should move build environment recommendations to another topic, probably where the library build is discussed. 
Also, we should decide if we want to provide the full list of supported (aka tested) systems or minimal requirements. 
@AlexVeprev @alexey-katranov

alexandraepan:
Deleted changes from index.rst to create a separate pr.

INSTALL.md
alexandraepan: Makes sense! I can delete it from there and update a PR with System Requirements topic.

PR #569, opened 2021-09-07
Workaround gcc bug in test_limiter_node

Iliamish:
Signed-off-by: Mishin, Ilya <ilya.mishin@intel.com>

### Description 
GNU compiler generates incorrect code for the call of limiter.register_successor in release (-03)
The function pointer to make_edge workarounds the issue.

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [ ] No
- [x] Unknown

### Other information
Linked issue https://github.com/oneapi-src/oneTBB/issues/342

### DIFF
--- a/test/tbb/test_limiter_node.cpp
+++ b/test/tbb/test_limiter_node.cpp
@@ -445,7 +445,14 @@ void test_try_put_without_successors() {
     int try_put_num{3};
     tbb::flow::buffer_node<int> bn(g);
     tbb::flow::limiter_node<int> ln(g, try_put_num);
-    tbb::flow::make_edge(bn, ln);
+#if __GNUC__ && __GNUC__ < 12 && !TBB_USE_DEBUG
+    // Seemingly, GNU compiler generates incorrect code for the call of limiter.register_successor in release (-03)
+    // The function pointer to make_edge workarounds the issue for unknown reason
+    auto make_edge_ptr = tbb::flow::make_edge<int>;
+    make_edge_ptr(bn, ln); //putting the successor back
+#else
+    tbb::flow::make_edge(bn, ln); //putting the successor back
+#endif
     int i = 1;
     for (; i <= try_put_num; i++)
         bn.try_put(i);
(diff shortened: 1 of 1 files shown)


test/tbb/test_limiter_node.cpp
```diff
@@ -445,7 +445,14 @@ void test_try_put_without_successors() {
     int try_put_num{3};
     tbb::flow::buffer_node<int> bn(g);
     tbb::flow::limiter_node<int> ln(g, try_put_num);
-    tbb::flow::make_edge(bn, ln);
+#if __GNUC__ && __GNUC__ < 12 && !TBB_USE_DEBUG
+    // Seemingly, GNU compiler generates incorrect code for the call of limiter.register_successor in release (-03)
+    // The function pointer to make_edge workarounds the issue for unknown reason
+    auto make_edge_ptr = tbb::flow::make_edge<int>;
+    make_edge_ptr(bn, ln); //putting the successor back
+#else
```
ivankochin, line 453: What about unifying this workaround with [this code](https://github.com/oneapi-src/oneTBB/blob/master/test/tbb/test_limiter_node.cpp#L358) by unifying into a separate function?

test/tbb/test_limiter_node.cpp
```diff
@@ -445,7 +445,14 @@ void test_try_put_without_successors() {
     int try_put_num{3};
     tbb::flow::buffer_node<int> bn(g);
     tbb::flow::limiter_node<int> ln(g, try_put_num);
-    tbb::flow::make_edge(bn, ln);
+#if __GNUC__ && __GNUC__ < 12 && !TBB_USE_DEBUG
```
ivankochin, line 448: And do we really need the first check?

```suggestion
#if __GNUC__ < 12 && !TBB_USE_DEBUG
```

test/tbb/test_limiter_node.cpp
```diff
@@ -445,7 +450,9 @@ void test_try_put_without_successors() {
     int try_put_num{3};
     tbb::flow::buffer_node<int> bn(g);
     tbb::flow::limiter_node<int> ln(g, try_put_num);
+
     tbb::flow::make_edge(bn, ln);
+
```
ivankochin, line 455: Why don't you place the `make_edge_impl` here?

Iliamish: We have bug in case where limiter_node is predecessor. But in this case limiter_node is successor. In previous commit I don't sight it.

phprus: This PR fixes the issue.
1000 launches completed successfully in my tests on CentOS 7 + GCC 7.4.

Iliamish: Sounds great!

ivankochin (approved):

PR #571, opened 2021-09-10
Remove old tms configurations

outoftardis:
This information is outdated because there is now a `oneapi-tbb` section specifically for oneTBB docs. Having this script here causing an error because the same scripts are loading twice:

image


outoftardis:
@omalyshe could you please review?

omalyshe (approved):

outoftardis:
@omalyshe could you merge this as well? I don't have write access

PR #576, opened 2021-09-14
Move enumerable_thread_specific Windows.h usage to cpp

### LINKED ISSUE #573: Remove <Windows.h> include from headers
The tbb currently [includes `Windows.h` in `enumerable_thread_specific.h`](https://github.com/oneapi-src/oneTBB/blob/4624e4ff597fc8c9d205f015a966ee6b61dc860c/include/oneapi/tbb/enumerable_thread_specific.h#L39). Through [transitive includes](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/oneapi-src/oneTBB%24+%28windows.h+OR+enumerable_thread_specific.h+OR+_concurrent_skip_list.h+OR+combinable.h+OR+concurrent_map.h+OR+concurrent_set.h+OR+_concurrent_skip_list.h%29+file:include&patternType=literal), this also affects `combinable.h`, `concurrent_map.h` and `concurrent_set.h`.

Since windows.h is massive and declares thousands of symbols, this include has significantly negative impact on build times. The windows include does not need to be in a public header. We should either:
* Copy the declarations of the few required symbols from windows.h or
* Move the calls to the Win32 API to a cpp

Chronial:
### Description 
This change gets rid of the Windows.h include in the public `enumerable_thread_specific .h` by moving it a cpp.

Fixes #573

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/include/oneapi/tbb/enumerable_thread_specific.h
+++ b/include/oneapi/tbb/enumerable_thread_specific.h
@@ -35,9 +35,7 @@
 
 #include "task.h" // for task::suspend_point
 
-#if _WIN32 || _WIN64
-#include <windows.h>
-#else
+#if !(_WIN32 || _WIN64)
 #include <pthread.h>
 #endif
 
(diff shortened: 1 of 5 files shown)


alexey-katranov (commented):
The patch raises two questions:
1. Can it affect performance?
2. Are we ready to pay maintenance cost for additional entry points?

Before we answering above questions, can we consider the option to use `processthreadsapi.h` (or `fibersapi.h` with `__TBB_WIN8UI_SUPPORT`)?

Chronial:
> Can it affect performance?

The `Tls*` and `Fls*` functions are themselves dllimport, so I wouldn't expect this change to have significant impact.

>  Before answering above questions, can we consider the option to use processthreadsapi.h (or fibersapi.h with __TBB_WIN8UI_SUPPORT)?

I'm not sure whether you're supposed to include `processthreadsapi.h` without `windows.h`. If we want to do that, we would have replicate the [architecture detection block](https://github.com/tpn/winsdk-10/blob/master/Include/10.0.10240.0/um/Windows.h#L106-L134) from `windows.h`, because `processthreadsapi.h` needs one of these symbols to be defined.

If you would prefer that, I can also create a PR for using `processthreadsapi.h`. But I do feel a bit uneasy about copying that architecture code.

anton-potapov:
@Chronial is there any estimates or real measurements of build times reduction effect of this patch ?

Chronial:
This change won't have a massive impact on our build times. In places where this is a serious issue, `enumerable_thread_specific.h` is simply added to the precompiled headers.

This is more about usability:
* `Windows.h` includes are affected by various macros, so it should be included before some and after some other headers. This now affects `enumerable_thread_specific.h`, which makes it more error-prone to use.
* `Windows.h` defines countless names as macros, like `SetEvent`, `SetWindowText`, `RegisterClass` and many more. This makes these identifiers unusable in any file that includes Windows.h. This doesn't just affect global names, but also as method names and names in any namespace. This property is now inherited by `enumerable_thread_specific`.
* `Windows.h` notoriously defines the names `min` and `max` as macros, breaking `std::min` and `std::max` [with nice error messages](https://stackoverflow.com/questions/13416418/define-nominmax-using-stdmin-max).
* Due to the points above and the size of Windows.h, you are strongly discouraged from including `enumerable_thread_specific` in any of our headers. So enumerable_thread_specific has to be hidden with pimpl or simply can't be used in templates.

Chronial:
@alexey-katranov Is there anything I can do to move this forward?

Chronial:
ping :)

anton-potapov:
@pavelkumbrasev , I believe now it's your chance to review it :)

Chronial:
Superseded by #1932

PR #618, opened 2021-10-13
Fix build with MinGW

jschueller:
Just a few fixes to be able to build on mingw (successfully tested on i686 and x86_64 / gcc 11.2 / mingw 9.0.0):
- on i686 sse2 should be explicitely enabled to use the sse intrinsics (_mm_getcsr) and al (fine on x86_64)
- windows.h should be lowercase if you compile from linux
- @longnguyen2004 I had to drop your `__STRICT_ANSI__` undefine, dont you get a warning with that ?
If its due to the gcc version I'm using gcc 11, maybe you're using some old version ?
- dont redefine NOMINAX (mingw headers redefine it)

### DIFF
--- a/src/tbbbind/tbb_bind.cpp
+++ b/src/tbbbind/tbb_bind.cpp
@@ -405,7 +405,7 @@ class binding_handler {
     affinity_masks_container affinity_backup;
     system_topology::affinity_mask handler_affinity_mask;
 
-#if WIN32
+#ifdef _WIN32
     affinity_masks_container affinity_buffer;
     int my_numa_node_id;
     int my_core_type_id;
@@ -415,7 +415,7 @@ class binding_handler {
 public:
     binding_handler( std::size_t size, int numa_node_id, int core_type_id, int max_threads_per_core )
         : affinity_backup(size)
-#if WIN32
+#ifdef _WIN32
         , affinity_buffer(size)
         , my_numa_node_id(numa_node_id)
         , my_core_type_id(core_type_id)
@@ -424,7 +424,7 @@ public:
     {
         for (std::size_t i = 0; i < size; ++i) {
             affinity_backup[i] = system_topology::instance().allocate_process_affinity_mask();
-#if WIN32
+#ifdef _WIN32
             affinity_buffer[i] = system_topology::instance().allocate_process_affinity_mask();
 #endif
         }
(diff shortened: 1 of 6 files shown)


jschueller:
thanks

anton-potapov:
@longnguyen2004 you opinion,as the original port author, is highly required whether patch should be merged. Please be so kind to provide the feedback.

longnguyen2004:
This LGTM, let's merge it

longnguyen2004 (approved):

jschueller:
thanks

MehdiChinoune:
@jschueller Could you amend the changes in #642 to your PR.

jschueller:
@MehdiChinoune yes #642 looks good, done

MehdiChinoune:
> @MehdiChinoune yes #642 looks good, done

Thanks

src/tbbbind/tbb_bind.cpp
```diff
@@ -405,7 +405,7 @@ class binding_handler {
     affinity_masks_container affinity_backup;
     system_topology::affinity_mask handler_affinity_mask;
 
-#if WIN32
+#ifdef _WIN32
```
alexey-katranov, line 408: In other places `#if` is preferred over `#ifdef` (even for `_WIN32`), is it critical to have `#ifdef`?

jschueller: ifdef is best to avoid warnings when its undefined on other platforms but I dont think its critical

MehdiChinoune: > ifdef is best to avoid warnings when its undefined on other platforms but I dont think its critical

It was an error not a warning.
If they prefer #if then they should include "windows.h" (or whatever the file that define WIN32=1), I think!

alexey-katranov: > ifdef is best to avoid warnings when its undefined on other platforms but I dont think its critical

Are all the warnings are eliminated with the patch? (it seems there are a lot of `#if something_is_undefined` expressions across the code base)

jschueller: no, you will see that tbb throws a lot of warnings if you enable the gcc warning "-Wundef"
but at least it will build for mingw

alexey-katranov (dismissed):
LGTM, @anton-potapov?

anton-potapov (dismissed):
**LGTM**

anton-potapov:
@jschueller please add "signed-of-by" to git commit message , as described in the [How to contribute]( https://github.com/oneapi-src/oneTBB/blob/master/CONTRIBUTING.md#pull-requests)

jschueller:
ok, done

anton-potapov (approved):

PR #637, opened 2021-11-02
Update spin_mutex implementation lock implementation

vinser52:
### Description 
Reducing cache coherency traffic under contention

### Type of change

- [x ] bug fix - _change which fixes an issue_

### Tests

- [x ] not needed

### Documentation

- [ ] needs to be updated
- [x ] not needed

- [ ] Yes
- [x ] No
- [ ] Unknown

### DIFF
--- a/include/oneapi/tbb/spin_mutex.h
+++ b/include/oneapi/tbb/spin_mutex.h
@@ -66,9 +66,11 @@ public:
     //! Acquire lock
     /** Spin if the lock is taken */
     void lock() {
-        atomic_backoff backoff;
         call_itt_notify(prepare, this);
-        while (m_flag.exchange(true)) backoff.pause();
+        for (atomic_backoff backoff; ; backoff.pause()) {
+            if(!m_flag.load(std::memory_order_relaxed) && !m_flag.exchange(true))
+                break;
+        }
         call_itt_notify(acquired, this);
     }
 


include/oneapi/tbb/spin_mutex.h
```diff
@@ -66,9 +66,11 @@ class spin_mutex {
     //! Acquire lock
     /** Spin if the lock is taken */
     void lock() {
-        atomic_backoff backoff;
         call_itt_notify(prepare, this);
-        while (m_flag.exchange(true)) backoff.pause();
+        for (atomic_backoff backoff; ; backoff.pause()) {
+            if(!m_flag.load(std::memory_order_relaxed) && !m_flag.exchange(true))
```
pavelkumbrasev, line 71: Extra read add overhead in non-contended cases(where using of spin_mutex is justified). In contended cases is better to use tbb::mutex

pavelkumbrasev: @alexey-katranov what do you think?

alexey-katranov: In almost all cases `tbb::mutex` should be considered because it is not expected pathetically degrade like `tbb::spin_mutex` with heavy contention (even with hundreds of threads). While `tbb::spin_mutex` has several ticks less overhead that might be useful in non-contented cases (e.g. when only few threads work with this mutex), however, I do not know any cases when `tbb::spin_mutex` is better than `tbb::mutex`. I think we can consider deprecation of `tbb::spin_mutex`.
As for proposed optimization, it is applied in [`tbb::mutex`](https://github.com/oneapi-src/oneTBB/blob/master/include/oneapi/tbb/mutex.h#L58-L73) and I would remain `tbb::spin_mutex` as is not to increase its overhead (and not to lose its last advantage if any).

pavelkumbrasev:
Closing this PR since tbb::mutex should be used in contended cases.

PR #659, opened 2021-11-16
Suppress data race during ittnotify initialization

vlserov:
### Description 
Suppress data race during ittnotify initialization.

### Type of change
Adding suppression file for TSAN with corresponding changes.

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/cmake/suppressions/tsan.suppressions
+++ b/cmake/suppressions/tsan.suppressions
@@ -1,0 +1,3 @@
+# TSAN suppression for known issues.
+# Possible data race during ittnotify initialization. Low impact.
+race:__itt_nullify_all_pointers
--- a/cmake/sanitize.cmake
+++ b/cmake/sanitize.cmake
@@ -34,7 +34,9 @@ if (NOT ${FLAG_DISPLAY_NAME})
     "please try another compiler or omit TBB_SANITIZE variable")
 endif()
 
-set(TBB_TESTS_ENVIRONMENT ${TBB_TESTS_ENVIRONMENT} "LSAN_OPTIONS=suppressions=${CMAKE_CURRENT_SOURCE_DIR}/cmake/suppressions/lsan.suppressions")
+set(TBB_TESTS_ENVIRONMENT ${TBB_TESTS_ENVIRONMENT}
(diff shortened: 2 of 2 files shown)


cmake/suppressions/tsan.suppressions
```diff
@@ -0,0 +1,3 @@
+# TSAN suppression for known issues.
+# Possible data race during ittnotify initialization. Low impact.
+race:__itt_nullify_all_pointers
```
anton-potapov, line 3: from this line it is not clear that problem is library initialization. Can we narrow it ?

vlserov: ITT API consists of two parts: a static part and a dynamic part. The dynamic part is specific for a tool and distributed only with a particular tool. The static part is integrated into oneTBB.
Initially all pointers to ITT API are initialized with similar to `__itt_someApiCall_init` function. In case collecting data, dynamic part are loaded and pointer will be initialized with correct function from dynamic part, otherwise it will be initialized with `nullprt` and function call would be ignored.
The read/write to a word-sized in aligned location will be atomic on the most systems, but it's not guaranteed by standard.
```
static void __cdecl __itt_someApiCall_init( params... );
typedef void __cdecl __itt_someApiCall_t( params... );
__itt_someApiCall_t* __itt_someApiCall_ptr_ = __itt_someApiCall_init;
static void __cdecl __itt_someApiCall_init( params... ) {
    if (!__itt__ittapi_global.api_initialized && __itt__ittapi_global.thread_list == 0) {
        __itt_init_ittlib_name(...);
    }
    if (__itt_someApiCall_ptr_ && __itt_someApiCall_ptr_ != __itt_someApiCall_init) {
        __itt_someApiCall_ptr_( params... );
    }
}
```
[TSAN Logs](https://github.com/oneapi-src/oneTBB/files/7553321/tsan_logs.txt)

anton-potapov: @vlserov thanks for the clarification

anton-potapov (approved):
_LGTM_

PR #667, opened 2021-11-22
Deallocate object with based on actual type

### LINKED ISSUE #639: Assertion m_private_counter >= 0 failed (located in the destroy function, line in file: 141)
When using flow graph, the following assert failed with the following stderr output:

**Assertion m_private_counter >= 0 failed (located in the destroy function, line in file: 141)
Detailed description: Private counter may not be less than 0**

in the destroy function:

Call Stack:
image

_raise 0x000014a314b6637f
abort 0x000014a314b50db5
tbb::detail::r1::assertion_failure_impl assert_impl.h:56
operator() assert_impl.h:73
tbb::detail::d0::run_initializer<tbb::detail::r1::assertion_failure(char const*, int, char const*, char const*)::<lambda()> >(const struct {...} &, std::atomic<tbb::detail::d0::do_once_state> &) _utils.h:288
tbb::detail::d0::atomic_do_once<tbb::detail::r1::assertion_failure(char const*, int, char const*, char const*)::<lambda()> >(const struct {...} &, std::atomic<tbb::detail::d0::do_once_state> &) _utils.h:277
tbb::detail::r1::assertion_failure assert_impl.h:73
tbb::detail::r1::small_object_pool_impl::destroy small_object_pool.cpp:141
tbb::detail::r1::thread_data::~thread_data thread_data.h:120
tbb::detail::r1::governor::auto_terminate governor.cpp:221
tbb::detail::r1::market::cleanup market.cpp:611
tbb::detail::r1::rml::private_worker::run private_server.cpp:283
tbb::detail::r1::rml::private_worker::thread_routine private_server.cpp:221
start_thread 0x000014a3156a815a
clone 0x000014a314c2bdd3_

vlserov:
### Description 
`graph_task` deallocates another type than allocated. Provides actual type of object for `small_object_allocator` allocator.

Fixes #639

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

@alexey-katranov

### DIFF
--- a/include/oneapi/tbb/detail/_flow_graph_impl.h
+++ b/include/oneapi/tbb/detail/_flow_graph_impl.h
@@ -135,9 +135,11 @@ public:
     graph& my_graph; // graph instance the task belongs to
     // TODO revamp: rename to my_priority
     node_priority_t priority;
+    template <typename DerivedType>
     void destruct_and_deallocate(const execution_data& ed);
     task* cancel(execution_data& ed) override;
 protected:
+    template <typename DerivedType>
     void finalize(const execution_data& ed);
 private:
     // To organize task_list
@@ -359,21 +361,23 @@ private:
 
 };  // class graph
 
+template<typename DerivedType>
 inline void graph_task::destruct_and_deallocate(const execution_data& ed) {
     auto allocator = my_allocator;
     // TODO: investigate if direct call of derived destructor gives any benefits.
     this->~graph_task();
-    allocator.deallocate(this, ed);
+    allocator.deallocate(static_cast<DerivedType*>(this), ed);
 }
 
+template<typename DerivedType>
 inline void graph_task::finalize(const execution_data& ed) {
     graph& g = my_graph;
-    destruct_and_deallocate(ed);
+    destruct_and_deallocate<DerivedType>(ed);
     g.release_wait();
 }
 
 inline task* graph_task::cancel(execution_data& ed) {
-    finalize(ed);
+    finalize<graph_task>(ed);
     return nullptr;
 }
 
--- a/include/oneapi/tbb/detail/_flow_graph_impl.h
+++ b/include/oneapi/tbb/detail/_flow_graph_impl.h
@@ -137,7 +137,6 @@ public:
     node_priority_t priority;
     template <typename DerivedType>
     void destruct_and_deallocate(const execution_data& ed);
-    task* cancel(execution_data& ed) override;
 protected:
     template <typename DerivedType>
     void finalize(const execution_data& ed);
@@ -374,11 +373,6 @@ inline void graph_task::finalize(const execution_data& ed) {
     graph& g = my_graph;
     destruct_and_deallocate<DerivedType>(ed);
     g.release_wait();
-}
-
-inline task* graph_task::cancel(execution_data& ed) {
-    finalize<graph_task>(ed);
-    return nullptr;
 }
 
 //********************************************************************************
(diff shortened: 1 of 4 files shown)


include/oneapi/tbb/detail/_flow_graph_impl.h
```diff
@@ -359,21 +361,23 @@ class graph : no_copy, public graph_proxy {
 
 };  // class graph
 
+template<typename DerivedType>
 inline void graph_task::destruct_and_deallocate(const execution_data& ed) {
     auto allocator = my_allocator;
     // TODO: investigate if direct call of derived destructor gives any benefits.
     this->~graph_task();
-    allocator.deallocate(this, ed);
+    allocator.deallocate(static_cast<DerivedType*>(this), ed);
```
aleksei-fedotov, line 369: As far as I understand, the root cause of the issue is the automatic upcast of the `this` pointer to `graph_task` during the call to function template `allocator.deallocate()`.

Can we introduce an overload to `deallocate` function such as:
```cpp
    void deallocate(void* ptr, size_t size, const execution_data& ed);
```
to have:
```suggestion
    allocator.deallocate(this, sizeof(this), ed);
```
instead? 

This way allows us avoiding introduction of one more template parameter.

alexey-katranov: You cannot remove `static_cast` to the derived type because it is not guaranteed (in a general case) that the derived object has the same address as the parent object.

As for one more templated parameter, `sizeof(this)` in the parent class code is `sizeof` of the base class, in other words, `sizeof(this) != sizeof(static_cast<Derived*>(this))`. (C++ does not have reflection (yet?), i.e. we cannot get size information of the real object from the pointer to the parent object)

include/oneapi/tbb/detail/_flow_graph_impl.h
```diff
@@ -359,21 +361,23 @@ class graph : no_copy, public graph_proxy {
...
-    destruct_and_deallocate(ed);
+    destruct_and_deallocate<DerivedType>(ed);
     g.release_wait();
 }
 
 inline task* graph_task::cancel(execution_data& ed) {
-    finalize(ed);
+    finalize<graph_task>(ed);
     return nullptr;
 }
```
alexey-katranov, line 382: Do we really need to override `cancel` in the `graph_task` now?

alexey-katranov:
Why test is not needed? It is a bug fix and #639 contains reproducer.

include/oneapi/tbb/detail/_flow_graph_impl.h
aleksei-fedotov: Yeah. I agree.

phprus:
Do you have any updates?

alexey-katranov:
> Do you have any updates?

@vlserov is on vacation, if the issue is urgent, we can try to reprioritize the task to some else?

phprus:
@alexey-katranov, No this issue is not urgent for me. No need to change priorities.

include/oneapi/tbb/detail/_flow_graph_impl.h
vlserov: Removed

include/oneapi/tbb/detail/_flow_graph_impl.h
```diff
@@ -359,24 +360,21 @@ class graph : no_copy, public graph_proxy {
 
 };  // class graph
 
+template<typename DerivedType>
 inline void graph_task::destruct_and_deallocate(const execution_data& ed) {
     auto allocator = my_allocator;
     // TODO: investigate if direct call of derived destructor gives any benefits.
     this->~graph_task();
```
alexey-katranov, line 367: Maybe we will resolve the TODO? Check that all derived are final (does it make sense to mark them as `final`)? Or you can address it in a separate PR.
```suggestion
    this->DerivedType::~graph_task();
```

alexey-katranov (approved):

PR #671, opened 2021-11-25
Use system functions to allocate aligned memory

Lastique:
### Description 

This uses the standard allocator for allocating and freeing memory with higher alignment requirements instead of the hand-rolled implementation.

### Type of change

- [x] optimization

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/src/tbb/allocator.cpp
+++ b/src/tbb/allocator.cpp
@@ -33,6 +33,24 @@
 #include <dlfcn.h>
 #endif
 
+#if (!defined(_WIN32) && !defined(_WIN64)) || defined(__CYGWIN__)
+#include <stdlib.h> // posix_memalign, free
+// With glibc, uClibc and musl on Linux and bionic on Android it is safe to use memalign(), as the allocated memory
+// can be freed with free(). It is also better to use memalign() since posix_memalign() is just a wrapper on top of
+// memalign() and it offers nothing but overhead due to inconvenient interface. This is likely the case with other
+// standard libraries as well, and more libraries can be added to the preprocessor check below. Unfortunately, we
+// can't detect musl, so we simply enable memalign() on Linux and Android in general.
+#if defined(linux) || defined(__linux) || defined(__linux__) || defined(__ANDROID__)
+#include <malloc.h> // memalign
+#define __TBB_USE_MEMALIGN
+#else
+#define __TBB_USE_POSIX_MEMALIGN
+#endif
+#elif defined(_MSC_VER) || defined(__MINGW32__)
+#include <malloc.h> // _aligned_malloc, _aligned_free
+#define __TBB_USE_MSVC_ALIGNED_MALLOC
+#endif
+
 #if __TBB_WEAK_SYMBOLS_PRESENT
 
 #pragma weak scalable_malloc
@@ -67,10 +85,10 @@ static void  (*deallocate_handler)(void* pointer) = nullptr;
 //! Initialization routine used for first indirect call via cache_aligned_allocate_handler.
 static void* initialize_cache_aligned_allocate_handler(std::size_t n, std::size_t alignment);
 
-//! Allocates memory using standard malloc. It is used when scalable_allocator is not available
+//! Allocates overaligned memory using standard memory allocator. It is used when scalable_allocator is not available.
 static void* std_cache_aligned_allocate(std::size_t n, std::size_t alignment);
 
-//! Allocates memory using standard free. It is used when scalable_allocator is not available
+//! Deallocates overaligned memory using standard memory allocator. It is used when scalable_allocator is not available.
 static void  std_cache_aligned_deallocate(void* p);
 
 //! Handler for padded memory allocation
@@ -185,6 +203,17 @@ void __TBB_EXPORTED_FUNC cache_aligned_deallocate(void* p) {
 }
 
 static void* std_cache_aligned_allocate(std::size_t bytes, std::size_t alignment) {
+#if defined(__TBB_USE_MEMALIGN)
+    return memalign(alignment, bytes);
+#elif defined(__TBB_USE_POSIX_MEMALIGN)
+    void* p = nullptr;
+    int res = posix_memalign(&p, alignment, bytes);
+    if (res != 0)
+        p = nullptr;
+    return p;
+#elif defined(__TBB_USE_MSVC_ALIGNED_MALLOC)
+    return _aligned_malloc(bytes, alignment);
+#else
     // TODO: make it common with cache_aligned_resource
     std::size_t space = alignment + bytes;
     std::uintptr_t base = reinterpret_cast<std::uintptr_t>(std::malloc(space));
@@ -199,9 +228,15 @@ static void* std_cache_aligned_allocate(std::size_t bytes, std::size_t alignment
     // Record where block actually starts.
     (reinterpret_cast<std::uintptr_t*>(result))[-1] = base;
     return reinterpret_cast<void*>(result);
+#endif
 }
 
 static void std_cache_aligned_deallocate(void* p) {
+#if defined(__TBB_USE_MEMALIGN) || defined(__TBB_USE_POSIX_MEMALIGN)
+    free(p);
+#elif defined(__TBB_USE_MSVC_ALIGNED_MALLOC)
+    _aligned_free(p);
+#else
     if (p) {
         __TBB_ASSERT(reinterpret_cast<std::uintptr_t>(p) >= 0x4096, "attempt to free block not obtained from cache_aligned_allocator");
         // Recover where block actually starts
@@ -209,6 +244,7 @@ static void std_cache_aligned_deallocate(void* p) {
         __TBB_ASSERT(((base + nfs_size) & ~(nfs_size - 1)) == reinterpret_cast<std::uintptr_t>(p), "Incorrect alignment or not allocated by std_cache_aligned_deallocate?");
         std::free(reinterpret_cast<void*>(base));
     }
+#endif
 }
 
 void* __TBB_EXPORTED_FUNC allocate_memory(std::size_t size) {


Lastique:
This patch originated from https://github.com/oneapi-src/oneTBB/pull/326.

src/tbb/allocator.cpp
```diff
@@ -67,11 +85,15 @@ static void  (*deallocate_handler)(void* pointer) = nullptr;
...
 
+#if !defined(__TBB_USE_MSVC_ALIGNED_MALLOC)
 //! Allocates memory using standard malloc. It is used when scalable_allocator is not available
 static void* std_cache_aligned_allocate(std::size_t n, std::size_t alignment);
+#endif
 
+#if !defined(__TBB_USE_POSIX_MEMALIGN) && !defined(__TBB_USE_MEMALIGN) && !defined(__TBB_USE_MSVC_ALIGNED_MALLOC)
 //! Allocates memory using standard free. It is used when scalable_allocator is not available
 static void  std_cache_aligned_deallocate(void* p);
+#endif
```
alexey-katranov, line 96: What about the approach like in `std_cache_aligned_allocate`?
```suggestion
#if !defined(__TBB_USE_MSVC_ALIGNED_MALLOC)
//! Allocates memory using standard malloc. It is used when scalable_allocator is not available
static void* std_cache_aligned_allocate(std::size_t n, std::size_t alignment);

#if !defined(__TBB_USE_POSIX_MEMALIGN) && !defined(__TBB_USE_MEMALIGN)
//! Allocates memory using standard free. It is used when scalable_allocator is not available
static void  std_cache_aligned_deallocate(void* p);
#endif
#endif // !defined(__TBB_USE_MSVC_ALIGNED_MALLOC)
```

alexey-katranov (approved):

src/tbb/allocator.cpp
Lastique: Done.

src/tbb/allocator.cpp
```diff
@@ -67,11 +85,15 @@ static void  (*deallocate_handler)(void* pointer) = nullptr;
...
 
+#if !defined(__TBB_USE_MSVC_ALIGNED_MALLOC)
 //! Allocates memory using standard malloc. It is used when scalable_allocator is not available
 static void* std_cache_aligned_allocate(std::size_t n, std::size_t alignment);
 
+#if !defined(__TBB_USE_POSIX_MEMALIGN) && !defined(__TBB_USE_MEMALIGN)
 //! Allocates memory using standard free. It is used when scalable_allocator is not available
 static void  std_cache_aligned_deallocate(void* p);
+#endif // !defined(__TBB_USE_POSIX_MEMALIGN) && !defined(__TBB_USE_MEMALIGN)
+#endif // !defined(__TBB_USE_MSVC_ALIGNED_MALLOC)
```
alexey-katranov, line 96:
```suggestion
//! Allocates memory using standard malloc. It is used when scalable_allocator is not available
static void* std_cache_aligned_allocate(std::size_t n, std::size_t alignment);

//! Allocates memory using standard free. It is used when scalable_allocator is not available
static void  std_cache_aligned_deallocate(void* p);
```

src/tbb/allocator.cpp
```diff
@@ -123,8 +145,16 @@ void initialize_handler_pointers() {
...
+#if defined(__TBB_USE_POSIX_MEMALIGN) || defined(__TBB_USE_MEMALIGN)
+        cache_aligned_allocate_handler_unsafe = &std_cache_aligned_allocate;
+        cache_aligned_deallocate_handler = &::free;
+#elif defined(__TBB_USE_MSVC_ALIGNED_MALLOC)
+        cache_aligned_allocate_handler_unsafe = &::_aligned_malloc;
+        cache_aligned_deallocate_handler = &::_aligned_free;
+#else
         cache_aligned_allocate_handler_unsafe = &std_cache_aligned_allocate;
         cache_aligned_deallocate_handler = &std_cache_aligned_deallocate;
+#endif
```
alexey-katranov, line 157:
```suggestion
        cache_aligned_allocate_handler_unsafe = &std_cache_aligned_allocate;
        cache_aligned_deallocate_handler = &std_cache_aligned_deallocate;
```

src/tbb/allocator.cpp
```diff
@@ -184,7 +214,17 @@ void __TBB_EXPORTED_FUNC cache_aligned_deallocate(void* p) {
...
 static void* std_cache_aligned_allocate(std::size_t bytes, std::size_t alignment) {
+#if defined(__TBB_USE_MEMALIGN)
+    return memalign(alignment, bytes);
+#elif defined(__TBB_USE_POSIX_MEMALIGN)
+    void* p = nullptr;
+    int res = posix_memalign(&p, alignment, bytes);
+    if (res != 0)
+        p = nullptr;
+    return p;
+#else
```
alexey-katranov, line 227:
```suggestion
#elif defined(__TBB_USE_MSVC_ALIGNED_MALLOC)
    return _aligned_malloc(bytes, alignment);
#else
```

src/tbb/allocator.cpp
```diff
@@ -184,7 +214,17 @@ void __TBB_EXPORTED_FUNC cache_aligned_deallocate(void* p) {
     (*cache_aligned_deallocate_handler)(p);
 }
 
+#if !defined(__TBB_USE_MSVC_ALIGNED_MALLOC)
```
alexey-katranov, line 217:
```suggestion
```

src/tbb/allocator.cpp
```diff
@@ -199,8 +239,11 @@ static void* std_cache_aligned_allocate(std::size_t bytes, std::size_t alignment
     // Record where block actually starts.
     (reinterpret_cast<std::uintptr_t*>(result))[-1] = base;
     return reinterpret_cast<void*>(result);
+#endif
 }
+#endif // !defined(__TBB_USE_MSVC_ALIGNED_MALLOC)
```
alexey-katranov, line 244:
```suggestion
```

src/tbb/allocator.cpp
```diff
@@ -199,8 +239,11 @@ static void* std_cache_aligned_allocate(std::size_t bytes, std::size_t alignment
...
 
+#if !defined(__TBB_USE_POSIX_MEMALIGN) && !defined(__TBB_USE_MEMALIGN) && !defined(__TBB_USE_MSVC_ALIGNED_MALLOC)
 static void std_cache_aligned_deallocate(void* p) {
```
alexey-katranov, line 247:
```suggestion
static void std_cache_aligned_deallocate(void* p) {
#if defined(__TBB_USE_POSIX_MEMALIGN) || defined(__TBB_USE_MEMALIGN)
    free(p);
#elif defined(__TBB_USE_MSVC_ALIGNED_MALLOC)
    _aligned_free(p);
#else
```

src/tbb/allocator.cpp
```diff
@@ -210,6 +253,7 @@ static void std_cache_aligned_deallocate(void* p) {
         std::free(reinterpret_cast<void*>(base));
     }
 }
+#endif // !defined(__TBB_USE_POSIX_MEMALIGN) && !defined(__TBB_USE_MEMALIGN) && !defined(__TBB_USE_MSVC_ALIGNED_MALLOC)
```
alexey-katranov, line 256:
```suggestion
#endif // defined(__TBB_USE_POSIX_MEMALIGN) || defined(__TBB_USE_MEMALIGN)
}
```

alexey-katranov (commented):
Overall, I am Ok with your approach. However, can you consider a set of proposed changes to reduce the code slicing and make macro dependent code more isolated inside the functions? What do you think? (I understand that it will cause one more call in the chain but I hope modern compilers can perform tail call optimization)

Lastique:
> Overall, I am Ok with your approach. However, can you consider a set of proposed changes to reduce the code slicing and make macro dependent code more isolated inside the functions? What do you think? (I understand that it will cause one more call in the chain but I hope modern compilers can perform tail call optimization)

I'm quite certain the compiler won't optimize away the forwarding function, even if it amounts to a `jmp` instruction, because it has to have a unique address. That's unnecessary overhead that I'd prefer to avoid. The resulting code doesn't improve much in terms if simplicity, IMHO, and I value runtime performance more than code simplicity anyway. However, if you insist I can change the code as you suggest. Please, confirm.

alexey-katranov:
> The resulting code doesn't improve much in terms if simplicity

At rough estimation, I removed 6 condition blocks but added only 2.

> I value runtime performance more than code simplicity anyway.

It seems we have slightly different views here. The code is not on critical path in term of performance but any correctness issue should be avoided. However, code complexity wastes engineering time for debug and maintaining while it is not critical for performance of the whole library and we can spend this time to improve something else. As for optimization, it seems possible only for `__TBB_USE_MSVC_ALIGNED_MALLOC`, while other platforms will have excessive `jmp`. So, why not have a generic approach for all platforms until we face performance issue due to this inefficiency?

>  However, if you insist I can change the code as you suggest. Please, confirm.

I do not like to insist and prefer finding consensus. In any case, good balance is consist of multiple opinions. Taking into consideration that you contribute relatively often and reasonably, your opinion is valuable.

Lastique:
> > The resulting code doesn't improve much in terms if simplicity
> 
> At rough estimation, I removed 6 condition blocks but added only 2.

Yes, it is better, I agree, but I just don't find the change *that much* better. I mean, the original code is not hard to decipher, and all conditions are uniform and easy to locate if they need to be updated. In my personal code I would probably be fine with the extra preprocessor checks if it makes the code more optimal.

> > I value runtime performance more than code simplicity anyway.
> 
> It seems we have slightly different views here. The code is not on critical path in term of performance but any correctness issue should be avoided. However, code complexity wastes engineering time for debug and maintaining while it is not critical for performance of the whole library and we can spend this time to improve something else. As for optimization, it seems possible only for `__TBB_USE_MSVC_ALIGNED_MALLOC`, while other platforms will have excessive `jmp`. So, why not have a generic approach for all platforms until we face performance issue due to this inefficiency?

Because MSVC is lucky enough for this optimization to be possible, and it seems wasteful to not take this opportunity.

> > However, if you insist I can change the code as you suggest. Please, confirm.
> 
> I do not like to insist and prefer finding consensus. In any case, good balance is consist of multiple opinions. Taking into consideration that you contribute relatively often and reasonably, your opinion is valuable.

The thing is, this project is not my personal code, and you, the maintainers, may have different priorities than I do, and this is totally fine. I'm just stating my position, and if you still think you prefer it the other way, I can change accordingly, and that is also fine. If I'm adamantly against something, you can be sure I will say so and refuse to change. :)

In fact in this case I don't care about MSVC as my main target platform is Linux. I just added support for it because it was the straightforward and easy thing to do. My main goal here is to get this optimization merged, primarily for Linux, and I don't want this PR to hang indefinitely in limbo or get bogged down in discussions about minor things like this. So I will change the code as you suggest, no problem.

Lastique:
> Because MSVC is lucky enough for this optimization to be possible, and it seems wasteful to not take this opportunity.

Actually, it's not only MSVC - `free` is used directly on other platforms.

Lastique:
> So I will change the code as you suggest, no problem.

Done.

alexey-katranov (approved):
Thank you for the contribution.

PR #673, opened 2021-11-29
Fix unordered and ordered containers ranges for empty containers

### LINKED ISSUE #641: empty ranges for concurrent_map and concurrent_unordered_map cause crash / assert
Ranges can be empty, right?

I'm seeing strange behaviour of `range()` called on empty containers of type `tbb::concurrent_map()` and `tbb::concurrent_unordered_map` leading to crashes or infinite loops. Is that expected?

In the following minimal working example an assert is triggered in Debug mode and an infinite loop occurs in Release:

```cpp
#include <tbb/concurrent_unordered_map.h>
//#include <tbb/concurrent_map.h>

#include <iostream>

int main() {
  tbb::concurrent_unordered_map<int, int> map;
  // tbb::concurrent_map<int, int> map;

  // map[1] = 2;

  auto range = map.range();

  std::cout << range.empty() << std::endl;
}
```

Assert:
```
Assertion bucket != 0 failed (located in the get_parent function, line in file: 1445)
Detailed description: Unable to get_parent of the bucket 0
```

If I uncomment the line `map[1] = 2;`, it runs fine. If I switch to `tbb::concurrent_map`, I get a segfault somewhere in `_concurrent_skip_list.h`.

This is on macOS 10.15 (Catalina) with TBB `2021.4.0` from Homebrew. Same behaviour with Apple clang version 12.0.0 from XCode or clang 13 from Homebrew.

kboyarinov:
### Description 
Fix for empty ordered and unordered containers range_type and const_range_type

Fixes #641 

- [ ] - git commit message contains appropriate signed-off-by string _(see [CONTRIBUTING.md](https://github.com/oneapi-src/oneTBB/blob/master/CONTRIBUTING.md#pull-requests) for details)_

### Type of change

### Tests

- [x] added - _required for new features and for some bug fixes_
- [ ] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown


test/common/concurrent_associative_common.h
```diff
@@ -425,6 +425,26 @@ void test_comparison_operators() {
...
+template <typename Range, typename Container>
+void test_empty_container_range(Container&& cont) {
+    REQUIRE(cont.empty());
+    Range r = cont.range();
+    REQUIRE_MESSAGE(r.empty(), "Empty container range should be empty");
+    REQUIRE_MESSAGE(!r.is_divisible(), "Empty container range should not be divisible");
+    REQUIRE_MESSAGE(r.begin() == r.end(), "Incorrect iterators on empty range");
+    REQUIRE_MESSAGE(r.begin() == cont.begin(), "Incorrect iterators on empty range");
+
+    Range r2(r, tbb::split{});
```
alexey-katranov, line 437: It is not stated if split ctor can be used for non-divisible range; however, [the spec](https://spec.oneapi.io/versions/latest/elements/oneTBB/source/named_requirements/algorithms/range.html) says about `is_divisible`: "_True if the range can be partitioned into two subranges._". I.e. range cannot be partitioned if `false`. Does it mean that we cannot use splitting ctor to partition the empty range?

anton-potapov:
@kboyarinov please fix the test failures

PR #676, opened 2021-12-02
Fix incorrect definition of assignment operator in containers

### LINKED ISSUE #372: Using-declaration of assignment operators results in incorrect return type
There are a number of instances in oneTBB where the code imports `operator=` from a base class with a `using`-declaration. This is incorrect because the imported `operator=` that are not copy or move assignment operators will have the return type of a reference to the base class instead of the derived class. (For copy and move assignment operators the standard makes an exception, as in this case it forces the implicit generation of the assignment operators in the derived class instead of importing the operators from the base class, see [class.copy.assign]/8.) [Here](https://gcc.godbolt.org/z/orPPMEEjf) is a demonstration.

The solution is to define `operator=` in the derived classes explicitly for all argument types, and forward to the base class `operator=`. https://github.com/oneapi-src/oneTBB/pull/318 implements this for concurrent containers.

isaevil:
### Description 
Is a part of https://github.com/oneapi-src/oneTBB/pull/318

Fixes #372 

### Type of change

### Tests

- [x] added - _required for new features and for some bug fixes_
- [ ] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

@anton-potapov @Lastique

### DIFF
--- a/include/oneapi/tbb/concurrent_map.h
+++ b/include/oneapi/tbb/concurrent_map.h
@@ -92,7 +92,6 @@ public:
 
     // Include constructors of base type
     using base_type::base_type;
-    using base_type::operator=;
 
     // Required for implicit deduction guides
     concurrent_map() = default;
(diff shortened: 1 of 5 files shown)


anton-potapov (approved):
*LGTM*

anton-potapov:
@Lastique , @isaevil  thanks you very much!

PR #695, opened 2021-12-22
Preliminary version of EMSCRIPTEN port

### LINKED ISSUE #460: How to compile oneTBB with Emscritpen?
Is it possible to compile oneTBB with Emscripten? There are third party libraries using oneTBB and it would be nice to get them to compile without changing the source code.

anton-potapov:
### Description 
This patch adds preliminary support of [Emscripten](https://emscripten.org/) compiler 

Fixes #460

### Type of change

- [x] new feature - _change which adds functionality_

### Tests

- [ ] not needed

### Documentation

- [x] needs to be updated
- [ ] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### Other information
this is preliminary version, and most likely it's content will change

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,6 +62,10 @@ include(CheckCXXCompilerFlag)
 include(GNUInstallDirs)
 include(CMakeDependentOption)
 
+if(CMAKE_CXX_COMPILER MATCHES "/em\\+\\+(-[a-zA-Z0-9.])?$")
+  message(" * C++ compiler: Emscripten")
+  set(CMAKE_EXECUTABLE_SUFFIX .html)
+endif()
 # ---------------------------------------------------------------------------------------------------------
 # Handle C++ standard version.
 if (NOT MSVC)  # no need to cover MSVC as it uses C++14 by default.
--- a/include/oneapi/tbb/detail/_config.h
+++ b/include/oneapi/tbb/detail/_config.h
@@ -182,7 +182,7 @@
 
 /** __TBB_DYNAMIC_LOAD_ENABLED describes the system possibility to load shared libraries at run time **/
 #ifndef __TBB_DYNAMIC_LOAD_ENABLED
-    #define __TBB_DYNAMIC_LOAD_ENABLED 1
+    #define __TBB_DYNAMIC_LOAD_ENABLED (!__EMSCRIPTEN__)
 #endif
 
 /** __TBB_WIN8UI_SUPPORT enables support of Windows* Store Apps and limit a possibility to load
@@ -195,7 +195,7 @@
 
 /** __TBB_WEAK_SYMBOLS_PRESENT denotes that the system supports the weak symbol mechanism **/
 #ifndef __TBB_WEAK_SYMBOLS_PRESENT
-    #define __TBB_WEAK_SYMBOLS_PRESENT ( !_WIN32 && !__APPLE__ && !__sun && (__TBB_GCC_VERSION >= 40000 || __INTEL_COMPILER ) )
+    #define __TBB_WEAK_SYMBOLS_PRESENT (!__EMSCRIPTEN__ && !_WIN32 && !__APPLE__ && !__sun && (__TBB_GCC_VERSION >= 40000 || __INTEL_COMPILER ) )
 #endif
 
 /** Presence of compiler features **/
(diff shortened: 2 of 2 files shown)


cmake/compilers/Clang.cmake
```diff
@@ -12,6 +12,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+if(CMAKE_CXX_COMPILER MATCHES "/em\\+\\+(-[a-zA-Z0-9.])?$")
```
estollnitz, line 15: I recommend building the project using `emcmake`, which will set the `EMSCRIPTEN` variable for you. Then this line could be written simply as:
```
if(EMSCRIPTEN)
```

cmake/compilers/Clang.cmake
```diff
@@ -12,6 +12,14 @@
...
+  message(" * C++ compiler: Emscripten")
```
estollnitz, line 16: This message may be useful while getting things working, but for consistency with the configuration of other compilers in oneTBB, it should probably be removed at some point.

cmake/compilers/Clang.cmake
```diff
@@ -12,6 +12,14 @@
...
+  set(CMAKE_EXECUTABLE_SUFFIX .js)
```
estollnitz, line 17: Another reason to use `emcmake` to build: it sets the executable suffix to `.js` for you.

cmake/compilers/Clang.cmake
```diff
@@ -12,6 +12,14 @@
...
+  set(TBB_EMSCRIPTEN 1)
```
estollnitz, line 18: Again, by using `emcmake` to build, you can rely on `EMSCRIPTEN` being defined, so `TBB_EMSCRIPTEN` is not really needed.

cmake/compilers/Clang.cmake
```diff
@@ -12,6 +12,14 @@
...
+  set(CMAKE_CXX_FLAGS ${CMAKE_CXX_FLAGS} " -pthread -s TOTAL_MEMORY=128MB -s EXIT_RUNTIME=1 -s PROXY_TO_PTHREAD=1")
```
estollnitz, line 19: I don't recommend setting the link-time flags `-s TOTAL_MEMORY=128MB -s EXIT_RUNTIME=1 -s PROXY_TO_PTHREAD=1` here. Client apps that consume oneTBB may prefer different settings. Only `-pthread` is truly needed here. The other flags should probably be set where the executable targets are defined. For example, the `tbb_add_test()` function in `test/CMakeLists.txt` could apply these settings to each executable test using CMake's `target_link_options()` function.

Also, instead of guessing how much total memory is required, you may wish to use `-s ALLOW_MEMORY_GROWTH=1`. In that case, you can suppress the compiler warning about using threads and growable memory with the `-Wno-pthreads-mem-growth` flag.

test/CMakeLists.txt
```diff
@@ -368,7 +370,9 @@ if (TARGET TBB::tbb)
     tbb_add_test(SUBDIR tbb NAME test_tick_count DEPENDENCIES TBB::tbb)
     tbb_add_test(SUBDIR tbb NAME test_allocators DEPENDENCIES TBB::tbb)
     tbb_add_test(SUBDIR tbb NAME test_arena_priorities DEPENDENCIES TBB::tbb)
-    tbb_add_test(SUBDIR tbb NAME test_dynamic_link DEPENDENCIES TBB::tbb)
+    if (NOT TBB_EMSCRIPTEN)    
+        tbb_add_test(SUBDIR tbb NAME test_dynamic_link DEPENDENCIES TBB::tbb)
+    endif()
```
estollnitz, line 375: I think the tests are already protected by `__TBB_DYNAMIC_LOAD_ENABLED`, so hopefully you don't need to test for `TBB_EMSCRIPTEN` here. And likewise for other tests in this file that are currently being avoided under Emscripten.

estollnitz (commented):
I've added several specific comments that I hope will improve this PR.

In general, I think the build process will be greatly improved by relying on `emcmake`, which configures many CMake settings for Emscripten compilation. Client applications that rely on oneTBB will probably use `emcmake` to build.

spiffmon:
I'd just like to express support for this contribution on behalf of the Universal Scene Description (USD) project at Pixar, as lack of emscripten support for OneTBB has become a blocking point for USD to move forward to OneTBB, and USD in turn cascades as a blocking point for the majority of the DCC's in the VFX industry (Maya, Houdini, Katana, Nuke, etc) because they depend deeply on USD, and we all rely on TBB to arbitrate resources between host and subsystem.

Thank you, and thanks @anton-potapov for this work!

PR #741, opened 2022-01-20
Split tbb::flow::join_node testing into separate test files

isaevil:
### Description 
In order to reduce GOT size on MIPS that leads to linkage errors (`relocation truncated to fit: R_MIPS_GOT_PAGE`) in debug, `tbb::flow::join_node` testing was split into separate files.

New testing files added:
* `test_join_node_preview` for `join_node` preview functionality
* `test_join_node_msg_key_matching_n_args` to separate regular TEST_CASE's and TEST_CASE_TEMPLATE

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

@aleksei-fedotov @vlserov 
@AlexVeprev to review CMake part


cmake/compilers/GNU.cmake
```diff
@@ -56,7 +56,7 @@ endif()
 
 # Workaround for heavy tests and too many symbols in debug (rellocation truncated to fit: R_MIPS_CALL16)
 if ("${CMAKE_SYSTEM_PROCESSOR}" MATCHES "mips")
-    set(TBB_TEST_COMPILE_FLAGS ${TBB_TEST_COMPILE_FLAGS} -DTBB_TEST_LOW_WORKLOAD $<$<CONFIG:DEBUG>:-mxgot>)
+    set(TBB_TEST_COMPILE_FLAGS ${TBB_TEST_COMPILE_FLAGS} -DTBB_TEST_LOW_WORKLOAD $<$<CONFIG:DEBUG>:-fPIE -mxgot>)
```
isaevil, line 59: According to [MIPS Options (GCC)](https://gcc.gnu.org/onlinedocs/gcc/MIPS-Options.html) "These options (-mxgot, -mno-xgot) have no effect unless GCC is generating position independent code."

cmake/toolchains/mips.cmake
```diff
@@ -35,6 +35,12 @@ set(THREADS_PTHREAD_ARG "0" CACHE STRING "Result from TRY_RUN" FORCE)
...
 set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
 
+if (CMAKE_BUILD_TYPE STREQUAL "Debug")
+    set(_tbb_mips_pie -pie)
+endif()
+
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -EL -mabi=64 -march=mips64r6 -mcrc -mfp64 -mmt -mtune=mips64r6 -ggdb -ffp-contract=off -mhard-float" CACHE INTERNAL "")
 set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -mvirt -mxpa" CACHE INTERNAL "")
-set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -mvirt -mxpa" CACHE INTERNAL "")  # for tests
+set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -mvirt -mxpa ${_tbb_mips_pie}" CACHE INTERNAL "")  # for tests
```
vlserov, line 44:
```suggestion
set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -mvirt -mxpa $<$<CONFIG:Debug>:-pie>" CACHE INTERNAL "")  # for tests
```

Is it possible to use suggestion above instead of using temp variable? Or is there  some problem with this approach?

isaevil: I have tried this approach but CMake just inserts this generator expression to the command line as link option...

vlserov:
```suggestion
set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS}" -mvirt -mxpa $<$<CONFIG:Debug>:-pie> CACHE INTERNAL "")  # for tests
```
How about that?

isaevil: Still doesn't work. Fails on configuration stage with:
```
mips-img-linux-gnu-g++: error: ;-mvirt;-mxpa;$<$<CONFIG:Debug>:-pie>: No such file or directory
```

cmake/compilers/GNU.cmake
AlexVeprev, line 59: Consider setting `-fPIE` in `cmake/toolchains/mips.cmake` as you do for linker flag

test/tbb/test_join_node_preview.cpp
```diff
@@ -0,0 +1,98 @@
...
+    test_follows_and_precedes_api();
+}
+
+//! Test maintaining correct count of ports without input
+//! \brief \ref error_guessing
+TEST_CASE("Test removal of the predecessor while having none") {
+    using namespace multiple_predecessors;
+
+    test(connect_join_via_follows);
+    test(connect_join_via_precedes);
```
aleksei-fedotov, line 97: Is it (`follows` and `precedes`)-specific?

test/tbb/test_join_node_preview.cpp
```diff
@@ -0,0 +1,98 @@
+/*
```
aleksei-fedotov, line 1: I see this file is supposed to test preview features of the join node. Yet, it seems to test a regression (not necessarily connected with preview features such as `follows` and `precedes`). Also, I see message key matching is placed in separate test files, though it is also a preview feature. I wonder how can we organize tests for join node better?

cmake/compilers/GNU.cmake
isaevil: @AlexVeprev As I understand, there is no such variable as `CMAKE_EXE_CXX_FLAGS`. We need to set `-fPIE` together with `-mxgot` only for tests only in debug. We can leave this as is or move this whole line to `cmake/toolchains/mips.cmake`.

aleksei-fedotov: According to the usage of `-DTBB_TEST_LOW_WORKLOAD` macro there is already a reduction in tests for MIPS. Perhaps, it may be better to reuse it in a more aggressive way even without splitting tests on files and yet test everything we need and still fit in GOT. Have not we considered this option as well?

test/tbb/test_join_node_preview.cpp
isaevil: Do you suggest move it to the test case "Test follows and precedes API"?

test/tbb/test_join_node_preview.cpp
isaevil: I admit the possibility that I overdid the splitting because only splitting message key matching test was really helpful to solve the issue and maybe this `test_join_node_preview` is not needed...

isaevil: As for `-DTBB_TEST_LOW_WORKLOAD` approach (I can't reply to thread thread below) I guess we just reduce scope of testing on MIPS but the problem is not high workload but a high number of symbols. Splitting testing into separate file doesn't affect test scope at all.

cmake/compilers/GNU.cmake
isaevil: As for -DTBB_TEST_LOW_WORKLOAD approach I guess we just reduce scope of testing on MIPS but the problem is not high workload but a high number of symbols. Splitting testing into separate file doesn't affect test scope at all.

test/tbb/test_join_node_preview.cpp
aleksei-fedotov: No, I am trying to understand if this test is preview-specific or not, and judging by the name of the test case, it should not be. However, we need to check that. And, if it is not preview-specific, I suggest rewriting it by not using preview features, and then move out of this file to better separate preview and non-preview testing of the join node.

test/tbb/test_join_node_preview.cpp
aleksei-fedotov: What if we move all the test of preview features of the join_node inside this file. Will that work?

isaevil: This file already contains code parts under PREVIEW macro from other *join_node* testing files. Exception is `test_join_node_msg_key_matching`.

test/tbb/test_join_node_preview.cpp
isaevil: I believe it is because functions connect_join_via_follows and connect_join_via_precedes contain calls of `follows(...)` and `precedes(...)` functions. And in original testing this part was also under PREVIEW macro.

AlexVeprev (dismissed):

AlexVeprev (approved):

aleksei-fedotov (approved):

test/tbb/test_join_node_preview.cpp
aleksei-fedotov: As per offline discussion, moving all the tests of preview features inside this single file will presumably break linkage again.

test/tbb/test_join_node_preview.cpp
aleksei-fedotov: Decided to leave a TODO about further analysis into this.

PR #748, opened 2022-01-24
Check if getcontext() exists to enable resumable tasks

rui314:
getcontext/setcontext functions are deprecated in the recent versions
of POSIX and may not exist. musl libc does not provide these functions.

Signed-off-by: Rui Ueyama <ruiu@cs.stanford.edu>

### Description 
Currently, if you build libtbb.so with musl libc, the resulting libtbb.so is not loadable because `getcontext` and `setcontext` are undefined. These functions are deprecated POSIX functions and because of that reason, musl indeed doesn't define them.

This patch disables resumable tasks so that libtbb.so built with musl become usable.

### Type of change

Bug fix though the bug has not been filed.

### Tests

Not needed

### Documentation

Not needed

No

### Other information

Related to https://github.com/rui314/mold/issues/281 and https://github.com/rui314/mold/issues/295

### DIFF
--- a/include/oneapi/tbb/detail/_config.h
+++ b/include/oneapi/tbb/detail/_config.h
@@ -268,7 +268,7 @@
     #define __TBB_CPP20_COMPARISONS_PRESENT __TBB_CPP20_PRESENT
 #endif
 
-#define __TBB_RESUMABLE_TASKS                           (!__TBB_WIN8UI_SUPPORT && !__ANDROID__ && !__QNXNTO__)
+#define __TBB_RESUMABLE_TASKS                           (!__TBB_WIN8UI_SUPPORT && !__ANDROID__ && !__QNXNTO__ && (!__linux__ || __GLIBC__))
 
 /* This macro marks incomplete code or comments describing ideas which are considered for the future.
  * See also for plain comment with TODO and FIXME marks for small improvement opportunities.


stefson:
I tested the patch with tbb-2021.5.0, here is the output of ldd for libtbb.so.12.5 at runtime: 

```
        ldd (0x7faeb9d000)
        libstdc++.so.6 => /usr/lib/gcc/aarch64-gentoo-linux-musl/10.3.0/libstdc++.so.6 (0x7fae976000)
        libgcc_s.so.1 => /usr/lib/gcc/aarch64-gentoo-linux-musl/10.3.0/libgcc_s.so.1 (0x7fae952000)
        libc.so => ldd (0x7faeb9d000)
```

rui314:
I pushed a new change to attempt to fix the workflows.

stefson:
@rui314 care to explain why you apparently removed the check for getcontext in CMakeLists.txt?

rui314:
@stefson All existing macros in this file are defined this way. It looks like TBB doesn't use CMake features that much and instead relies on C macros.

My initial PR was actually completely broken. It didn't even compile because it didn't define HAVE_GETCONTEXT at all in any .h.

rui314:
Ping?

alexey-katranov:
@rui314, will the patch affect MinGW?

Notify: @jschueller, @longnguyen2004

rui314:
@alexey-katranov
I don't think this patch affects MinGW, as `__linux__` should be 0 on MinGW and therefore `(!__linux__ || __GLIBC__)` is true on MinGW.

alexey-katranov:
> @alexey-katranov I don't think this patch affects MinGW, as `__linux__` should be 0 on MinGW and therefore `(!__linux__ || __GLIBC__)` is true on MinGW.

Yes, you are right!

alexey-katranov (approved):

rui314:
Thanks!

PR #758, opened 2022-01-31
Big market refactoring

pavelkumbrasev:
### Description 
This patch propose refactoring of the market(threading dispatching layer). 
The main idea is to split market on the set off classes with unique responsibility for each class. Also finally uncouple market from arena and make future development easer.
  
### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [] Unknown

@aleksei-fedotov @isaevil @kboyarinov


src/tbb/thread_dispatcher.h
```diff
@@ -0,0 +1,271 @@
...
+
+    //! Pointer to the RML server object that services this TBB instance.
+    rml::tbb_server* my_server{nullptr};
+};
+
+} // namespace r1
+} // namespace detail
+} // namespace tbb
+
+#endif // _TBB_thread_dispatcher_H
```
alexey-katranov, line 271:
```suggestion
#endif // _TBB_thread_dispatcher_H

```

pavelkumbrasev:
This diagram can be used for review class responsibilities and relationship (connections).
image

pavelkumbrasev:
image

aleksei-fedotov (changes requested):
Publishing comments for partially reviewed code, to start the process of their earlier incorporation. @isaevil @kboyarinov I suggest you do the same.

@pavelkumbrasev Please do not update the patch before people finish with the review of current version. Otherwise, I think people might get lost with what part they have reviewed and what is still to be reviewed.

kboyarinov (commented):
Publishing the first part of comments

src/tbb/threading_control.cpp
```diff
@@ -0,0 +1,397 @@
...
+    if (my_thread_dispatcher->try_unregister_client(deleter.my_td_client, deleter.aba_epoch, deleter.priority_level)) {
+        my_permit_manager->destroy_client(*deleter.my_pm_client);
+        return true;
+    }
+    return false;
+}
+
+void threading_control_impl::publish_client(threading_control_impl::threading_control_client client) {
+    my_permit_manager->register_client(static_cast<pm_client*>(client.first));
+    my_thread_dispatcher->register_client(client.second);
```
aleksei-fedotov, line 133: I would recommend to avoid using pairs since they are hard to read. One would need to jump back and forth in the code in order to understand what this `client.second` means. As for me, the whole lines says to thread dispatcher "register a second client for me", which is, obviously, not correct.
But how to change it so that it is read correctly?

src/tbb/threading_control.cpp
```diff
@@ -0,0 +1,397 @@
...
+
+void threading_control_impl::adjust_demand(threading_control_impl::threading_control_client tc_client, int mandatory_delta, int workers_delta) {
+    pm_client& c = *static_cast<pm_client*>(tc_client.first);
+    my_thread_request_serializer->register_mandatory_request(mandatory_delta);
+    my_permit_manager->adjust_demand(c, mandatory_delta, workers_delta);
+}
+
+thread_control_monitor& threading_control_impl::get_waiting_threads_monitor() {
+    return *my_waiting_threads_monitor;
+}
```
aleksei-fedotov, line 169: What is the reason to have a getter of the field? Is it supposed to do something more intelligent in the future? Why not put it into `public` section instead?

src/tbb/threading_control.cpp
```diff
@@ -0,0 +1,397 @@
...
+        global_control_lock();
+        global_mutex_type::scoped_lock lock(g_threading_control_mutex);
+
+        thr_control = get_threading_control(/*public = */ true);
+        if (thr_control != nullptr) {
+            return;
+        }
+
+        // POSSIBLE MEMORY LEAK THERE!!!!!!!
+        thr_control = new (cache_aligned_allocate(sizeof(threading_control))) threading_control(1, 1);
```
aleksei-fedotov, line 221: What is `threading_control(1, 1)`?

src/tbb/threading_control.cpp
```diff
@@ -0,0 +1,397 @@
...
+        my_pimpl->release(blocking_terminate);
+        return blocking_terminate;
+    }
+    return false;
+}
+
+threading_control::threading_control(unsigned public_ref, unsigned ref) : my_public_ref_count(public_ref), my_ref_count(ref)
+{}
+
+threading_control* threading_control::register_public_reference() {
```
aleksei-fedotov, line 283: Seems strange that the `register_public_reference` returns something.

src/tbb/threading_control.cpp
```diff
@@ -0,0 +1,397 @@
...
+    threading_control* control{nullptr};
+    global_mutex_type::scoped_lock lock(g_threading_control_mutex);
+    control = get_threading_control(/*public = */ true);
```
aleksei-fedotov, line 286: It reads as if there is private threading control.
I would actually name this function something like "get_threading_control" since it returns it, and add the ref along the way or create it if necessary.

src/tbb/threading_control.cpp
```diff
@@ -0,0 +1,397 @@
...
+    return my_pimpl->publish_client(client);
+}
+
+bool threading_control::check_client_priority(threading_control::threading_control_client client) {
+    return my_pimpl->check_client_priority(client);
+}
+
+threading_control::client_deleter threading_control::prepare_client_destruction(threading_control::threading_control_client client) {
+    return my_pimpl->prepare_client_destruction(client);
+}
```
aleksei-fedotov, line 321: Again, it seems there will be more value if we implement inheritance rather than pointer to implementation. Am I wrong? For example, we will just not have these methods implemented.

src/tbb/threading_control.cpp
```diff
@@ -0,0 +1,397 @@
...
+        global_mutex_type::scoped_lock lock(g_threading_control_mutex);
+        if (blocking_terminate) {
+            __TBB_ASSERT(is_public, "Only an object with a public reference can request the blocking terminate");
+            wait_last_reference(lock);
+        }
+        do_release = remove_ref(is_public);
+    }
+
+    if (do_release) {
+        __TBB_ASSERT(!my_public_ref_count.load(std::memory_order_relaxed), "No public references remain if we remove the threading control.");
```
aleksei-fedotov, line 272: Shall here be the word "must"?
```suggestion
        __TBB_ASSERT(!my_public_ref_count.load(std::memory_order_relaxed), 
                     "No public references must remain if we remove the threading control.");
```

src/tbb/threading_control.cpp
```diff
@@ -0,0 +1,397 @@
...
+
+    if (workers_soft_limit >= workers_hard_limit) {
+        workers_soft_limit = workers_hard_limit - 1;
+    }
+
+    return workers_soft_limit;
+}
+
+d1::cache_aligned_unique_ptr<permit_manager> threading_control_impl::make_permit_manager(unsigned workers_soft_limit) {
+    return d1::make_cache_aligned_unique<market>(workers_soft_limit);
```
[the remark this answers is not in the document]
pavelkumbrasev: We might have different implementations of `PM` and only one for  `thread_request_serializer` thats why I used such approach.

src/tbb/threading_control.cpp
pavelkumbrasev: one public and one private reference the go in pair on creation. I just want to expose this logic from default ctor and make it more obvious.

src/tbb/threading_control.cpp
pavelkumbrasev: Applied

src/tbb/threading_control.cpp
pavelkumbrasev: So order should be -> if obj exist - yes -> inc ref - no > create it -> inc ref -> return obj
Do you propose create two different function? One for creation and another just getter? They should follow each other everywhere

src/tbb/threading_control.cpp
pavelkumbrasev: I tried to expose public and private references. get_threading_control is definitely suitable but kills this idea.

src/tbb/threading_control.cpp
pavelkumbrasev: Inheritance is really powerful tool and should be used accurately (should follow generalization principles e.g. [LSP](https://en.wikipedia.org/wiki/Liskov_substitution_principle)). Since you propose to use it only to remove code duplication I prefer not to do so.

src/tbb/threading_control.h
```diff
@@ -0,0 +1,157 @@
...
+namespace detail {
+namespace r1 {
+
+class arena;
+class thread_data;
+
+class threading_control;
+
+class threading_control_impl {
+    using global_mutex_type = d1::mutex;
```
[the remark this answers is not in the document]
pavelkumbrasev: Removed

src/tbb/resource_manager.h
```diff
@@ -0,0 +1,64 @@
...
+    virtual void destroy_client(permit_manager_client& c) = 0;
+
+    virtual void request_demand(unsigned min, unsigned max, permit_manager_client&) = 0;
+    virtual void release_demand(permit_manager_client&) = 0;
+
+    // Remove this trash from PM
+    virtual bool try_destroy_arena(permit_manager_client*, uintptr_t aba_epoch, unsigned priority_level) = 0;
+    virtual void adjust_demand (permit_manager_client&, int delta, bool mandatory ) = 0;
+    virtual void enable_mandatory_concurrency(permit_manager_client* c) = 0;
+    virtual void mandatory_concurrency_disable(permit_manager_client* c) = 0;
```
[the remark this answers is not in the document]
pavelkumbrasev: Where did you find this file ? :)

include/oneapi/tbb/mutex.h
```diff
@@ -36,9 +36,7 @@ class mutex {
     };
 
     //! Destructor
-    ~mutex() {
-        __TBB_ASSERT(!my_flag.load(std::memory_order_relaxed), "destruction of an acquired mutex");
-    }
+    ~mutex() = default;
```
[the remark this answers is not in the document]
aleksei-fedotov: > the destructor of the mutex might be called before real mutex EOL

How come is that? Can you please elaborate on this?

include/oneapi/tbb/task_group.h
```diff
@@ -51,7 +51,7 @@ class task_group_base;
 namespace r1 {
 // Forward declarations
 class tbb_exception_ptr;
-class market;
+class cancellation_disseminator;
```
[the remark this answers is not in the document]
aleksei-fedotov: I mean using the generalization of a particular pieces of data the class store.
`car_factory` is a good name because it says that the class with that name manages all the logic and states related to manufacturing of cars.

I would expect to name it something like `cancellation_thread_list` or similarly, with the accent on the list of threads since it is what this class actually maintains. Perhaps, even move the word "cancellation" in the corresponding method name, with possible extension of the method list so that the class does not only cancellation propagation.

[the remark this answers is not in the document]
pavelkumbrasev: The real propose of this class is to collect inside it external and worker threads. It doesn't matter how they are stored inside the class. The name `cancellation_disseminator` here have similar properties with `car_factry` since it manage cancelation logic across TBB via storing all types of threads and disseminate the signal across them - this is a abstraction for this class.

src/tbb/arena.cpp
```diff
@@ -274,80 +341,93 @@ bool arena::has_enqueued_tasks() {
...
+    return no_available_tasks;
+}
+
+void arena::is_out_of_work() {
+    bool disable_mandatory = false;
+    bool release_workers = false;
+
+    // We should try unset my_pool_state first due to keep arena invariants in consistent state
+    // Otherwise, we might have my_pool_state = false and my_mandatory_concurrency = true that is broken invariant
+    release_workers = my_pool_state.try_clear_if([this] { return is_arena_empty(); });
```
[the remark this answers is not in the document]
pavelkumbrasev: Ok, since there are no logic between variables initialization and real logic let's merge them

src/tbb/arena.h
```diff
@@ -372,8 +343,7 @@ class arena: public padded<arena_base>
 #endif
 
     //! Check if there is job anywhere in arena.
-    /** Return true if no job or if arena is being cleaned up. */
-    bool is_out_of_work();
+    void is_out_of_work();
```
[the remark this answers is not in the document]
pavelkumbrasev: Removed is, now it `out_of_work`

src/tbb/cancellation_disseminator.h
```diff
@@ -0,0 +1,85 @@
...
+            return false;
+        }
+
+        // Advance global state propagation epoch
+        ++the_context_state_propagation_epoch;
+        // Propagate to all workers and external threads and sync up their local epochs with the global one
+        // The whole propagation sequence is locked, thus no contention is expected
+        for (thread_data_list_type::iterator it = my_threads_list.begin(); it != my_threads_list.end(); it++) {
+            it->propagate_task_group_state(mptr_state, src, new_state);
+        }
```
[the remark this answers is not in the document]
pavelkumbrasev: Applied

src/tbb/cancellation_disseminator.h
```diff
@@ -0,0 +1,85 @@
...
+        threads_list_mutex_type::scoped_lock lock(my_threads_list_mutex);
+        my_threads_list.remove(td);
+    }
+
+private:
+    using thread_data_list_type = intrusive_list<thread_data>;
+    using threads_list_mutex_type = d1::mutex;
+
+    threads_list_mutex_type my_threads_list_mutex;
+    thread_data_list_type my_threads_list;
```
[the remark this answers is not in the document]
pavelkumbrasev: Basically, `thread_data` is a representation for threads and we disseminate signal across threads.
In my mind current is suitable but we might rename it to `my_thread_data_list`. What do you think?

src/tbb/thread_dispatcher.cpp
```diff
@@ -0,0 +1,231 @@
...
+    insert_client(*client);
+}
+
+bool thread_dispatcher::try_unregister_client(thread_dispatcher_client* client, std::uint64_t aba_epoch, unsigned priority) {
+    __TBB_ASSERT(client, nullptr);
+    // we hold reference to the server, so market cannot be destroyed at any moment here
+    __TBB_ASSERT(!is_poisoned(my_server), nullptr);
+    my_list_mutex.lock();
+    client_list_type::iterator it = my_client_list[priority].begin();
+    for ( ; it != my_client_list[priority].end(); ++it ) {
```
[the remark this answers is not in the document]
pavelkumbrasev: Implemented

src/tbb/thread_dispatcher.cpp
```diff
@@ -0,0 +1,231 @@
...
+    __TBB_ASSERT(client.priority_level() < num_priority_levels, nullptr);
+    my_client_list[client.priority_level()].remove(client);
+
+    if (my_next_client == &client) {
+        my_next_client = nullptr;
+    }
+    my_next_client = select_next_client(my_next_client);
+}
+
+bool thread_dispatcher::is_client_in_list(client_list_type& clients, thread_dispatcher_client* client) {
```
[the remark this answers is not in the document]
pavelkumbrasev: It would be inconsistent with other methods. Do you think it is worth it?

test/tbb/test_task_arena.cpp
```diff
@@ -1764,8 +1766,48 @@ struct enqueue_test_helper {
...
+    barrier.wait();
+}
+
+void test_threads_sleep(int concurrency, int reserved_slots) {
+    test_threads_sleep(concurrency, reserved_slots, reserved_slots);
+    test_threads_sleep(concurrency, reserved_slots, 2 * concurrency);
+}
 
+//--------------------------------------------------//
+#if 1
```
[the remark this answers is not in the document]
pavelkumbrasev: Removed

test/tbb/test_task_arena.cpp
```diff
@@ -1884,10 +1926,13 @@ TEST_CASE("Exception thrown during tbb::task_arena::execute call") {
     }(), std::exception );
 }
 #endif // TBB_USE_EXCEPTIONS
+#endif
 //! \brief \ref stress
 TEST_CASE("Stress test with mixing functionality") {
     StressTestMixFunctionality();
 }
+
+#if 1
```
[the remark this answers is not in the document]
pavelkumbrasev: Removed

src/tbb/thread_request_serializer.cpp
```diff
@@ -0,0 +1,136 @@
...
+        if (my_num_mandatory_requests > 0 && !my_is_mandatory_concurrency_enabled) {
+            my_is_mandatory_concurrency_enabled = true;
+            my_serializer.set_active_num_workers(1);
+        }
+    }
+}
+
+void thread_request_serializer_proxy::update(int delta) { my_serializer.update(delta); }
+
+void thread_request_serializer_proxy::try_enable_mandatory_concurrency(mutex_type::scoped_lock& lock, bool should_enable) {
```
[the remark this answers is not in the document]
isaevil: It looks like that we'll never know, was mandatory enable/disable attempt successful/unsuccessful after we call those functions. I guess in current logic we don't need to know. But still I agree with Konstantin.

src/tbb/market.cpp
```diff
@@ -14,394 +14,72 @@
...
-    return my_num_workers_requested - old_request;
+void market::register_client(pm_client* c) {
+    mutex_type::scoped_lock lock(my_mutex);
+    my_clients[c->priority_level()].push_back(c);
 }
 
-void market::set_active_num_workers ( unsigned soft_limit ) {
-    market *m;
-
+void market::destroy_client(pm_client& c) {
```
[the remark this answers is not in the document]
pavelkumbrasev: `unregister_and_destroy_client`

src/tbb/pm_client.h
```diff
@@ -0,0 +1,78 @@
...
+        return my_is_top_priority.store(b, std::memory_order_relaxed);
+    }
+
+    int min_workers() const {
+        __TBB_ASSERT(my_min_workers >= 0, nullptr);
+        return my_min_workers;
+    }
+
+    int max_workers() const {
+        __TBB_ASSERT(my_max_workers >= 0, nullptr);
```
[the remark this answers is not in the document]
pavelkumbrasev: Consider remove asserts in getters

src/tbb/threading_control.cpp
```diff
@@ -0,0 +1,397 @@
...
+}
+
+d1::cache_aligned_unique_ptr<thread_dispatcher> threading_control_impl::make_thread_dispatcher(threading_control& tc,
+                                                                                               unsigned workers_soft_limit,
+                                                                                               unsigned workers_hard_limit)
```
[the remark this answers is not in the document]
aleksei-fedotov: Why did you split on 114th column then? Also, the alignment of the arguments has been broken.

The problem here is that everyone has their own screens. Desktop monitors usually have wider screens than laptop screens. However, even on wider screens people prefer having two or more files open side-by-side, which diminishes the width advantage of the screens. In addition, for larger monitors people increase font size to reduce eye focus exertion when reading the code. Many people also use vertical screens.

One of the extremes is 120 line length. However, even that many characters make the lines wrap when doing side-by-side review on the GitHub using 2.5k resolution, which is the most widely used display nowadays. 
Another extreme is the margin of 72 characters. This is used to be the maximum line length because old-school terminal windows have 72 characters, and punch cards had 80 characters. So, if we frequently read the code on such screens, it makes sense to follow that length of lines. Fortunately, we are not. Unfortunately, we are not aware about how others people read it. After all, we are writing not only for yourself.
I found many proofs on the Internet that people still prefer using 80 characters limit. Here are a couple of them: [Google C++ Style Guide](https://google.github.io/styleguide/cppguide.html#Line_Length) keeps that limit describing pros and cons of that decision, Linux kernel [used to enforce that limit](https://www.phoronix.com/news/Linux-Kernel-Deprecates-80-Col), but now only recommends sticking to it.

I believe there should be some reasonable tradeoff. Let's try to find it. I think the most important aspects of the line length are:
- Be able to read the code continuously without side scrolling of the screen. The horizontal/side scrolling distracts reader's attention. Especially, if the amount of side scrolling is huge. Consider doing horizontal scrolling back and forth for every line of code.
- Limiting complexity of a single line
- Limiting indentation, which in turn limiting the cyclomatic complexity by prohibiting too much of a conditional blocks, loops, and other type of  nesting.

For the 2.5K desktop monitor and FHD laptop the golden mean is ~100 characters.

src/tbb/arena.h
```diff
@@ -397,158 +372,58 @@ class arena: public padded<arena_base>
...
+
+    bool is_arena_workerless() { return my_max_num_workers == 0; }
+
+    bool is_top_priority();
+
+    bool try_join();
+
+    void set_allotment(unsigned allotment);
+
+    std::pair<int, int> update_request(int mandatory_delta, int workers_delta);
```
aleksei-fedotov, line 389: It is unclear what `std::pair<int, int>` semantically means. Consider improve code readability.

src/tbb/arena.h
```diff
@@ -397,158 +372,58 @@ class arena: public padded<arena_base>
 
     std::uintptr_t calculate_stealing_threshold();
 
+    unsigned priority_level() { return my_priority_level; }
+
+    int has_request() { return my_total_num_workers_requested; }
```
aleksei-fedotov, line 377: Isn't the method should return `bool`? Otherwise, consider renaming to explicitly return the number, e.g.:
```suggestion
    int num_requested_workers() const { return my_total_num_workers_requested; }
```

src/tbb/arena.h
```diff
@@ -397,158 +372,58 @@ class arena: public padded<arena_base>
...
+
+    unsigned references() { return my_references.load(std::memory_order_acquire); }
+
+    bool is_arena_workerless() { return my_max_num_workers == 0; }
+
+    bool is_top_priority();
```
aleksei-fedotov, line 383: Consider making these methods (and similar ones) to possess constness:
```suggestion
    unsigned references() const { return my_references.load(std::memory_order_acquire); }

    bool is_arena_workerless() const { return my_max_num_workers == 0; }

    bool is_top_priority() const;
```

src/tbb/threading_control.cpp
```diff
@@ -0,0 +1,396 @@
...
+
+void threading_control_impl::register_thread(thread_data& td) {
+    my_cancellation_disseminator->register_thread(td);
+}
+void threading_control_impl::unregister_thread(thread_data& td) {
+    my_cancellation_disseminator->unregister_thread(td);
+}
+
+void threading_control_impl::propagate_task_group_state(std::atomic<uint32_t> d1::task_group_context::*mptr_state,
+                                d1::task_group_context& src, uint32_t new_state)
```
aleksei-fedotov, line 143:
```suggestion
void threading_control_impl::propagate_task_group_state(std::atomic<uint32_t> d1::task_group_context::*mptr_state,
                                                        d1::task_group_context& src, uint32_t new_state)
```

src/tbb/threading_control.h
```diff
@@ -0,0 +1,155 @@
...
+
+    thread_control_monitor& get_waiting_threads_monitor();
+
+private:
+    static unsigned calc_workers_soft_limit(unsigned workers_soft_limit, unsigned workers_hard_limit);
+    static std::pair<unsigned, unsigned> calculate_workers_limits();
+    static cache_aligned_unique_ptr<permit_manager> make_permit_manager(unsigned workers_soft_limit);
+    static cache_aligned_unique_ptr<thread_dispatcher> make_thread_dispatcher(threading_control& control,
+                                                                                  unsigned workers_soft_limit,
+                                                                                  unsigned workers_hard_limit);
```
aleksei-fedotov, line 82:
```suggestion
    static cache_aligned_unique_ptr<thread_dispatcher> make_thread_dispatcher(threading_control& control,
                                                                              unsigned workers_soft_limit,
                                                                              unsigned workers_hard_limit);
```

src/tbb/arena.cpp
```diff
@@ -199,15 +276,12 @@ arena::arena ( market& m, unsigned num_slots, unsigned num_reserved_slots, unsig
 #if __TBB_PREVIEW_CRITICAL_TASKS
     my_critical_task_stream.initialize(my_num_slots);
 #endif
-#if __TBB_ENQUEUE_ENFORCED_CONCURRENCY
-    my_local_concurrency_requests = 0;
-    my_local_concurrency_flag.clear();
-    my_global_concurrency_mode.store(false, std::memory_order_relaxed);
-#endif
+    my_mandatory_requests = 0;
+    my_mandatory_concurrency.clear();
```
aleksei-fedotov, line 280: AFAIR, `my_mandatory_concurrency` is already in the clear state.
```suggestion
```

src/tbb/arena.cpp
```diff
@@ -254,15 +324,12 @@ void arena::free_arena () {
...
-    // remove an internal reference
-    my_market->release( /*is_public=*/false, /*blocking_terminate=*/false );
-
     // Clear enfources synchronization with observe(false)
     my_observers.clear();
 
     void* storage  = &mailbox(my_num_slots-1);
     __TBB_ASSERT( my_references.load(std::memory_order_relaxed) == 0, nullptr);
-    __TBB_ASSERT( my_pool_state.load(std::memory_order_relaxed) == SNAPSHOT_EMPTY || !my_max_num_workers, nullptr);
+    __TBB_ASSERT( my_pool_state.test() == false|| !my_max_num_workers, nullptr);
```
aleksei-fedotov, line 332: Debug vs Release memory semantics discrepancy detected. Perhaps, add `test_debug()` method under the debug `ifdef` if the assert is that useful?

aleksei-fedotov (commented):
A bunch of new comments.

src/tbb/arena.cpp
kboyarinov: Agreed on adding memory semantics argument to test

src/tbb/arena.h
kboyarinov: Should be `bool` because `thread_dispatcher_client::has_request` returns `bool`

src/tbb/arena.cpp
pavelkumbrasev: Done

src/tbb/arena.h
pavelkumbrasev: Done

src/tbb/cancellation_disseminator.h
```diff
@@ -0,0 +1,85 @@
...
+        propagation cannot be generally localized to one arena only.
+    */
+    bool propagate_task_group_state(std::atomic<uint32_t> d1::task_group_context::*mptr_state, d1::task_group_context& src, uint32_t new_state) {
+        if (src.my_may_have_children.load(std::memory_order_relaxed) != d1::task_group_context::may_have_children) {
+            return true;
+        }
+
+        // The whole propagation algorithm is under the lock in order to ensure correctness
+        // in case of concurrent state changes at the different levels of the context tree.
+        // See comment at the bottom of scheduler.cpp
```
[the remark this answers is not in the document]
pavelkumbrasev: Removed

src/tbb/market.h
```diff
@@ -17,291 +17,61 @@
...
-    int my_num_workers_requested;
+    using mutex_type = d1::rw_mutex;
+    mutex_type my_mutex;
 
-    //! First unused index of worker
-    /** Used to assign indices to the new workers coming from RML, and busy part
-        of my_workers array. **/
-    std::atomic<unsigned> my_first_unused_worker_idx;
+    //! Current application-imposed limit on the number of workers
+    int my_num_workers_soft_limit{0};
```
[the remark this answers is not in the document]
pavelkumbrasev: Done

src/tbb/pm_client.h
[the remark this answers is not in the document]
pavelkumbrasev: Done

src/tbb/arena.h
pavelkumbrasev: Added comment

src/tbb/market.cpp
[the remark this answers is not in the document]
pavelkumbrasev: Applied

src/tbb/thread_dispatcher.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+    return hint;
+}
+
+thread_dispatcher_client* thread_dispatcher::create_client(arena& a) {
+    return new (cache_aligned_allocate(sizeof(thread_dispatcher_client))) thread_dispatcher_client(a, my_clients_aba_epoch);
+}
+
+
+void thread_dispatcher::register_client(thread_dispatcher_client* client) {
```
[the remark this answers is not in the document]
pavelkumbrasev: Added comment

src/tbb/thread_dispatcher.cpp
```diff
@@ -0,0 +1,231 @@
...
+
+void thread_dispatcher::insert_client(thread_dispatcher_client& client) {
+    __TBB_ASSERT(client.priority_level() < num_priority_levels, nullptr);
+    my_client_list[client.priority_level()].push_front(client);
+
+    __TBB_ASSERT(!my_next_client || my_next_client->priority_level() < num_priority_levels, nullptr);
+    my_next_client = select_next_client(my_next_client);
+}
+
+void thread_dispatcher::remove_client(thread_dispatcher_client& client) {
```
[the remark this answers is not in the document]
pavelkumbrasev: Done

src/tbb/thread_request_serializer.cpp
```diff
@@ -0,0 +1,136 @@
...
+    constexpr std::uint64_t delta_mask = 0xFFFF;
+    constexpr std::uint64_t counter_value = delta_mask + 1;
+
+    int prev_pending_delta = my_pending_delta.fetch_add(counter_value + delta);
+
+    // There is a pseudo request aggregator, so only thread that see pending_delta_base in my_pending_delta
+    // Will enter to critical section and call adjust_job_count_estimate
+    if (prev_pending_delta == pending_delta_base) {
+        mutex_type::scoped_lock lock(my_mutex);
+        delta = int(my_pending_delta.exchange(pending_delta_base) & delta_mask) - int(pending_delta_base);
```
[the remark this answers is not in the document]
pavelkumbrasev: Agreed keep it under the mutex for the simplicity

src/tbb/threading_control.cpp
```diff
@@ -0,0 +1,396 @@
...
+    threading_control* control = g_threading_control;
+    if (control) {
+        control->add_ref(is_public);
+    }
+
+    return control;
+}
+
+threading_control* threading_control::create_threading_control() {
+    threading_control* thr_control{ nullptr };
```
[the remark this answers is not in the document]
pavelkumbrasev: Rewrote this function

src/tbb/arena.cpp
```diff
@@ -274,80 +341,93 @@ bool arena::has_enqueued_tasks() {
...
+    int max_workers_request = 0;
+
+    // Calculate min request
+    my_mandatory_requests += mandatory_delta;
+    min_workers_request = my_mandatory_requests > 0 ? 1 : 0;
+
+    // Calculate max request
+    my_total_num_workers_requested += workers_delta;
+    // Clamp worker request into interval [0, my_max_num_workers]
+    max_workers_request = clamp(my_total_num_workers_requested, 0,
```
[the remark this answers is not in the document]
pavelkumbrasev: We would increase mandatory_delta only if arena has a slot for worker, that mean only the case with 1,1 interested

src/tbb/thread_dispatcher.cpp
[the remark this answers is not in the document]
pavelkumbrasev: Agreed to refactor the method since it is hard to apply find to intrusive list

src/tbb/threading_control.cpp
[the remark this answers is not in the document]
pavelkumbrasev: Not only this patch is affected but a bunch of different files. Do you agree to move such improvement to separate activity?

test/tbb/test_arena_priorities.cpp
```diff
@@ -1,5 +1,5 @@
 /*
-    Copyright (c) 2020-2021 Intel Corporation
+    Copyright (c) 2020-2022 Intel Corporation
```
kboyarinov, line 2: Seems like changes in this file should be discarded

src/tbb/threading_control_client.h
```diff
@@ -0,0 +1,56 @@
...
+namespace tbb {
+namespace detail {
+namespace r1 {
+
+class pm_client;
+class thread_dispatcher_client;
+
+class threading_control_client {
+public:
+    threading_control_client() = default;
```
kboyarinov, line 31: Is it really required?

src/tbb/thread_request_serializer.cpp
```diff
@@ -0,0 +1,142 @@
...
+    // case CROSS_LIMIT:
+    //     delta = delta > 0 ? limit - prev_value : new_value - limit;
+    // case BELOW_LIMIT:
+    //     // No changes to delta
+    // }
+
+   int prev_value = new_value - delta;
+
+    // actual new_value and prev_value cannot exceed the limit
+    new_value = std::min(limit, new_value);
```
kboyarinov, line 77: As far as I remember, we already have our own `min` somewhere in src, lets use it here and remove `#include <algorithm>`.

src/tbb/thread_dispatcher_client.h
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    bool has_request() {
+        return my_arena.has_request();
+    }
+
+private:
+    arena& my_arena;
+    std::uint64_t my_aba_epoch;
+    // mask_type m_mask;
```
kboyarinov, line 58: Should be removed?

src/tbb/scheduler_common.h
```diff
@@ -68,6 +69,22 @@ template<task_stream_accessor_type> class task_stream;
...
+        cache_aligned_deallocate(ptr);
+    }
+};
+
+template <typename T>
+using cache_aligned_unique_ptr = std::unique_ptr<T, cache_aligned_deleter>;
+
+template <typename T, typename ...Args>
+cache_aligned_unique_ptr<T> make_cache_aligned_unique(Args&& ...args) {
+    return cache_aligned_unique_ptr<T>(new (r1::cache_aligned_allocate(sizeof(T))) T(std::forward<Args>(args)...));
```
kboyarinov, line 85:
```suggestion
    return cache_aligned_unique_ptr<T>(new (cache_aligned_allocate(sizeof(T))) T(std::forward<Args>(args)...));
```

src/tbb/market.cpp
```diff
@@ -14,394 +14,73 @@
...
-#include "market.h"
-#include "main.h"
-#include "governor.h"
 #include "arena.h"
-#include "thread_data.h"
-#include "itt_notify.h"
+#include "market.h"
 
-#include <cstring> // std::memset()
+#include <algorithm>
```
kboyarinov, line 20:
```suggestion
#include <algorithm> // std::find
```

src/tbb/governor.h
```diff
@@ -29,6 +29,8 @@ namespace r1 {
 class market;
 class thread_data;
 class __TBB_InitOnce;
+class thread_dispatcher;
+class threading_control_impl;
```
kboyarinov, line 33: Seems unused. `friend` will provide forward declaration

kboyarinov (commented):
Minor comments from my side. I guess that last ones

src/tbb/threading_control_client.h
pavelkumbrasev: On arena creation default ctor is used.

test/tbb/test_arena_priorities.cpp
pavelkumbrasev: The empty line was removed :)

src/tbb/thread_request_serializer.cpp
pavelkumbrasev: Done

src/tbb/thread_dispatcher_client.h
pavelkumbrasev: Removed

src/tbb/governor.h
pavelkumbrasev: Removed

test/tbb/test_arena_priorities.cpp
kboyarinov: The last change in the file was removed, so lets roll back the copyright and exclude the whole file from the patch

src/tbb/governor.h
pavelkumbrasev: Removed

src/tbb/thread_dispatcher.cpp
```diff
@@ -0,0 +1,224 @@
...
+
+thread_dispatcher_client* thread_dispatcher::select_next_client(thread_dispatcher_client* hint) {
+    unsigned next_client_priority_level = num_priority_levels;
+    if (hint) {
+        next_client_priority_level = hint->priority_level();
+    }
+
+    for (unsigned idx = 0; idx < next_client_priority_level; ++idx) {
+        if (!my_client_list[idx].empty()) {
+            return &*my_client_list[idx].begin();
```
isaevil, line 45:
```suggestion
            return &my_client_list[idx].front();
```

src/tbb/arena.cpp
```diff
@@ -274,80 +340,94 @@ bool arena::has_enqueued_tasks() {
...
-        my_market->adjust_demand(*this, /* delta = */ -1, /* mandatory = */ true);
+void arena::request_workers(int mandatory_delta, int workers_delta, bool wakeup_threads) {
+    my_threading_control->adjust_demand(my_tc_client, mandatory_delta, workers_delta);
+
+    if (wakeup_threads) {
+        // Notify all sleeping threads that work has appeared in the arena.
+        get_waiting_threads_monitor().notify([&] (market_context context) {
+            return this == context.my_arena_addr;
+        });
     }
```
aleksei-fedotov, line 351: I would think that this part better belong to threading control rather than arena itself. Moreover, if we looked at the implementation of the [`get_waiting_threads_monitor()`](https://github.com/oneapi-src/oneTBB/pull/758/files#diff-277ea5199fd9ffb60eb8432bd91d0b65b7b58bb13ea06e0856b4f5330d337d56R427), we would see that it is a getter from the `threading_control`.
What do you think if we move it to the `threading_control` entity?

Also, is not that waking up the threads decision can be made based on the difference between current and desired number of the worker threads? I mean we should notify iff current < desired, should not we? Are there other conditions for notification?

src/tbb/cancellation_disseminator.h
```diff
@@ -0,0 +1,84 @@
...
+    */
+    bool propagate_task_group_state(std::atomic<uint32_t> d1::task_group_context::*mptr_state, d1::task_group_context& src, uint32_t new_state) {
+        if (src.my_may_have_children.load(std::memory_order_relaxed) != d1::task_group_context::may_have_children) {
+            return true;
+        }
+
+        // The whole propagation algorithm is under the lock in order to ensure correctness
+        // in case of concurrent state changes at the different levels of the context tree.
+        threads_list_mutex_type::scoped_lock lock(my_threads_list_mutex);
+        if ((src.*mptr_state).load(std::memory_order_relaxed) != new_state) {
```
aleksei-fedotov, line 46: Do we need to have double check idiom here? Can it be useful here and why?

include/oneapi/tbb/mutex.h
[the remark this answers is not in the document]
kboyarinov: @aleksei-fedotov @pavelkumbrasev,
It seems like Aleksei is right - I have double checked and `mutex` is not trivial type and strictly speaking, accessing memory before the constructor and after the destructor is UB.
`mutex` is not trivial because its default ctor is not trivial. First of all, we have itt call inside of the body and secondly, we initialize `waitable_atomic<bool>` inside of the mutex with `0` which is not trivial.
`waitable_atomic<bool>` is also not a trivial type but it can be easily made trivial by removing `{}` from the atomic variable inside.
I have also checked standard traits `std::is_trivial_v<mutex>` and `std::is_trivial_v<waitable_atomic<bool>>` - both of them are `false`.

It seems like we need to add a TODO comment and resolve this issue separately because it is more complex then just making the destructor trivial. I think we also can return the assert and add a TODO comment because it is not related to this patch.

src/tbb/cancellation_disseminator.h
aleksei-fedotov: That's the whole point of double check idiom, is not that? Theoretically, in the contended cases it may transfer threads to other, non-cancelled, tasks quicker. See [the cancellation point in the `any_of` algorithm](https://github.com/oneapi-src/oneDPL/blob/c697fac0b51ce2a36f3824bb9063dfaf6aee88ac/include/oneapi/dpl/pstl/parallel_impl.h#L83).
I suggest writing a TODO comment to try out double check idiom and measure task cancellation "responsiveness" in contended cases.

src/tbb/arena.h
```diff
@@ -361,7 +361,9 @@ class arena: public padded<arena_base>
     bool has_enqueued_tasks();
 
     //! Check for the presence of any tasks
-    bool is_arena_empty();
+    bool check_work_absence();
+
+    bool is_arena_empty() { return my_pool_state.test() == /* EMPTY */ false; }
```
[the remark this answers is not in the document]
pavelkumbrasev: Done

src/tbb/arena.cpp
```diff
@@ -369,7 +370,7 @@ void arena::out_of_work() {
     // We should try unset my_pool_state first due to keep arena invariants in consistent state
     // Otherwise, we might have my_pool_state = false and my_mandatory_concurrency = true that is broken invariant
     bool disable_mandatory = my_mandatory_concurrency.try_clear_if([this] { return !has_enqueued_tasks(); });
-    bool release_workers = my_pool_state.try_clear_if([this] { return is_arena_empty(); });
+    bool release_workers = my_pool_state.try_clear_if([this] { return check_work_absence(); });
```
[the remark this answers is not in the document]
pavelkumbrasev: Done

aleksei-fedotov (approved):

omalyshe (approved):
LGTM.

PR #760, opened 2022-02-01
Code coverage for overwrite_node

vlserov:
### Description 
Test for coverage cancel function of `oneapi::tbb::flow::overwrite_node::register_predecessor_task`

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

@aleksei-fedotov

### DIFF
--- a/test/tbb/test_overwrite_node.cpp
+++ b/test/tbb/test_overwrite_node.cpp
@@ -230,3 +230,33 @@ TEST_CASE("try_release"){
 
     CHECK_MESSAGE ((on.try_release()== true), "try_release should return true");
 }
+
+template< typename Type >
+struct fake_receiver : public tbb::flow::receiver<Type>, utils::NoAssign {
+    tbb::flow::graph& my_graph;
+
+    fake_receiver(tbb::flow::graph& g) : my_graph(g) {}
+
+    tbb::detail::d1::graph_task* try_put_task(const Type&) override {
+        return nullptr;
+    }
+
+    tbb::flow::graph& graph_reference() const override {
+        return my_graph;
+    }
+};
+
+//! Test for cancel register_predecessor_task
+//! \brief \ref error_guessing
+TEST_CASE("Cancel register_predecessor_task") {
+    tbb::flow::graph g;
+
+    fake_receiver<size_t> r(g);
+    oneapi::tbb::flow::overwrite_node<size_t> node(g);
+
+    node.try_put(1);
+    tbb::flow::make_edge(node, r);
+
+    g.cancel();
+    g.wait_for_all();
+}
--- a/test/tbb/test_overwrite_node.cpp
+++ b/test/tbb/test_overwrite_node.cpp
@@ -231,31 +231,16 @@ TEST_CASE("try_release"){
     CHECK_MESSAGE ((on.try_release()== true), "try_release should return true");
 }
 
-template< typename Type >
-struct fake_receiver : public tbb::flow::receiver<Type>, utils::NoAssign {
-    tbb::flow::graph& my_graph;
-
-    fake_receiver(tbb::flow::graph& g) : my_graph(g) {}
-
-    tbb::detail::d1::graph_task* try_put_task(const Type&) override {
-        return nullptr;
-    }
-
-    tbb::flow::graph& graph_reference() const override {
-        return my_graph;
-    }
-};
-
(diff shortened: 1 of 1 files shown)


test/tbb/test_overwrite_node.cpp
```diff
@@ -230,3 +230,33 @@ TEST_CASE("try_release"){
...
+
+template< typename Type >
+struct fake_receiver : public tbb::flow::receiver<Type>, utils::NoAssign {
+    tbb::flow::graph& my_graph;
+
+    fake_receiver(tbb::flow::graph& g) : my_graph(g) {}
+
+    tbb::detail::d1::graph_task* try_put_task(const Type&) override {
+        return nullptr;
+    }
```
aleksei-fedotov, line 242: If I remember correctly, returning `nullptr` is the very reason to turn over the edge state from push to pull, which implies spawning of a `register_predecessor_task`. (BTW, since `overwrite_node` here is a predecessor and `fake_receiver` is its successor, why not it is called a `register_successor_task`?)
However, I cannot understand why that spawned task becomes cancelled? Could you please remind to me?
Writing a detailed comment for this test to clarify and explain the details to its future readers would even be better.

alexey-katranov (dismissed):

test/tbb/test_overwrite_node.cpp
vlserov: Updated with using `join_node` instead of custom receiver.

To spawn `register_predecessor_task` of `overwrite_node` internal buffer should be valid (calling `overwrite_node::try_put`) and successor should return `nullptr` during call `try_put_task`.

Since group context is canceled by calling `graph::cancel()`, all spawned tasks will be cancelled during `graph::wait_for_all`.  `graph::wait_for_all` may be directly called like in test or as part of `graph` destructor.

aleksei-fedotov: > Since group context is canceled by calling graph::cancel(), all spawned tasks will be cancelled 

Is it guaranteed? It seems to me that they can also be executed once spawned, even before graph.cancel() or even graph.wait_for_all(), cannot they?

vlserov: Race is possible. Updating review by moving `graph.cancel()` before task spawn. Task will be cancelled by worker or external thread.

test/tbb/test_overwrite_node.cpp
```diff
@@ -230,3 +230,27 @@ TEST_CASE("try_release"){
...
+    // should be valid and successor should failed during putting an item to it
+    oneapi::tbb::flow::overwrite_node<size_t> node(g);
+    // join_node always fails during putting an item to it
+    tbb::flow::join_node<std::tuple<size_t>, tbb::flow::reserving> j_node(g);
+
+    // Make internal buffer of overwrite_node valid
+    node.try_put(1);
+    // Make edge between overwrite_node and join_node, register join_node as
+    // overwrite_node successor and spanw register_predecessor_task
+    tbb::flow::make_edge(node, tbb::flow::input_port<0>(j_node));
```
aleksei-fedotov, line 252: I would rewrite a bit to clarify when certain things exactly happen:
```suggestion
    // To spawn register_predecessor_task internal buffer of overwrite_node
    // should be valid and successor should failed during putting an item to it
    oneapi::tbb::flow::overwrite_node<size_t> node(g);
    // Reserving join_node always fails during putting an item to it
    tbb::flow::join_node<std::tuple<size_t>, tbb::flow::reserving> j_node(g);

    // Make internal buffer of overwrite_node valid
    node.try_put(1);
    // Making an edge attempts pushing an item to join_node
    // that immediately fails and tries to reverse an edge into PULL state
    // by spawning register_predecessor_task, which will be cancelled
    // during execution
    tbb::flow::make_edge(node, tbb::flow::input_port<0>(j_node));
```

test/tbb/test_overwrite_node.cpp
```diff
@@ -230,3 +230,27 @@ TEST_CASE("try_release"){
...
+
+    // Wait for cancellation spawned tasks
```
aleksei-fedotov, line 254: A typo:
```suggestion
    // Wait for cancellation of spawned tasks
```

aleksei-fedotov (approved):

PR #842, opened 2022-05-31
Tbb numa bug fix

AnuyaWelling2801:
### Description 

### Type of change

- [X ] bug fix - _change that fixes an issue_

### Tests

- [X ] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [X ] No
- [ ] Unknown

### DIFF
--- a/src/tbb/misc_ex.cpp
+++ b/src/tbb/misc_ex.cpp
@@ -150,7 +150,7 @@ static void initialize_hardware_concurrency_info () {
         int pid = getpid();
         err = sched_getaffinity( pid, curMaskSize, processMask );
         if ( !err || errno != EINVAL || curMaskSize * CHAR_BIT >= 256 * 1024 )
-            break;
+             break;
 #endif
         delete[] processMask;
         numMasks <<= 1;
@@ -257,7 +257,8 @@ int ProcessorGroupInfo::NumGroups = 1;
 int ProcessorGroupInfo::HoleIndex = 0;
 
 ProcessorGroupInfo theProcessorGroups[MaxProcessorGroups];
-
+int calculate_numa[MaxProcessorGroups];  //Array needed for FindProcessorGroupIndex
+int numaSum;
 struct TBB_GROUP_AFFINITY {
     DWORD_PTR Mask;
     WORD   Group;
@@ -312,15 +313,27 @@ static void initialize_hardware_concurrency_info () {
             if ( TBB_GetThreadGroupAffinity( GetCurrentThread(), &ga ) )
                 ProcessorGroupInfo::HoleIndex = ga.Group;
             int nprocs = 0;
+            int min_procs = INT_MAX;
             for ( WORD i = 0; i < ProcessorGroupInfo::NumGroups; ++i ) {
                 ProcessorGroupInfo  &pgi = theProcessorGroups[i];
                 pgi.numProcs = (int)TBB_GetActiveProcessorCount(i);
+                if (pgi.numProcs < min_procs) min_procs = pgi.numProcs;  //Finding the minimum number of processors in the Processor Groups
+                calculate_numa[i] = pgi.numProcs;
                 __TBB_ASSERT( pgi.numProcs <= (int)sizeof(DWORD_PTR) * CHAR_BIT, nullptr);
                 pgi.mask = pgi.numProcs == sizeof(DWORD_PTR) * CHAR_BIT ? ~(DWORD_PTR)0 : (DWORD_PTR(1) << pgi.numProcs) - 1;
                 pgi.numProcsRunningTotal = nprocs += pgi.numProcs;
             }
             __TBB_ASSERT( nprocs == (int)TBB_GetActiveProcessorCount( TBB_ALL_PROCESSOR_GROUPS ), nullptr);
+
+            for (WORD i = 0; i < ProcessorGroupInfo::NumGroups; ++i) {
+                if (i == 0) calculate_numa[i] = (calculate_numa[i] / min_procs) - 1;
+                else calculate_numa[i] = calculate_numa[i-1] + (calculate_numa[i] / min_procs);
+                
+            }
+            numaSum = calculate_numa[ProcessorGroupInfo::NumGroups - 1]+1;
+
         }
+
     }
 #endif /* __TBB_WIN8UI_SUPPORT */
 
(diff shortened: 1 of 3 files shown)


src/tbb/misc_ex.cpp
```diff
@@ -369,6 +370,48 @@ int FindProcessorGroupIndex ( int procIdx ) {
...
+            // shift the worker_id index to position it on processor group same as master thread
+            proc_id_tracker = procIdx + (theProcessorGroups[ProcessorGroupInfo::HoleIndex - 1].numProcsRunningTotal);
+    } else {
+            // master thread is in group 0 => No shifting required
+            proc_id_tracker = procIdx;
+    }
+
+    if (procIdx > theProcessorGroups[ProcessorGroupInfo::NumGroups - 1].numProcsRunningTotal) {
+        // over subscription - Round Robin.
+        current_grp_idx = (current_grp_idx + 1) % (ProcessorGroupInfo::NumGroups);
```
vossmjp, line 397: Does this achieve round robin? current_group_idx is a local variable that is set to HoleIndex. I think HoleIndex does not change and so the initial value of the local current_grp_idx will not change regardless of the procIdx.  And so if there is oversubscription, the current_grp_idx will always be (HoleIndex + 1)%NumGroups. Or am I missing something...?

src/tbb/misc_ex.cpp
```diff
@@ -369,6 +370,48 @@ int FindProcessorGroupIndex ( int procIdx ) {
...
+    } else {
+        // check if there is space in current Numa node (processor group)
+        if (proc_id_tracker < theProcessorGroups[current_grp_idx].numProcsRunningTotal) {
```
vossmjp, line 400: I'd suggest changing the condition so there is not an empty if body, for example change the condition to >= and move the else block here.

src/tbb/misc_ex.cpp
```diff
@@ -335,6 +335,7 @@ int NumberOfProcessorGroups() {
     return ProcessorGroupInfo::NumGroups;
 }
 
+#if 0 // old-code for win-numa(Spread thread affinity)
```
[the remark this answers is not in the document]
AnuyaWelling2801: Fixed.

src/tbb/misc_ex.cpp
```diff
@@ -369,6 +370,48 @@ int FindProcessorGroupIndex ( int procIdx ) {
     return i;
 }
 
+#endif
```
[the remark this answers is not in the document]
AnuyaWelling2801: Fixed

src/tbb/misc_ex.cpp
AnuyaWelling2801: Discussed offline and fixed.

src/tbb/misc_ex.cpp
AnuyaWelling2801: Fixed.

src/tbb/misc_ex.cpp
```diff
@@ -335,38 +348,29 @@ int NumberOfProcessorGroups() {
...
-    // Approximate the likely group index assuming all groups are of the same size
-    int i = procIdx / theProcessorGroups[0].numProcs;
-    // Make sure the approximation is a valid group index
-    if (i >= ProcessorGroupInfo::NumGroups) i = ProcessorGroupInfo::NumGroups-1;
-    // Now adjust the approximation up or down
-    if ( theProcessorGroups[i].numProcsRunningTotal > HoleAdjusted(procIdx, i) ) {
-        while ( theProcessorGroups[i].numProcsRunningTotal - theProcessorGroups[i].numProcs > HoleAdjusted(procIdx, i) ) {
-            __TBB_ASSERT( i > 0, nullptr);
-            --i;
+    else if (procIdx > theProcessorGroups[ProcessorGroupInfo::NumGroups - 1].numProcsRunningTotal) {
```
dnmokhov, line 361: \> numProcsRunningTotal ***- 1***, right?

src/tbb/misc_ex.cpp
```diff
@@ -335,38 +348,29 @@ int NumberOfProcessorGroups() {
...
-
 int FindProcessorGroupIndex ( int procIdx ) {
-    // In case of oversubscription spread extra workers in a round robin manner
-    int holeIdx;
-    const int numProcs = theProcessorGroups[ProcessorGroupInfo::NumGroups - 1].numProcsRunningTotal;
-    if ( procIdx >= numProcs - 1 ) {
-        holeIdx = INT_MAX;
-        procIdx = (procIdx - numProcs + 1) % numProcs;
+    int current_grp_idx = ProcessorGroupInfo::HoleIndex;
+    if (procIdx > theProcessorGroups[ProcessorGroupInfo::HoleIndex].numProcs-1  && procIdx <= theProcessorGroups[ProcessorGroupInfo::NumGroups - 1].numProcsRunningTotal-1 ) {
```
dnmokhov, line 353: `theProcessorGroups[ProcessorGroupInfo::HoleIndex]` → `theProcessorGroups[current_grp_idx]` would be more clear, since it is used two more times below.

src/tbb/misc_ex.cpp
```diff
@@ -335,38 +348,29 @@ int NumberOfProcessorGroups() {
...
-    if (i >= ProcessorGroupInfo::NumGroups) i = ProcessorGroupInfo::NumGroups-1;
-    // Now adjust the approximation up or down
-    if ( theProcessorGroups[i].numProcsRunningTotal > HoleAdjusted(procIdx, i) ) {
-        while ( theProcessorGroups[i].numProcsRunningTotal - theProcessorGroups[i].numProcs > HoleAdjusted(procIdx, i) ) {
-            __TBB_ASSERT( i > 0, nullptr);
-            --i;
+    else if (procIdx > theProcessorGroups[ProcessorGroupInfo::NumGroups - 1].numProcsRunningTotal) {
+        int temp_grp_index = 0;
+        procIdx = procIdx - (theProcessorGroups[ProcessorGroupInfo::NumGroups - 1].numProcsRunningTotal); //starts from 0
+        procIdx = procIdx % (numaSum + 1);
```
dnmokhov, line 364: Shouldn't it just be `numaSum` instead of `numaSum + 1`?

dnmokhov (changes requested):

src/tbb/misc_ex.cpp
AnuyaWelling2801: Changed the logic a bit to accommodate this.

src/tbb/misc_ex.cpp
```diff
@@ -257,7 +257,8 @@ int ProcessorGroupInfo::NumGroups = 1;
 int ProcessorGroupInfo::HoleIndex = 0;
 
 ProcessorGroupInfo theProcessorGroups[MaxProcessorGroups];
-
+int calculate_numa[MaxProcessorGroups];  //Array needed for FindProcessorGroupIndex
```
vossmjp, line 260: Maybe add a better comment. Not that is just needed for that function but also what its values mean.

src/tbb/misc_ex.cpp
```diff
@@ -312,15 +313,28 @@ static void initialize_hardware_concurrency_info () {
...
+                if (pgi.numProcs < min_procs) min_procs = pgi.numProcs;  //Finding the minimum number of processors in the Processor Groups
+                calculate_numa[i] = pgi.numProcs;
                 __TBB_ASSERT( pgi.numProcs <= (int)sizeof(DWORD_PTR) * CHAR_BIT, nullptr);
                 pgi.mask = pgi.numProcs == sizeof(DWORD_PTR) * CHAR_BIT ? ~(DWORD_PTR)0 : (DWORD_PTR(1) << pgi.numProcs) - 1;
                 pgi.numProcsRunningTotal = nprocs += pgi.numProcs;
             }
             __TBB_ASSERT( nprocs == (int)TBB_GetActiveProcessorCount( TBB_ALL_PROCESSOR_GROUPS ), nullptr);
+
+            for (WORD i = 0; i < ProcessorGroupInfo::NumGroups; ++i) {
+                if (i == 0) calculate_numa[i] = (calculate_numa[i] / min_procs);
```
vossmjp, line 329: You could put this before the loop and then run the for loop from i =1.  It basically initializes calculate_numa[0]. The loop body would be simpler then.

src/tbb/misc_ex.cpp
```diff
@@ -312,15 +313,28 @@ static void initialize_hardware_concurrency_info () {
...
+                else calculate_numa[i] = calculate_numa[i-1] + (calculate_numa[i] / min_procs);
```
vossmjp, line 330: Does the (calculate_numa[i] / min_procs) assume that the number of cores in each processor group is an exact multiple of min_procs. Since this is integer division it will round down if its not an exact multiple -- is that ok?

src/tbb/misc_ex.cpp
```diff
@@ -312,15 +313,28 @@ static void initialize_hardware_concurrency_info () {
...
+                
+            }
+
+            numaSum = calculate_numa[ProcessorGroupInfo::NumGroups - 1]+1;
```
vossmjp, line 334: Why +1?  Maybe this is right, but it seems like you are calculating a prefix sum of the number of min_procs groups. I'm not sure why you are adjusting by 1.

src/tbb/misc_ex.cpp
AnuyaWelling2801: Yes, that's ok

src/tbb/misc_ex.cpp
AnuyaWelling2801: Should be +1 so that the ProcIdx in FindProcessorGroupIndex stays between 1 and calculate_numa[ProcessorGroupInfo::NumGroups - 1], but makes sense to have this later in the function. Fixed.

dnmokhov (approved):

PR #845, opened 2022-06-15
Build tbbmalloc also on mips

AdrianBunk:
Commit 51c0b2f742920535178560f31c6e91065bf87b41 does not give any information why tbbmalloc is not being built on mips, it would be useful if someone could check the rationale in the original git tree.

There is an (unused/obsolete?) https://github.com/oneapi-src/oneTBB/blob/master/cmake/toolchains/mips.cmake and I wonder whether not building tbbmalloc might have been for this toolchain instead of the processor.

On Debian tbbmalloc builds with gcc for both 32bit and 64bit (little endian) mips.
Enabling all currently disabled tests in `test/CMakeLists.txt` resulted in 133 passes for 64bit and only `test_malloc_pools` failing on 32bit, mips tests will be a separate pull request.

Signed-off-by: Adrian Bunk <bunk@debian.org>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright (c) 2020-2022 Intel Corporation
+# Copyright (c) 2020-2023 Intel Corporation
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


pavelkumbrasev (approved):

pavelkumbrasev:
Hi @AdrianBunk, could you please update a year in copyright? (from 2022 to 2023)

AdrianBunk:
done

AdrianBunk:
> Hi @AdrianBunk, could you please update a year in copyright? (from 2022 to 2023)

@pavelkumbrasev Done.

pavelkumbrasev (approved):

PR #846, opened 2022-06-15
Add Win Numa test-cases.

sarathnandu:
### Description 
Add an initial version of test-case for testing Win Numa behavior.

### Type of change

### Tests

- [x ] added - _required for new features and some bug fixes_
- [ ] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown


test/tbb/test_numa_dist.cpp
```diff
@@ -0,0 +1,109 @@
...
+        {
+            // dummy work
+            double& b = tls_dummy.local();
+            for (int k = 0; k < 1000000; ++k)
+                b += pow(i * k, 1. / 3.);
+
+            PROCESSOR_NUMBER proc;
+            if (GetThreadIdealProcessorEx(GetCurrentThread(), &proc))
+            {
+                tls.local() = std::pair<int, int>(proc.Group, proc.Number);
```
vossmjp, line 92: I think you should check if tls.local() is already set. If so, its .first should also be equal to proc.Group. Since you are only keeping the last value in tls.local, the test would pass if a thread were to execute a chunk of iterations from the wrong processor group before moving to the correct one for the final chunk it executes. While I doubt that would happen, I think the test could be slightly modified to test for that case anyway.

test/tbb/test_numa_dist.cpp
```diff
@@ -0,0 +1,109 @@
...
+            }
+        });
+    for (const auto it : tls) {
+		validateProcgrp[it.first]++;
+	}
+	REQUIRE(validateProcgrp[master_thread_proc_grp] == requested_parallelism);
```
vossmjp, line 98: Another unlikely corner case is that the master_thread_proc_grp has the requested_parallelism count but there are counts for other processor groups too.  All other groups should have zeros but that is not checked by the test.

test/tbb/test_numa_dist.cpp
```diff
@@ -0,0 +1,109 @@
...
+    numa() : processorGroupCount(GetMaximumProcessorGroupCount()), maxProcessors(GetActiveProcessorCount(ALL_PROCESSOR_GROUPS)){   
+        numaProcessors.resize(processorGroupCount);
+       for (WORD i = 0; i < processorGroupCount; i++) {
+            this->numaProcessors[i] = GetActiveProcessorCount((i));
+        }        
+    }
+};
+
+
+void TestNumaDistribution(){
```
AnuyaWelling2801, line 70: Maybe change this to 
int TestNumaDistribution(std::vector<DWORD> &validateProcgrp), so that the function can be reused

test/tbb/test_numa_dist.cpp
AnuyaWelling2801, line 98: Move to TEST_CASE so that the function can be reused.

test/tbb/test_numa_dist.cpp
```diff
@@ -0,0 +1,109 @@
...
+	std::vector<DWORD> validateProcgrp;
+	validateProcgrp.resize(GetMaximumProcessorGroupCount());
+	PROCESSOR_NUMBER proc;
+	struct numa nodes;
+	GetThreadIdealProcessorEx(GetCurrentThread(), &proc);
+	int master_thread_proc_grp = proc.Group;
+	auto requested_parallelism = nodes.numaProcessors.at(master_thread_proc_grp);
```
AnuyaWelling2801, line 77: auto requested_parallelism = nodes.numaProcessors.at(master_thread_proc_grp) + additionalParallelism;
If additionalParallelism = 0 - we have a test case for threads = cores on the node
if additionalParallelism > 0 - we can check for overflow

test/tbb/test_numa_dist.cpp
```diff
@@ -0,0 +1,109 @@
...
+    tbb::enumerable_thread_specific< std::pair<int, int> > tls;
+	tbb::enumerable_thread_specific< double > tls_dummy;
+	
+    tbb::parallel_for(0, 1000, [&](int i)
+        {
+            // dummy work
+            double& b = tls_dummy.local();
+            for (int k = 0; k < 1000000; ++k)
+                b += pow(i * k, 1. / 3.);
+
```
dnmokhov, line 88: It is probably not a good idea to assume that every thread will be engaged because of this. Something more reliable, e.g., a barrier might be better.

dnmokhov (changes requested):
Please resolve the CI issues:

1. Copyright.
2. If this will stay a Windows-only test, we should only add it there. Otherwise, a Linux equivalent is needed.

test/tbb/test_numa_dist.cpp
AnuyaWelling2801: Do we still need this if we are using a static partitioner and a barrier?

test/tbb/test_numa_dist.cpp
AnuyaWelling2801: Fixed.

test/tbb/test_numa_dist.cpp
```diff
@@ -0,0 +1,109 @@
...
+#include "common/spin_barrier.h"
+#include "common/exception_handling.h"
+#include "common/concepts_common.h"
+#include "test_partitioner.h"
+
+#include <cstddef>
+#include <vector>
+
+//! \file test_parallel_for.cpp
+//! \brief Test for [algorithms.parallel_for] specification
```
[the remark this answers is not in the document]
AnuyaWelling2801: Fixed.

test/tbb/test_numa_dist.cpp
AnuyaWelling2801: Fixed.

test/tbb/test_numa_dist.cpp
```diff
@@ -0,0 +1,146 @@
+/*
+    Copyright (c) 2005-2021 Intel Corporation
```
dnmokhov, line 2: update

test/tbb/test_numa_dist.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+TEST_CASE("Numa overflow") {
+    numa example;
+    std::vector<DWORD> validateProcgrp;
+
+    int numaGrp = TestNumaDistribution(validateProcgrp, 1, 0);
+    std::vector<DWORD> result(GetMaximumProcessorGroupCount(), 0);
+    result[numaGrp] = example.numaProcessors[numaGrp];
+    result[(numaGrp+1)% GetMaximumProcessorGroupCount()] = 1;
+    REQUIRE(validateProcgrp == result);
```
dnmokhov, line 135: This is failing CI. For 1 NUMA node, `additionalParallelism=1` results in `maxProcessors+1` total threads.

dnmokhov (changes requested):

test/tbb/test_numa_dist.cpp
sarathnandu: Fixed in commit 3.

test/tbb/test_numa_dist.cpp
AnuyaWelling2801: Fixed

phprus:
``SYNCHRONIZATION_BARRIER`` required "Windows 8 / Windows Server 2012" but all other WinAPI calls required "Windows 7 / Windows Server 2008 R2".

Maybe wrap the test in an ``#if _WIN32_WINNT >= 0x0602`` check or reuse existing ``SpinBarrier`` (https://github.com/oneapi-src/oneTBB/blob/801df0edeb6ed19283744f1e4a41653ed58e7513/test/common/spin_barrier.h#L88)?

phprus:
Thanks @AnuyaWelling2801!
All this branch tests run successfully on Windows 7.

dnmokhov (approved):

PR #868, opened 2022-08-05
Make alias input_type protected for join_node

dbiswas2808:
Currently the public aliases exposed by `flow_graph::join_node` includes `input_type`. This is a bit confusing since `join_node` has multiple input ports and I believe other nodes(with multiple inputs) don't have the same issue. I have been leveraging substitution failure with `input_ports_type` and `input_type` and similarly for `output_ports_type` and `output_type` for implementing a higher level interface for flow_graph. AFAIT only `join_node` breaks this distinction. Also, having `join_node` expose `input_type` as a public alias could be confusing to users.

### DIFF
--- a/include/oneapi/tbb/detail/_flow_graph_join_impl.h
+++ b/include/oneapi/tbb/detail/_flow_graph_join_impl.h
@@ -669,12 +669,12 @@
 
     template<typename InputTuple, typename OutputTuple>
     class join_node_FE<reserving, InputTuple, OutputTuple> : public reserving_forwarding_base {
-    public:
+    private:
         static const int N = std::tuple_size<OutputTuple>::value;
         typedef OutputTuple output_type;
         typedef InputTuple input_type;
         typedef join_node_base<reserving, InputTuple, OutputTuple> base_node_type; // for forwarding
-
+    public:
         join_node_FE(graph &g) : reserving_forwarding_base(g), my_node(nullptr) {
             ports_with_no_inputs = N;
             join_helper<N>::set_join_node_pointer(my_inputs, this);
@@ -1028,7 +1028,6 @@
                            public sender<OutputTuple> {
     protected:
         using graph_node::my_graph;
-        using input_type = typename join_node_FE<JP, InputTuple, OutputTuple>::input_type;
     public:
         typedef OutputTuple output_type;
 
--- a/include/oneapi/tbb/detail/_flow_graph_join_impl.h
+++ b/include/oneapi/tbb/detail/_flow_graph_join_impl.h
@@ -1028,6 +1028,7 @@
                            public sender<OutputTuple> {
     protected:
         using graph_node::my_graph;
+        using input_type = typename join_node_FE<JP, InputTuple, OutputTuple>::input_type;
     public:
         typedef OutputTuple output_type;
 
--- a/include/oneapi/tbb/detail/_flow_graph_join_impl.h
+++ b/include/oneapi/tbb/detail/_flow_graph_join_impl.h
@@ -1214,10 +1214,10 @@
     //  using tuple_element.  The class PT is the port type (reserving_port, queueing_port, key_matching_port)
     //  and should match the typename.
 
-    template<int N, template<class> class PT, typename OutputTuple, typename JP>
+    template<int M, template<class> class PT, typename OutputTuple, typename JP>
     class unfolded_join_node : public join_base<N,PT,OutputTuple,JP>::type {
     public:
-        typedef typename wrap_tuple_elements<N, PT, OutputTuple>::type input_ports_type;
+        typedef typename wrap_tuple_elements<M, PT, OutputTuple>::type input_ports_type;
         typedef OutputTuple output_type;
     private:
         typedef join_node_base<JP, input_ports_type, output_type > base_type;
--- a/include/oneapi/tbb/detail/_flow_graph_join_impl.h
+++ b/include/oneapi/tbb/detail/_flow_graph_join_impl.h
@@ -1215,7 +1215,7 @@
     //  and should match the typename.
 
     template<int M, template<class> class PT, typename OutputTuple, typename JP>
-    class unfolded_join_node : public join_base<N,PT,OutputTuple,JP>::type {
+    class unfolded_join_node : public join_base<M,PT,OutputTuple,JP>::type {
     public:
         typedef typename wrap_tuple_elements<M, PT, OutputTuple>::type input_ports_type;
         typedef OutputTuple output_type;


include/oneapi/tbb/detail/_flow_graph_join_impl.h
```diff
@@ -1026,6 +1026,7 @@
                            public sender<OutputTuple> {
     protected:
         using graph_node::my_graph;
+        using input_type = typename join_node_FE<JP, InputTuple, OutputTuple>::input_type;
```
kboyarinov, line 1029: I have maid a quick check and it seems like this alias it is not needed outside the `join_node_FE` class - as well as other public aliases defined there. It seems like it would be better to make `input_type` and other aliases in `join_node_FE` specializations private. @aleksei-fedotov, what do you think?

akukanov: For what it`s worth. a public class like `join_node` definitely should hide all internal interfaces inherited from the implementation.

kboyarinov:
@dbiswas2808, could you please remove the line related to another PR from this patch and it would be approved and merged.
Thank you

include/oneapi/tbb/detail/_flow_graph_join_impl.h
aleksei-fedotov: > I have maid a quick check and it seems like this alias it is not needed outside the `join_node_FE` class - as well as other public aliases defined there. It seems like it would be better to make `input_type` and other aliases in `join_node_FE` specializations private. @aleksei-fedotov, what do you think?

Sorry for a late reply. 
The idea behind input and output type is to make node interfaces common so that users can reference the corresponding type easily by writing something like: `<node_type>::input_type` or `<node_type>::output_type`.
There are exceptions for multi-input/output node types (such as `join_node`, `split_node`). 
Yes, internal interfaces should be hidden. So, the changes are correct (except for the line with static assert).

dbiswas2808:
> @dbiswas2808, could you please remove the line related to another PR from this patch and it would be approved and merged.
> Thank you

Done!

kboyarinov (dismissed):
LGTM

dbiswas2808:
@kboyarinov 
```
    template<typename JP, typename InputTuple, typename OutputTuple>
    class join_node_base : public graph_node, public join_node_FE<JP, InputTuple, OutputTuple>,
                           public sender<OutputTuple> {
    protected:
        using graph_node::my_graph;
    public:
        typedef OutputTuple output_type;
```
I think making the aliases protected might fix the failures. If you look at the above snippet the `OutputTuple` typedef  uses the `OutputType` from `join_node_FE`. What do you think?

kboyarinov:
@dbiswas2808, as I understand the testing failure, the problem is the following ( I have simplified a code a bit):

```
template <typename Tuple>
class join_node_FE {
    static const int N = std::tuple_size<Tuple>::value;
};

template <int N, typename Tuple>
class unfolded_join_node : public join_node_FE<Tuple> {
public:
    typedef typename wrap_tuple_elements<N, Tuple>::type input_ports_type; // problem is here
};
```

Visual Studio compiler for some reason interprets the usage of `N` on the last line as an access to private field of the base class, not the local non-type template parameter `N`.
Unfortunately, I failed to reproduce this on my local machine, I would try on other machines to figure out how this can be fixed

dbiswas2808:
> @dbiswas2808, as I understand the testing failure, the problem is the following ( I have simplified a code a bit):
> 
> ```
> template <typename Tuple>
> class join_node_FE {
>     static const int N = std::tuple_size<Tuple>::value;
> };
> 
> template <int N, typename Tuple>
> class unfolded_join_node : public join_node_FE<Tuple> {
> public:
>     typedef typename wrap_tuple_elements<N, Tuple>::type input_ports_type; // problem is here
> };
> ```
> 
> Visual Studio compiler for some reason interprets the usage of `N` on the last line as an access to private field of the base class, not the local non-type template parameter `N`. Unfortunately, I failed to reproduce this on my local machine, I would try on other machines to figure out how this can be fixed

@kboyarinov That is strange indeed I haven't seen that myself before. Any updates on how to fix this or get around this issue?

include/oneapi/tbb/detail/_flow_graph_join_impl.h
```diff
@@ -667,12 +667,12 @@
 
     template<typename InputTuple, typename OutputTuple>
     class join_node_FE<reserving, InputTuple, OutputTuple> : public reserving_forwarding_base {
-    public:
+    private:
         static const int N = std::tuple_size<OutputTuple>::value;
```
BenFrantzDale, line 671: Should there be a comment here saying why these are private?

kboyarinov: As discussed [above](https://github.com/oneapi-src/oneTBB/pull/868#discussion_r950049713), this parameter is part of the internal implementation details of `join_node_FE` so it should not be part of `oneapi::tbb::join_node` public API. 
Do you think we really need an extra comment in the code for this?

kboyarinov:
@dbiswas2808, as a simplest way to workaround a VS issue, I would propose to rename the template parameter of `unfolded_join_node` from `N` to something different. I guess it should fix the problem.

include/oneapi/tbb/detail/_flow_graph_join_impl.h
BenFrantzDale: 🤷  I'm just trying to consider a future reader looking here without much context thinking "Why are `N`, `output_type`, `input_type`, and `base_node_type` private?" I'm thinking that would be clarified if they saw `// These parameters are private because they are part of the internal implementation details of join_node_FE: After all, a join node doesn't have a single input port.`

Just a thought.

dbiswas2808:
> @dbiswas2808, as a simplest way to workaround a VS issue, I would propose to rename the template parameter of `unfolded_join_node` from `N` to something different. I guess it should fix the problem.

@kboyarinov I just Made the suggested change. Hopefully it compiles fine on MSVC.

dbiswas2808:
@kboyarinov I've fixed the msvc compile issue as you suggested in the last commit. LMK if it looks right.

kboyarinov (approved):

dbiswas2808:
@kboyarinov I fixed the issue I think. Seems to build on my MacOS system

kboyarinov (dismissed):
Thanks for the contribution! LGTM!

kboyarinov:
@dbiswas2808, one last request from my side. Could you please add a sign-off string to the commit message? See [Contribution.md](https://github.com/oneapi-src/oneTBB/blob/master/CONTRIBUTING.md#pull-requests:~:text=No%20anonymous%20contributions%20are%20accepted.%20The%20name%20in%20the%20commit%20message%20Signed%2Doff%2Dby%20line%20and%20your%20email%20must%20match%20the%20change%20authorship%20information.%20Make%20sure%20your%20.gitconfig%20is%20set%20up%20correctly%20so%20you%20can%20use%20git%20commit%20%2Ds%20for%20signing%20your%20patches%3A) for details. After that, the PR can be merged to master.
Thank you once again for your contribution.

dbiswas2808:
@kboyarinov I just signed off all the commits. Will need your approval for the pr-builder to start I think

kboyarinov (approved):

PR #880, opened 2022-08-22
Use of std::invoke to run the body in algorithms and graph nodes

kboyarinov:
### Description 
Change the execution of the body in TBB algorithms and Flow Graph nodes to use ``std::invoke`` instead of ``operator()``.
It allows to pass pointers to non-static member functions and non-static member objects as a body.
Affected components:

- ``parallel_for`` 
- ``parallel_reduce``
- ``parallel_deterministic_reduce``
- ``parallel_for_each``
- ``parallel_scan``
- ``parallel_pipeline``
- ``function_node``
- ``multifunction_node``
- ``async_node``
- ``sequencer_node``
- ``join_node`` with ``key_matching`` policy

- [ ] - git commit message contains an appropriate signed-off-by string _(see [CONTRIBUTING.md](https://github.com/oneapi-src/oneTBB/blob/master/CONTRIBUTING.md#pull-requests) for details)_

### Type of change

### Tests

- [ ] not needed

### Documentation

- [x] needs to be updated
- [ ] not needed

- [ ] Yes
- [x] No
- [ ] Unknown


include/oneapi/tbb/parallel_for.h
```diff
@@ -310,7 +305,7 @@ void parallel_for_impl(Index first, Index last, Index step, const Function& f, P
         throw_exception(exception_id::nonpositive_step); // throws std::invalid_argument
     else if (first < last) {
         // Above "else" avoids "potential divide by zero" warning on some platforms
-        Index end = (last - first - Index(1)) / step + Index(1);
+        Index end = Index(last - first - 1ul) / step + Index(1);
```
vossmjp, line 308: This seems unrelated to std::invoke.

vossmjp (commented):
Partial review. So far have looked at _utils.h, parallel_for.h, parallel_for_each.h, parallel_scan.h, _pipeline_filters.h and parallel_reduce.h. Mostly looks good to me. A couple of small comments/questions. Still need to look at tests, conformance tests & flow graph.

include/oneapi/tbb/parallel_for.h
kboyarinov: Yes, it is not related for `std::invoke`, but I failed to implement custom `ParallelForIndex` as it defined in the spec to make this line work. I faced a lot of problems with the operation priority, conversions and wrong return values for mathematical operations. This workaround can be considered as partial fix for the implementation to better match the spec.

include/oneapi/tbb/detail/_pipeline_filters.h
```diff
@@ -281,7 +281,7 @@ class concrete_filter<InputType, void, Body>: public base_filter {
 
     void* operator()(void* input) override {
         input_pointer temp_input = input_helper::cast_from_void_ptr(input);
-        my_body(std::move(input_helper::token(temp_input)));
+        tbb::detail::invoke(my_body, std::move(input_helper::token(temp_input)));
```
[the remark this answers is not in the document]
kboyarinov: Yes, it is - to have any profit from using `std::invoke`, the user will need to provide pointer-to-member function or member object from the input object. And currently we don't have any useful functions in `flow_control` that can be used as as a filter body.

include/oneapi/tbb/detail/_pipeline_filters_deduction.h
```diff
@@ -26,18 +26,18 @@ namespace detail {
 namespace d1 {
 
 template <typename Input, typename Output>
-struct declare_fitler_types {
+struct declare_filter_types {
```
omalyshe, line 29: Copyright year should be fixed

include/oneapi/tbb/parallel_scan.h
```diff
@@ -1,6 +1,5 @@
 /*
-    Copyright (c) 2005-2021 Intel Corporation
-
+    Copyright (c) 2005-2023 Intel Corporation
```
omalyshe, line 2: Insert the empty line after Copyright

test/tbb/test_task.cpp
```diff
@@ -785,7 +785,7 @@ TEST_CASE("Test with priority inversion") {
     // take first core on execute
     utils::SpinBarrier barrier(thread_number + 1);
     test_arena.execute([&] {
-        tbb::parallel_for(std::uint32_t(0), thread_number + 1, [&] (std::uint32_t&) {
+        tbb::parallel_for(std::uint32_t(0), thread_number + 1, [&] (std::uint32_t) {
```
omalyshe, line 788: Copyright year should be fixed

include/oneapi/tbb/detail/_pipeline_filters_deduction.h
kboyarinov: Done

test/tbb/test_task.cpp
kboyarinov: Done

include/oneapi/tbb/parallel_scan.h
kboyarinov: Done

include/oneapi/tbb/detail/_flow_graph_body_impl.h
```diff
@@ -218,7 +218,7 @@ template <typename Input, typename Output, typename B>
 class type_to_key_function_body_leaf : public type_to_key_function_body<Input, Output> {
 public:
     type_to_key_function_body_leaf( const B &_body ) : body(_body) { }
-    Output operator()(const Input &i) override { return body(i); }
+    Output operator()(const Input &i) override { return tbb::detail::invoke(body, i); }
```
vossmjp, line 221: I suppose you intend to exclude use of the invoke from the Output& specialization's operator(), since it would used to return a reference to something in the input message, which would be unsafe. Is that right?

include/oneapi/tbb/detail/_flow_graph_node_impl.h
```diff
@@ -496,7 +496,7 @@ class multifunction_input : public function_input_base<Input, Policy, A, multifu
     // constructor
     template<typename Body>
     multifunction_input(graph &g, size_t max_concurrency,Body& body, node_priority_t a_priority )
-      : base_type(g, max_concurrency, a_priority, noexcept(body(input_type(), my_output_ports)))
+      : base_type(g, max_concurrency, a_priority, noexcept(tbb::detail::invoke(body, input_type(), my_output_ports)))
```
vossmjp, line 499: I suppose you intend to exclude the use of invoke in the apply_body_impl_bypass. Is there a chance that the input message could define a function that would accept the output ports, and so be a valid use case?

vossmjp (commented):
The flow graph modifications look ok to me. Made a couple of comments to make sure I understand why certain invocation points are excluded.

include/oneapi/tbb/detail/_flow_graph_body_impl.h
kboyarinov: Actually, I missed that use-case. I checked for its safety and I don't see any unsafe actions, because our interface forces the body to return constant reference for some object inside the input message and never modifies it. I have updated the implementation and added the test for it.

include/oneapi/tbb/detail/_flow_graph_node_impl.h
kboyarinov: I have excluded `apply_body_impl_bypass` because we call operator() of `multifunction_body_leaf` there and it is always a callable object defined inside oneTBB. It supports `std::invoke` inside, so use-case you described would work and we have a test for it.

vossmjp (approved):
Did a quick look through tests and they look ok to me.

pavelkumbrasev (approved):

PR #900, opened 2022-09-09
Fix build on systems with ucontext.h implemented outside the libc

ismaell:
### Description 
Check for the `libucontext` package. The `ucontext.h` API may be implemented outside the libc.

Without this patch the build seems to succeed but results in under-linking.

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/cmake/ucontext.cmake
+++ b/cmake/ucontext.cmake
@@ -1,0 +1,20 @@
+# The ucontext API may be implemented outside the default library set
+include(CheckSymbolExists)
+if (NOT DEFINED UCONTEXT_LIBRARIES
+    AND CMAKE_SYSTEM_NAME MATCHES "^Linux")
+    check_symbol_exists("getcontext" "ucontext.h" HAVE_UCONTEXT)
+    if (NOT HAVE_UCONTEXT)
+        message(STATUS "Checking for libucontext")
+        find_library(LIBUCONTEXT "ucontext")
+        if (LIBUCONTEXT STREQUAL "LIBUCONTEXT-NOTFOUND")
+            message(WARNING "No ucontext API available")
+            set(LIBUCONTEXT "")
+        else()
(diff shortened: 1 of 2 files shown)


ismaell:
Ah, that's even better.

pavelkumbrasev:
Could I ask to do such change in your patch? You simply need to define this macro - `__TBB_RESUMABLE_TASKS_USE_THREADS` in cmake if ucontext not present.
Also, please add the copyright to cmake file you can copy it from our cmake files.

pavelkumbrasev:
@ismaell do you have a chance to update PR with proposed changes or we should create separate PR for this?

brad0:
> Could I ask to do such change in your patch? You simply need to define this macro - `__TBB_RESUMABLE_TASKS_USE_THREADS` in cmake if ucontext not present. Also, please add the copyright to cmake file you can copy it from our cmake files.

Please do. I ran into this trying to build TBB on OpenBSD, which does not have ucontext. Also I see Haiku does not as well.

https://github.com/oneapi-src/oneTBB/issues/795

brad0:
> Could I ask to do such change in your patch? You simply need to define this macro - `__TBB_RESUMABLE_TASKS_USE_THREADS` in cmake if ucontext not present. Also, please add the copyright to cmake file you can copy it from our cmake files.

Can you please try to come up with a diff  that checks for ucontext and enables the thread-based implementation if it does not exist? I can help test if you would like.

pavelkumbrasev:
@brad0 Basically, you could submit a separate PR that implements proposed approach since there is no progress in this PR.
Are you ok with this?

pavelkumbrasev:
Closing it since #1055 was merged.

PR #904, opened 2022-09-13
Correct behavior for proportional_mode

pavelkumbrasev:
### Description 
This patch correct the behavior of static_partitioner when it used with custom range that does not support proportional splitting. 
For example such combination (static_partitioner + custom range) on machine with 6 threads leads to tasks with: [0,32], [32,64], [64,128], [128,256], [256,512], [512,1024] range distribution. Hence, leads to huge disbalance and performance degradation. 

It happens because during partition at [function](https://github.com/oneapi-src/oneTBB/blob/8155aaeb6baead9bd571bbb76a9b5dc56a1a9403/include/oneapi/tbb/partitioner.h#L283) we lost partitioner dividing proportions.

This patch propose always use proportional_split with calculated proportions with respect to available partitioner resources. The correct constructor for range will be called via [line](https://github.com/oneapi-src/oneTBB/blob/8155aaeb6baead9bd571bbb76a9b5dc56a1a9403/include/oneapi/tbb/parallel_for.h#L87).

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

@kboyarinov @aleksei-fedotov 

### Other information

### DIFF
--- a/include/oneapi/tbb/partitioner.h
+++ b/include/oneapi/tbb/partitioner.h
@@ -314,25 +314,6 @@ struct adaptive_mode : partition_type_base<Partition> {
 //! Helper type for checking availability of proportional_split constructor
 template <typename T> using supports_proportional_splitting = typename std::is_constructible<T, T&, proportional_split&>;
 
-//! A helper class to create a proportional_split object for a given type of Range.
-/** If the Range has proportional_split constructor,
-    then created object splits a provided value in an implemenation-defined proportion;
-    otherwise it represents equal-size split. */
-// TODO: check if this helper can be a nested class of proportional_mode.
-template <typename Range, typename = void>
-struct proportion_helper {
-    static proportional_split get_split(std::size_t) { return proportional_split(1,1); }
-};
-
-template <typename Range>
-struct proportion_helper<Range, typename std::enable_if<supports_proportional_splitting<Range>::value>::type> {
-    static proportional_split get_split(std::size_t n) {
-        std::size_t right = n / 2;
-        std::size_t left  = n - right;
-        return proportional_split(left, right);
-    }
-};
-
 //! Provides proportional splitting strategy for partition objects
 template <typename Partition>
 struct proportional_mode : adaptive_mode<Partition> {
@@ -358,7 +339,10 @@ struct proportional_mode : adaptive_mode<Partition> {
     template <typename Range>
     proportional_split get_split() {
         // Create a proportion for the number of threads expected to handle "this" subrange
-        return proportion_helper<Range>::get_split( self().my_divisor / my_partition::factor );
+        std::size_t n = self().my_divisor / my_partition::factor;
+        std::size_t right = n / 2;
+        std::size_t left  = n - right;
+        return proportional_split(left, right);
     }
 };
 
--- a/test/tbb/test_partitioner.cpp
+++ b/test/tbb/test_partitioner.cpp
@@ -19,6 +19,7 @@
 #include "tbb/parallel_for.h"
 #include "tbb/task_arena.h"
 #include "tbb/global_control.h"
+#include "oneapi/tbb/mutex.h"
 
 #include "common/utils.h"
 #include "common/utils_concurrency_limit.h"
@@ -28,6 +29,7 @@
 #include <cstddef>
 #include <utility>
 #include <vector>
+#include <algorithm>
 
 //! \file test_partitioner.cpp
 //! \brief Test for [internal] functionality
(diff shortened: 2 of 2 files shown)


include/oneapi/tbb/partitioner.h
```diff
@@ -314,25 +314,6 @@ struct adaptive_mode : partition_type_base<Partition> {
 //! Helper type for checking availability of proportional_split constructor
 template <typename T> using supports_proportional_splitting = typename std::is_constructible<T, T&, proportional_split&>;
```
pavelkumbrasev, line 315:
```suggestion
```

include/oneapi/tbb/partitioner.h
```diff
@@ -312,26 +312,6 @@ struct adaptive_mode : partition_type_base<Partition> {
 };
 
 //! Helper type for checking availability of proportional_split constructor
```
pavelkumbrasev, line 314:
```suggestion
```

test/tbb/test_partitioner.cpp
```diff
@@ -138,3 +140,53 @@ TEST_CASE("Threads respect task affinity") {
...
+
+        std::size_t min_size = *std::min_element(res.begin(), res.end());
+        for (auto elem : res) {
+            REQUIRE(min_size * 2 + 1 >= elem);
+        }
+    }
+}
+
+//! \brief \ref regression
+TEST_CASE("Test ranges without proportional split") {
```
kboyarinov, line 169:
```suggestion
TEST_CASE("Test partitioned tasks count and size for static_partitioner") {
```

kboyarinov (dismissed):
LGTM

aleksei-fedotov:
I believe `get_range_split_object` entity becomes unnecessary as well as `explicit`-ness of type conversion operator for `proportional_split` class. Could you please update the patch accordingly?

kboyarinov:
@aleksei-fedotov, it is not true.
This helper is still required because the implicit conversion from `proportional_split` to `split` is prohibited because of `explicit` operator. The helper is to make such a conversion explicit.
See the short reproducer: https://godbolt.org/z/eqK57EYK1
If the macro USE_RANGE_SPLIT_OBJ is defined to 0, the code fails to compile

aleksei-fedotov:
> @aleksei-fedotov, it is not true. This helper is still required because the implicit conversion from `proportional_split` to `split` is prohibited because of `explicit` operator. The helper is to make such a conversion explicit. See the short reproducer: https://godbolt.org/z/eqK57EYK1 If the macro USE_RANGE_SPLIT_OBJ is defined to 0, the code fails to compile

Yeah, sure! And the explicitness of the type conversion operator is necessary since we don't want to accidentally lose the original proportion.

test/tbb/test_partitioner.cpp
```diff
@@ -138,3 +140,53 @@ TEST_CASE("Threads respect task affinity") {
...
+            results[i].push_back(r.size());
+        }, oneapi::tbb::static_partitioner{});
+    }
+
+    for (auto& res : results) {
+        REQUIRE(res.size() == utils::get_platform_max_threads());
+
+        std::size_t min_size = *std::min_element(res.begin(), res.end());
+        for (auto elem : res) {
+            REQUIRE(min_size * 2 + 1 >= elem);
```
aleksei-fedotov, line 163: I think for the static partitioner it would be good to add test that checks tasks distribution with a known tolerance. And, for the ranges that support proportional split, to test that task sizes do not differ on more than 1 or maybe 2 iterations.

test/tbb/test_partitioner.cpp
```diff
@@ -19,6 +19,7 @@
 #include "tbb/parallel_for.h"
 #include "tbb/task_arena.h"
 #include "tbb/global_control.h"
+#include "oneapi/tbb/mutex.h"
```
akukanov, line 22: Is `oneapi` needed when referring to TBB public headers in the tests? I mean, what is the convention? Looks like for other headers above it's not specified.

test/tbb/test_partitioner.cpp
```diff
@@ -28,6 +29,7 @@
 #include <cstddef>
 #include <utility>
 #include <vector>
+#include <algorithm>
```
akukanov, line 32: When including "big" headers with lots of functions of which you only need a couple, it makes sense to say in a comment which function(s) are used.

include/oneapi/tbb/partitioner.h
```diff
@@ -358,7 +337,10 @@ struct proportional_mode : adaptive_mode<Partition> {
     template <typename Range>
     proportional_split get_split() {
         // Create a proportion for the number of threads expected to handle "this" subrange
```
akukanov, line 339: I think this comment needs to be updated, so that it explains the logic better, and - ideally - how it works for ranges with and without proportional split.

test/tbb/test_partitioner.cpp
pavelkumbrasev: There is still no convention. After we changed the layout of directories we create copy of all original header and placed them into include/tbb folder (made them include implementation only) and implementation into include/oneapi/tbb.
As the mutex.h after revamp did not exist we decided to move it to include/oneapi/tbb only.

Actually, we might create some convention to use <oneapi/tbb/*.h> in tests to be consistent. (also create a test that would use <tbb/*.h> headers)

test/tbb/test_partitioner.cpp
pavelkumbrasev: What do you mean by "that checks tasks distribution with a known tolerance" ?

aleksei-fedotov: Well, I meant to have the test that checks task distribution and execution more accurately for a given partitioner. For example, for the `static_partitioner` we know that all subranges are of the same size, except for one subrange that should not differ by one iteration. We also know that ideally these subranges should be executed by different threads and we also know in what circumstances that ideal execution might not happen so we can detect such situations in the test and still make it pass, and then, perhaps, check that non-ideal execution does not happen too often. This is what I wanted to have a test for - to detect incorrect "range <=> partitioner <=> execution engine"-relationship as early as possible in the SDL.

I think we need to take back (some of the) tests from [the classic TBB partitioner testing](https://github.com/oneapi-src/oneTBB/blob/v2020.3/src/test/test_partitioner_whitebox.cpp).

pavelkumbrasev: Are you agree that for the proposed change the proposed test is enough? And we could create separate issue to add such testing for partitioners.

test/tbb/test_partitioner.cpp
pavelkumbrasev: done

test/tbb/test_partitioner.cpp
aleksei-fedotov: As far as I remember, there is (or at least was) a convention for unit tests to use include/tbb and for conformance tests to use include/oneapi/tbb. In this case we check oneAPI layout and make sure the old one also works fine.
For new headers there is only one way of inclusion though.

test/tbb/test_partitioner.cpp
aleksei-fedotov: Yes, I agree.

aleksei-fedotov (dismissed):
Looks good to me. Please consider proposed changes though.

aleksei-fedotov (approved):

Issue #358, opened 2021-03-05
Clang thread sanitizer (TSan) reports many errors, most tests fail


seanm:
I built TBB from git master using Clang's `-fsanitize=thread` option:

https://clang.llvm.org/docs/ThreadSanitizer.html

The result is that almost all the unit tests fail.

Are these false positives or true positives?

If false positives, is there an 'official' suppression list?

alexey-katranov:
There are a lot of true positives (especially, in `tbbmalloc`). We are fixing them from time to time. Some basics tests, (e.g. for `parallel_for`, `parallel_reduce` and `task_group`) are not expected to issue hits without `tbbmalloc`.

seanm:
>There are a lot of true positives

Oh.

>We are fixing them from time to time.

I would have thought threading correctness issues in a threading library would be high priority. :) In my experience, TSan makes these otherwise hard to debug and hard to reproduce issues much easier to fix, has that not been the case with TBB?

>Some basics tests, (e.g. for parallel_for...

I just tried `ctest -R parallel_for -V` and TSan reports data races there.  The backtraces do contains a lot of `libtbbmalloc_debug.dylib`.

I'm not sure what `tbbmalloc` is, is it some build option that can be disabled?

alexey-katranov:
> I would have thought threading correctness issues in a threading library would be high priority. :) In my experience, TSan makes these otherwise hard to debug and hard to reproduce issues much easier to fix, has that not been the case with TBB?

Definitely threading correctness issues is high priority; however, C++ races reported with Thread Sanitizer are not always correctness issues on real hardware. Back to 2006 when the first version of TBB was released, C++ did not know anything about threading, i.e. any threaded application had UB from pre-C++11 perspective. In such environment, TBB (and other threading solutions) was build based on experience, practice and knowledge about target architecture. The first supported platform was x86 with its [TSO memory model](https://en.wikipedia.org/wiki/Processor_consistency#Similarity_to_SPARC_V8_TSO,_IBM-370,_and_x86-TSO_Memory_Models). It is relatively strong model, i.e. any properly aligned `int` provides release-acquire semantics for any store-load pair. This basic assumption became fundamental thing inside TBB, especially in tbbmalloc. With new platforms (like IA-64, ARM and others) in pre-C++11 environment, the implementation became even more complicated but remained correct for each particular platform.

[...]

PR #911, opened 2022-09-18
Enable ThreadSanitizer on macOS

pavelkumbrasev:

### Description 
To enable ThreadSanitizer on __APPLE__ (basically on macOS) replaced  semaphore_t on dispatch_semaphore_t which supports via ThreadSanitizer.
In accordance to scenarios where and how we use semaphore the performance should not change, but still need to prove this.

Fixes # - #358

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

@phprus


akukanov:
Does this change introduce any new dependencies for either compilation/link time or execution time which need to be documented?

Do we know that the dispatch semaphore has no spurious wakeups?

pavelkumbrasev:
> Does this change introduce any new dependencies for either compilation/link time or execution time which need to be documented?
> 
> Do we know that the dispatch semaphore has no spurious wakeups?

I currently researching both questions. For now I checked that dispatch_semaphore_t introduced with macOS 10.6. Also I did not observe any new flags during the compilation. (I compiled example from #358 using dispatch_semaphore_t without any additional flags)

pavelkumbrasev:
It seems that from dispatch_semaphore_t [documentation](https://developer.apple.com/documentation/dispatch/1452919-dispatch_semaphore_signal) there should not be spurious wakeups during `wait` calls.  Also from the [implementation](https://opensource.apple.com/source/libdispatch/libdispatch-84.5/src/semaphore.c) point of view they have similar while loop on case DISPATCH_TIME_FOREVER.

test/tbb/test_partitioner.cpp
```diff
@@ -1,5 +1,5 @@
 /*
-    Copyright (c) 2021 Intel Corporation
+    Copyright (c) 2022 Intel Corporation
```
aleksei-fedotov, line 2: Why replacing? Isn't it should be extended instead?
```suggestion
    Copyright (c) 2021-2022 Intel Corporation
```

omalyshe: Agree with @aleksei-fedotov . While changing the existing file the year range should be used.

aleksei-fedotov (approved):
Looks good to me!

natale-p:
Tested on mac os 12.6, and I can easily see that I don't have many false positives as I had without this patch.

natale-p:
Hello, any news on this PR? Can it be merged?

pavelkumbrasev:
Hi @natale-p, we are going to merge it after we will satisfy all test requirements for it. (I hope it will be soon)

PR #936, opened 2022-10-09
Workaround for a bug in GCC. If LTO is enabled, GCC may not link application with pthread library.

### LINKED ISSUE #935: Test test_dynamic_link failed on gcc-10 with LTO
Commit: 6bfd29bde7e9c47bdfc3ae8a1670ea961c95c78c (current master)

g++:
```
Using built-in specs.
COLLECT_GCC=x86_64-alt-linux-g++
COLLECT_LTO_WRAPPER=/usr/libexec/gcc/x86_64-alt-linux/10/lto-wrapper
Target: x86_64-alt-linux
Configured with: ../configure --host=x86_64-alt-linux --build=x86_64-alt-linux --target=x86_64-alt-linux --prefix=/usr --exec-prefix=/usr --bindir=/usr/bin --sbindir=/usr/sbin --sysconfdir=/etc --datadir=/usr/share --includedir=/usr/include --libdir=/usr/lib64 --libexecdir=/usr/libexec --localstatedir=/var/lib --sharedstatedir=/var/lib --mandir=/usr/share/man --infodir=/usr/share/info --disable-dependency-tracking --disable-silent-rules --without-included-gettext --enable-shared --program-suffix=-10 --with-slibdir=/lib64 --with-bugurl=http://bugzilla.altlinux.org --enable-__cxa_atexit --enable-threads=posix --enable-checking=release --with-system-zlib --without-included-gettext --enable-multilib --enable-default-pie --enable-gnu-unique-object --enable-linker-build-id --with-linker-hash-style=gnu --with-arch_32=i586 --with-tune_32=generic --with-multilib-list=m64,m32,mx32 --with-gcc-major-version-only --enable-vtable-verify --enable-bootstrap --enable-languages=c,c++,fortran,objc,obj-c++,ada,go,d,lto --enable-plugin
Thread model: posix
Supported LTO compression algorithms: zlib
gcc version 10.3.1 20210703 (ALT Sisyphus 10.3.1-alt2) (GCC)
```

CMake options:
```
cmake -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_CXX_STANDARD=20 -DCMAKE_INTERPROCEDURAL_OPTIMIZATION=ON ../..
```

Error:
```
# ./gnu_10.3_cxx20_64_relwithdebinfo/test_dynamic_link
terminate called after throwing an instance of 'std::system_error'
  what():  Unknown error -1
Aborted (core dumped)
```

``test_dynamic_link`` stacktrace:
```
#0  0x00007fbe698472f1 in raise () from /lib64/libc.so.6
#1  0x00007fbe69830538 in abort () from /lib64/libc.so.6
#2  0x00007fbe69bd2861 in ?? () from /usr/lib64/libstdc++.so.6
#3  0x00007fbe69c00b68 in ?? () from /usr/lib64/libstdc++.so.6
#4  0x00007fbe69c00bd3 in std::terminate() () from /usr/lib64/libstdc++.so.6
#5  0x00007fbe69c00e64 in __cxa_throw () from /usr/lib64/libstdc++.so.6
#6  0x00007fbe69bd55b1 in std::__throw_system_error(int) () from /usr/lib64/libstdc++.so.6
#7  0x0000560a4f665adc in std::call_once<void (&)()> (__once=..., __f=<optimized out>) at /usr/include/c++/10/mutex:737
#8  tbb::detail::r1::init_dynamic_link_data () at /root/onetbb/oneTBB-6bfd29bde7e9c47bdfc3ae8a1670ea961c95c78c/test/../src/tbb/dynamic_link.cpp:331
#9  tbb::detail::r1::static_init_dl_data_t::static_init_dl_data_t (this=<optimized out>)
    at /root/onetbb/oneTBB-6bfd29bde7e9c47bdfc3ae8a1670ea961c95c78c/test/../src/tbb/dynamic_link.cpp:341
#10 __static_initialization_and_destruction_0 (__priority=<optimized out>, __initialize_p=<optimized out>)
    at /root/onetbb/oneTBB-6bfd29bde7e9c47bdfc3ae8a1670ea961c95c78c/test/../src/tbb/dynamic_link.cpp:343
#11 0x0000560a4f67f745 in __libc_csu_init (argc=1, argv=0x7ffc916f7c78, envp=0x7ffc916f7c88) at elf-init.c:89
#12 0x00007fbe69831dbd in __libc_start_main () from /lib64/libc.so.6
#13 0x0000560a4f665b2a in _start () at ../sysdeps/x86_64/start.S:120
```

phprus:
### Description 
See issue #935 for detailed description.

Fixes #935

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -361,17 +361,29 @@ unset(_sde_find_name)
 # Common target for the tbbbind related tests
 add_custom_target(test_suite_arena_constraints)
 
+# Check support for --no-as-needed linker option
+if (MINGW OR NOT WIN32)
+    include(CheckCXXSourceCompiles)
+    set(CMAKE_REQUIRED_LIBRARIES "-Wl,--no-as-needed")
+    check_cxx_source_compiles("int main(int, char*[]) { return 0; }" LINKER_HAS_NO_AS_NEEDED)
+    unset(CMAKE_REQUIRED_LIBRARIES)
+endif()
+
 if (TARGET TBB::tbb)
     # Define the tests
     tbb_add_test(SUBDIR tbb NAME test_tick_count DEPENDENCIES TBB::tbb)
     tbb_add_test(SUBDIR tbb NAME test_allocators DEPENDENCIES TBB::tbb)
     tbb_add_test(SUBDIR tbb NAME test_arena_priorities DEPENDENCIES TBB::tbb)
     tbb_add_test(SUBDIR tbb NAME test_dynamic_link DEPENDENCIES TBB::tbb)
-    if (WIN32)
-        tbb_add_test(SUBDIR tbb NAME test_numa_dist DEPENDENCIES TBB::tbb)
+    if (LINKER_HAS_NO_AS_NEEDED)
+        # The linker may not detect a dependency on pthread in static variable constructors.
+        target_link_libraries(test_dynamic_link PRIVATE "-Wl,--no-as-needed")
     endif()
     if (APPLE OR ANDROID_PLATFORM)
-        target_link_libraries(test_dynamic_link PRIVATE -rdynamic) # for the test_dynamic_link
+        target_link_libraries(test_dynamic_link PRIVATE -rdynamic)
+    endif()
+    if (WIN32)
+        tbb_add_test(SUBDIR tbb NAME test_numa_dist DEPENDENCIES TBB::tbb)
     endif()
     tbb_add_test(SUBDIR tbb NAME test_collaborative_call_once DEPENDENCIES TBB::tbb)
     tbb_add_test(SUBDIR tbb NAME test_concurrent_lru_cache DEPENDENCIES TBB::tbb)


phprus:
@pavelkumbrasev 
My guess is that LTO affects the order in which static variables are initialized.
Using a local static variable removes the dependency between global static variables.
This code has been thread safe since C++11.

If LTO is completely turn off, everything works without a patch:
```
cmake -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_CXX_STANDARD=20 -DCMAKE_INTERPROCEDURAL_OPTIMIZATION=OFF ../..
```

src/tbb/dynamic_link.cpp
```diff
@@ -328,7 +326,11 @@ namespace r1 {
 #endif  // __TBB_DYNAMIC_LOAD_ENABLED
     void init_dynamic_link_data() {
     #if __TBB_DYNAMIC_LOAD_ENABLED
-        std::call_once( init_dl_data_state, init_dl_data );
+        static bool init = []() -> bool {
+            init_dl_data();
+            return true;
+        }();
+        (void)init;
```
pavelkumbrasev, line 333:
```suggestion
        static bool init = [] {
            init_dl_data();
            return true;
        }();
       suppress_unused_warning(init);
```
Could you also add comment why this approach is used?

phprus:
@aleksei-fedotov 
All tests are linked with ``TBB`` and ``Threads::Threads`` (line 77):
https://github.com/oneapi-src/oneTBB/blob/6bfd29bde7e9c47bdfc3ae8a1670ea961c95c78c/test/CMakeLists.txt#L17-L82

build log:
```
[ 28%] Building CXX object test/CMakeFiles/test_dynamic_link.dir/tbb/test_dynamic_link.cpp.o
cd /root/oneTBB-6bfd29bde7e9c47bdfc3ae8a1670ea961c95c78c/build/cxx20rd/test && /usr/bin/c++  -I/root/oneTBB-6bfd29bde7e9c47bdfc3ae8a1670ea961c95c78c/test/.. -I/root/oneTBB-6bfd29bde7e9c47bdfc3ae8a1670ea961c95c78c/test -I/root/oneTBB-6bfd29bde7e9c47bdfc3ae8a1670ea961c95c78c/src/tbb/../../include -O2 -g -DNDEBUG -flto -fno-fat-lto-objects -Wall -Wextra -Werror -Wfatal-errors -Wshadow -Wcast-qual -Woverloaded-virtual -Wnon-virtual-dtor -mrtm -pthread -std=c++2a -MD -MT test/CMakeFiles/test_dynamic_link.dir/tbb/test_dynamic_link.cpp.o -MF CMakeFiles/test_dynamic_link.dir/tbb/test_dynamic_link.cpp.o.d -o CMakeFiles/test_dynamic_link.dir/tbb/test_dynamic_link.cpp.o -c /root/oneTBB-6bfd29bde7e9c47bdfc3ae8a1670ea961c95c78c/test/tbb/test_dynamic_link.cpp

[ 30%] Linking CXX executable ../gnu_10.3_cxx20_64_relwithdebinfo/test_dynamic_link
cd /root/oneTBB-6bfd29bde7e9c47bdfc3ae8a1670ea961c95c78c/build/cxx20rd/test && /usr/bin/cmake -E cmake_link_script CMakeFiles/test_dynamic_link.dir/link.txt --verbose=1
/usr/bin/c++ -O2 -g -DNDEBUG -flto -fno-fat-lto-objects -rdynamic CMakeFiles/test_dynamic_link.dir/tbb/test_dynamic_link.cpp.o -o ../gnu_10.3_cxx20_64_relwithdebinfo/test_dynamic_link  -Wl,-rpath,/root/oneTBB-6bfd29bde7e9c47bdfc3ae8a1670ea961c95c78c/build/cxx20rd/gnu_10.3_cxx20_64_relwithdebinfo ../gnu_10.3_cxx20_64_relwithdebinfo/libtbb.so.12.8 -ldl -pthread
```

The ``-pthread`` flag is present on the compiler and linker.

pavelkumbrasev:
This is a kind strange. I have use gdb on this test case:
```bash
(gdb) l
697     __gthread_once (__gthread_once_t *__once, void (*__func) (void))
698     {
699       if (__gthread_active_p ())
700         return __gthrw_(pthread_once) (__once, __func);
701       else
702         return -1;
703     }
```
```__gthread_active_p``` returns `false` in `test_dynamic_link`, but in same configuration on `test_task` it return `true`.
I checked compilation and link flags and they are the same.

**Comment by phprus** [comment]
@pavelkumbrasev , @aleksei-fedotov
PR updated.

File: test/CMakeLists.txt [hunk 2]
```diff
@@ -367,6 +367,11 @@ if (TARGET TBB::tbb)
     tbb_add_test(SUBDIR tbb NAME test_allocators DEPENDENCIES TBB::tbb)
     tbb_add_test(SUBDIR tbb NAME test_arena_priorities DEPENDENCIES TBB::tbb)
     tbb_add_test(SUBDIR tbb NAME test_dynamic_link DEPENDENCIES TBB::tbb)
+    if (CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
```
pavelkumbrasev, line 370: Should we check on LTO also?

phprus: It doesn't break the non-LTO build.

For correct check it is necessary to check TBB flags: ``TBB_IPO_PROPERTY, TBB_IPO_FLAGS`` and Cmake flags: ``CMAKE_INTERPROCEDURAL_OPTIMIZATION, DCMAKE_INTERPROCEDURAL_OPTIMIZATION_<config>``.
It's necessary?

phprus:
Added check support for ``--no-as-needed`` linker option.

phprus:
@pavelkumbrasev , @aleksei-fedotov

This may be not a bug, but the documented behavior of ld:

https://wiki.gentoo.org/wiki/Project:Quality_Assurance/As-needed#Initializers_and_deconstructors

> There exists a class of applications at the moment that break when using --as-needed . These applications are not at fault, but rather the linker itself. The linker is unable to detect dependencies between the initializers and deconstructors (.init/.fini ELF sections) when working with C++ code. As such, it may discard libraries when none of the symbols are used from it, thus mistakenly changing the initialization and deconstruction code paths.

Maybe it's better to use ``-Wl,--as-needed`` for all libraries and tests if the linker supports this flag? Not only for GCC, because this is ld issue.
What do you think?

pavelkumbrasev:
> @pavelkumbrasev , @aleksei-fedotov
> 
> This may be not a bug, but the documented behavior of ld:
> 
> https://wiki.gentoo.org/wiki/Project:Quality_Assurance/As-needed#Initializers_and_deconstructors
> 
> > There exists a class of applications at the moment that break when using --as-needed . These applications are not at fault, but rather the linker itself. The linker is unable to detect dependencies between the initializers and deconstructors (.init/.fini ELF sections) when working with C++ code. As such, it may discard libraries when none of the symbols are used from it, thus mistakenly changing the initialization and deconstruction code paths.
> 
> Maybe it's better to use `-Wl,--as-needed` for all libraries and tests if the linker supports this flag? Not only for GCC, because this is ld issue. What do you think?

So, with LTO gcc should enable `--as-needed` on link stage, am I right? My question is which flags will introduce `--as-needed`.

aleksei-fedotov:
IMHO, the patch looks good except for a few things:
1. The elements related to "test_dynamic_link" becoming sort of spread across the file rather than being kept next to each other. Consider moving the part related to "test_numa_dist" out of "test_dynamic_link" scope.
2. The use of "CMAKE_REQUIRED_LIBRARIES" instead of "CMAKE_REQUIRED_LINK_OPTIONS". However, the latter was added in CMake 3.14, which is newer than the project minimally supports. So, AFAIR, cannot do anything regarding this.
3. I still don't understand why the same thing does not happen with the TBB library itself (and other tests)? It does include dynamic_link.cpp on the linker invocation command and `-flto` flag, but lacks pthread library (at least explicitly). Ldd also shows no explicit dependency on pthread. Is it hidden somewhere among other dependencies, like libstdc++ or something? Perhaps, it is machine-dependent because there is invocation line above that includes pthread library in the linker command.
4. The last but not the least: Since the use of `std::call_once` brings about more dependency on a "third-party", I think it might be somewhat an overkill comparing to the language-only solution given in the first version of the patch and then overridden by a consecutive ones. Though, I admit, `std::call_once` is an explicit and hence more readable solution.

phprus:
@aleksei-fedotov 

3. ``test_dynamic_link`` (``dynamic_link.cpp``) use pthread (via ``std::call_once``) only in a global static variable constructor. 

Why no explicit dependency on pthread?
```
ldd ./libtbb.so.*
        ...
        libstdc++.so.6 => /usr/lib64/libstdc++.so.6 (0x00007fb47fa69000)
        libpthread.so.0 => /lib64/libpthread.so.0 (0x00007fb47f2e2000)
        ...
ldd /usr/lib64/libstdc++.so.6
        linux-vdso.so.1 (0x00007ffc93eea000)
        libm.so.6 => /lib64/libm.so.6 (0x00007f239edfd000)
        libc.so.6 => /lib64/libc.so.6 (0x00007f239ea08000)
        /lib64/ld-linux-x86-64.so.2 (0x00007f239f55b000)
        libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007f239e7ef000)
```

4. Using a local static variable and a lambda function might not help.

Can I write comment in Russian?
I'm not sure I can formulate the explanation correctly in English :(

phprus:
@aleksei-fedotov

In Russian:

Если ld по умолчанию использует флаг ``--as-needed``, то ld будет подключать только те библиотеки, символы из которых реально используются. При этом, ld не умеет отслеживать используемые символы из секций ``.init/.fini`` ELF-файлов. Таким образом, если ``pthread`` функции вызываются только из конструктора глобальной ``static`` переменной, то этот вызов ld не обнаружит.

``std::call_once`` вызывается из функции ``init_dynamic_link_data``, которая вызывается из конструктора глобальной ``static`` переменной ``static_init_dl_data`` и из функции ``dynamic_link``. При включенном LTO для теста ``test_dynamic_link`` компилятор может отследить, что вызов функции ``init_dynamic_link_data`` из конструктора ``static_init_dl_data_t`` всегда происходит раньше, чем из функции ``dynamic_link`` и вызов из функции ``dynamic_link`` может быть удален. Таким образом, если это предположение верно (я не дизассемблировал исполняемый файл), то при включенном LTO вызов ``std::call_once`` (а через него зависимость от ``pthread``) остается только в конструкторе ``static_init_dl_data_t``. Эту зависимость ld не видит и не добавляет ссылку на библиотеку ``pthread``.

В связи с этим, решение через локальную статическую переменную и лямбду только маскирует проблему (при более агрессивном inline в случае LTO компилятор может прийти к тем же самым выводам), а добавлении флага ``--no-as-needed`` ее решает.

Для остальных тестов и самой библиотеки TBB проблем нет, так как они используют функции ``pthread`` из кода, а не только из секций ``.init/.fini`` ELF-файлов.

aleksei-fedotov:
> @aleksei-fedotov
> 
> 3. `test_dynamic_link` (`dynamic_link.cpp`) use pthread (via `std::call_once`) only in a global static variable constructor.
> 
> Why no explicit dependency on pthread?
> 
> ```
> ldd ./libtbb.so.*
>         ...
>         libstdc++.so.6 => /usr/lib64/libstdc++.so.6 (0x00007fb47fa69000)
>         libpthread.so.0 => /lib64/libpthread.so.0 (0x00007fb47f2e2000)
>         ...
> ldd /usr/lib64/libstdc++.so.6
>         linux-vdso.so.1 (0x00007ffc93eea000)
>         libm.so.6 => /lib64/libm.so.6 (0x00007f239edfd000)
>         libc.so.6 => /lib64/libc.so.6 (0x00007f239ea08000)
>         /lib64/ld-linux-x86-64.so.2 (0x00007f239f55b000)
>         libgcc_s.so.1 => /lib64/libgcc_s.so.1 (0x00007f239e7ef000)
> ```
> 
> 4. Using a local static variable and a lambda function might not help.
> 

3. Not only there. allocator.cpp also has similar thing. As for pthread dependency on my machine (Fedora 35) `ldd` shows no dependency on it neither for test_parallel_for, test_dynamic_link, libtbb, libstdc++, libgcc_s, libc.
4. I agree, your explanation makes sense. Still I would not say that using local static variable is a way of hiding the problem, I would say that approach avoids the problem. Since that solution does not require to link against libpthread (or any other library, cause it is independent, language-only solution), I don't see how compiler can break the code if it comes to the same conclusions that you mentioned. With providing detailed commentary (for example, like you wrote in Russian) around the code block, I think this can be a justified way to solve this issue. However, I don't insist on it. I believe each one can exist in the code. So, either way is okay to me.

> Can I write comment in Russian? I'm not sure I can formulate the explanation correctly in English :(

I think it is better to try sticking to English. This information is public and hence can be helpful to others, not to mention that more people can collaborate on this. Fortunately, there is a bunch of online resources that may help with translation either side. Hopefully, one day technology would allow us speaking any language we prefer and still clearly understand each other.

phprus:
@aleksei-fedotov

3. Fedora 35 uses glibc 2.34.
In glibc 2.34 libpthread is merged into libc. Starting with glibc 2.34 this issue will not exist.

See https://sourceware.org/glibc/wiki/Release/2.34:
> 2.3. Libraries merged into libc
> Several libraries including libpthread, libdl, libutil, and libanl have all been merged into libc in order to provide a smoother in-place-upgrade experience, to support safe use of threading at any time by a process, and to simplify the internal implementation. Parts of libresolv have been moved into libc, and this was done to support moving the NSS files and dns plugins directly into libc. The NSS files and dns plugins are now builtin to libc and they can be used during an upgrade or across a chroot or container boundary. Use across a chroot or container boundary supports safely querying of IdM data from those sources. In a future release the rest of libresolv will be moved into libc.


4. The initialization of a local static variable in C++11 is thread safe. The language implicitly uses locks to ensure thread safety. On UNIX-like OS, this is most likely pthread.

If the compiler proves that the call to the ``init_dynamic_link_data`` from the ``dynamic_link `` can be removed, but do not remove the initialization lock, then we get a repeat of this issue.

> I think it is better to try sticking to English.

I am sorry :(


PR updated.
``test_numa_dist`` and ``test_dynamic_link`` options are ordered.

aleksei-fedotov (approved):
Looks good to me.

pavelkumbrasev (approved):

isaevil (approved):
LGTM!

PR #939, opened 2022-10-11
Handle case when LRU history capacity is zero

### LINKED ISSUE #265: tbb::concurrent_lru_cache fails when my_number_of_lru_history_items == 0
1 line fix: https://github.com/oneapi-src/oneTBB/pull/263

```
#define TBB_PREVIEW_CONCURRENT_LRU_CACHE 1
#include "tbb/concurrent_lru_cache.h"

int main() {
    tbb::concurrent_lru_cache<int, int> cache([](int i) { return i; }, 0);
    cache[2];
}
```

https://godbolt.org/z/h4xT4Y

isaevil:
Signed-off-by: Isaev, Ilya <ilya.isaev@intel.com>

### Description 
Current implementation of `concurrent_lru_cache` doesn't handle case when `number_of_lru_history_items` equals zero. This patch is inspired by #263

Fixes #265 

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

@kboyarinov @pavelkumbrasev @HenryHeffanFormlabs

### DIFF
--- a/test/common/utils.h
+++ b/test/common/utils.h
@@ -462,13 +462,11 @@ public:
     }
 
     static bool is_alive(const LifeTrackableObject& object) {
-        auto it = alive_objects.find(&object);
-        return it != alive_objects.end();
+        return is_alive(&object);
     }
 
     static bool is_alive(const LifeTrackableObject* object) {
-        auto it = alive_objects.find(object);
-        return it != alive_objects.end();
+        return alive_objects.find(object) != alive_objects.end();
     }
 
     static const set_type& set() {
--- a/test/common/utils.h
+++ b/test/common/utils.h
@@ -1,5 +1,5 @@
 /*
-    Copyright (c) 2005-2021 Intel Corporation
+    Copyright (c) 2005-2022 Intel Corporation
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
@@ -32,6 +32,7 @@
 #include <algorithm>
 #include <cstring>
 #include <chrono>
+#include <unordered_set>
 
 #if HARNESS_TBBMALLOC_THREAD_SHUTDOWN && __TBB_SOURCE_DIRECTLY_INCLUDED && (_WIN32 || _WIN64)
 #include "../../src/tbbmalloc/tbbmalloc_internal_api.h"
@@ -436,6 +437,45 @@ concept well_formed_instantiation = requires {
     typename Template<Types...>;
 };
 #endif // __TBB_CPP20_CONCEPTS_PRESENT
+
+class LifeTrackableObject {
+    using set_type = std::unordered_set<const LifeTrackableObject*>;
+    static set_type alive_objects;
+public:
+    LifeTrackableObject() {
+        alive_objects.insert(this);
+    }
+
+    LifeTrackableObject(const LifeTrackableObject&) {
+        alive_objects.insert(this);
+    }
+
+    LifeTrackableObject(LifeTrackableObject&&) {
+        alive_objects.insert(this);
+    }
+
+    LifeTrackableObject& operator=(const LifeTrackableObject&) = default;
+    LifeTrackableObject& operator=(LifeTrackableObject&&) = default;
+
+    ~LifeTrackableObject() {
+        alive_objects.erase(this);
+    }
+
+    static bool is_alive(const LifeTrackableObject& object) {
+        auto it = alive_objects.find(&object);
+        return it != alive_objects.end();
+    }
+
+    static bool is_alive(const LifeTrackableObject* object) {
+        auto it = alive_objects.find(object);
+        return it != alive_objects.end();
+    }
+
+    static const set_type& set() {
+        return alive_objects;
+    }
+};
+std::unordered_set<const LifeTrackableObject*> LifeTrackableObject::alive_objects{};
 
 } // namespace utils
 
--- a/test/tbb/test_concurrent_lru_cache.cpp
+++ b/test/tbb/test_concurrent_lru_cache.cpp
@@ -1,5 +1,5 @@
 /*
-    Copyright (c) 2005-2021 Intel Corporation
+    Copyright (c) 2005-2022 Intel Corporation
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
--- a/test/tbb/test_concurrent_lru_cache.cpp
+++ b/test/tbb/test_concurrent_lru_cache.cpp
@@ -21,6 +21,7 @@
 #endif
 
 #include "common/test.h"
+#include "common/utils.h"
 #include <tbb/concurrent_lru_cache.h>
 #include <common/concurrent_lru_cache_common.h>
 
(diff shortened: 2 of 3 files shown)


BenFrantzDale (dismissed):
Good to see this revisited. It looks good.

BenFrantzDale:
If you are revisiting the LRU cache, I'd recommend reworking it to use `tbb::collaborative_once_flag` (that Henry and I instigated) to avoid a potential moonlighting deadlock in `operator[]`:
```
        if (op.is_new_value_needed()){
             op.result().second.my_value = my_value_function(k);
             __TBB_store_with_release(op.result().second.my_is_ready, true);
        }else{
            tbb::internal::spin_wait_while_eq(op.result().second.my_is_ready,false); //< Could cause moonlighting deadlock, IIRc.
        }
```
I think that would mean changing
```
struct map_value_type {
    value_type my_value;
    ref_counter_type my_ref_counter;
    typename lru_list_type::iterator my_lru_list_iterator;
    bool my_is_ready;
    map_value_type(...
};
```
to 
```
struct map_value_type {
    value_type my_value;
    ref_counter_type my_ref_counter;
    typename lru_list_type::iterator my_lru_list_iterator;
    tbb::collaborative_once_flag my_once_flag;

    map_value_type(...
};
```
and then `operator[]` is more like
```
        auto& map_value = op.result().second;
        tbb::collaborative_call_once(map_value.my_once_flag, [&map_value, &my_value_function, &k] {
             map_value.my_value = my_value_function(k);
        });
```

test/tbb/test_concurrent_lru_cache.cpp
```diff
@@ -125,3 +125,27 @@ TEST_CASE("basic test for eviction of only unused items 2") {
...
+        call_count++;
+        return key*2;
+    };
+    
+    using cache_type =  tbb::concurrent_lru_cache<int, int, decltype(counter_func)>;
+    cache_type cache{counter_func, 0};
+
+    for(int i = 1; i <= 10; ++i) {
+        cache[1];
+        REQUIRE_MESSAGE(call_count == i, "when number_of_lru_history_items is zero, element must be erased after use");
```
kboyarinov, line 141: As for me, this check does not guarantee that the element is erased from the cache. Since we don't have the API that checks for the element presence in the cache, I would recommend to use an object with custom destructor (or check other tests and use some pre-existing type with the static counter) to check if the object was actually deleted by calling the dtor.

isaevil:
> If you are revisiting the LRU cache, I'd recommend reworking it to use `tbb::collaborative_once_flag` (that Henry and I instigated) to avoid a potential moonlighting deadlock in `operator[]`:
> 
> ```
>         if (op.is_new_value_needed()){
>              op.result().second.my_value = my_value_function(k);
>              __TBB_store_with_release(op.result().second.my_is_ready, true);
>         }else{
>             tbb::internal::spin_wait_while_eq(op.result().second.my_is_ready,false); //< Could cause moonlighting deadlock, IIRc.
>         }
> ```
> 
> I think that would mean changing
> ...

@BenFrantzDale thanks for you recommendations! I think it is more reasonable to create a different issue with more detailed problem description (having examples would also be really great). This PR solves different issue so we definitely would like to track that in some separate place. Thank you!

BenFrantzDale:
> > If you are revisiting the LRU cache, I'd recommend reworking it to use `tbb::collaborative_once_flag` (that Henry and I instigated) to avoid a potential moonlighting deadlock in `operator[]`:
> > ```
> >         if (op.is_new_value_needed()){
> >              op.result().second.my_value = my_value_function(k);
> >              __TBB_store_with_release(op.result().second.my_is_ready, true);
> >         }else{
> >             tbb::internal::spin_wait_while_eq(op.result().second.my_is_ready,false); //< Could cause moonlighting deadlock, IIRc.
> >         }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think that would mean changing
> > ...
> 
> @BenFrantzDale thanks for you recommendations! I think it is more reasonable to create a different issue with more detailed problem description (having examples would also be really great). This PR solves different issue so we definitely would like to track that in some separate place. Thank you!

Yeah, I wasn't suggesting changing that in this PR. I'll make an issue.

test/common/utils.h
```diff
@@ -437,6 +438,45 @@ concept well_formed_instantiation = requires {
...
+    LifeTrackableObject& operator=(const LifeTrackableObject&) = default;
+    LifeTrackableObject& operator=(LifeTrackableObject&&) = default;
+
+    ~LifeTrackableObject() {
+        alive_objects.erase(this);
+    }
+
+    static bool is_alive(const LifeTrackableObject& object) {
+        auto it = alive_objects.find(&object);
+        return it != alive_objects.end();
```
kboyarinov, line 466:
```suggestion
        return is_alive(&object);
```

test/common/utils.h
```diff
@@ -437,6 +438,45 @@ concept well_formed_instantiation = requires {
...
+    }
+
+    static bool is_alive(const LifeTrackableObject* object) {
+        auto it = alive_objects.find(object);
+        return it != alive_objects.end();
```
kboyarinov, line 471:
```suggestion
        return alive_objects.find(object) != alive_objects.end();
```

test/tbb/test_concurrent_lru_cache.cpp
```diff
@@ -125,3 +126,25 @@ TEST_CASE("basic test for eviction of only unused items 2") {
     REQUIRE_MESSAGE(is_correct, "cache should not evict items in use");
 }
 
+//! \brief \ref error_guessing
+TEST_CASE("basic test for handling case when number_of_lru_history_items is zero") {
+    auto counter_func = [] (int) {
```
kboyarinov, line 131: This is no more counting in this function, isn't it:) Please rename

test/tbb/test_concurrent_lru_cache.cpp
```diff
@@ -125,3 +126,25 @@ TEST_CASE("basic test for eviction of only unused items 2") {
...
+        return utils::LifeTrackableObject{};
+    };
+    using cache_type =  tbb::concurrent_lru_cache<int, utils::LifeTrackableObject, decltype(counter_func)>;
+    cache_type cache{counter_func, 0};
+    
+    for(int i = 0; i < 10; ++i) {
+        const utils::LifeTrackableObject* obj_addr = &cache[1].value();
```
kboyarinov, line 138: Consider commenting what is going on:

```suggestion
        const utils::LifeTrackableObject* obj_addr = &cache[1].value();
        // Destructor of the handle returned by the operator[] calls immediately
```

test/tbb/test_concurrent_lru_cache.cpp
```diff
@@ -125,3 +126,25 @@ TEST_CASE("basic test for eviction of only unused items 2") {
...
+        REQUIRE_MESSAGE(utils::LifeTrackableObject::is_alive(obj_addr) == false, "when number_of_lru_history_items is zero, element must be erased after use");
+    }
+
+    cache_type::handle h = cache[1];
+    const utils::LifeTrackableObject* obj_addr = &h.value();
+    auto& object_set = utils::LifeTrackableObject::set();
+    for(int i = 0; i < 10; ++i) {
+        cache[1];
+        REQUIRE_MESSAGE(utils::LifeTrackableObject::is_alive(obj_addr), "there is still a reference for key=1");
```
kboyarinov, line 147: It seems like the message does not report what is going wrong. Consider something like: "The object with the key=1 was destroyed but should not"

test/tbb/test_concurrent_lru_cache.cpp
isaevil: Didn't noticed suggestion. But seem like mine says similar thing.

kboyarinov (approved):

BenFrantzDale:
Great!

PR #981, opened 2022-12-08
Fix compilation against modern MacOS SDKs

georgthegreat:
While ucontext routines are marked as deprecated on MacOS, modern SDKs trigger #error during compilation:

```
/usr/include/ucontext.h:51:2: error: The deprecated ucontext routines require _XOPEN_SOURCE to be defined
```

This PR workarounds the problem.

### DIFF
--- a/src/tbb/co_context.h
+++ b/src/tbb/co_context.h
@@ -47,6 +47,8 @@
         #pragma clang diagnostic push
         #pragma clang diagnostic ignored "-Wdeprecated-declarations"
     #endif
+    // Workaround error "The deprecated ucontext routines require _XOPEN_SOURCE to be defined" found in modern MacOS SDKs
+    #define _XOPEN_SOURCE 700
 #endif // __APPLE__
 
 #include <ucontext.h>


pavelkumbrasev:
Seems that [PR#900 ](https://github.com/oneapi-src/oneTBB/pull/900/files) with proposed updates might cover this problem as well. (Just enable thread based implementation on newer version)

georgthegreat:
Using external ucontext library is another option which we can not easily adopt.

Can this PR be merged disregarding #900?

pavelkumbrasev:
We don't need to use external ucontext. In that PR I suggest:

>We have thread-based implementation of resumable tasks. What do you think if we enable such implementation in case where ucontext is absented? (Use libc or thread-based)
https://github.com/oneapi-src/oneTBB/blob/f2af74732ba7b303f23a2f453051a52a63560a75/src/tbb/co_context.h#LL27C5-L27C38

So we can just use our thread-based implementation where ucontext is deprecated.

phprus:
``_XOPEN_SOURCE`` is already defined for Apple in cmake:

AppleClang (Xcode):
https://github.com/oneapi-src/oneTBB/blob/989ede96616703421499f87f03849a057c317a5e/cmake/compilers/AppleClang.cmake#L21

Clang:
https://github.com/oneapi-src/oneTBB/blob/989ede96616703421499f87f03849a057c317a5e/cmake/compilers/Clang.cmake#L22-L23

GCC:
https://github.com/oneapi-src/oneTBB/blob/989ede96616703421499f87f03849a057c317a5e/cmake/compilers/GNU.cmake#L22-L23

CMake define ``APPLE`` if **target** platform is macOS, iOS, tvOS or watchOS (https://cmake.org/cmake/help/v3.25/variable/APPLE.html)

georgthegreat:
Ok, I see.
Maybe defining `_XOPEN_SOURCE` will work out.

I am closing this PR for now and return if it will not.

georgthegreat:
I have conducted some research on the subject.

Unfortunately, we can not define `_XOPEN_SOURCE` universally, as it removes certain funtions, not listed in `X/Open 7, incorporating POSIX 2008` (quote from [stackoverflow](https://stackoverflow.com/questions/5378778/what-does-d-xopen-source-do-mean)). I. e. a lot of third party components use `getpagesize()` method without checking its availability.

Having it defined on per-module basis is hardly an option: we operate a huge monorepo and we do not want user code to be dependant os these defines.

Can we proceed with merging this PR taking the above into account?
I think it is even better to do the #undef after `#include <ucontext.h>`.

What do you think?

phprus:
Describe your build environment, please.
Native build from master on macOS (macOS 12.6 + Xcode 14.2 + macOS SDK 13.1 on Apple Silicon) completes without errors.

georgthegreat:
We use custom (proprietary) build system, which translates original CMakeLists.

It consistently copies all the defines from the original build, except for the list of exceptions, including `_XOPEN_SOURCE`.
Later on, we use this system to cross-compile from linux to darwin using clang14.

phprus:
This is not CMake problem.
With this PR we define ``_XOPEN_SOURCE`` by compiler options and in source file.
The macro must be defined only once.

georgthegreat:
Ok, I will keep this patch applied to a local copy.

Issue #982, opened 2022-12-08
test_collaborative_call_once fails on 32-Bit Linux PowerPC


glaubitz:
Tested on git master after using the workaround from #980:

```
Test project /home/glaubitz/oneTBB/build
    Start 5: test_collaborative_call_once
1/1 Test #5: test_collaborative_call_once .....Subprocess aborted***Exception:   0.07 sec
terminate called without an active exception
[doctest] doctest version is "2.4.7"
[doctest] run with "--help" for options
===============================================================================
/home/glaubitz/oneTBB/test/tbb/test_collaborative_call_once.cpp:211:
TEST CASE:  only calls once - stress test

/home/glaubitz/oneTBB/test/tbb/test_collaborative_call_once.cpp:211: FATAL ERROR: test case CRASHED: SIGABRT - Abort (abnormal termination) signal

===============================================================================
[doctest] test cases:  4 |  3 passed | 1 failed | 5 skipped
[doctest] assertions: 12 | 12 passed | 0 failed |
[doctest] Status: FAILURE!


0% tests passed, 1 tests failed out of 1

Total Test time (real) =   0.09 sec

The following tests FAILED:
          5 - test_collaborative_call_once (Subprocess aborted)
Errors while running CTest
```

pavelkumbrasev:
Could you please something similar with https://github.com/oneapi-src/oneTBB/pull/955?

glaubitz:
I tested the workaround from #955 with the following change:

```diff
diff --git a/include/oneapi/tbb/collaborative_call_once.h b/include/oneapi/tbb/collaborative_call_once.h
index db082f89..d8a8a734 100644
--- a/include/oneapi/tbb/collaborative_call_once.h
+++ b/include/oneapi/tbb/collaborative_call_once.h
@@ -1,5 +1,5 @@
 /*
-    Copyright (c) 2021 Intel Corporation
+    Copyright (c) 2021-2022 Intel Corporation
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
@@ -128,6 +128,24 @@ public:
 
 };
 
+// Workaround: MSVC does not apply alignment to variables on the stack
+class collaborative_once_runner_holder : no_copy {
+    collaborative_once_runner *m_runner;
+    char m_storage[max_nfs_size + sizeof(collaborative_once_runner)];
+public:
+    collaborative_once_runner_holder() {
[...]

Issue #26, opened 2017-08-28
Build project with TBB using Intel Compiler 16 is failed


Semen52:
Build the project on Windows using Intel Compiler 16:
tbb/machine/msvc_ia32_common.h(69): error #1879: unimplemented pragma ignored 
               __TBB_MACHINE_DEFINE_ATOMICS(1,8,char,__int8)
With VC this problem isn't observed.

alexey-katranov:
Could you provide a reproducer and/or more details about versions of used products, please? E.g. Intel TBB version, VC version and so on. The issue seems untypical because [Intel(R) C++ Compiler 16.0](https://github.com/01org/tbb/blob/tbb_2017/doc/Release_Notes.txt#L84) is an officially supported compiler and [test_atomic.cpp](https://github.com/01org/tbb/blob/tbb_2017/src/test/test_atomic.cpp) should work with it.

Semen52:
I've build TBB using this version: Commit: eb6336ad29450f2a64af5123ca1b9429ff6bc11d [eb6336a] (Committing Intel(R) TBB 2017 Update 7 source code) and visual studio 2015 (Update 3).
After that I'm linking tbb.lib to my project in VS2015 and trying to build using Intel(R) C++ Compiler 16.0.
It fails:
msvc_ia32_common.h(69): error #1879: unimplemented pragma ignored
__TBB_MACHINE_DEFINE_ATOMICS(1,8,char,__int8)
msvc_ia32_common.h(70): error #1879: unimplemented pragma ignored
__TBB_MACHINE_DEFINE_ATOMICS(2,16,short,__int16)
...
msvc_ia32_common.h(197): error #1879: unimplemented pragma ignored
#pragma intrinsic(__rdtsc)
icl: : error #10298: problem during post processing of parallel object compilation

Semen52:
I've found one topic https://software.intel.com/en-us/forums/intel-c-compiler/topic/278366, but it doesn't contains the solution.

ntfshard:
Hello
Could you please check compatibility of your installed update of Intel(R) C++ Compiler 16.0 with your Visual Studio 2015 update 3. They can be incompatible (VS can be newer than Intel(R) C++ Compiler).

Try to install the latest update of Intel(R) C++ Compiler OR try to use it with older VS compiler.

vladimir-polin:
Hello,

[...]

PR #988, opened 2022-12-09
Fix test_collaborative_call_once SIGABRT for all 32-bit targets (#982)

glaubitz:
Signed-off-by: John Paul Adrian Glaubitz <glaubitz@physik.fu-berlin.de>

### DIFF
--- a/test/tbb/test_collaborative_call_once.cpp
+++ b/test/tbb/test_collaborative_call_once.cpp
@@ -211,12 +211,12 @@ TEST_CASE("only calls once - move only argument") {
 TEST_CASE("only calls once - stress test") {
 #if TBB_TEST_LOW_WORKLOAD
     constexpr std::size_t N = 32;
-#elif __TBB_x86_32 || __arm__  || __ANDROID__
+#elif SIZE_MAX <= 4294967295 // 32-bit targets
     // Some C++ implementations allocate 8MB stacks for std::thread on 32 bit platforms
     // that makes impossible to create more than ~500 threads.
     // Android has been added to decrease testing time.
     constexpr std::size_t N = tbb::detail::d0::max_nfs_size * 2;
-#elif __TBB_USE_THREAD_SANITIZER
+#elif __TBB_USE_THREAD_SANITIZER // 64-bit targets
     // Reduce execution time under Thread Sanitizer
     constexpr std::size_t N = tbb::detail::d0::max_nfs_size + 64;
 #else 
@@ -285,12 +285,12 @@ TEST_CASE("handles exceptions - state reset") {
 TEST_CASE("handles exceptions - stress test") {
 #if TBB_TEST_LOW_WORKLOAD
     constexpr std::size_t N = 32;
-#elif __TBB_x86_32 || __arm__ || __ANDROID__
+#elif SIZE_MAX <= 4294967295 // 32-bit targets
     // Some C++ implementations allocate 8MB stacks for std::thread on 32 bit platforms
     // that makes impossible to create more than ~500 threads.
     // Android has been added to decrease testing time.
     constexpr std::size_t N = tbb::detail::d0::max_nfs_size * 2;
-#else 
+#else // 64-bit targets
     constexpr std::size_t N = tbb::detail::d0::max_nfs_size * 4;
 #endif
 


glaubitz:
FWIW, I noticed that this workaround is required on at least hppa and m68k as well. So we might just enable this workaround for all 32-bit targets.

phprus:
Maybe check the value of the macro ``SIZE_MAX`` (https://en.cppreference.com/w/c/types/limits)?

```c++
#include <stdint.h>
#if SIZE_MAX <= 4294967295
// 32 bit
#else
// 64 bit
#endif
```
32-bit demo: https://godbolt.org/z/1bdWe67qa
64-bit demo: https://godbolt.org/z/aT9zv3cdo

glaubitz:
Yeah, I actually wanted to send a more generic version later after this was merged.

But we can already do that now as tests have shown that every exotic 32-bit target I have tested so far (hppa, m68k, sh4) is affected by this issue as well.

glaubitz:
I think it might be better to check for `__SIZEOF_POINTER__ == 4` as it's easier to read.

Your version doesn't seem to work properly with clang for powerpc64 when testing for 64 bit.

phprus:
``__SIZEOF_POINTER__`` - is non standard and not a portable solution.
MSVC does not define this macro.

glaubitz:
Ah, right, I forgot about MSVC, sorry. I will check what's available on MSVC.

glaubitz:
Shall I use the suggested test for `SIZE_MAX`? Alternatively, we could test for `_WIN32` and `_WIN64` on Windows and `_LP64` on Unix targets.

See: https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170

glaubitz:
> Maybe check the value of the macro `SIZE_MAX` (https://en.cppreference.com/w/c/types/limits)?
> 
> ```c++
> #include <stdint.h>
> #if SIZE_MAX <= 4294967295
> // 32 bit
> #else
> // 64 bit
> #endif
> ```
> 
> 32-bit demo: https://godbolt.org/z/1bdWe67qa 64-bit demo: https://godbolt.org/z/aT9zv3cdo

I have opted for this approach now. Should be good to merge (hopefully) ;).

glaubitz:
Ping. Any update on this?

test/tbb/test_collaborative_call_once.cpp
```diff
@@ -211,12 +211,12 @@ TEST_CASE("only calls once - move only argument") {
 TEST_CASE("only calls once - stress test") {
 #if TBB_TEST_LOW_WORKLOAD
     constexpr std::size_t N = 32;
-#elif __TBB_x86_32 || __arm__  || __ANDROID__
+#elif SIZE_MAX <= 4294967295 // 32-bit targets
```
pavelkumbrasev, line 214: Is there similar macro __arm__ for arm64? (or __arm__ general macro for both configuration and code behavior is changed?)
Similar question with android

pavelkumbrasev:
Could you please add description to PR? 
In general such solution looks kinda odd at least for me. Could you define macro in test config.h with meaningful name and use it here.
Is sizeof(void*) == 4 acceptable?

glaubitz:
> Could you please add description to PR? In general such solution looks kinda odd at least for me. Could you define macro in test config.h with meaningful name and use it here. Is sizeof(void*) == 4 acceptable?

pavelkumbrasev:
> Edit: The answer is simple: `SIZE_MAX` can be evaluated by the pre-processor while `sizeof(void*)` can be evaluated at runtime only.
If I'm not mistaken sizeof is compile time evaluation. https://godbolt.org/z/1Phzbc6nh

glaubitz:
> > Edit: The answer is simple: `SIZE_MAX` can be evaluated by the pre-processor while `sizeof(void*)` can be evaluated at runtime only.
> 
> If I'm not mistaken sizeof is compile time evaluation. https://godbolt.org/z/1Phzbc6nh

You're right, of course. But it's still a difference compared to when the expression is evaluated by the pre-processor what the rest of the code seems to do in these situations, no?

I mean, we can use a macro as suggested to make the use of `SIZE_MAX` more readable.

phprus:
> If I'm not mistaken sizeof is compile time evaluation. https://godbolt.org/z/1Phzbc6nh

At compile time, but not at preprocessing time.

pavelkumbrasev:
Could we use this approach? (I personally don't like the idea comparing with magical constant) https://stackoverflow.com/a/1505839

phprus:
@pavelkumbrasev
I think yes. We can.

Offtopic:
The ``test_task_arena`` test has timeout on Windows.

```
26/138 Test  #26: test_task_arena ..........................***Timeout 185.82 sec
TBB Warning: The number of workers is currently limited to 1. The request for 3 workers is ignored. Further requests for more workers will be silently ignored until the limit changes.
```

https://github.com/oneapi-src/oneTBB/actions/runs/4112056617/jobs/7096513775

I have reproduced this timeout in following envirinment:
1) Tests are run in a virtual machine.
2) The number of virtual cores is small.
3) The physical machine is heavily overloaded - virtual cores do not work simultaneously, but one after another.

pavelkumbrasev:
> @pavelkumbrasev I think yes. We can.
> 
> Offtopic: The `test_task_arena` test has timeout on Windows.
> 
> ```
> 26/138 Test  #26: test_task_arena ..........................***Timeout 185.82 sec
> TBB Warning: The number of workers is currently limited to 1. The request for 3 workers is ignored. Further requests for more workers will be silently ignored until the limit changes.
> ```
> 
> https://github.com/oneapi-src/oneTBB/actions/runs/4112056617/jobs/7096513775
> 
> I have reproduced this timeout in following envirinment:
> 
> 1. Tests are run in a virtual machine.
> 2. The number of virtual cores is small.
> 3. The physical machine is heavily overloaded - virtual cores do not work simultaneously, but one after another.

Currently we are limited with resources and don't have time :(
Could you please create Issue for this problem?

phprus:
@pavelkumbrasev 
Issue already created: https://github.com/oneapi-src/oneTBB/issues/952
My comment in issue: https://github.com/oneapi-src/oneTBB/issues/952#issuecomment-1336448704

I think the problem of hangs on ARM and multiple x86 test slowdowns on an overloaded system has a similar cause.
On x86, these tests don't fully hang, but run 10-100-1000 times slower.

test/tbb/test_collaborative_call_once.cpp
glaubitz: No, there isn't.

(sid_arm64-dchroot)glaubitz@amdahl:~$ echo | gcc -E -dM - |grep -i arm
#define __ARM_SIZEOF_WCHAR_T 4
#define __ARM_FEATURE_IDIV 1
#define __ARM_FP 14
#define __ARM_SIZEOF_MINIMAL_ENUM 4
#define __ARM_ALIGN_MAX_PWR 28
#define __ARM_FP16_FORMAT_IEEE 1
#define __ARM_FP16_ARGS 1
#define __ARM_FEATURE_CLZ 1
#define __ARM_64BIT_STATE 1
#define __ARM_FEATURE_FMA 1
#define __ARM_ARCH_PROFILE 65
#define __ARM_PCS_AAPCS64 1
#define __ARM_ARCH 8
#define __ARM_ARCH_8A 1
#define __ARM_FEATURE_UNALIGNED 1
#define __ARM_NEON 1
#define __ARM_ALIGN_MAX_STACK_PWR 16
#define __ARM_FEATURE_NUMERIC_MAXMIN 1
#define __ARM_ARCH_ISA_A64 1
(sid_arm64-dchroot)glaubitz@amdahl:~$

```
(sid_armhf-dchroot)glaubitz@abel:~$ echo | gcc -E -dM - |grep -i arm
#define __ARM_SIZEOF_WCHAR_T 4
#define __ARM_FEATURE_SAT 1
#define __ARM_ARCH_ISA_ARM 1
#define __ARMEL__ 1
#define __ARM_FP 12
#define __ARM_ASM_SYNTAX_UNIFIED__ 1
#define __ARM_SIZEOF_MINIMAL_ENUM 4
#define __ARM_PCS_VFP 1
#define __ARM_FEATURE_LDREX 15
#define __ARM_FEATURE_UNALIGNED 1
#define __ARM_FEATURE_QBIT 1
#define __ARM_ARCH_PROFILE 65
#define __ARM_32BIT_STATE 1
#define __ARM_FEATURE_CLZ 1
#define __ARM_ARCH_ISA_THUMB 2
#define __ARM_ARCH 7
#define __arm__ 1
#define __ARM_ARCH_7A__ 1
#define __ARM_FEATURE_SIMD32 1
#define __ARM_FEATURE_COPROC 15
#define __ARM_FEATURE_DSP 1
#define __ARM_EABI__ 1
(sid_armhf-dchroot)glaubitz@abel:~$
```

pavelkumbrasev:
> In general such solution looks kinda odd at least for me. Could you define macro in test config.h with meaningful name and use it here.

> Could we use this approach? (I personally don't like the idea comparing with magical constant) https://stackoverflow.com/a/1505839

@glaubitz Could you please address these comments and I think we can move forward and merge this patch.

glaubitz:
> > In general such solution looks kinda odd at least for me. Could you define macro in test config.h with meaningful name and use it here.
> 
> > Could we use this approach? (I personally don't like the idea comparing with magical constant) https://stackoverflow.com/a/1505839
> 
> @glaubitz Could you please address these comments and I think we can move forward and merge this patch.

I just did. The `__arm__` macro is not defined on arm64. I just verified that.

pavelkumbrasev:
Could you please apply these two changes:
- Define macro in test config.h with meaningful name: like UTILS_ARCH_32 and UTILS_ARCH_64
- Use this approach to define these constants https://stackoverflow.com/a/1505839

glaubitz:
> Could you please apply these two changes:
> 
>     * Define macro in test config.h with meaningful name: like UTILS_ARCH_32 and UTILS_ARCH_64
> 
>     * Use this approach to define these constants https://stackoverflow.com/a/1505839

OK, I'll have a look.

test/tbb/test_collaborative_call_once.cpp
```diff
@@ -285,12 +285,12 @@ TEST_CASE("handles exceptions - state reset") {
 TEST_CASE("handles exceptions - stress test") {
 #if TBB_TEST_LOW_WORKLOAD
     constexpr std::size_t N = 32;
-#elif __TBB_x86_32 || __arm__ || __ANDROID__
+#elif SIZE_MAX <= 4294967295 // 32-bit targets
```
Millak, line 288: The comment below says that Android was added to decrease testing time. It'd probably be better then to do:
(SIZE_MAX <= 4294967295) || __ANDROID__ // 32-bit targets or Android

PR #1017, opened 2023-01-23
Remove shebang lines from Python files

### LINKED ISSUE #1014: Useless(?) shebangs in python modules
https://github.com/oneapi-src/oneTBB/blob/master/python/TBB.py and all the files in https://github.com/oneapi-src/oneTBB/tree/master/python/tbb/ have a `#!/usr/bin/env python3` shebang, but none of them is executable. That means the shebang is useless, because you can never execute them.

Are those python files actually *supposed* to be executed? Or are they only intended to be used as modules, e.g. `python -m TBB user_app.py`?

I note that in tbb-2020.3 not all those files had a shebang, only `TBB.py`, `tbb/__init__.py` and `tbb/__main__.py`. But in oneTBB the others have gained a shebang too.

This causes problems for downstream packagers, e.g. Fedora has packaging checks that do not allow non-executable files with shebangs, because it doesn't make sense (if they're supposed to be executable they should have the 'x' bit set in their file mode, and if they're not supposed to be executable they shouldn't have a shebang). So currently the Fedora package for `tbb` uses `chmod` to make those files executable, and if the package is updated to oneTBB then the packaging config will need to be updated to `chmod` the other files that didn't used to have a shebang. Maybe we should really be stripping the shebang from them instead.

If I'm right that those shebangs are useless then I can submit a PR to remove them, but I wanted to check if I'm misunderstanding something.

jwakely:
### Description 
These files do not have the executable bit set in their file mode, and they only need to be run as modules, not as executable scripts.

Fixes #1014

### Type of change

### Tests

- [x] not needed

100% tests passed, 0 tests failed out of 136

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/python/tbb/__main__.py
+++ b/python/tbb/__main__.py
@@ -1,6 +1,4 @@
-#!/usr/bin/env python3
-#
-# Copyright (c) 2016-2021 Intel Corporation
+# Copyright (c) 2016-2023 Intel Corporation
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/python/TBB.py
+++ b/python/TBB.py
@@ -1,6 +1,4 @@
-#!/usr/bin/env python3
-#
-# Copyright (c) 2016-2021 Intel Corporation
+# Copyright (c) 2016-2023 Intel Corporation
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/python/setup.py
+++ b/python/setup.py
@@ -1,6 +1,4 @@
-#!/usr/bin/env python3
-#
-# Copyright (c) 2016-2021 Intel Corporation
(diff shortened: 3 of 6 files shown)


python/tbb/__main__.py
```diff
@@ -1,6 +1,4 @@
-#!/usr/bin/env python3
-#
-# Copyright (c) 2016-2021 Intel Corporation
+# Copyright (c) 2016-2022 Intel Corporation
```
isaevil, line 1:
```suggestion
# Copyright (c) 2016-2023 Intel Corporation
```

jwakely: Yeah, sorry for the typo. I've force-pushed a fix.

isaevil (approved):

pavelkumbrasev (approved):

PR #1033, opened 2023-02-23
Adding helper methods to concurrent_queue and concurrent_bounded_queue

JhaShweta1:
### Description 
Align interface of concurrent_queue and concurrent_bounded_queue with oneTBB spec

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown


test/tbb/test_concurrent_queue.cpp
```diff
@@ -281,3 +281,29 @@ TEST_CASE("Test clear and dtor with TrackableItem") {
     test_tracking_dtors_on_clear<oneapi::tbb::concurrent_queue<TrackableItem>>();
     test_tracking_dtors_on_clear<oneapi::tbb::concurrent_bounded_queue<TrackableItem>>();
 }
+
+template<typename CQ, typename T>
+void test_iterator_ctor(){
+    int size = 5;
+    T vec_1(size, 0), vec_2(size, 0), vec_3(size, 0), vec_4(size, 0);
+    srand(time(0));
```
dnmokhov, line 289: Seems like you need `static_cast<unsigned>(time(0))` to make the Microsoft compiler happy.

include/oneapi/tbb/concurrent_queue.h
```diff
@@ -131,8 +133,43 @@ class concurrent_queue {
...
+      my_queue_representation->assign(*src.my_queue_representation, my_allocator, move_construct_item);
+    }
+
+    void assign( std::initializer_list<value_type> init ){
+      assign(init.begin(), init.end());
+    }
+
+    void swap( concurrent_queue& other ){
+      if (my_allocator == other.my_allocator)
+	internal_swap(other);
```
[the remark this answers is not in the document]
kboyarinov: Actually, `swap` operation with unequal allocators is undefined behavior. I would just add `__TBB_ASSERT` that allocators are equal and then perform unconditioned `internal_swap`. It is done the same way for other containers.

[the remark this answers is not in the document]
kboyarinov: Another question is - do we need non-member swap? Is it required by the spec?

include/oneapi/tbb/concurrent_queue.h
```diff
@@ -131,8 +133,43 @@ class concurrent_queue {
...
+    }
 
+    template <typename InputIterator>
+    void assign( InputIterator first, InputIterator last ){
+      concurrent_queue src(first, last);
+      my_queue_representation->assign(*src.my_queue_representation, my_allocator, move_construct_item);
+    }
+
+    void assign( std::initializer_list<value_type> init ){
+      assign(init.begin(), init.end());
```
kboyarinov, line 163: It seems like an issue here and in `assign(Iterator, Iterator)`. I have checked `my_queue_representation->assign` and it seems like it does not remove all of the items from the container and immediately starts to assign. I did not deeply investigated `micro_queue::assign` method. Could you please check that such an assignment would not result in memory leak?

include/oneapi/tbb/concurrent_queue.h
```diff
@@ -131,8 +133,43 @@ class concurrent_queue {
         queue_allocator_traits::destroy(my_allocator, my_queue_representation);
         r1::cache_aligned_deallocate(my_queue_representation);
     }
+  
+    concurrent_queue& operator=( const concurrent_queue& other ){
+      if (my_queue_representation != other.my_queue_representation) {
+	my_queue_representation = other.my_queue_representation;
```
kboyarinov, line 139: Such an implementation would not be correct. Since `my_queue_representation` is a pointer, we would just copy a pointer into another instance of queue. We would actually have two queues `other` and `*this` point to the same representation, not two independent queues with the same elements inside. Such an approach can be OK for move assignment, but in case of copy assignment, it can result in double-free than both queues would be destroyed.

include/oneapi/tbb/concurrent_queue.h
```diff
@@ -131,8 +133,43 @@ class concurrent_queue {
         queue_allocator_traits::destroy(my_allocator, my_queue_representation);
         r1::cache_aligned_deallocate(my_queue_representation);
     }
+  
+    concurrent_queue& operator=( const concurrent_queue& other ){
```
kboyarinov, line 137: Ideally, special allocator support should be done for copy/move assignment, copy constructor and swap, but it can be a complicated in case of `concurrent_queue` that uses PIMPL implementation. But I guess we should at least add some TODO comments. Something like:
`//TODO: implement support for std::allocator_traits::propogate_on_container_copy_assignment` and so on.

include/oneapi/tbb/concurrent_queue.h
```diff
@@ -215,6 +251,16 @@ class concurrent_queue {
...
+
+    friend bool operator==( const concurrent_queue<T, Allocator>& lhs,
+                            const concurrent_queue<T, Allocator>& rhs ){
+      return lhs.my_queue_representation == rhs.my_queue_representation;
+    }
+  
+    friend bool operator!=( const concurrent_queue<T, Allocator>& lhs,
+                            const concurrent_queue<T, Allocator>& rhs ){
+      return lhs.my_queue_representation != rhs.my_queue_representation;
+    }
```
kboyarinov, line 263:
```suggestion
#if !__TBB_CPP20_COMPARISONS_PRESENT
    friend bool operator!=( const concurrent_queue<T, Allocator>& lhs,
                            const concurrent_queue<T, Allocator>& rhs ){
      return lhs.my_queue_representation != rhs.my_queue_representation;
    }
#endif // __TBB_CPP20_COMPARISONS_PRESENT
```

Since C++20, `!=` operation would be automatically done by the compiler using the `==` primary operator

include/oneapi/tbb/concurrent_queue.h
```diff
@@ -215,6 +251,16 @@ class concurrent_queue {
 
     queue_allocator_type my_allocator;
     queue_representation_type* my_queue_representation;
+
+    friend bool operator==( const concurrent_queue<T, Allocator>& lhs,
+                            const concurrent_queue<T, Allocator>& rhs ){
+      return lhs.my_queue_representation == rhs.my_queue_representation;
```
kboyarinov, line 257: Such an implementation seems incorrect. I guess two queues are considered equal if they contains the same elements, e.g.
```
   concurrent_queue<int> q1, q2;
   q1.push(1);
   q2.push(1);
   bool expect_true = (q1 == q2);
```

`q1` and `q2` both contains only one element equal to `1`, so these queues are considered equal. But they does not point to the same representation, so your implementation of `operator==` would return `false`. I believe that we should not have the case when two queues points to the same representation - it can lean in double-free problems.

include/oneapi/tbb/concurrent_queue.h
```diff
@@ -346,6 +395,40 @@ class concurrent_bounded_queue {
                                          sizeof(queue_representation_type));
     }
 
+      concurrent_bounded_queue& operator=( const concurrent_bounded_queue& other ){
```
kboyarinov, line 398: Actually, all of the comments I left for `concurrent_queue` are applicable for `concurrent_bounded_queue` as well.

test/tbb/test_concurrent_queue.cpp
```diff
@@ -281,3 +281,29 @@ TEST_CASE("Test clear and dtor with TrackableItem") {
     test_tracking_dtors_on_clear<oneapi::tbb::concurrent_queue<TrackableItem>>();
     test_tracking_dtors_on_clear<oneapi::tbb::concurrent_bounded_queue<TrackableItem>>();
 }
+
+template<typename CQ, typename T>
+void test_iterator_ctor(){
```
kboyarinov, line 286: It seems like the name of the test should be changed, because it does not only tests iterator constructor

include/oneapi/tbb/concurrent_queue.h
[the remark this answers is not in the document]
dnmokhov: https://spec.oneapi.io/versions/latest/elements/oneTBB/source/containers/concurrent_queue_cls.html#non-member-functions

include/oneapi/tbb/concurrent_queue.h
```diff
@@ -135,37 +136,42 @@ class concurrent_queue {
...
     }
 
     concurrent_queue& operator=( concurrent_queue&& other ){
-      if (my_queue_representation != other.my_queue_representation) {
-	my_queue_representation = std::move(other.my_queue_representation);
-      }
-      return *this;
+        //TODO: implement support for std::allocator_traits::propogate_on_container_move_assignment
+        if (my_queue_representation != other.my_queue_representation) {
+            my_queue_representation = std::move(other.my_queue_representation);
```
kboyarinov, line 150: Since `my_queue_representation` is a pointer, such a move is incorrect, because it will just copy pointers and result in two queues with the same representation. 
The correct move assignment would be something like:
```
if (my_allocator == other.my_allocator) {
    my_queue_representation = other.my_queue_representation;
    other.my_queue_representation = nullptr;
} else {
    // Deep per-element move, e.g. with assign (again if assign re-allocates everything)
    clear();
    my_queue_representation->assign(*other.my_queue_representation, my_allocator, move_construct_item);
}
```

For move construction/assignment, we should reuse existing move support tests that covers tricky use-cases with equal/unequal allocators. See how it is done for other containers as a reference.

include/oneapi/tbb/concurrent_queue.h
```diff
@@ -254,13 +261,41 @@ class concurrent_queue {
...
+	if(lhs.unsafe_size() == rhs.unsafe_size()){
+	    ret = true;
+	    while(lhs_ptr != lhs.unsafe_end()){
+	        if(*lhs_ptr != *rhs_ptr)
+		  ret = false;
+		lhs_ptr++;
+		rhs_ptr++;
+	    }
+	}
+	return ret;
```
kboyarinov, line 276: May be we can simplify using algorithms:

```suggestion
        return lhs.unsafe_size() == rhs.unsafe_size() && std::equal(lhs.unsafe_begin(), lhs.unsafe_end(), rhs.unsafe_begin());
```

include/oneapi/tbb/concurrent_queue.h
```diff
@@ -135,37 +136,42 @@ class concurrent_queue {
...
   
     concurrent_queue& operator=( const concurrent_queue& other ){
-      if (my_queue_representation != other.my_queue_representation) {
-	my_queue_representation = other.my_queue_representation;
-      }
-      return *this;
+        //TODO: implement support for std::allocator_traits::propogate_on_container_copy_assignment
+        if (my_queue_representation != other.my_queue_representation) {
+            clear();
+	    my_queue_representation->assign(*other.my_queue_representation, my_allocator, copy_construct_item);
```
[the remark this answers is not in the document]
JhaShweta1: micro_queue's assign reallocates the memory and then copy and move constructor items are constructed and placed in this preallocated memory using new. I tried the test that you had mentioned (I believe you meant test_scoped_allocator) but that test fail because '"Allocator should propagate to the data during copy construction'. Do you think it is something that will get resolved when we implement support of propagate_on_container or is it something that you mentioned earlier.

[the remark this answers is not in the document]
kboyarinov: I meant not exactly the scoped allocator test. It actually tests only the support for `propagate_on_container_*`. I meant a test that checks the following:
Suppose that we have two queues with unequal allocator instances (`q1.get_allocator() != q2.get_allocator()`). 
I guess counting allocators would be good for this test (we have the implementation somewhere in tests).
First of all - save the number of allocations/deallocations done for creating the container.
Then do `q1 = q2` and check that `q1` destroys all of the elements and deallocates using the previous allocator from `q1` and then re-allocates the memory using the allocator from `q2`.
If the description seems overcomplicated, we can discuss it offline:)

include/oneapi/tbb/concurrent_queue.h
```diff
@@ -131,8 +134,52 @@ class concurrent_queue {
...
+	}
+	return *this;
+    }
+
+    concurrent_queue& operator=( concurrent_queue&& other ){
+        //TODO: implement support for std::allocator_traits::propogate_on_container_move_assignment
+        if (my_allocator == other.my_allocator){
+	    my_queue_representation = other.my_queue_representation;
```
kboyarinov, line 150: We need to store `other` into the "empty" state as well:
```suggestion
	    my_queue_representation = other.my_queue_representation;
	    other.my_queue_representation = nullptr;
```

include/oneapi/tbb/concurrent_queue.h
```diff
@@ -131,8 +134,52 @@ class concurrent_queue {
...
+	}
+	else{
+	    clear();
```
kboyarinov, line 153: I guess `clear()` should be done in the beginning of the method unconditionally - it is also required for equal allocators, because if `clear` would not be done before assigning the representations - previous memory would be lost.

include/oneapi/tbb/concurrent_queue.h
```diff
@@ -131,8 +134,54 @@ class concurrent_queue {
         queue_allocator_traits::destroy(my_allocator, my_queue_representation);
         r1::cache_aligned_deallocate(my_queue_representation);
     }
+  
+    concurrent_queue& operator=( const concurrent_queue& other ){
+        //TODO: implement support for std::allocator_traits::propogate_on_container_copy_assignment
+        if (my_queue_representation != other.my_queue_representation) {
+            clear();
+	    my_queue_representation->assign(*other.my_queue_representation, my_allocator, copy_construct_item);
```
kboyarinov, line 142:
```suggestion
            clear();
            my_allocator = other.my_allocator;
	    my_queue_representation->assign(*other.my_queue_representation, my_allocator, copy_construct_item);
```
Since we don't implement POCCA support now, it would be better to always copy assign allocators.

include/oneapi/tbb/concurrent_queue.h
```diff
@@ -131,8 +134,54 @@ class concurrent_queue {
...
+	}
+	return *this;
+    }
+
+    concurrent_queue& operator=( concurrent_queue&& other ){
+        //TODO: implement support for std::allocator_traits::propogate_on_container_move_assignment
+        if (my_allocator == other.my_allocator){
+	    clear();
```
kboyarinov, line 150: As I see, this `clear()` call is not in both `if` and `else` branches - lets move it into the common part before the `if`.
Also, we need to add self-assignment check (either `if (this != &other)` or as it is done for copy assignment).

include/oneapi/tbb/concurrent_queue.h
```diff
@@ -131,8 +134,54 @@ class concurrent_queue {
...
+	    my_queue_representation = other.my_queue_representation;
+	}
+	else{
+	    clear();
+            my_queue_representation->assign(*other.my_queue_representation, my_allocator, move_construct_item);
+	    other.clear();
```
kboyarinov, line 156: I thought one more time about this `clear` and it seems like it should not be done here, because we definitely need to move-assign allocators before doing the `my_queue_representation->assign` to re-allocate everything using `other.my_allocator`. In this case, `other.clear()` can result in an issue because `other.my_allocator` would be moved-from.

May be we can call `assign` method with `other.my_allocator`, then call `other.clear()` and only after that call allocators move assignment? Could you please check it using `container_move_support` tests?

test/tbb/test_concurrent_queue.cpp
```diff
@@ -281,3 +282,63 @@ TEST_CASE("Test clear and dtor with TrackableItem") {
...
+    CHECK(q2 == q3);
+}
+
+TEST_CASE("Test iterator queue"){
+    test_queue_helper<tbb::concurrent_queue<std::vector<int>>, std::vector<int>>();
+    test_queue_helper<tbb::concurrent_bounded_queue<std::vector<int>>, std::vector<int>>();
+}
+
+template <typename queue_type, typename allocator_type>
+void TestMoveQueue(){
```
kboyarinov, line 315: Ideally, we would need to test move support as it is done for other containers. 
For example, https://github.com/oneapi-src/oneTBB/blob/master/test/conformance/conformance_concurrent_unordered_map.cpp#:~:text=TEST_CASE(%22concurrent_unordered_map%20move,%7D

include/oneapi/tbb/concurrent_queue.h
dnmokhov: I think you forgot to apply this change: `other` does not become empty.

test/tbb/test_concurrent_queue.cpp
```diff
@@ -281,3 +282,63 @@ TEST_CASE("Test clear and dtor with TrackableItem") {
...
+    CQ q4({vec_1, vec_2, vec_3});
+    CQ q3 = {vec_4, vec_2, vec_3};
+
+    q1 = q3;
+    q2 = std::move(q3);
+
+    CHECK(q1 != q4);
+    q1.swap(q4);
+    CHECK(q2 == q4);
+    CHECK(q2 == q3);
```
dnmokhov, line 306: After line 301, `q3` should be empty and not equal to `q2`. Maybe `CHECK(q3.empty())`?

JhaShweta1: If allocator is equal, I just modify the pointer. So q3 is not empty in the end, but q2 == q3 is also irrelevant - removed it.

test/tbb/test_concurrent_queue.cpp
JhaShweta1: I added the test. Because container's member naming is different I rewrote a small test. I think with these two tests I am covering all the cases that are tested in other containers.

include/oneapi/tbb/concurrent_queue.h
```diff
@@ -346,6 +412,53 @@ class concurrent_bounded_queue {
...
+	      my_allocator = other.my_allocator;
+	      my_queue_representation->assign(*other.my_queue_representation, my_allocator, copy_construct_item);
+	  }
+	  return *this;
+    }
+
+    concurrent_bounded_queue& operator=( concurrent_bounded_queue&& other ){
+        //TODO: implement support for std::allocator_traits::propogate_on_container_move_assignment
+        clear();
+	if (my_queue_representation != other.my_queue_representation) {
```
dnmokhov, line 428: Same here.

include/oneapi/tbb/concurrent_queue.h
```diff
@@ -215,6 +266,18 @@ class concurrent_queue {
...
+    friend bool operator==( const concurrent_queue<T, Allocator>& lhs,
+                            const concurrent_queue<T, Allocator>& rhs ){
+        return lhs.unsafe_size() == rhs.unsafe_size() && std::equal(lhs.unsafe_begin(), lhs.unsafe_end(), rhs.unsafe_begin());
+    }
+
+#if !__TBB_CPP20_COMPARISONS_PRESENT
+    friend bool operator!=( const concurrent_queue<T, Allocator>& lhs,
+                            const concurrent_queue<T, Allocator>& rhs ){
+        return !(lhs == rhs);
+    }
```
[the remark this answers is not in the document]
JhaShweta1: It is inside for other containers.

test/tbb/test_concurrent_queue.cpp
```diff
@@ -1,5 +1,5 @@
 /*
-    Copyright (c) 2005-2022 Intel Corporation
+    Copyright (c) 2005-2023 Intel Corporation
```
kboyarinov, line 2: Since new helpers are parts of the spec now - all of the tests should be moved to conformance part and each test case should be marked as `//! \brief \ref interface \ref requirement`

test/tbb/test_concurrent_queue.cpp
```diff
@@ -281,3 +282,211 @@ TEST_CASE("Test clear and dtor with TrackableItem") {
...
+    CHECK(q2 == q4);
+
+    swap(q2, q3);
+    CHECK(q2.empty());
+    CHECK(q3 == q4);
+}
+
+//! basic test for copy, move and swap
+//! \brief \ref error_guessing
+TEST_CASE("Test iterator queue") {
```
kboyarinov, line 315: Could you please change the name of the test case to better explain what is tested.

test/tbb/test_concurrent_queue.cpp
```diff
@@ -281,3 +282,211 @@ TEST_CASE("Test clear and dtor with TrackableItem") {
...
+    test_queue_helper<tbb::concurrent_queue<std::vector<int>>, std::vector<int>>();
+    test_queue_helper<tbb::concurrent_bounded_queue<std::vector<int>>, std::vector<int>>();
+}
+
+template <typename queue_type, typename allocator_type>
+void TestMoveQueue() {
+    allocator_type::set_limits(300);
```
kboyarinov, line 322: Why do we need to set limits for this test case? I thought limits are intended to throw `bad_alloc` if the number of bytes extends the limit. It seems like it is not used for this test case. Could you please check if we can drop this line?

test/tbb/test_concurrent_queue.cpp
```diff
@@ -281,3 +282,211 @@ TEST_CASE("Test clear and dtor with TrackableItem") {
...
+    TestMoveQueue<tbb::concurrent_bounded_queue<move_support_tests::Foo, allocator_type>, allocator_type>();
+}
+
+template<class T>
+struct stateful_allocator {
+    typedef T value_type;
+    stateful_allocator () = default;
+    int state = 0;
+    template<class U>
+    constexpr stateful_allocator (const stateful_allocator <U>&) noexcept {}
```
kboyarinov, line 360:
```suggestion
    constexpr stateful_allocator (const stateful_allocator<U>&) noexcept {}
```

test/tbb/test_concurrent_queue.cpp
```diff
@@ -281,3 +282,211 @@ TEST_CASE("Test clear and dtor with TrackableItem") {
...
+
+    T* allocate(std::size_t n) {
+        return static_cast<T*>(std::malloc(n * sizeof(T)));
+    }
+
+    void deallocate(T* p, std::size_t n) noexcept {
+        static_cast<void>(n);
+        std::free(p);
```
kboyarinov, line 368:
```suggestion
    void deallocate(T* p, std::size_t) noexcept {
        ::operator delete(p);
```

test/tbb/test_concurrent_queue.cpp
```diff
@@ -281,3 +282,211 @@ TEST_CASE("Test clear and dtor with TrackableItem") {
...
+    }
+};
+
+template<class T, class U>
+bool operator==(const stateful_allocator <T>& lhs, const stateful_allocator <U>& rhs) { return &lhs.state == &rhs.state; }
+
+template<class T, class U>
+bool operator!=(const stateful_allocator <T>& lhs, const stateful_allocator <U>& rhs) { return &lhs.state != &rhs.state; }
+
+template <typename queue_type, typename allocator_type>
```
kboyarinov, line 378: Same as above: replace `queue_type` -> `QueueType`, drop `allocator_type` and replace it with `typename QueueType::allocator_type`

test/tbb/test_concurrent_queue.cpp
```diff
@@ -281,3 +282,211 @@ TEST_CASE("Test clear and dtor with TrackableItem") {
...
+    REQUIRE_MESSAGE( (allocator_type::items_allocated - q2_items_allocated) <= 0, "More then excepted memory allocated ?" );
+    REQUIRE_MESSAGE(std::all_of(q1.unsafe_begin(), q1.unsafe_end(), is_state_predicate<move_support_tests::Foo::MoveInitialized>()),
+                    "Container did not move construct some elements");
+    REQUIRE_MESSAGE(std::all_of(q2.unsafe_begin(), q2.unsafe_end(), is_state_predicate<move_support_tests::Foo::MovedFrom>()),
+                    "Container did not move all the elements");
+}
+
+//! move assignment test for unequal allocator
+//! \brief \ref error_guessing
+TEST_CASE("Test move queue-unequal allocator") {
```
kboyarinov, line 406:
```suggestion
TEST_CASE("Test move assignment with unequal allocators") {
```

test/tbb/test_concurrent_queue.cpp
```diff
@@ -281,3 +282,211 @@ TEST_CASE("Test clear and dtor with TrackableItem") {
...
+    test_check_move_allocator<tbb::concurrent_bounded_queue<std::vector<int, stateful_allocator<int>>, stateful_allocator<int>>>(src_bnd, dst_bnd, cpy_bnd);
+    REQUIRE_MESSAGE(src.get_allocator() != dst.get_allocator(), "Incorrect test setup: allocators should be unequal");
+    REQUIRE_MESSAGE(src_bnd.size() == 0, "Moved from container should not contain any elements");
+    REQUIRE_MESSAGE(dst_bnd.size() == cpy_bnd.size(), "Queues are not equal");
+    REQUIRE_MESSAGE(std::equal(dst_bnd.unsafe_begin(), dst_bnd.unsafe_end(), cpy_bnd.unsafe_begin()), "Elements are not equal");
+}
+
+//! move assignment test for equal and unequal allocator
+//! \brief \ref error_guessing
+TEST_CASE("concurrent_queue") {
```
kboyarinov, line 489: Please provide the better name for test case

kboyarinov (commented):
The implementation part looks good, but some changes are required for the test part before merging.

test/tbb/test_concurrent_queue.cpp
```diff
@@ -281,3 +282,210 @@ TEST_CASE("Test clear and dtor with TrackableItem") {
...
+    T vec_1(size, 0), vec_2(size, 0), vec_3(size, 0), vec_4(size, 0);
+    srand(static_cast<unsigned>(time(0)));
+    generate(vec_1.begin(), vec_1.end(), rand);
+    generate(vec_2.begin(), vec_2.end(), rand);
+    generate(vec_3.begin(), vec_3.end(), rand);
+    generate(vec_4.begin(), vec_4.end(), rand);
+    
+    CQ q1, q2;
+    CQ q4({vec_1, vec_2, vec_3});
+    CQ q3 = {vec_4, vec_2, vec_3};
```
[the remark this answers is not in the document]
dnmokhov: changed

test/tbb/test_concurrent_queue.cpp
```diff
@@ -281,3 +282,210 @@ TEST_CASE("Test clear and dtor with TrackableItem") {
     test_tracking_dtors_on_clear<oneapi::tbb::concurrent_queue<TrackableItem>>();
     test_tracking_dtors_on_clear<oneapi::tbb::concurrent_bounded_queue<TrackableItem>>();
 }
+
+template<typename CQ, typename T>
```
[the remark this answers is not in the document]
dnmokhov: removed

test/tbb/test_concurrent_queue.cpp
```diff
@@ -281,3 +282,210 @@ TEST_CASE("Test clear and dtor with TrackableItem") {
...
+    CHECK(q2 == q4);
+}
+
+//basic test for copy, move and swap
+TEST_CASE("Test iterator queue") {
+    test_queue_helper<tbb::concurrent_queue<std::vector<int>>, std::vector<int>>();
+    test_queue_helper<tbb::concurrent_bounded_queue<std::vector<int>>, std::vector<int>>();
+}
+
+template <typename queue_type, typename allocator_type>
```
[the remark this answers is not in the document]
dnmokhov: removed

test/tbb/test_concurrent_queue.cpp
dnmokhov: moved

test/tbb/test_concurrent_queue.cpp
dnmokhov: renamed

test/tbb/test_concurrent_queue.cpp
dnmokhov: removed

test/tbb/test_concurrent_queue.cpp
```diff
@@ -281,3 +282,211 @@ TEST_CASE("Test clear and dtor with TrackableItem") {
...
+    q1 = std::move(q2);
+
+    CHECK(q1_items_allocated == allocator_type::items_freed);
+    CHECK(q1_items_constructed == allocator_type::items_destroyed);
+    CHECK(q2_items_allocated >= allocator_type::items_allocated);
+}
+
+//! move assignment test for equal allocator
+//! \brief \ref error_guessing
+TEST_CASE("Test move queue") {
```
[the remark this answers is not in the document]
dnmokhov: renamed

test/tbb/test_concurrent_queue.cpp
```diff
@@ -281,3 +282,211 @@ TEST_CASE("Test clear and dtor with TrackableItem") {
     test_tracking_dtors_on_clear<oneapi::tbb::concurrent_queue<TrackableItem>>();
     test_tracking_dtors_on_clear<oneapi::tbb::concurrent_bounded_queue<TrackableItem>>();
 }
+
+template<typename CQ, typename T>
+void test_queue_helper() {
```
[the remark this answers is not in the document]
dnmokhov: added

test/tbb/test_concurrent_queue.cpp
```diff
@@ -281,3 +282,211 @@ TEST_CASE("Test clear and dtor with TrackableItem") {
...
+
+template<class T, class U>
+bool operator==(const stateful_allocator <T>& lhs, const stateful_allocator <U>& rhs) { return &lhs.state == &rhs.state; }
+
+template<class T, class U>
+bool operator!=(const stateful_allocator <T>& lhs, const stateful_allocator <U>& rhs) { return &lhs.state != &rhs.state; }
+
+template <typename queue_type, typename allocator_type>
+void TestMoveQueueUnequal() {
+    allocator_type::set_limits(300);
```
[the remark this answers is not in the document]
dnmokhov: removed

test/tbb/test_concurrent_queue.cpp
dnmokhov: changed

test/tbb/test_concurrent_queue.cpp
```diff
@@ -281,3 +282,211 @@ TEST_CASE("Test clear and dtor with TrackableItem") {
...
+TEST_CASE("Test move queue") {
+    using allocator_type = StaticSharedCountingAllocator<std::allocator<move_support_tests::Foo>>;
+    TestMoveQueue<tbb::concurrent_queue<move_support_tests::Foo, allocator_type>, allocator_type>();
+    TestMoveQueue<tbb::concurrent_bounded_queue<move_support_tests::Foo, allocator_type>, allocator_type>();
+}
+
+template<class T>
+struct stateful_allocator {
+    typedef T value_type;
+    stateful_allocator () = default;
```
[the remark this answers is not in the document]
dnmokhov: changed

test/tbb/test_concurrent_queue.cpp
```diff
@@ -281,3 +282,211 @@ TEST_CASE("Test clear and dtor with TrackableItem") {
...
+    int state = 0;
+    template<class U>
+    constexpr stateful_allocator (const stateful_allocator <U>&) noexcept {}
+
+    T* allocate(std::size_t n) {
+        return static_cast<T*>(std::malloc(n * sizeof(T)));
```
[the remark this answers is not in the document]
dnmokhov: changed

test/tbb/test_concurrent_queue.cpp
dnmokhov: changed

test/tbb/test_concurrent_queue.cpp
```diff
@@ -281,3 +282,211 @@ TEST_CASE("Test clear and dtor with TrackableItem") {
...
+    }
+
+    void deallocate(T* p, std::size_t n) noexcept {
+        static_cast<void>(n);
+        std::free(p);
+    }
+};
+
+template<class T, class U>
+bool operator==(const stateful_allocator <T>& lhs, const stateful_allocator <U>& rhs) { return &lhs.state == &rhs.state; }
```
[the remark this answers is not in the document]
dnmokhov: changed

test/tbb/test_concurrent_queue.cpp
```diff
@@ -281,3 +282,211 @@ TEST_CASE("Test clear and dtor with TrackableItem") {
...
+
+template<class T, class U>
+bool operator!=(const stateful_allocator <T>& lhs, const stateful_allocator <U>& rhs) { return &lhs.state != &rhs.state; }
```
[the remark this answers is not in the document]
dnmokhov: changed

test/tbb/test_concurrent_queue.cpp
dnmokhov: removed

test/tbb/test_concurrent_queue.cpp
dnmokhov: renamed

test/tbb/test_concurrent_queue.cpp
dnmokhov: renamed

kboyarinov (approved):
LGTM

PR #1054, opened 2023-03-29
Avoid CMake warning if CPack is already included

ZeronSix:
### Description 

One of the ways to use oneTBB in a project is to use CMake `add_subdirectory` or tools like [`cpm-cmake`](https://github.com/cpm-cmake/CPM.cmake).

oneTBB includes `CPack` in `packaging.cmake`. However, the upper-level project CMake files could have already included it.
This leads to a warning:
```
CMake Warning at /home/zeronsix/.local/lib/python3.8/site-packages/cmake/data/share/cmake-3.26/Modules/CPack.cmake:507 (message):
  CPack.cmake has already been included!!
Call Stack (most recent call first):
  oneTBB/cmake/packaging.cmake:22 (include)
  oneTBB/CMakeLists.txt:199 (include)
```

A simple `CMakeLists.txt` to reproduce:
```cmake
cmake_minimum_required(VERSION 3.26)

project(MyProject)

include(CPack)
add_subdirectory(oneTBB)
```

This PR includes CPack only if it has not been included already.

### Type of change

### Tests

- [ ] added
- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown


isaevil:
@ZeronSix I am wondering, where can I find information about this variable `CPack_CMake_INCLUDED` in CMake documentation?

ZeronSix:
Looks like it is not present in [CPack docs](https://cmake.org/cmake/help/latest/module/CPack.html). It is a non-documented "inner" variable defined inside CPack [itself](https://github.com/Kitware/CMake/blob/e159b5b2ae0c42ac493c71a9ef1fbc7cc385db55/Modules/CPack.cmake) at line 505. I don't know if there is a way to avoid this warning without using it.

isaevil:
> Looks like it is not present in [CPack docs](https://cmake.org/cmake/help/latest/module/CPack.html). It is a non-documented "inner" variable defined inside CPack [itself](https://github.com/Kitware/CMake/blob/e159b5b2ae0c42ac493c71a9ef1fbc7cc385db55/Modules/CPack.cmake) at line 505. I don't know if there is a way to avoid this warning without using it.

Could you please specify that this is an internal non-documented variable as a comment to this code in this case?

isaevil (approved):

PR #1055, opened 2023-03-29
Fix build on systems without ucontext.h in libc

pavelkumbrasev:
### Description 
This patch is based on the @ismaell contribution in [PR#900](https://github.com/oneapi-src/oneTBB/pull/900).
Enable thread based resumable tasks if ucontext is absent in libc.
The performance of thread based implementation is questionable but at least oneTBB can be built. 

### Type of change

- [] bug fix - _change that fixes an issue_

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

@ismaell @brad0


pavelkumbrasev:
@vossmjp Do you ok with such approach?

cmake/resumable_tasks.cmake
```diff
@@ -0,0 +1,22 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+include(CheckSymbolExists)
+
+if (CMAKE_SYSTEM_NAME MATCHES "^Linux")
```
isaevil, line 17:
```suggestion
if (UNIX)
```

cmake/resumable_tasks.cmake
```diff
@@ -0,0 +1,22 @@
...
+    check_symbol_exists("getcontext" "ucontext.h" HAVE_UCONTEXT)
+    if (HAVE_UCONTEXT)
```
isaevil, line 19: I would rename `HAVE_UCONTEXT` to `_tbb_have_ucontext` since it is not meant to be used globally.

sarathnandu (approved):
LGTM!
The CI failures seems to be non-related.

sarathnandu:
https://github.com/oneapi-src/oneTBB/blob/master/include/oneapi/tbb/detail/_config.h#L459
IS this still correct?

brad0:
Builds. Ran the test suite..

```
93% tests passed, 9 tests failed out of 135

Total Test time (real) = 424.06 sec

The following tests FAILED:
          4 - test_dynamic_link (Failed)
         17 - test_parallel_invoke (Subprocess aborted)
         23 - test_task_group (Subprocess aborted)
         68 - test_arena_constraints (Subprocess aborted)
        125 - test_malloc_regression (Subprocess aborted)
        127 - test_malloc_compliance (Subprocess aborted)
        128 - test_malloc_used_by_lib (Subprocess aborted)
        129 - test_malloc_lib_unload (Subprocess aborted)
        131 - test_malloc_whitebox (Subprocess aborted)
Errors while running CTest
```

Had to workaround two minor issues..

https://github.com/oneapi-src/oneTBB/issues/1015

brad0:
You should remove the APPLE bits from src/tbb/co_context.h and have the thread implementation for macOS as well.

brad0:
ping.

brad0:
ping.

pavelkumbrasev:
@brad0 Just to clarify. 
The real issue with resumable task for you, is that you can not build oneTBB because of ucontext absence? 
I curious because while thread-based implementation will definitely solve build problem it will contain comparably worse (vs ucontext based) performance for resumable tasks.

brad0:
> @brad0 Just to clarify. The real issue with resumable task for you, is that you can not build oneTBB because of ucontext absence? I curious because while thread-based implementation will definitely solve build problem it will contain comparably worse (vs ucontext based) performance for resumable tasks.

Correct. No ucontext API at all. Modern POSIX compliant OS's have no requirement to implement said API as it is deprecated.

brad0:
This isn't the first time dealing with something like this. QEMU is another example with their use of ucontext for co-routines.
The threaded backend is used on macOS and OpenBSD.

```
coroutine_backend = get_option('coroutine_backend')
ucontext_probe = '''
  #include <ucontext.h>
  #ifdef __stub_makecontext
  #error Ignoring glibc stub makecontext which will always fail
  #endif
  int main(void) { makecontext(0, 0, 0); return 0; }'''

# On Windows the only valid backend is the Windows specific one.
# For POSIX prefer ucontext, but it's not always possible. The fallback
# is sigcontext.
supported_backends = []
if targetos == 'windows'
  supported_backends += ['windows']
else
  if targetos != 'darwin' and cc.links(ucontext_probe)
    supported_backends += ['ucontext']
  endif
  supported_backends += ['sigaltstack']
endif
```

vossmjp (approved):
Looks ok to me. Seems better to compile than not for platforms where the only option is the thread version.

isaevil (approved):
LGTM

PR #1063, opened 2023-04-11
Add install of Windows debugger files (pdb)

### LINKED ISSUE #730: Kindly add installation of Windows debugger files (pdb)?
This is a kind feature request. Thanks for this great library! I've seen that when built on Windows, it generates debugger information. However I think the `*.pdb` files are not installed with `make install`. Therefore it can not be found during debugging, and i.e. it is lost when the build tree is removed.

Would it be possible to add this file to the install target on Windows? I guess something like this may be a good start for shared library builds:
```
if(MSVC AND BUILD_SHARED_LIBS)
    install(FILES $<TARGET_PDB_FILE:${PROJECT_NAME}>
        DESTINATION "${CMAKE_INSTALL_BINDIR}"
        OPTIONAL)
endif()
```
It works a bit different for static libraries, though, and I do not have copy-paste-instructions for that case. See https://stackoverflow.com/a/40860436/7200132 for a good start...

isaevil:
### Description 
*.pdb files are shipped in GH prebuilt binaries but they are not installed when trying to do `make install`

Fixes #730 

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

@emmenlau

### DIFF
--- a/cmake/utils.cmake
+++ b/cmake/utils.cmake
@@ -43,6 +43,12 @@ macro(tbb_install_target target)
                 NAMELINK_ONLY
                 COMPONENT devel)
     endif()
+    if (MSVC AND BUILD_SHARED_LIBS)
+        install(FILES $<TARGET_PDB_FILE:${target}>
+            DESTINATION ${CMAKE_INSTALL_BINDIR}
+            COMPONENT devel
+            OPTIONAL)
+    endif()
 endmacro()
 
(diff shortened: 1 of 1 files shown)


pavelkumbrasev (approved):
LGTM!

PR #1122, opened 2023-06-05
Add additional synchronization around [soft/hard]CachesCleanup function

lplewa:
### Description 
Please check commits descriptions.

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/src/tbbmalloc/backend.cpp
+++ b/src/tbbmalloc/backend.cpp
@@ -319,9 +319,17 @@ inline bool BackendSync::waitTillBlockReleased(intptr_t startModifiedCnt)
         if (currCoalescQInFlyBlocks > 0 && backend->scanCoalescQ(/*forceCoalescQDrop=*/false))
             break;
         // 4) when there are no blocks
-        if (!currBinsInFlyBlocks && !currCoalescQInFlyBlocks)
+        if (!currBinsInFlyBlocks && !currCoalescQInFlyBlocks) {
             // re-scan make sense only if bins were modified since scanned
+            auto pool = backend->extMemPool;
+            if (pool->hardCachesCleanupInProgress.load(std::memory_order_acquire) ||
+                pool->softCachesCleanupInProgress.load(std::memory_order_acquire)) {
+                backoff.pause();
+                continue;
+            }
+
             return startModifiedCnt != getNumOfMods();
+        }
         myBinsInFlyBlocks = currBinsInFlyBlocks;
         myCoalescQInFlyBlocks = currCoalescQInFlyBlocks;
         backoff.pause();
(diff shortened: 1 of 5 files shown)


src/tbbmalloc/backend.h
```diff
@@ -265,6 +265,7 @@ class Backend {
     IndexedBins freeLargeBlockBins,
                 freeSlabAlignedBins;
 
+    std::atomic<intptr_t> backendCleanCnt;
```
igchor, line 268: it might be better to use unsigned integer to have a well-defined overflow

src/tbbmalloc/backend.cpp
```diff
@@ -317,11 +319,19 @@ inline bool BackendSync::waitTillBlockReleased(intptr_t startModifiedCnt)
         if (currCoalescQInFlyBlocks > 0 && backend->scanCoalescQ(/*forceCoalescQDrop=*/false))
             break;
         // 4) when there are no blocks
-        if (!currBinsInFlyBlocks && !currCoalescQInFlyBlocks)
+        if (!currBinsInFlyBlocks && !currCoalescQInFlyBlocks) {
             // re-scan make sense only if bins were modified since scanned
+            auto pool = backend->extMemPool;
+            if (pool->hardCachesCleanupInProgress || pool->softCachesCleanupInProgress) {
```
igchor, line 325: You should be able to use acquire/release memory order instead of sequential consistency (at least on `*CleanupInProgress`): std::memory_order_acquire for load operations , std::memory_order_release for store operations and std::memory_order_acq_rel for read-modify-write operations (`exchange`)

src/tbbmalloc/backend.cpp
```diff
@@ -317,11 +319,19 @@ inline bool BackendSync::waitTillBlockReleased(intptr_t startModifiedCnt)
...
+                backoff.pause();
+                continue;
+            }
+
             return startModifiedCnt != getNumOfMods();
```
igchor, line 330: What if `cleanup` starts after you checked the condition (line 325) but before executing this `return` statement? Is this OK?

src/tbbmalloc/backend.h
lplewa: It would be. But i kept this similar to other counters in which is for some reason also intptr_t

src/tbbmalloc/backend.cpp
lplewa: I don't care - in theory we could wait infinitely, until someone will free memory for us.
What is important is to not return ENOMEM while there was a free memory when user called malloc()

So in your case, if we are in this line - minimum one cleanup* has finished (it was done by this thread, or other)


* for smaller allocation waitTillBlockReleased is called twice (ones before cleanup, and second time after cleanup) so after first call, this sentence is not true - but it doesn't matter anyway

src/tbbmalloc/backend.cpp
lplewa: Done

KFilipek (approved):
LGTM

ldorau (approved):
LGTM

PR #1151, opened 2023-07-28
Fix constraint arena construction

pavelkumbrasev:
### Description 
Currently constrained arena might be initialized (first touch to fields) on inappropriate constraints because initializing thread doesn't apply constraints during initialization.
This PR reuses `binding_observer` to apply constraints on thread that performs initialization. 

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/src/tbb/arena.cpp
+++ b/src/tbb/arena.cpp
@@ -543,13 +543,6 @@ void task_arena_impl::initialize(d1::task_arena_base& ta) {
         .set_core_type(ta.core_type())
         .set_max_threads_per_core(ta.max_threads_per_core())
         .set_numa_id(ta.my_numa_id);
-
-    numa_binding_observer* observer = construct_binding_observer(
-        static_cast<d1::task_arena*>(&ta), arena::num_arena_slots(ta.my_max_concurrency, ta.my_num_reserved_slots),
-        ta.my_numa_id, ta.core_type(), ta.max_threads_per_core());
-    if (observer) {
-        observer->on_scheduler_entry(true);
-    }
 #endif /*__TBB_ARENA_BINDING*/
 
     if (ta.my_max_concurrency < 1) {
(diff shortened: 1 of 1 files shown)


pavelkumbrasev:
> Alternative approach would be deferring initialization of the fields until the task is spawned in the arena or the thread is joined the arena for task execution, essentially adding more laziness to the process. In this case, we can save on superfluous calls to enter and exit observer. It will even save on arena initialization if the task will not be spawned and thread would not be entering arena for execution at all.

This is an interesting suggestion. I believe it will require extra synchronization on each arena entry (we already have one for `task_arena::initialize`) so I'm not sure if it will be effective.
Added TODO comment.

src/tbb/arena.cpp
```diff
@@ -554,15 +553,27 @@ void task_arena_impl::initialize(d1::task_arena_base& ta) {
...
     }
 
+#if __TBB_ARENA_BINDING
+    numa_binding_observer* observer = construct_binding_observer(
+        static_cast<d1::task_arena*>(&ta), arena::num_arena_slots(ta.my_max_concurrency, ta.my_num_reserved_slots),
+        ta.my_numa_id, ta.core_type(), ta.max_threads_per_core());
+    if (observer) {
+        observer->on_scheduler_entry(true);
+    }
+#endif /*__TBB_ARENA_BINDING*/
```
pavelkumbrasev, line 563:
```suggestion
#if __TBB_CPUBIND_PRESENT
    numa_binding_observer* observer = construct_binding_observer(
        static_cast<d1::task_arena*>(&ta), arena::num_arena_slots(ta.my_max_concurrency, ta.my_num_reserved_slots),
        ta.my_numa_id, ta.core_type(), ta.max_threads_per_core());
    if (observer) {
        // TODO: Consider lazy initialization for internal arena so the direct calls to observer might be omitted until actual initialization. 
        observer->on_scheduler_entry(true);
    }
#endif /*__TBB_CPUBIND_PRESENT*/
```

isaevil (approved):
LGTM

PR #1173, opened 2023-08-11
Extend migration example

pavelkumbrasev:
### Description 
Extend migration example to support `recycle_as_continuation` and dynamic children creation. Add ability to allocate "compute" to prolong their life time as a continuation.

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/examples/migration/recursive_fibonacci/fibonacci_single_task.h
+++ b/examples/migration/recursive_fibonacci/fibonacci_single_task.h
@@ -56,19 +56,19 @@ struct single_fib_task : task_emulation::base_task {
             *x = serial_fib_1(n);
         }
         else {
-            auto bypass = this->allocate_child_of_continuation_safe<single_fib_task>(n - 2, &x_r);
-            task_emulation::run_task(this->allocate_child_of_continuation_safe<single_fib_task>(n - 1, &x_l));
+            auto bypass = this->allocate_child_and_increment<single_fib_task>(n - 2, &x_r);
+            task_emulation::run_task(this->allocate_child_and_increment<single_fib_task>(n - 1, &x_l));
 
             // Recycling
             this->s = state::sum;
-            this->recycle_as_continuation();
+            this->recycle_as_predecessor();
 
             // Bypass is not supported by task_emulation and next_task executed directly.
             // However, the old-TBB bypass behavior can be achieved with
             // `return task_group::defer()` (check Migration Guide).
             // Consider submit another task if recursion call is not acceptable
-            // i.e. instead of Recycling + Direct Body call
-            // submit task_emulation::run_task(c.create_child_of_continuation<fib_computation>(n - 2, &c.y));
+            // i.e. instead of Direct Body call
+            // submit task_emulation::run_task(this->allocate_child_and_increment<single_fib_task>(n - 2, &x_r));
             bypass->operator()();
         }
     }
--- a/examples/migration/recursive_fibonacci/fibonacci_single_task.h
+++ b/examples/migration/recursive_fibonacci/fibonacci_single_task.h
@@ -56,8 +56,8 @@ struct single_fib_task : task_emulation::base_task {
             *x = serial_fib_1(n);
         }
         else {
-            task_emulation::run_task(this->create_child_of_continuation_safe<single_fib_task>(n - 1, &x_l));
-            auto bypass = this->create_child_of_continuation_safe<single_fib_task>(n - 2, &x_r);
+            auto bypass = this->allocate_child_of_continuation_safe<single_fib_task>(n - 2, &x_r);
+            task_emulation::run_task(this->allocate_child_of_continuation_safe<single_fib_task>(n - 1, &x_l));
 
             // Recycling
             this->s = state::sum;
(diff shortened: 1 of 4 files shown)


examples/migration/recursive_fibonacci/fibonacci_single_task.h
```diff
@@ -0,0 +1,91 @@
...
+        if (n < cutoff) {
+            *x = serial_fib_1(n);
+        }
+        else {
+            auto bypass = this->allocate_child_of_continuation_safe<single_fib_task>(n - 2, &x_r);
+            task_emulation::run_task(this->allocate_child_of_continuation_safe<single_fib_task>(n - 1, &x_l));
+
+            // Recycling
+            this->s = state::sum;
+            this->recycle_as_continuation();
```
vossmjp, line 64: This is safe here because the bypass task is not spawned, and so it cannot complete before this->s is set. In general though, we are missing the recycle_as_safe_continuation function which would prevent a race in the case where both children were passed to run_task before compute_impl is done.  So recycling as continuation here assumes the developer guards against this race.

examples/migration/recursive_fibonacci/task_emulation_layer.h
```diff
@@ -55,49 +55,98 @@ class base_task {
...
+        return create_child_of_continuation_impl<F>(std::forward<Args>(args)...);
+    }
+
+    template <typename F, typename... Args>
+    F* allocate_child_of_continuation(Args&&... args) {
+        return allocate_child_of_continuation_impl<F>(std::forward<Args>(args)...);
+    }
+
+    template <typename F, typename... Args>
+    F* allocate_child_of_continuation_safe(Args&&... args) {
```
vossmjp, line 98: I think the "safe" part of the allocate_child_of_continuation_safe name is not correct since it solves a different problem than the recycle_as_safe_continuation in the old API. This function is more like allocate_additional_child_of in the old API. The "safe" in old API was about concurrent execution of about-to-be-recycled task with its children / predecessors.

examples/migration/recursive_fibonacci/task_emulation_layer.h
```diff
@@ -55,49 +55,98 @@ class base_task {
     void operator() () const {
         base_task* parent_snapshot = m_parent;
         const_cast<base_task*>(this)->execute();
-        if (m_parent && parent_snapshot == m_parent) {
-            m_parent->release();
+        if (m_parent && parent_snapshot == m_parent && m_child_counter == 0) {
+            if (m_parent->remove_reference() == 0) {
+                m_parent->operator()();
+                delete m_parent;
```
vossmjp, line 61: Any chance, now that we have recycle_as_continuation, that a stack allocated task becomes a continuation?

examples/migration/recursive_fibonacci/fibonacci_single_task.h
pavelkumbrasev: Does `recycle_as_safe_continuation` assume that it should be called before any `spawn`?
Do you think we need it? It might be problematic because tasks doesn't have reference to itself only a children references.

examples/migration/recursive_fibonacci/task_emulation_layer.h
pavelkumbrasev: It definitely possible because we introduce 2 sets of APIs one for stack created task and another for allocated tasks.
Not sure if it is really a problem, however  it is also error prone as old Task API :)

examples/migration/recursive_fibonacci/task_emulation_layer.h
pavelkumbrasev: Do you have any suggestion for the names of these functions?

examples/migration/recursive_fibonacci/fibonacci_single_task.h
vossmjp: Not necessarily. recycle_as_safe_continuation makes *this a continuation and adds an implicit decrement of the ref_count after the task completes its initial execute (the one that does the recycle call). So it assumes that the reference count is N+1, where N is the number of children. Or at least that the reference count is incremented N+1 times (it might not actually reach N+1 if some children are spawned and complete). With continuation passing you don't wait inside the task body for the children to complete, so you just need to make sure that the reference counting works out and the continuation is not prematurely started (unless that's ok).  Since it may not be ok, the safe version is there to prevent it.

examples/migration/recursive_fibonacci/task_emulation_layer.h
vossmjp: I really don't like the old naming "allocate_additional_child_of" because that doesn't explain how it is different from "allocate_child_of", so let's not copy that.  And "safe" is not right either. Do we need the "of_continuation" on all of these calls; there's no other kind of child to allocate. What about "allocate_child_and_increment"? And call the other just "allocate_child".  Or how about "allocate_predecessor_and_increment" and "allocate_predecessor" -- perhaps these more clearly combines the concepts of child and continuation???

vossmjp (dismissed):
Looks good.

isaevil (approved):

PR #1183, opened 2023-08-24
Fix _WIN32 define

Vertexwahn:
By accident, the define was called `WIN32` - should be `_WIN32`.

The string `#if _WIN32` can be found more than 100 times in this repo - `#if WIN32` only one time.

It breaks the Bazel build of oneTBB on Windows (`bazel build //...`).

Was introduced by https://github.com/oneapi-src/oneTBB/pull/1163 @pavelkumbrasev @isaevil


dnmokhov (approved):
LGTM

isaevil (approved):

PR #1184, opened 2023-08-25
Add fuzz testing

dnmokhov:
### Description 
* Add a CMake option: `TBB_FUZZ_TESTING` (default: off)
* Add a `test_fuzzing`, which uses libFuzzer to fuzz all the environment variables and launches `test_task` each time.

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown


test/CMakeLists.txt
```diff
@@ -446,6 +446,20 @@ if (TARGET TBB::tbb)
     tbb_add_test(SUBDIR tbb NAME test_eh_thread DEPENDENCIES TBB::tbb)
     tbb_add_test(SUBDIR tbb NAME test_global_control DEPENDENCIES TBB::tbb)
     tbb_add_test(SUBDIR tbb NAME test_task DEPENDENCIES TBB::tbb)
+
+    if (TBB_FUZZ_TESTING)
+        if (NOT CMAKE_CXX_COMPILER_ID STREQUAL IntelLLVM)
+            message(FATAL_ERROR "Fuzzing requires IntelLLVM compiler.")
+        endif()
```
isaevil, line 453: Clang is applicable here as well.

test/CMakeLists.txt
```diff
@@ -446,6 +446,20 @@ if (TARGET TBB::tbb)
...
+        if (MSVC AND NOT (CMAKE_MSVC_RUNTIME_LIBRARY STREQUAL MultiThreaded))
+            message(FATAL_ERROR "Fuzzing on Windows requires CMAKE_MSVC_RUNTIME_LIBRARY to be set to MultiThreaded.")
+        endif()
+        tbb_add_test(SUBDIR tbb NAME test_fuzzing)
+        add_dependencies(test_fuzzing test_task)
+        target_compile_definitions(test_fuzzing PRIVATE CMD="$<TARGET_FILE:test_task>")
```
isaevil, line 459: Maybe change `CMD` to `TEST_EXECUTABLE`?

test/CMakeLists.txt
isaevil: It is also supported by MSVC.
https://learn.microsoft.com/en-us/cpp/build/reference/fsanitize?view=msvc-170

test/tbb/test_fuzzing.cpp
```diff
@@ -0,0 +1,42 @@
...
+    _putenv_s(var, val.c_str());
+#else
+    setenv(var, val.c_str(), 1);
+#endif
+  }
+
+  if (std::system(CMD) != 0)
+    __builtin_trap();
+
+  return 0;
```
isaevil, line 41: Why not return just a status of the command?
```suggestion
  return std::system(CMD)
```

test/CMakeLists.txt
dnmokhov: I will add Clang. MSVC support, however, is experimental and incomplete. In fact, since this is a special test option, I will enable it only on Linux with IntelLLVM and Clang.

test/tbb/test_fuzzing.cpp
dnmokhov: A fuzz target is only supposed to return [0 or -1](https://llvm.org/docs/LibFuzzer.html#rejecting-unwanted-inputs), based on whether the input is acceptable. A way to "fail" the input is either via one of the sanitizers (e.g., address) or manually signalling ([example](https://llvm.org/docs/LibFuzzer.html#toy-example)).

test/CMakeLists.txt
dnmokhov: I will add output redirects to the command, so `CMD` might be a better name.

dnmokhov: Done.

test/CMakeLists.txt
dnmokhov: Done.

test/CMakeLists.txt
```diff
@@ -446,6 +446,17 @@ if (TARGET TBB::tbb)
...
     tbb_add_test(SUBDIR tbb NAME test_global_control DEPENDENCIES TBB::tbb)
     tbb_add_test(SUBDIR tbb NAME test_task DEPENDENCIES TBB::tbb)
+
+    if (TBB_FUZZ_TESTING AND NOT WIN32)
+        if (NOT ((CMAKE_CXX_COMPILER_ID STREQUAL Clang) OR (CMAKE_CXX_COMPILER_ID STREQUAL IntelLLVM)))
+            message(FATAL_ERROR "Fuzzing requires Clang or IntelLLVM compiler.")
+        endif()
+        tbb_add_test(SUBDIR tbb NAME test_fuzzing)
+        add_dependencies(test_fuzzing test_task)
+        target_compile_definitions(test_fuzzing PRIVATE CMD="$<TARGET_FILE:test_task> >/dev/null 2>&1")
```
isaevil, line 456:
```suggestion
        target_compile_definitions(test_fuzzing PRIVATE CMD="$<TARGET_FILE:test_task> > /dev/null 2>&1")
```
If we redirect to /dev/null/, I think we need to add one more check, that this is Unix.

isaevil (approved):

PR #1201, opened 2023-09-07
Fix test_task crash/hang on RT Linux systems

isaevil:
### Description 
"Test priority inversion" test case manually increases threads' priority (including main thread) to test priority inversion problem but never restore it.  That could lead to busy-waiting inside oneTBB on RT Linux kernels which is not expected. This patch restores previous priority after the test case completes.

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/test/common/utils_concurrency_limit.h
+++ b/test/common/utils_concurrency_limit.h
@@ -1,5 +1,5 @@
 /*
-    Copyright (c) 2020-2022 Intel Corporation
+    Copyright (c) 2020-2023 Intel Corporation
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


test/common/utils_concurrency_limit.h
```diff
@@ -309,6 +309,36 @@ void decrease_thread_priority() {
...
+        pthread_t this_thread = pthread_self();
+        int err = pthread_setschedparam(this_thread, 
+            /*policy*/ m_backup.first, /*sched_param*/ &m_backup.second);
+        ASSERT(err == 0, nullptr);
+    }
+private:
+    std::pair<int, sched_param> get_current_schedparam() {
+        pthread_t this_thread = pthread_self();
+        sched_param params;
+        int policy;
```
kboyarinov, line 330:
```suggestion
        int policy = 0;
```

test/common/utils_concurrency_limit.h
```diff
@@ -309,6 +309,36 @@ void decrease_thread_priority() {
 #endif
 }
 
+#if __unix__
+class increased_priority_guard {
+public:
+    increased_priority_guard() : m_backup(get_current_schedparam()) {
+        increase_thread_priority();
```
kboyarinov, line 316: I have checked the oneTBB repo and it seems like `increase_thread_priority` only used in test_task in the places you changed. And `decrease_thread_priority` is never used. 
May be it makes sense to remove `decrease_thread_priority` and move `increase_thread_priority` implementation into the guard to prevent its usage outside of the guard?

isaevil: Yeah, makes sense. Thank you.

kboyarinov (approved):
LGTM

PR #1209, opened 2023-09-19
Fix race in single task Fibonacci example

pavelkumbrasev:
### Description 
There was a possibility that parent task will be deleted before it exit its body. This patch insures the lifetime of parent task and unblocks ability to recycle continuation several times.

This patch was validated with Thread Sanitizer.

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/examples/migration/recursive_fibonacci/fibonacci_single_task.h
+++ b/examples/migration/recursive_fibonacci/fibonacci_single_task.h
@@ -24,7 +24,7 @@
 #include <utility>
 
 extern int cutoff;
-extern bool tesing_enabled;
+extern bool testing_enabled;
 
 long serial_fib_1(int n) {
     return n < 2 ? n : serial_fib_1(n - 1) + serial_fib_1(n - 2);
@@ -49,7 +49,7 @@ struct single_fib_task : task_emulation::base_task {
             case state::sum : {
                 *x = x_l + x_r;
 
-                if (tesing_enabled) {
+                if (testing_enabled) {
                     if (n == cutoff && num_recycles > 0) {
                         --num_recycles;
                         bypass = compute_impl();
--- a/examples/migration/recursive_fibonacci/fibonacci_single_task.h
+++ b/examples/migration/recursive_fibonacci/fibonacci_single_task.h
@@ -32,7 +32,8 @@ long serial_fib_1(int n) {
 struct single_fib_task : task_emulation::base_task {
     enum class state {
         compute,
-        sum
+        sum,
+        print_status
     };
 
     single_fib_task(int n, int* x) : n(n), x(x), s(state::compute)
@@ -46,6 +47,16 @@ struct single_fib_task : task_emulation::base_task {
             }
             case state::sum : {
                 *x = x_l + x_r;
+
+                // Recycling
+                this->s = state::print_status;
+                this->recycle_as_continuation();
+                this->operator()();
+
+                break;
+            }
+            case state::print_status: {
+                std::cout << std::to_string(n) + "\n";
                 break;
             }
         }
--- a/examples/migration/recursive_fibonacci/fibonacci_single_task.h
+++ b/examples/migration/recursive_fibonacci/fibonacci_single_task.h
@@ -24,6 +24,7 @@
 #include <utility>
 
 extern int cutoff;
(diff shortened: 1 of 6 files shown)


examples/migration/recursive_fibonacci/task_emulation_layer.h
```diff
@@ -47,22 +47,25 @@ class base_task {
...
-    base_task(const base_task& t) : m_parent(t.m_parent), m_child_counter(t.m_child_counter.load())
+    base_task(const base_task& t) : m_parent(t.m_parent), m_ref_counter(t.m_ref_counter.load())
     {}
 
     virtual ~base_task() = default;
 
     void operator() () const {
         base_task* parent_snapshot = m_parent;
+        task_type type_snapshot = m_type;
+
```
vossmjp, line 58: What about creating another m_type of "executing". Store the snapshot and then always set m_type to "executing" before calling this->execute(). On return from this->execute(), you can then detect if that task was recycled as continuation during that execution (if so, you do not decrement parent or delete this). If it remained in "executing" state, it was not recycled during that execution, so you can decrement its parent, remove its self_ref and if it was restored to the "allocated" state, then you can delete it.

examples/migration/recursive_fibonacci/fibonacci_single_task.h
```diff
@@ -46,6 +47,16 @@ struct single_fib_task : task_emulation::base_task {
...
+                this->s = state::print_status;
+                this->recycle_as_continuation();
+                this->operator()();
+
+                break;
+            }
+            case state::print_status: {
+                std::cout << std::to_string(n) + "\n";
                 break;
             }
```
dnmokhov, line 61: In this example, recycling the task again results in doing `m_parent->remove_child_reference()` after `m_parent` has destroyed itself.

examples/migration/recursive_fibonacci/task_emulation_layer.h
pavelkumbrasev: The problem is that approach is: "once recycled task enters `operator()()` and move state to `executing` the original task will read `m_state` and thinks that it was not recycled". Basically. it is another race, probably we can solve it with monotonic epoch.

dnmokhov (approved):
LGTM

dnmokhov (dismissed):

examples/migration/recursive_fibonacci/task_emulation_layer.h
```diff
@@ -47,32 +47,49 @@ class base_task {
...
-        if (m_parent && parent_snapshot == m_parent && m_child_counter == 0) {
-            if (m_parent->remove_reference() == 0) {
+        task_type type_snapshot = m_type;
+
+        base_task* bypass = const_cast<base_task*>(this)->execute();
+
+        bool is_task_recycled_as_child = parent_snapshot != m_parent;
+        bool is_task_recycled_as_continuation = type_snapshot != m_type;
+
+        if (m_parent && !is_task_recycled_as_child && !is_task_recycled_as_continuation) {
```
vossmjp, line 64: After simplifying the bypass approach, I think you should be able to replace ``m_parent && !is_task_recycled_as_child && !is_task_recycled_as_continuation`` with just ``m_type != task_type::recycled``.

examples/migration/recursive_fibonacci/task_emulation_layer.h
```diff
@@ -47,32 +47,49 @@ class base_task {
...
+            if (m_parent->remove_child_reference() == 0) {
                 m_parent->operator()();
-                delete m_parent;
             }
         }
 
-        if (m_child_counter == 0 && m_type == task_type::allocated) {
+        if (type_snapshot != task_type::stack_based && !is_task_recycled_as_child && !is_task_recycled_as_continuation) {
```
vossmjp, line 70: Similarly, I think you should be able to now replace ``type_snapshot != task_type::stack_based && !is_task_recycled_as_child && !is_task_recycled_as_continuation`` with just ``m_type == task_type::allocated``

dnmokhov (approved):

PR #1219, opened 2023-09-28
Fix build failure when using TBB_USE_EXCEPTIONS=0

GertyP:
### Description 
This is just an alternative (arguably simplifying) fix in comparison to [PR 864](https://github.com/oneapi-src/oneTBB/pull/864) and also discussed in [issue 1208](https://github.com/oneapi-src/oneTBB/issues/1208).  I did comment in PR 864, to describe what I've done here, but since 864 has been open for a long time, I thought it might be useful to show my suggested changes here and see if it gains any traction.

We get a link error from 'task_dispatcher.h's reference to 'do_throw_noexcept(...)', which is not defined when building with 'TBB_USE_EXCEPTIONS=0'.  E.g. -
```
cmake ... -DTBB_COMMON_COMPILE_FLAGS:STRING="/DTBB_USE_EXCEPTIONS=0" ...
```

Since any throw from a 'noexcept' function results in a 'std::terminate', and all current uses of 'do_throw_noexcept' do nothing more than throw a custom exception, 'do_throw_noexcept' is equivalent to 'std::terminate', and replacing with 'std::terminate' makes things a bit clearer & cleaner.

Fixes # - 1208

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/src/tbb/exception.cpp
+++ b/src/tbb/exception.cpp
@@ -1,5 +1,5 @@
 /*
-    Copyright (c) 2005-2022 Intel Corporation
+    Copyright (c) 2005-2023 Intel Corporation
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
@@ -41,27 +41,12 @@ const char* user_abort::what() const noexcept(true) { return "User-initiated abo
 const char* missing_wait::what() const noexcept(true) { return "wait() was not called on the structured_task_group"; }
 
 #if TBB_USE_EXCEPTIONS
-    template <typename F>
-    /*[[noreturn]]*/ void do_throw_noexcept(F throw_func) noexcept {
-        throw_func();
-    }
-
-    /*[[noreturn]]*/ void do_throw_noexcept(void (*throw_func)()) noexcept {
-        throw_func();
-#if __GNUC__ == 7
-        // In release, GCC 7 loses noexcept attribute during tail call optimization.
-        // The following statement prevents tail call optimization.
-        volatile bool reach_this_point = true;
-        suppress_unused_warning(reach_this_point);
-#endif
-    }
-
     bool terminate_on_exception(); // defined in global_control.cpp and ipc_server.cpp
 
     template <typename F>
     /*[[noreturn]]*/ void do_throw(F throw_func) {
         if (terminate_on_exception()) {
-            do_throw_noexcept(throw_func);
+            std::terminate();
         }
         throw_func();
     }
--- a/src/tbb/task_dispatcher.h
+++ b/src/tbb/task_dispatcher.h
@@ -54,9 +54,6 @@ inline d1::task* get_self_recall_task(arena_slot& slot) {
     return t;
 }
 
-// Defined in exception.cpp
-/*[[noreturn]]*/void do_throw_noexcept(void (*throw_exception)()) noexcept;
-
 //------------------------------------------------------------------------
 // Suspend point
 //------------------------------------------------------------------------
@@ -356,7 +353,7 @@ d1::task* task_dispatcher::local_wait_for_all(d1::task* t, Waiter& waiter ) {
             break; // Exit exception loop;
         } catch (...) {
             if (global_control::active_value(global_control::terminate_on_exception) == 1) {
-                do_throw_noexcept([] { throw; });
+                std::terminate();
             }
             if (ed.context->cancel_group_execution()) {
                 /* We are the first to signal cancellation, so store the exception that caused it. */


pavelkumbrasev:
It might make sense to keep `do_throw_noexcept` because it gives more information to user. 
Exception message + terminate vs just terminate. 
I would say it make sense to use [DO_THROW](https://github.com/oneapi-src/oneTBB/blob/7b927dc4a828cb7b2e2a68796c8c1966e4a83028/src/tbb/exception.cpp#L76) in task_dispatcher instead of `do_throw_noexcept`.

GertyP:
OK.  That is going a bit further than fixing the issue and keeping the same behaviour;  this initial change just keeps the exact same behaviour but fixes the TBB_USE_EXCEPTIONS==0 problem.

If you prefer, I don't object to going a step further and writing out a message in the cases that have otherwise always quietly terminated.

See the recent [Better exception error reporting](https://github.com/oneapi-src/oneTBB/pull/1219/commits/d0adc35b742afa56c69258dc6f7575311c8c0824) commit for my interpretation of your suggestion.

... I've just seen it's upset the tests
```
TEST CASE:  terminate_on_exception: enabled
  internal exception

D:\a\oneTBB\oneTBB\test\conformance\conformance_global_control.cpp(359): FATAL ERROR: test case CRASHED: SIGABRT - Abort (abnormal termination) signal
```
This looks like the test is expecting something like a terminate instead of an abort.  I'll look in to that when I get time but, ignoring that for now, do the added reporting mechanisms look like what you're after?

A couple of things to note about the additions -
- task_dispatcher.h's 'local_wait_for_all' catch block does (and did) nothing with any caught exception, so I'm reluctant to complicate it further with additional catch clauses, just to get some more info from a `e.what()`
- We now have the following set of different `DO_THROW` call uses against 3 different build/run-time configurations to handle -

|                                                                         | terminate_on_exception==true | terminate_on_exception==false | TBB_USE_EXCEPTIONS==0 |
| ------------- | ------------- | ------------- | ------------- |
| DO_THROW(except_type,"string literal")  | do_terminate_on_exc("except_type","string literal") | throw except_type("string literal") | do_exceptions_disabled_error("except_type","string literal") |
| DO_THROW(except_type,runtime_buf)  | do_terminate_on_exc("except_type",runtime_buf) | throw except_type(runtime_buf) | do_exceptions_disabled_error("except_type",runtime_buf) |
| DO_THROW(except_type2)  | do_terminate_on_exc("except_type2") | throw except_type2() | do_exceptions_disabled_error("except_type2") |

GertyP:
... Yes, the tests are understandably expecting terminate, not abort, when using the `terminate_on_exception` control 😄 .  So, for consistency, all exception-related fatal error reporting (whether though use of `terminate_on_exception` or `TBB_USE_EXCEPTIONS=0`) and now always ends in a `std::terminate()`.

... Whether that's preferable from a user's point of view is a separate question that I don't think should be addressed here;  changing that in this PR would go even further beyond the intended scope of just fixing a broken build configuration while keeping the originally intended behaviour (i.e. it's always essentially been a std::terminate in all cases).

pavelkumbrasev:
@GertyP Thank you for the update. It turned out this place is tricky. (It will take some time for me to figure out how we can proceed with this)

GertyP:
Thanks for having a look at this @pavelkumbrasev 
As mentioned above, the first commit ('Fix build failure when using TBB_USE_EXCEPTIONS=0') is the lowest impact change that strictly fixes the issue in question, while preserving exactly the same existing behaviour ('i.e. quietly terminating when hitting exceptions when the run-time `terminate_on_exception` is enabled, and now the same for when `TBB_USE_EXCEPTIONS` is disabled).

It only got a little more tricky to consider the matrix of uses and build/run combinations at your request to try to get more useful error info out to the user 😄 , which I think is perfectly reasonable and which hopefully I've managed to do in the latest commit... But if you're unwilling to take both together (and we can't agree on what else might need further changes), then perhaps consider just taking just the first change.

pavelkumbrasev:
> Thanks for having a look at this @pavelkumbrasev As mentioned above, the first commit ('Fix build failure when using TBB_USE_EXCEPTIONS=0') is the lowest impact change that strictly fixes the issue in question, while preserving exactly the same existing behaviour ('i.e. quietly terminating when hitting exceptions when the run-time `terminate_on_exception` is enabled, and now the same for when `TBB_USE_EXCEPTIONS` is disabled).
> 
> It only got a little more tricky to consider the matrix of uses and build/run combinations at your request to try to get more useful error info out to the user 😄 , which I think is perfectly reasonable and which hopefully I've managed to do in the latest commit... But if you're unwilling to take both together (and we can't agree on what else might need further changes), then perhaps consider just taking just the first change.

I have tried to clarify whether oneTBB (library part) should be build with  `TBB_USE_EXCEPTIONS` disabled. Seems there was a decision that we don't support such build configuration because it would increase test scope.
So there is a question: Do you need such configuration and why?

GertyP:
> there was a decision that we don't support such build configuration because it would increase test scope

This is a documented feature macro [here](https://spec.oneapi.io/versions/latest/elements/oneTBB/source/configuration/feature_macros.html#).

I can sympathise that it would suggest the need for additional build tests if you really wanted to be rigorous ... but a phrase also comes to mind: "Don't let perfection be the enemy of good".  In other words, I fear that there's a possibility that you find testing a full and complete set of tests with configuration 'TBB_USE_EXCEPTIONS=0' far too much of a burden and so decide it's easer to drop it than to implement an exhaustive set of extra tests.  But there is another practical compromise alternative, which is to add minimal testing that hits this link error at least.  E.g.  -
```
cmake ... -DCMAKE_BUILD_TYPE=Debug -DTBB_CPF=OFF -DTBB_BUILD=on -DTBB_TEST=off -DTBBMALLOC_BUILD=on -DTBB4PY_BUILD=off -DTBB_COMMON_COMPILE_FLAGS:STRING="/DTBB_USE_EXCEPTIONS=0" ...
cmake --build ... --config Debug
```
which takes just a few seconds before revealing this issue as a link error.

But even if there aren't any auto-tests to cover the build and/or run-time testing of this configuration, it's still preferable, from my point of view, if you're still willing to take fixes for cases which you've decided to no longer officially support but which can technically still be used.  I think the vast majority of TBB users would not blame you for not covering absolutely 100% of all build and run-time config permutations, so it would be a shame to not take small fixes when they are discovered.  It would also be a shame to remove a fairly simple, quite well isolated feature like this (it's all quite nicely contained within `exception.cpp`) for fear of not having comprehensive enough auto-testing.  I don't mind that I've found an untested, documented feature issue... so long as you'll take fixes 😃 

> Do you need such configuration and why?

_Need_ may be too strong a word but, in addition to [another answer to this question](https://github.com/oneapi-src/oneTBB/pull/864#issuecomment-1322144738), I think that, in some circumstances, use of exceptions can facilitate worse code to read and understand as well as well as facilitating worse system design decisions.  It's a big topic and a bit philosophical but I think throwing and catching exceptions too easily promotes' _exceptional_ circumstances as normal cases' to be considered and handled throughout more code than usually necessary, encouraging programmers think at a very zoomed in level about how to try to gracefully handle and 'soldier on' with things like bad inputs, bad results, out-of-memory, misuse of an interface, etc. instead of being confronted by the clean simplicity of a hard break/abort/crash, where you're forced to better understand the systems you use, the assumptions you have, and as a result, usually come up with a better designed system (e.g. more comprehensive and careful up-front partitioning and budgeting of memory given to various systems instead of lower-level run-time code that's polluted with bloated retry flailing upon failure to allocate memory).

GertyP:
I'd be interested to know if there's any update on this.  At least a follow-up decision (with some rationale, depending on the decision) would be more satisfactory than an open, ignored PR.

As I mentioned above, the first [commit](https://github.com/oneapi-src/oneTBB/pull/1219/commits/f3f5ce4d87c65fd52308e880d9c65025f4772fef) really is a minimal change that fixes this build config while keeping the same behaviour in the usual `TBB_USE_EXCEPTIONS==1` case, so I think I'd struggle to see any reason not to take at least that.  

However, your suggestion to improve the error reporting has resulted in the [follow-up commit](https://github.com/oneapi-src/oneTBB/pull/1219/commits/983874b5265a88a720b3ea5c81ff21a581bc45ab), which I think not only improves the `TBB_USE_EXCEPTIONS==0` scenario (that you're reluctant to maintain / keen to drop), but it also improves error messages over the previous/existing code for cases that you are using/maintaining:  I.e. anything making use of `global_control::terminate_on_exception`;  anything going though `throw_exception(exception_id)`, `DO_THROW`, and the task_dispatcher's catch around the infinite loop, would all just quietly terminate when using `global_control::terminate_on_exception` but will now terminate with a much more informative message... So, if you're uncomfortable taking this change on the grounds of it fixing something you have that you're not keen to maintain, then please consider taking it on the grounds of it improving something I presume you are wanting to maintain 😃

pavelkumbrasev:
Hi @GertyP, there is still no decision regarding this problem. I'm concerned with such solution because it's just hides the problem.
Library doesn't support build without exception so `TBB_USE_EXCEPTIONS=0` during library build doesn't make sense. (TBB still will throw exception from other places)
So I don't know what is a real value of such fix. @vossmjp what do you think about it?

PR #1245, opened 2023-11-02
Support Windows on ARM64 (aarch64) when building with MinGW-w64

brechtsanders:
Support ARM64 (aarch64) when building with MinGW-w64


isaevil (approved):
LGTM

Issue #1299, opened 2024-01-20
Is there any reason we cannot implement a parallel_reduce for a rvalue range?


sukeya:
Hello,

I'm writing a program merging many `std::set`s. I would like to parallelize it with joining two sets by `std::set::merge`, but oneTBB doesn't seem to provide any move join function like `T join(T&&, T&&)` as far as I can see [Reduction](https://oneapi-src.github.io/oneTBB/main/tbb_userguide/design_patterns/Reduction.html#reduction), [parallel_reduce.h](https://github.com/oneapi-src/oneTBB/blob/4dfcb6d2e91ba240f3ab529cd80f5417d3cecac8/include/oneapi/tbb/parallel_reduce.h#L395) and [enumerable_thread_specific.h](https://github.com/oneapi-src/oneTBB/blob/4dfcb6d2e91ba240f3ab529cd80f5417d3cecac8/include/oneapi/tbb/enumerable_thread_specific.h#L1017).
Is there any reason we cannot implement it?

dnmokhov:
You should be able to use the lvalue-reference version of `std::set::merge`, e.g.,

```cpp
struct Body {
  std::vector<std::set<int>> &sets;
  std::set<int> result;

  Body(std::vector<std::set<int>> &data) : sets(data) {}
  Body(Body &left, tbb::split) : sets(left.sets) {}

  void operator()(tbb::blocked_range<int> range) {
    for (int i = range.begin(); i < range.end(); ++i) {
      result.merge(sets[i]);
    }
  }

  void join(Body &right) { result.merge(right.result); }
};
```

In the example above, `right` is an lvalue reference to the subtask and`right.result` is an lvalue reference to its result.

Specifying a `parallel_reduce` body is further documented at https://oneapi-src.github.io/oneTBB/main/tbb_userguide/parallel_reduce.html

sukeya:
I didn't know that an `Body` can have the `join` member function which argument is a lvalue-reference without `const`.
Thank you!

PR #1307, opened 2024-02-01
Better rvalues support for parallel_reduce algorithm

kboyarinov:
### Description 
Adding the optimization for reduction of rvalues using `parallel_reduce` and `parallel_deterministic_reduce`.
Based on #171 , #169 and #1299.

This patch is intended to optimize operating with Value class in parallel reduce algorithm (lambda version). Example from #1299

```
// We want to efficiently reduce a number of std::set objects into single std::set containing unique values from each set
std::vector<std::set<int>> sets = ...;

parallel_reduce(blocked_range<size_t>(0, sets.size()),
                         std::set<int>{}, // Identity - empty set
                         [&](const blocked_range<size_t>& range, const std::set<int>& value) { // Second argument is required to be const Value& by the spec
                             std::set<int> result = value; // We are forced to do copy here since value is constant
                             
                             for (size_t i = range.begin(); i < range.end(); ++i) {
                                 result.merge(sets[i]); // set::merge is efficient since it transfers required node from one set to another
                             }
                             return result;
                         },
                         [](const std::set<int>& value1, const std::set<int>& value2) { // Both arguments are required to be const Value&
                             std::set<int> result = value1; // We are forced to do copy
                             result.insert(value2.begin(), value2.end()); // unable to use set::merge here since value2 is constant  
                         }                  
                         );
```

The spec forces the user to use inefficient copying instead of "moving" even if the input range preservation is not important for the user. "Body" version of parallel reduce allows to drop unnecessary copying:

```
struct Body {
    std::vector<std::set<int>>& sets;
    std::set<int> result;
    
    void operator()(const blocked_range<size_t>& range) {
        for (size_t i = range.begin(); i != range.end(); ++i) {
            result.merge(sets[i]); // only moving nodes
        }
    }
    
    void join(Body& rhs) {
        result.merge(rhs.result); // also only moving nodes
    }
};
```

The main goal of this patch is to allow rvalues optimization for lambda version of reduce.
This patch is NOT intended to allow reduction of move-only objects (such as `std::unique_ptr`s) since we still need to copy the identity value into each parallel_reduce leaf.

Specification change is also needed for this PR:
* Relax `ParallelReduceFunc` named requirement to allow `Value operator()(const Range& range, Value&& value) const`
* Relax `ParallelReduceReduction` named requirement to allow `Value operator()(Value&& x, Value&& y) const`.

This PR also contains simple test that I guess should be extended to better cover all overloads and algorithms.

### Type of change

### Tests

- [ ] not needed

### Documentation

- [x] needs to be updated
- [ ] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

@BenFrantzDale
@rfschtkt
Feel free to participate in review and discussions for this patch:)

### DIFF
--- a/include/oneapi/tbb/parallel_reduce.h
+++ b/include/oneapi/tbb/parallel_reduce.h
@@ -35,7 +35,7 @@ inline namespace d0 {
 
 template <typename Body, typename Range>
 concept parallel_reduce_body = splittable<Body> &&
-                               requires( Body& body, const Range& range, Body&& rhs ) {
+                               requires( Body& body, const Range& range, Body& rhs ) {
                                    body(range);
                                    body.join(rhs);
                                };
--- a/include/oneapi/tbb/parallel_reduce.h
+++ b/include/oneapi/tbb/parallel_reduce.h
@@ -35,9 +35,9 @@ inline namespace d0 {
 
 template <typename Body, typename Range>
 concept parallel_reduce_body = splittable<Body> &&
-                               requires( Body& body, const Range& range, Body& rhs ) {
+                               requires( Body& body, const Range& range, Body&& rhs ) {
                                    body(range);
-                                   body.join(rhs);
+                                   body.join(std::move(rhs));
                                };
 
 template <typename Function, typename Range, typename Value>
@@ -390,14 +390,15 @@ public:
         , my_value(other.my_identity_element)
     { }
     void operator()(Range& range) {
-        my_value = tbb::detail::invoke(my_real_body, range, const_cast<const Value&>(my_value));
+        my_value = tbb::detail::invoke(my_real_body, range, std::move(my_value));
     }
+
     void join( lambda_reduce_body& rhs ) {
-        my_value = tbb::detail::invoke(my_reduction, const_cast<const Value&>(my_value),
-                                                     const_cast<const Value&>(rhs.my_value));
+        my_value = tbb::detail::invoke(my_reduction, std::move(my_value), std::move(rhs.my_value));
     }
-    Value result() const {
-        return my_value;
+
+    Value&& result() noexcept {
+        return std::move(my_value);
     }
 };
 
@@ -411,7 +412,7 @@ public:
     - \code Body::~Body(); \endcode                     Destructor
     - \code void Body::operator()( Range& r ); \endcode Function call operator applying body to range \c r
                                                         and accumulating the result
-    - \code void Body::join( Body& b ); \endcode        Join results.
+    - \code void Body::join( Body&& b ); \endcode        Join results.
                                                         The result in \c b should be merged into the result of \c this
 **/
 
@@ -514,7 +515,7 @@ Value parallel_reduce( const Range& range, const Value& identity, const RealBody
     lambda_reduce_body<Range,Value,RealBody,Reduction> body(identity, real_body, reduction);
     start_reduce<Range,lambda_reduce_body<Range,Value,RealBody,Reduction>,const __TBB_DEFAULT_PARTITIONER>
                           ::run(range, body, __TBB_DEFAULT_PARTITIONER() );
-    return body.result();
+    return std::move(body.result());
 }
 
 //! Parallel iteration with reduction and simple_partitioner.
@@ -527,7 +528,7 @@ Value parallel_reduce( const Range& range, const Value& identity, const RealBody
     lambda_reduce_body<Range,Value,RealBody,Reduction> body(identity, real_body, reduction);
     start_reduce<Range,lambda_reduce_body<Range,Value,RealBody,Reduction>,const simple_partitioner>
                           ::run(range, body, partitioner );
-    return body.result();
+    return std::move(body.result());
 }
 
 //! Parallel iteration with reduction and auto_partitioner
@@ -540,7 +541,7 @@ Value parallel_reduce( const Range& range, const Value& identity, const RealBody
     lambda_reduce_body<Range,Value,RealBody,Reduction> body(identity, real_body, reduction);
     start_reduce<Range,lambda_reduce_body<Range,Value,RealBody,Reduction>,const auto_partitioner>
                           ::run( range, body, partitioner );
-    return body.result();
+    return std::move(body.result());
 }
 
 //! Parallel iteration with reduction and static_partitioner
@@ -553,7 +554,7 @@ Value parallel_reduce( const Range& range, const Value& identity, const RealBody
     lambda_reduce_body<Range,Value,RealBody,Reduction> body(identity, real_body, reduction);
     start_reduce<Range,lambda_reduce_body<Range,Value,RealBody,Reduction>,const static_partitioner>
                                         ::run( range, body, partitioner );
-    return body.result();
+    return std::move(body.result());
 }
 
 //! Parallel iteration with reduction and affinity_partitioner
@@ -566,7 +567,7 @@ Value parallel_reduce( const Range& range, const Value& identity, const RealBody
     lambda_reduce_body<Range,Value,RealBody,Reduction> body(identity, real_body, reduction);
     start_reduce<Range,lambda_reduce_body<Range,Value,RealBody,Reduction>,affinity_partitioner>
                                         ::run( range, body, partitioner );
-    return body.result();
+    return std::move(body.result());
 }
 
 //! Parallel iteration with reduction, default partitioner and user-supplied context.
@@ -579,7 +580,7 @@ Value parallel_reduce( const Range& range, const Value& identity, const RealBody
     lambda_reduce_body<Range,Value,RealBody,Reduction> body(identity, real_body, reduction);
     start_reduce<Range,lambda_reduce_body<Range,Value,RealBody,Reduction>,const __TBB_DEFAULT_PARTITIONER>
                           ::run( range, body, __TBB_DEFAULT_PARTITIONER(), context );
-    return body.result();
+    return std::move(body.result());
 }
 
 //! Parallel iteration with reduction, simple partitioner and user-supplied context.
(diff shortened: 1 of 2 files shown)


include/oneapi/tbb/parallel_reduce.h
```diff
@@ -390,14 +390,15 @@ class lambda_reduce_body {
...
-        my_value = tbb::detail::invoke(my_reduction, const_cast<const Value&>(my_value),
-                                                     const_cast<const Value&>(rhs.my_value));
+        my_value = tbb::detail::invoke(my_reduction, std::move(my_value), std::move(rhs.my_value));
     }
-    Value result() const {
-        return my_value;
+
+    Value&& result() noexcept {
+        return std::move(my_value);
     }
```
BenFrantzDale, line 402: Would it be desirable to `&&` qualify this member function? As in
```cpp
[[nodiscard]] Value&& result() noexcept && {
    return std::move(my_value);
}
```
? 
That would force you to say `return std::move(body).result();` elsewhere, which may be clearer. (In general `body.result()` isn't obvious that it is extracting the result from `body`.)

include/oneapi/tbb/parallel_reduce.h
```diff
@@ -514,7 +515,7 @@ Value parallel_reduce( const Range& range, const Value& identity, const RealBody
     lambda_reduce_body<Range,Value,RealBody,Reduction> body(identity, real_body, reduction);
     start_reduce<Range,lambda_reduce_body<Range,Value,RealBody,Reduction>,const __TBB_DEFAULT_PARTITIONER>
                           ::run(range, body, __TBB_DEFAULT_PARTITIONER() );
-    return body.result();
+    return std::move(body.result());
```
BenFrantzDale, line 518: I think this `std::move` does nothing since `body.result()` is an rvalue?

test/conformance/conformance_parallel_reduce.cpp
```diff
@@ -56,6 +57,37 @@ struct ReduceBody {
     }
 };
 
+template <class T>
```
BenFrantzDale, line 60: `//!` explaining what this struct is for?

test/conformance/conformance_parallel_reduce.cpp
```diff
@@ -174,3 +206,42 @@ TEST_CASE("parallel_[deterministic_]reduce and std::invoke") {
...
+            new_vector.reserve(vector_of_vectors[index].size());
+            new_vector.insert(new_vector.end(), std::make_move_iterator(vector_of_vectors[index].begin()), std::make_move_iterator(vector_of_vectors[index].end()));
+        }
+        return new_vector;
+    };
+
+    auto join_body = [](vector_wrapper<int>&& x, vector_wrapper<int>&& y) {
+        vector_wrapper<int> new_vector = std::move(x);
+        new_vector.reserve(new_vector.size() + y.size());
+        new_vector.insert(new_vector.end(), std::make_move_iterator(y.begin()), std::make_move_iterator(y.end()));
```
BenFrantzDale, line 234: A move iterator over a vector of `int`s doesn't buy much. Consider for this example making
```cpp
struct MoveOnlyInt {
    int x;
    MoveOnlyInt() = default;
    explicit MoveOnlyInt(int x) : x{x} {}
    MoveOnlyInt(const MoveOnlyInt&) = delete;
    MoveOnlyInt(MoveOnlyInt&& other) : x{std::exchange(other.x, 0)} {}
    MoveOnlyInt& operator=(const MoveOnlyInt&) = delete;
    MoveOnlyInt& operator=(MoveOnlyInt&& other) {
        this->x = std::exchange(other.x, 0);
        return *this;
    }
};
```
and use vectors of `MoveOnlyInt` so the move iterators are required?

BenFrantzDale: Could also consider a test case using `std::list` or `std::set` that moves the nodes across, like
```cpp
auto join_body = [](std::set<int>&& x, std::set<int>&& y) {
   while (!y.empty()) {
       x.insert(y.extract(y.begin()));
   }
   return std::move(x);
}
```
(That could use `MoveOnlyInt` too.)

BenFrantzDale (commented):
Great to see someone looking at this again! A few suggestions.

include/oneapi/tbb/parallel_reduce.h
```diff
@@ -35,9 +35,9 @@ inline namespace d0 {
 
 template <typename Body, typename Range>
 concept parallel_reduce_body = splittable<Body> &&
-                               requires( Body& body, const Range& range, Body& rhs ) {
+                               requires( Body& body, const Range& range, Body&& rhs ) {
                                    body(range);
-                                   body.join(rhs);
+                                   body.join(std::move(rhs));
```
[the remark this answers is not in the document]
kboyarinov: Since this part of the reduce API was not changed at all, it was a mistake to change this concept. This line was removed.
`parallel_reduce_reduction` and `parallel_reduce_combine` concepts were changed instead.

include/oneapi/tbb/parallel_reduce.h
kboyarinov: My first thought on this question was removing the `&&` qualifier because as an internal structure, the body would be only move-from the result. But now I guess having this qualifier would be more clear from the move semantics perspective. Thanks for catching that.

include/oneapi/tbb/parallel_reduce.h
kboyarinov: Changed to `std::move(body).result()`, as discussed in the previous thread.

include/oneapi/tbb/parallel_reduce.h
BenFrantzDale: Yeah. An alternative is to give it a different name so it’s clearer that it moves out without the `std::move`, but just qualifying it like this feels reasonable.

test/conformance/conformance_parallel_reduce.cpp
BenFrantzDale: It would probably be good to do a test with a type that isn’t default constructible and/or isn’t movable or copyable in a node-based container — basically all the weird corner cases of irregular types.

test/conformance/conformance_parallel_reduce.cpp
kboyarinov: Done

test/conformance/conformance_parallel_reduce.cpp
kboyarinov: Unfortunately, we still cannot use move-only types for parallel_reduce since we need to copy the identity into each reduction "leaf". And using `std::vector<MoveOnlyInt>` or list would cause compilation errors even if the identity element is empty. 
I have changed the value type into the integer wrapper that is copyable, but the actual copy constructor should never be called.

Regarding adding new test cases, I think it makes no sense to test both "vector of vectors" and "vector of lists" use-cases. Splicing the lists while doing parallel_reduce is more interesting IMHO, so I left only the vector of lists test.

test/conformance/conformance_parallel_reduce.cpp
```diff
@@ -56,6 +57,67 @@ struct ReduceBody {
...
+    NeverCopyWrapper& operator=(NeverCopyWrapper&&) = default;
+
+    NeverCopyWrapper(const NeverCopyWrapper&) {
+        REQUIRE_MESSAGE(false, "Copy constructor of NeverCopyWrapper should never be called");
+    }
+
+    NeverCopyWrapper& operator=(const NeverCopyWrapper&) {
+        REQUIRE_MESSAGE(false, "Copy assignment of NeverCopyWrapper should never be called");
+        return *this;
+    }
```
BenFrantzDale, line 78: Wouldn't it be better to `=delete;` these?

test/conformance/conformance_parallel_reduce.cpp
```diff
@@ -174,3 +236,108 @@ TEST_CASE("parallel_[deterministic_]reduce and std::invoke") {
...
+    auto operator()(Args&&... args) const -> decltype(oneapi::tbb::parallel_reduce(std::forward<Args>(args)...)) {
+        return oneapi::tbb::parallel_reduce(std::forward<Args>(args)...);
+    }
+};
+
+struct DeterministicReduceRunner {
+    template <typename... Args>
+    auto operator()(Args&&... args) const -> decltype(oneapi::tbb::parallel_deterministic_reduce(std::forward<Args>(args)...)) {
+        return oneapi::tbb::parallel_deterministic_reduce(std::forward<Args>(args)...);
+    }
```
BenFrantzDale, line 306: Is the trailing return type helpful versus just leave it to `auto` (or `decltype(auto)`)?

BenFrantzDale (commented):
Looks good! I'll leave it to tbb maintainers to approve.

test/conformance/conformance_parallel_reduce.cpp
kboyarinov: We are forced to use trailing return type, because both `auto` with automatic deduction and `decltype(auto)` are C++14 features and we still need to support C++11.

test/conformance/conformance_parallel_reduce.cpp
kboyarinov: I have investigated one more time, and since we don't use `std::vector` directly, only using the wrapper, we can control that actual copy constructor would not be called by the container. So I removed `NeverCopyWrapper` and changed it to be `MoveOnlyWrapper`

test/conformance/conformance_parallel_reduce.cpp
BenFrantzDale: Gotcha. That's annoying.

pavelkumbrasev (approved):

pavelkumbrasev:
LGTM!

PR #1319, opened 2024-02-29
cmake apple framework support

### LINKED ISSUE #1252: building an iOS XCFramework for oneTBB
I'm trying to build an xcframework (and Swift Package) for oneTBB. I'm aware that tbb shouldn't be built as a static library due to singletons. ([this iOS port](https://github.com/yorkie/ios-tbb) builds tbb as a static library)

I've got this script: https://github.com/audulus/tbb-spm/blob/main/xcframework.sh

The script generates an xcframework out of dylibs, but alas [Apple says](https://developer.apple.com/documentation/xcode/creating-a-multi-platform-binary-framework-bundle): 

```Avoid using dynamic library files (.dylib files) for dynamic linking. An XCFramework can include dynamic library files, but only macOS supports these libraries for dynamic linking. Dynamic linking on iOS, watchOS, and tvOS requires the XCFramework to contain .framework bundles.```

So I need to create frameworks for both macOS and iOS (because I can't mix dylibs and frameworks in an XCFramework).

How should I approach that? When I generate Xcode projects with cmake, it requires code signing. Also seems like the target is still a dylib, not a framework.

So I suppose I simply have to assemble the frameworks according to apple's format with a bunch of shell commands (making the appropriate directories, copying in plists, setting up symlinks). I haven't found a tool that will assemble a framework from a dylib.

Thanks for any help!

### PR DESCRIPTION
**Pull request by geertbleyen**
### Description 
Add cmake option TBB_BUILD_APPLE_FRAMEWORKS to build oneTBB as Apple Frameworks.
This is needed to be able to create iOS compatible app bundles which need to embed TBB as framework, as using dylibs is prohibited by Apple.

Fixes #1252

### Type of change

### Tests

- [x] not needed

### Documentation

- [x] updated in this PR
- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

@isaevil 

### Other information
I've also ran into various integer type mismatches when building (mainly long <-> int) when I was working on this (using a M2 MacBook Pro). Not sure if I should include these fixes (just static_cast'ing to the expected types) in this PR or not?

### DIFF
--- a/cmake/README.md
+++ b/cmake/README.md
@@ -19,6 +19,7 @@ TBB_INSTALL_VARS:BOOL - Enable auto-generated vars installation(packages generat
 TBB_VALGRIND_MEMCHECK:BOOL - Enable scan for memory leaks using Valgrind (OFF by default)
 TBB_DISABLE_HWLOC_AUTOMATIC_SEARCH - Disable HWLOC automatic search by pkg-config tool (OFF by default)
 TBB_ENABLE_IPO - Enable Interprocedural Optimization (IPO) during the compilation (ON by default)
+TBB_BUILD_APPLE_FRAMEWORKS - Enable building Apple .frameworks instead of dylibs, only available on Apple platform. (OFF by default)
 ```
 
 ## Configure, Build, and Test
--- a/cmake/README.md
+++ b/cmake/README.md
@@ -19,7 +19,7 @@ TBB_INSTALL_VARS:BOOL - Enable auto-generated vars installation(packages generat
 TBB_VALGRIND_MEMCHECK:BOOL - Enable scan for memory leaks using Valgrind (OFF by default)
 TBB_DISABLE_HWLOC_AUTOMATIC_SEARCH - Disable HWLOC automatic search by pkg-config tool (OFF by default)
 TBB_ENABLE_IPO - Enable Interprocedural Optimization (IPO) during the compilation (ON by default)
-TBB_BUILD_APPLE_FRAMEWORKS - Enable building Apple .frameworks instead of dylibs, only available on Apple platform. (OFF by default)
+TBB_BUILD_APPLE_FRAMEWORKS - Enable the Apple* frameworks instead of dylibs, only available on the Apple platform. (OFF by default)
 ```
 
 ## Configure, Build, and Test
--- a/cmake/vars_utils.cmake
+++ b/cmake/vars_utils.cmake
@@ -29,8 +29,8 @@ macro(tbb_gen_vars target)
     if (NOT TBB_BUILD_APPLE_FRAMEWORKS)
         set(BIN_PATH $<TARGET_FILE_DIR:${target}>)
     else()
-        # For Apple frameworks, the binaries are placed in a framework bundle. 
-        # When using an Apple framework, you refer to the bundle, not the binary inside, so we take the bundle's path and go up one level.
+        # For Apple* frameworks, the binaries are placed in a framework bundle. 
+        # When using an Apple* framework, you refer to the bundle, not the binary inside, so we take the bundle's path and go up one level.
         # This path will then be used to generate the vars file, and the contents of the vars file will use the bundle's parent directory.
         set(BIN_PATH $<TARGET_BUNDLE_DIR:${target}>/..)
     endif()
--- a/cmake/vars_utils.cmake
+++ b/cmake/vars_utils.cmake
@@ -1,4 +1,4 @@
-# Copyright (c) 2020-2021 Intel Corporation
+# Copyright (c) 2020-2024 Intel Corporation
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
--- a/cmake/vars_utils.cmake
+++ b/cmake/vars_utils.cmake
@@ -26,12 +26,20 @@ get_filename_component(TBB_VARS_TEMPLATE_NAME ${PROJECT_SOURCE_DIR}/integration/
 string(REPLACE ".in" "" TBB_VARS_NAME ${TBB_VARS_TEMPLATE_NAME})
 
 macro(tbb_gen_vars target)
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

**Comment by geertbleyen** [comment]
Oopsie, will sort out the correct CI failures first thing tomorrow.

**Comment by geertbleyen** [comment]
Missing copyright updates in the edited files is fixed now.

**Comment by geertbleyen** [comment]
It appears VS code auto-formatted and broke it :(.
Pushed a new commit to rollback those unintended changes.

File: cmake/README.md [hunk 1]
```diff
@@ -19,6 +19,7 @@ TBB_INSTALL_VARS:BOOL - Enable auto-generated vars installation(packages generat
 TBB_VALGRIND_MEMCHECK:BOOL - Enable scan for memory leaks using Valgrind (OFF by default)
 TBB_DISABLE_HWLOC_AUTOMATIC_SEARCH - Disable HWLOC automatic search by pkg-config tool (OFF by default)
 TBB_ENABLE_IPO - Enable Interprocedural Optimization (IPO) during the compilation (ON by default)
+TBB_BUILD_APPLE_FRAMEWORKS - Enable building Apple .frameworks instead of dylibs, only available on Apple platform. (OFF by default)
```
aepanchi, line 22:
```suggestion
TBB_BUILD_APPLE_FRAMEWORKS - Enable the Apple* frameworks instead of dylibs, only available on the Apple platform. (OFF by default)
```

test/common/utils_dynamic_libs.h
```diff
@@ -46,9 +46,17 @@ namespace utils {
 #endif
 #define EXT ".dll"
 #else
+#if TBB_USE_APPLE_FRAMEWORKS
+#define PREFIX // When build as Apple Framework, the binary has no lib prefix
```
aepanchi, line 50:
```suggestion
#define PREFIX // When built as Apple* Framework, the binary has no lib prefix
```

test/common/utils_dynamic_libs.h
```diff
@@ -46,9 +46,17 @@ namespace utils {
...
+#else
 #define PREFIX "lib"
+#endif
 #if __APPLE__
+#if TBB_USE_APPLE_FRAMEWORKS
+#define EXT // When build as Apple Framework, the binary has no lib prefix
```
aepanchi, line 56:
```suggestion
#define EXT // When built as Apple* Framework, the binary has no lib prefix
```

cmake/README.md
geertbleyen: @aepanchi I'm not sure why Apple needs a asterisk. Is it to mark it as copyright? If so, shouldn't the 2nd Apple in this string also have an asterisk?
Tbh, I'm also not 100% about the 'the' in front of Apple in this sentence. But then again, I'm not a native English speaker :).

aepanchi: @geertbleyen Yes, it is for the copyright purposes. Just one for the first mention in the file is enough

geertbleyen: Ok, fixed.

test/common/utils_dynamic_libs.h
geertbleyen: Fixed.

geertbleyen:
@isaevil Is anything else required from me to get this PR completed?
I'm waiting for this PR to complete to also update the onetbb conan recipe on conan center index to make this option for Apple frameworks available to Conan users.

geertbleyen:
Hi @isaevil , any progress on this PR?
Any more info or effort required here?

src/tbb/CMakeLists.txt
```diff
@@ -181,6 +191,7 @@ if (TBB_INSTALL)
             COMPONENT devel)
 endif()
 
-if (COMMAND tbb_gen_vars)
+# not useful when bundled into a framework
+if (COMMAND tbb_gen_vars AND NOT TBB_BUILD_APPLE_FRAMEWORKS)
```
isaevil, line 195: Could you please elaborate a bit more in this comment, why vars scripts are not useful here?

isaevil (commented):
Hi, sorry for late review. Generally, it looks good, left some minor comment.

src/tbb/CMakeLists.txt
geertbleyen: In hindsight, there is actually no harm/problem in doing this, but I do have to fix where the vars.sh file gets generated when using Apple frameworks, as the binary itself is located inside a framework, and the vars.sh file should be besides the framework bundle.
I've undone this change, and fixed the tbb_gen_vars macro to place the vars file correctly when using Apple frameworks (and that the variables in the vars.sh file don't point inside the tbb framework file structure).

cmake/vars_utils.cmake
isaevil, line 1: Please update the copyright year here as well.

cmake/vars_utils.cmake
```diff
@@ -26,12 +26,20 @@ get_filename_component(TBB_VARS_TEMPLATE_NAME ${PROJECT_SOURCE_DIR}/integration/
 string(REPLACE ".in" "" TBB_VARS_NAME ${TBB_VARS_TEMPLATE_NAME})
 
 macro(tbb_gen_vars target)
+    if (NOT TBB_BUILD_APPLE_FRAMEWORKS)
+        set(BIN_PATH $<TARGET_FILE_DIR:${target}>)
+    else()
+        # For Apple frameworks, the binaries are placed in a framework bundle. 
+        # When using an Apple framework, you refer to the bundle, not the binary inside, so we take the bundle's path and go up one level.
+        # This path will then be used to generate the vars file, and the contents of the vars file will use the bundle's parent directory.
```
isaevil, line 34: Following approach as @aepanchi suggested in previous comments.
```suggestion
        # For Apple* frameworks, the binaries are placed in a framework bundle. 
        # When using an Apple* framework, you refer to the bundle, not the binary inside, so we take the bundle's path and go up one level.
        # This path will then be used to generate the vars file, and the contents of the vars file will use the bundle's parent directory.
```

isaevil (approved):
LGTM

Issue #1343, opened 2024-04-13
doc/README.md is misleading


meator:
I wasn't able to build documentation of oneTBB while following [doc/README.md](https://github.com/oneapi-src/oneTBB/blob/master/doc/README.md).

Here is what I've tried:
[![asciicast](https://asciinema.org/a/654012.svg)](https://asciinema.org/a/654012)

(Here is a plaintext version: [654012.txt](https://github.com/oneapi-src/oneTBB/files/14967972/654012.txt))

The Makefile looks like it's in a wrong directory or it's inspecting wrong ones. There also seems to be a Doxygen file. Does this project use multiple documentation systems?

omalyshe:
@meator  doc/README.md describes how to build oneTBB documentation in https://oneapi-src.github.io/oneTBB/ format. 
It doesn't use Doxygen but use Sphinx. Doxygen file is used internally to generate oneTBB Test Specification. 
Regarding your concern about incorrect Makefile location - we'll check it.

aepanchi:
@meator thanks for raising the issue! https://github.com/oneapi-src/oneTBB/pull/1351 should fix it.

PR #1344, opened 2024-04-13
pool.py: Fix invalid escape sequences

meator:
### Description 
When running

```
python -m compileall -q python/
```

the following warnings are displayed:

```
python/tbb/pool.py:91: SyntaxWarning: invalid escape sequence '\p'
  """
python/tbb/pool.py:270: SyntaxWarning: invalid escape sequence '\p'
  """
python/tbb/pool.py:319: SyntaxWarning: invalid escape sequence '\p'
  """
python/tbb/pool.py:406: SyntaxWarning: invalid escape sequence '\p'
  """
python/tbb/pool.py:413: SyntaxWarning: invalid escape sequence '\p'
  """Used to identify which results we're waiting for. Will
python/tbb/pool.py:422: SyntaxWarning: invalid escape sequence '\p'
  """Called by the ApplyResult object (already registered via
python/tbb/pool.py:431: SyntaxWarning: invalid escape sequence '\p'
  """Called by the CollectorIterator object to retrieve the
python/tbb/pool.py:453: SyntaxWarning: invalid escape sequence '\p'
  """\param AbstractResultCollector instance"""
python/tbb/pool.py:488: SyntaxWarning: invalid escape sequence '\p'
  """
python/tbb/pool.py:498: SyntaxWarning: invalid escape sequence '\p'
  """Used to identify which results we're waiting for. Will
python/tbb/pool.py:507: SyntaxWarning: invalid escape sequence '\p'
  """Called by the CollectorIterator object to retrieve the
python/tbb/pool.py:535: SyntaxWarning: invalid escape sequence '\p'
  """Called by the ApplyResult object (already registered via
python/tbb/pool.py:562: SyntaxWarning: invalid escape sequence '\p'
  """
python/tbb/pool.py:577: SyntaxWarning: invalid escape sequence '\p'
  """Used to identify which results we're waiting for. Will
python/tbb/pool.py:587: SyntaxWarning: invalid escape sequence '\p'
  """Called by the CollectorIterator object to retrieve the
python/tbb/pool.py:600: SyntaxWarning: invalid escape sequence '\p'
  """Called by the ApplyResult object (already registered via
```

These are caused by the usage of `\param` inside Python docstrings. It looks like Doxygen is used for documentation. `\` commands shouldn't really be used in Python. Doxygen provides alternative syntax for this, `@command`. I believe that `@param` should be used instead. I would have made this change, but I would have to be able to test it out by building the documentation. I was unfortunately not able to do so: #1343, so I chose to switch `\param` to `\\param` to avoid any breaking changes (which probably wouldn't happen, but I can't test it, so I must assume that they will).

If someone else with more experience with this project wants to switch to `@param`, please do so.

- [x] - git commit message contains an appropriate signed-off-by string

### Type of change

- [x] bug fix
- [ ] new feature
- [ ] tests
- [ ] infrastructure
- [x] documentation

### Tests

- [ ] added
- [x] not needed

### Documentation

- [ ] updated in #
- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/python/tbb/pool.py
+++ b/python/tbb/pool.py
@@ -89,8 +89,8 @@ class Pool(object):
 
     def __init__(self, nworkers=0, name="Pool"):
         """
-        \param nworkers (integer) number of worker threads to start
-        \param name (string) prefix for the worker threads' name
+        :param nworkers (integer) number of worker threads to start
+        :param name (string) prefix for the worker threads' name
         """
         self._closed = False
         self._tasks = task_group()
@@ -268,8 +268,8 @@ class Job:
 
     def __init__(self, func, args, kwds, apply_result):
         """
-        \param func/args/kwds used to call the function
-        \param apply_result ApplyResult object that holds the result
+        :param func/args/kwds used to call the function
+        :param apply_result ApplyResult object that holds the result
         of the function call
         """
         self._func = func
@@ -317,10 +317,10 @@ class ApplyResult(object):
 
     def __init__(self, collector=None, callback=None):
         """
-        \param collector when not None, the notify_ready() method of
+        :param collector when not None, the notify_ready() method of
         the collector will be called when the result from the Job is
         ready
-        \param callback when not None, function to call when the
+        :param callback when not None, function to call when the
         result becomes available (this is the parameter passed to the
         Pool::*_async() methods.
         """
@@ -404,7 +404,7 @@ class AbstractResultCollector(object):
 
     def __init__(self, to_notify):
         """
-        \param to_notify ApplyResult object to notify when all the
+        :param to_notify ApplyResult object to notify when all the
         results we're waiting for become available. Can be None.
         """
         self._to_notify = to_notify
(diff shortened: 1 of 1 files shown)


omalyshe:
@meator What Python version do you use? I don't see any warnings with 3.10.12 (on Ubuntu 22.04).

meator:
I'm using Python version 3.12.2. Even if warnings aren't shown (which they are), using invalid escape sequences such as `\p` can lead to problems down the line.

python/tbb/pool.py
```diff
@@ -89,8 +89,8 @@ class Pool(object):
 
     def __init__(self, nworkers=0, name="Pool"):
         """
-        \param nworkers (integer) number of worker threads to start
-        \param name (string) prefix for the worker threads' name
+        \\param nworkers (integer) number of worker threads to start
```
omalyshe, line 92: Can you consider using sphinx-style syntax like in https://github.com/oneapi-src/oneTBB/blob/master/python/tbb/__init__.py#L172? 
\param -> :param

meator: I am more inclined to `@param` as described above, but I can change it to a colon.

omalyshe (approved):
LGTM.

PR #1346, opened 2024-04-15
Fix control variable initialization

### LINKED ISSUE #1341: Global task_schedule_observer leads to error on WASM and static lib x86
The following program:

```C++
#include <tbb/task_scheduler_observer.h>

class dummy_observer final : public tbb::task_scheduler_observer {
 public:
  dummy_observer() : tbb::task_scheduler_observer() {
    observe(true);  // activates the observer
  }

  void on_scheduler_entry(bool worker) override {}

  void on_scheduler_exit(bool worker) override {}
};

dummy_observer global_observer;

int main(int argc, char** argv) { return 0; }
```

When compiled with `em++` version 3.1.56 and run with node 19.9.0 yields:

```
/home/brian/Dev/cpp/tinystan/test.js:147
      throw ex;
      ^

RuntimeError: table index is out of bounds
    at test.wasm.tbb::detail::r1::threading_control::create_threading_control() (wasm://wasm/test.wasm-00c65a2a:wasm-function[846]:0x213be)
    at test.wasm.tbb::detail::r1::threading_control::register_public_reference() (wasm://wasm/test.wasm-00c65a2a:wasm-function[848]:0x21a6f)
    at test.wasm.tbb::detail::r1::governor::init_external_thread() (wasm://wasm/test.wasm-00c65a2a:wasm-function[485]:0xcfb4)
    at test.wasm.tbb::detail::r1::observe(tbb::detail::d1::task_scheduler_observer&, bool) (wasm://wasm/test.wasm-00c65a2a:wasm-function[481]:0xc700)
    at test.wasm.tbb::detail::d1::task_scheduler_observer::observe(bool) (wasm://wasm/test.wasm-00c65a2a:wasm-function[35]:0x119a)
    at test.wasm.dummy_observer::dummy_observer() (wasm://wasm/test.wasm-00c65a2a:wasm-function[32]:0xfe1)
    at test.wasm.__cxx_global_var_init (wasm://wasm/test.wasm-00c65a2a:wasm-function[31]:0xf36)
    at test.wasm._GLOBAL__sub_I_test.cpp (wasm://wasm/test.wasm-00c65a2a:wasm-function[56]:0x1848)
    at test.wasm.__wasm_call_ctors (wasm://wasm/test.wasm-00c65a2a:wasm-function[24]:0xe60)
    at callRuntimeCallbacks (/home/brian/Dev/cpp/tinystan/test.js:1519:26)

Node.js v19.9.0
```

SoilRos:
### Description 
Fixes initialization order bug stated in https://github.com/oneapi-src/oneTBB/issues/1341#issuecomment-2054095330

Fixes #1341

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

@pavelkumbrasev

### DIFF
--- a/src/tbb/global_control.cpp
+++ b/src/tbb/global_control.cpp
@@ -1,5 +1,5 @@
 /*
-    Copyright (c) 2005-2023 Intel Corporation
+    Copyright (c) 2005-2024 Intel Corporation
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
@@ -17,6 +17,7 @@
 #include "oneapi/tbb/detail/_config.h"
 #include "oneapi/tbb/detail/_template_helpers.h"
 
+#include "oneapi/tbb/cache_aligned_allocator.h"
 #include "oneapi/tbb/global_control.h"
 #include "oneapi/tbb/tbb_allocator.h"
 #include "oneapi/tbb/spin_mutex.h"
@@ -139,11 +140,22 @@ class alignas(max_nfs_size) lifetime_control : public control_storage {
     }
 };
 
-static allowed_parallelism_control allowed_parallelism_ctl;
-static stack_size_control stack_size_ctl;
-static terminate_on_exception_control terminate_on_exception_ctl;
-static lifetime_control lifetime_ctl;
-static control_storage *controls[] = {&allowed_parallelism_ctl, &stack_size_ctl, &terminate_on_exception_ctl, &lifetime_ctl};
+static control_storage* controls[] = {nullptr, nullptr, nullptr, nullptr};
+
+void global_control_acquire() {
+    controls[0] = new (cache_aligned_allocate(sizeof(allowed_parallelism_control))) allowed_parallelism_control{};
+    controls[1] = new (cache_aligned_allocate(sizeof(stack_size_control))) stack_size_control{};
+    controls[2] = new (cache_aligned_allocate(sizeof(terminate_on_exception_control))) terminate_on_exception_control{};
+    controls[3] = new (cache_aligned_allocate(sizeof(lifetime_control))) lifetime_control{};
+}
+
+void global_control_release() {
+    for (auto& ptr : controls) {
+        ptr->~control_storage();
+        cache_aligned_deallocate(ptr);
+        ptr = nullptr;
+    }
+}
 
 void global_control_lock() {
     for (auto& ctl : controls) {
--- a/src/tbb/governor.cpp
+++ b/src/tbb/governor.cpp
@@ -1,5 +1,5 @@
 /*
-    Copyright (c) 2005-2023 Intel Corporation
+    Copyright (c) 2005-2024 Intel Corporation
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
@@ -42,6 +42,8 @@ namespace detail {
 namespace r1 {
 
 void clear_address_waiter_table();
+void global_control_acquire();
+void global_control_release();
 
 //! global_control.cpp contains definition
 bool remove_and_check_if_empty(d1::global_control& gc);
@@ -60,6 +62,7 @@ namespace system_topology {
 //------------------------------------------------------------------------
 
 void governor::acquire_resources () {
+    global_control_acquire();
 #if __TBB_USE_POSIX
     int status = theTLS.create(auto_terminate);
 #else
@@ -85,6 +88,7 @@ void governor::release_resources () {
 
     system_topology::destroy();
     dynamic_unlink_all();
+    global_control_release();
 }
 
 rml::tbb_server* governor::create_rml_server ( rml::tbb_client& client ) {


src/tbb/global_control.cpp
```diff
@@ -139,13 +141,20 @@ class alignas(max_nfs_size) lifetime_control : public control_storage {
     }
 };
 
-static allowed_parallelism_control allowed_parallelism_ctl;
-static stack_size_control stack_size_ctl;
-static terminate_on_exception_control terminate_on_exception_ctl;
-static lifetime_control lifetime_ctl;
-static control_storage *controls[] = {&allowed_parallelism_ctl, &stack_size_ctl, &terminate_on_exception_ctl, &lifetime_ctl};
+static std::array<std::unique_ptr<control_storage>, 4> controls;
+static std::once_flag control_storage_flag;
```
kboyarinov, line 145: We have discussed the solution briefly within the team and it seems like it can result in more undefined behavior since `std::unique_ptr` and `std::once_flag` are not trivial structures and it can be unsafe to access them before initialization (in particular if `global_control_init()` would be called before constructing the `controls` and `control_storage_flag`.

We propose to have global static pointer to the storage containing all required global control variables and dynamically allocate it inside `governor::acquire_resources()` (which would be called during the first access to TBB) and deallocate inside `governor::release_resources ()`.
@pavelkumbrasev, feel free to add if I missed something.

What do you think about the proposed change?

SoilRos: I am _very_ unfamiliar with TBB internal code. I suspected that this solution may introduce more initialization problems as you pointed out, but I did not know where TBB initialized and destroyed things... so `unique_ptr` was the "easy" answer. With your hint we can give it a try on the `governor::[acquire|release]_resources` functions.

SoilRos:
Previous commit fails due to C++14 compatibility on `new` alignment. Now this should hopefully be fixed.

src/tbb/global_control.cpp
```diff
@@ -149,17 +152,20 @@ static control_storage_instance* control_storage_instance_ptr = nullptr;
 static control_storage *controls[4] = {nullptr, nullptr, nullptr, nullptr};
 
 void global_control_acquire() {
-    control_storage_instance_ptr = new control_storage_instance{};
+    auto ptr = aligned_alloc(max_nfs_size, sizeof(control_storage_instance));
```
phprus, line 155: MSVC does not support ``aligned_alloc``/``free``. Use ``_aligned_malloc`` and ``_aligned_free`` instead.

kboyarinov: I suggest to use special functions: `cache_aligned_allocate()` and `cache_aligned_deallocate()`

src/tbb/global_control.cpp
```diff
@@ -139,11 +142,31 @@ class alignas(max_nfs_size) lifetime_control : public control_storage {
...
+    controls[2] = &control_storage_instance_ptr->terminate_on_exception_ctl;
+    controls[3] = &control_storage_instance_ptr->lifetime_ctl;
+}
+
+void global_control_release() {
+    for (auto& ptr : controls) {
+        ptr = nullptr;
+    }
+    control_storage_instance_ptr->~control_storage_instance();
+    free(std::exchange(control_storage_instance_ptr, nullptr));
```
pavelkumbrasev, line 168: `std::exchange` is not available in C++11

src/tbb/global_control.cpp
```diff
@@ -139,11 +142,31 @@ class alignas(max_nfs_size) lifetime_control : public control_storage {
...
-static lifetime_control lifetime_ctl;
-static control_storage *controls[] = {&allowed_parallelism_ctl, &stack_size_ctl, &terminate_on_exception_ctl, &lifetime_ctl};
+struct alignas(max_nfs_size) control_storage_instance {
+    allowed_parallelism_control     allowed_parallelism_ctl;
+    stack_size_control              stack_size_ctl;
+    terminate_on_exception_control  terminate_on_exception_ctl;
+    lifetime_control                lifetime_ctl;
+};
+static control_storage_instance* control_storage_instance_ptr = nullptr;
+static control_storage *controls[4] = {nullptr, nullptr, nullptr, nullptr};
```
pavelkumbrasev, line 152: Do we really need an extra variable + extra class? Can we just use pointer to array?

SoilRos: It’s not really needed, I was trying to follow @kboyarinov’s instructions but I guess I understood them wrongly. So you meant that each control variable is allocated separately, right?

One argument for the struct is that it only needs one allocation so data is more compact.   With individual allocations you have to hard-code the derived class twice too anyways (construction/destruction). On the other hand, one can argue that the struct is slightly less maintainable, so I’d understand if you still prefer one allocation per class.

SoilRos: Nvm, I just realized that every class sits in it’s own cache line so the “compact” argument doesn’t really hold…

src/tbb/global_control.cpp
```diff
@@ -27,6 +28,7 @@
 #include "misc.h"
 
 #include <atomic>
+#include <new>
```
pavelkumbrasev, line 31: Do we really need this?

src/tbb/global_control.cpp
```diff
@@ -139,11 +141,22 @@ class alignas(max_nfs_size) lifetime_control : public control_storage {
     }
 };
 
-static allowed_parallelism_control allowed_parallelism_ctl;
-static stack_size_control stack_size_ctl;
-static terminate_on_exception_control terminate_on_exception_ctl;
-static lifetime_control lifetime_ctl;
-static control_storage *controls[] = {&allowed_parallelism_ctl, &stack_size_ctl, &terminate_on_exception_ctl, &lifetime_ctl};
+static control_storage *controls[4] = {nullptr, nullptr, nullptr, nullptr};
```
pavelkumbrasev, line 144:
```suggestion
static control_storage* controls[] = {nullptr, nullptr, nullptr, nullptr};
```

src/tbb/global_control.cpp
```diff
@@ -17,6 +17,7 @@
 #include "oneapi/tbb/detail/_config.h"
```
kboyarinov, line 17: Please update the last copyright year to 2024 for both files to satisfy the copyright check.

pavelkumbrasev (commented):
Overall the patch looks good!

kboyarinov (commented):
Excluding the copyright comment - LGTM

SoilRos:
I addressed your reviews and squashed the commits, Thanks!

I was thinking that we may add a test for this with a global observer as presented in the original issue #1341. I just do not know where to put it. Do you have a hint where it can go? Or do you think it's not necessary?

pavelkumbrasev:
> I was thinking that we may add a test for this with a global observer as presented in the original issue #1341. I just do not know where to put it. Do you have a hint where it can go? Or do you think it's not necessary?

As I understand we cannot reproduce this failure without compiling TBB statically.  I don't know if we previously added separate tests for a specific platform.

SoilRos:
Alright, then this is ready from my side. I just amended my commit with the copyright comment (I hope it's correct). There is no changes in the code since the last CI run.

pavelkumbrasev (approved):

pavelkumbrasev:
> Maybe add tests for static build?

It might make sense since WASM compile TBB statically. Do you want to contribute it ? 😄

kboyarinov (approved):
LGTM

phprus:
> > Maybe add tests for static build?
> 
> It might make sense since WASM compile TBB statically. Do you want to contribute it ? 😄

PR: #1350

PR #1352, opened 2024-04-19
Add task arena worker block time

dnmokhov:
### Description 
Add `threading_control::is_any_other_client_active()` to verify if other arenas are requesting workers.

Add hybrid CPU detection at initialization.

On non-hybrid CPUs, make workers block for 1ms before leaving the arena, unless
- more work becomes available in the arena (&rArr; unblock and stay in the arena);
- other arenas are requesting workers (&rArr; unblock and leave the arena immediately).

Fixes performance degradation on large systems when CPU power state demotion is disabled.

### Type of change

### Tests

- [x] not needed (done manually)

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/src/tbb/arena.cpp
+++ b/src/tbb/arena.cpp
@@ -1,5 +1,5 @@
 /*
-    Copyright (c) 2005-2023 Intel Corporation
+    Copyright (c) 2005-2024 Intel Corporation
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
@@ -395,9 +395,11 @@ bool arena::is_top_priority() const {
     return my_is_top_priority.load(std::memory_order_relaxed);
 }
 
-bool arena::try_join() {
+bool arena::try_join(bool should_join) {
     if (num_workers_active() < my_num_workers_allotted.load(std::memory_order_relaxed)) {
-        my_references += arena::ref_worker;
+        if (should_join) {
+            my_references += arena::ref_worker;
+        }
         return true;
     }
     return false;
--- a/src/tbb/arena.cpp
+++ b/src/tbb/arena.cpp
@@ -395,11 +395,9 @@ bool arena::is_top_priority() const {
     return my_is_top_priority.load(std::memory_order_relaxed);
 }
 
-bool arena::try_join(bool should_join) {
-    if (num_workers_active() < my_num_workers_allotted.load(std::memory_order_relaxed)) {
-        if (should_join) {
-            my_references += arena::ref_worker;
-        }
+bool arena::try_join() {
+    if (is_joinable()) {
+        my_references += arena::ref_worker;
         return true;
     }
     return false;
--- a/src/tbb/thread_dispatcher.cpp
+++ b/src/tbb/thread_dispatcher.cpp
@@ -1,5 +1,5 @@
 /*
-    Copyright (c) 2022-2023 Intel Corporation
+    Copyright (c) 2022-2024 Intel Corporation
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
@@ -131,7 +131,7 @@ bool thread_dispatcher::is_client_alive(thread_dispatcher_client* client) {
     return false;
 }
 
-thread_dispatcher_client* thread_dispatcher::client_in_need(client_list_type* clients, thread_dispatcher_client* hint) {
+thread_dispatcher_client* thread_dispatcher::client_in_need(client_list_type* clients, thread_dispatcher_client* hint, bool should_join) {
     // TODO: make sure client with higher priority returned only if there are available slots in it.
     hint = select_next_client(hint);
     if (!hint) {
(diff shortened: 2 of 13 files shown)


src/tbb/thread_request_serializer.cpp
```diff
@@ -37,16 +37,16 @@ void thread_request_serializer::update(int delta) {
...
+        delta = limit_delta(delta, my_soft_limit, my_total_request.load(std::memory_order_relaxed));
         my_thread_dispatcher.adjust_job_count_estimate(delta);
     }
 }
 
 void thread_request_serializer::set_active_num_workers(int soft_limit) {
     mutex_type::scoped_lock lock(my_mutex);
     int delta = soft_limit - my_soft_limit;
-    delta = limit_delta(delta, my_total_request, soft_limit);
+    delta = limit_delta(delta, my_total_request.load(std::memory_order_relaxed), soft_limit);
```
pavelkumbrasev, line 49: Happens-before relationships are not built on top of `my_total_request`, thus all the operations should be `std::memory_order_relaxed`.

src/tbb/thread_request_serializer.h
```diff
@@ -39,6 +39,7 @@ class thread_request_serializer : public thread_request_observer {
 public:
     thread_request_serializer(thread_dispatcher& td, int soft_limit);
     void set_active_num_workers(int soft_limit);
+    int num_workers_requested() { return my_total_request.load(std::memory_order_acquire); }
```
pavelkumbrasev, line 42: Happens-before relationships are not built on top of `my_total_request`, thus all the operations should be `std::memory_order_relaxed`.

pavelkumbrasev (commented):
Overall the patch looks good.

src/tbb/thread_request_serializer.h
dnmokhov: Changed.

src/tbb/thread_request_serializer.cpp
dnmokhov: Changed.

src/tbb/arena.cpp
```diff
@@ -395,9 +395,11 @@ bool arena::is_top_priority() const {
     return my_is_top_priority.load(std::memory_order_relaxed);
 }
 
-bool arena::try_join() {
+bool arena::try_join(bool should_join) {
     if (num_workers_active() < my_num_workers_allotted.load(std::memory_order_relaxed)) {
-        my_references += arena::ref_worker;
+        if (should_join) {
+            my_references += arena::ref_worker;
+        }
```
isaevil, line 402: That functions now looks contradictory to itself. It basically said "try join arena by adding reference, if possible", but now there is a boolean argument that says to not do so which basically acts like some sort of `joinable()` function. 

Following my previous comment, shouldn't we introduce `arena::joinable()` with following implementation:
```cpp
bool arena::joinable() {
    return num_workers_active() < my_num_workers_allotted.load(std::memory_order_relaxed);
}
```
Thus implementation of `try_join` would be the following:
```cpp
bool arena::try_join() {
    if (joinable()) {
        my_references += arena::ref_worker;
        return true;
    }
    return false;
}
```

src/tbb/thread_dispatcher.cpp
```diff
@@ -149,19 +149,23 @@ thread_dispatcher_client* thread_dispatcher::client_in_need(client_list_type* cl
...
     if (is_client_alive(prev)) {
-        return client_in_need(my_client_list, prev);
+        return client_in_need(my_client_list, prev, should_join);
     }
-    return client_in_need(my_client_list, my_next_client);
+    return client_in_need(my_client_list, my_next_client, should_join);
+}
+
+bool thread_dispatcher::is_any_client_in_need() {
+    return client_in_need(nullptr, /* should_join = */ false) != nullptr;
```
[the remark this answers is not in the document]
pavelkumbrasev: But it will result in code duplication because it will be the same client list traverse approach.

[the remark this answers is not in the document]
isaevil: But is the same traverse approach really needed in that case? This function name clearly implies that it just needs to find out if **any** arena is joinable. But `client_in_need` function, besides that, tries to find arena with higher priority (which, after ). 
Current traverse logic is already a bit complicated (as for me) and in this function it served specifically for one purpose but boolean flag makes it double-purpose and introduces more complication (and confusion by `try_join(should_join)`).
Can't we use something like `std::any_of` for these 3 lists?
```cpp
bool thread_dispatcher::is_any_client_in_need() {
    for (int idx = 0; idx < num_priority_levels; ++idx) {
        bool is_in_need = std::any_of(my_client_list[idx].begin(), my_client_list[idx].end(), [] (const thread_dispatcher_client& client) {
            return client.joinable();
        });
        if (is_in_need) {
            return true;
        }
    }
    return false;
}
```
Yes, it is a slight code duplication, but is not very dramatic.

src/tbb/arena.cpp
dnmokhov: Makes sense. I added `arena::is_joinable()`.

src/tbb/thread_dispatcher.cpp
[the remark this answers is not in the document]
dnmokhov: I updated `thread_dispatcher::is_any_client_in_need()` to be similar to `thread_dispatcher::is_client_alive()`.

pavelkumbrasev (approved):
LGTM!

isaevil (approved):
LGTM

PR #1373, opened 2024-05-01
Ensure test_partitioner reliability by waiting between repeats for threads to leave

dnmokhov:
### Description 
The `test_partitioner` test was previously relying on worker threads leaving the arena "soon enough" between each `parallel_for`. Now we need to wait, because they block for 1ms before leaving,

Fixes sporadic `test_partitioner` hangs.

### Type of change

### Tests
- [x] not needed

### Documentation
- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

@omalyshe

### Other information

### DIFF
--- a/test/tbb/test_partitioner.cpp
+++ b/test/tbb/test_partitioner.cpp
@@ -101,6 +101,9 @@ template <typename PerBodyFunc> float test(PerBodyFunc&& body) {
                 tbb::static_partitioner()
             );
         });
+        // To avoid tasks stealing in the beginning of the parallel algorithm, the test waits for
+        // the threads to leave the arena, so that on the next iteration they have tasks assigned
+        // in their mailboxes and, thus, don't need to search for work to do in other task pools.
         observer.wait_leave();
     }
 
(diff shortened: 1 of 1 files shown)


isaevil (dismissed):
LGTM

test/tbb/test_partitioner.cpp
```diff
@@ -77,8 +101,7 @@ template <typename PerBodyFunc> float test(PerBodyFunc&& body) {
                 tbb::static_partitioner()
             );
         });
-        // TODO:
-        //   - Consider introducing an observer to guarantee the threads left the arena.
+        observer.wait_leave();
```
aleksei-fedotov, line 104: I would add the comment so that future readers better understand why the observer was introduced. Something like "_To avoid tasks stealing at a beginning of the parallel algorithm, the test waits when the threads leave the arena so that on the next iteration they have tasks assigned in their mailboxes and, thus, don't need to search for a work to do in other task pools_"

aleksei-fedotov (approved):
Looks good to me. Consider adding comment though.

test/tbb/test_partitioner.cpp
dnmokhov: Added.

pavelkumbrasev (approved):

PR #1386, opened 2024-05-23
Fix `task_alignment` for ARM

blonded04:
### Description 
`task_alignment` seem to be (as `git blame` says) added to _task.h in 2019 when moving from TBB to oneTBB. However, it happens so that `task_alignment` is 64 bytes by default (x86 cacheline), which doesn't make any sense for ARM architectures with 128 bytes cacheline.

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [ ] No
- [x] Unknown


include/oneapi/tbb/detail/_task.h
```diff
@@ -220,7 +224,11 @@ class alignas(task_alignment) task : public task_traits {
     virtual task* cancel(execution_data&) = 0;
 
 private:
+#if __ARM_ARCH_7A__ || __aarch64__
+    std::uint64_t m_reserved[14]{};
```
blonded04, line 228: i decided to put `#if` here, instead of `(task_alignment - sizeof(void*) - sizeof(task_traits)) / sizeof(std::uint64_t)`, because it gets weird on architectures where pointer isn't 8 bytes? idk, this seems cleaner than it was before

pavelkumbrasev:
This patch might introduce backward compatibility problems so it should be carefully investigated.
Have you seen any performance improvements with this change?

blonded04:
> This patch might introduce backward compatibility problems so it should be carefully investigated. Have you seen any performance improvements with this change?

I tried to come up with some benchmarks with different variations of system topology, however, no difference is spotted :(

PR #1391, opened 2024-06-01
Don't add -mrtm when building w/ GNU + Emscripten

ochafik:
Similar to the check done in https://github.com/oneapi-src/oneTBB/blob/d8c53ce141bada427a6eb56fc2fd43d7a210980e/cmake/compilers/Clang.cmake#L57
in GNU.cmake the `-mrtm` flag shouldn't be added when building w/ Emscripten.

I ran into this while trying to compile https://github.com/elalish/manifold w/ pyodide-build (as an out-of-tree package https://pyodide.org/en/stable/development/building-and-testing-packages.html).


isaevil (approved):
LGTM

PR #1394, opened 2024-06-08
deprecate aligned_storage

### LINKED ISSUE #1253: std::aligned_storage is deprecated in C++23 and should not be used anymore
https://eel.is/c++draft/depr.meta.types

blonded04:
### Description 
Removed deprecated `aligned_storage_t` from tbb

Fixes #1253

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown


include/oneapi/tbb/detail/_concurrent_unordered_base.h
```diff
@@ -179,8 +179,7 @@ class value_node : public list_node<SokeyType>
     }
 
 private:
-    using aligned_storage_type = typename std::aligned_storage<sizeof(value_type)>::type;
-    aligned_storage_type my_value;
+    alignas(value_type) char my_value[sizeof(value_type)];
```
blonded04, line 182: is `alignas(max_nfs_size)` missing here?

blonded04:
Hello @kboyarinov ! Can you please review it?

blonded04:
@kboyarinov Any chance it's merged?

kboyarinov (commented):
@blonded04,

Sorry for taking so long to review this. I am OK with the idea of the change.

But the problem is that for the previous implementation, the second argument for `std::aligned_storage` was missed and hence the `my_value` field alignment can change if the change would be applied that results in a breaking change (consider the use-case when the container is passed between the border of the translation units built with different oneTBB versions). 
To protect the user from this, we need to update the implementation version for all of the concurrent unordered containers by changing the internal namespace number. Currently it is `d1`. For this change, we need to update it to `d2`. Could you please do it as part of this PR to make it possible to merge it?

blonded04:
> @blonded04,
> 
> Sorry for taking so long to review this. I am OK with the idea of the change.
> 
> But the problem is that for the previous implementation, the second argument for `std::aligned_storage` was missed and hence the `my_value` field alignment can change if the change would be applied that results in a breaking change (consider the use-case when the container is passed between the border of the translation units built with different oneTBB versions). To protect the user from this, we need to update the implementation version for all of the concurrent unordered containers by changing the internal namespace number. Currently it is `d1`. For this change, we need to update it to `d2`. Could you please do it as part of this PR to make it possible to merge it?

Thanks for review. I'll finish this PR on this weekend.

blonded04:
Okay, thank you, I'll do it on the closest weekend!

blonded04:
> > @blonded04,
> > Sorry for taking so long to review this. I am OK with the idea of the change.
> > But the problem is that for the previous implementation, the second argument for `std::aligned_storage` was missed and hence the `my_value` field alignment can change if the change would be applied that results in a breaking change (consider the use-case when the container is passed between the border of the translation units built with different oneTBB versions). To protect the user from this, we need to update the implementation version for all of the concurrent unordered containers by changing the internal namespace number. Currently it is `d1`. For this change, we need to update it to `d2`. Could you please do it as part of this PR to make it possible to merge it?
> 
> Thanks for review. I'll finish this PR on this weekend.

next weekend* sorry

PR #1397, opened 2024-06-10
Fix concurrent unordered containers build on GCC14

kboyarinov:
### Description 
GCC14 still produces warnings for incorrect destructor or out-of-boundaries access while destructing the dummy_node since the correct destruction branch would be selected in run-time.
Adding `volatile` to the value_node pointer still fixes the issue, so this PR prologues the workaround until GCC15 release.

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown


pavelkumbrasev (approved):
LGTM!

Issue #1403, opened 2024-06-13
static_partitioner + global_control triggers an unbounded memory leak


lennoxho:
## Description
We found using `tbb::static_partitioner` while a `tbb::global_control` is active causes steady and seemingly unbounded memory leaks as tasks were executed.

A minimal repro is attached below.

The issue goes away if we do any of the following
- Replace `tbb::static_partitioner` with `tbb::auto_partitioner`
- Remove the call to `tbb::global_control`
- Wrap the `tbb::parallel_for` call with `tbb::task_arena::execute`

Reproduced the following setup
- GCC 13.1.0 & GCC 13.2.0
- Ubuntu 20.04LTS native & Ubuntu 22.04 LTS on WSL 2
- A range of x86-64 Intel/AMD workstation & server-grade hardware
- oneTBB 2021.12.0 & the latest commit (fdf1fdb)

### Makefile
```
TBB_DIR ?= /mnt/d/Users/LennoxHo/source/onetbb/oneTBB-2021.12.0-install

COMPILE_FLAGS := -g -O2 -Wall -Wextra -fPIC -isystem $(TBB_DIR)/include
LINK_FLAGS := -Wl,-rpath,$(TBB_DIR)/lib -L$(TBB_DIR)/lib -ltbb

TEST_NAME := tbb-leak-test

all : test
.PHONY : all test clean

$(TEST_NAME) : tbb-leak-test.o
	g++ $(COMPILE_FLAGS) $^ -o $@ $(LINK_FLAGS)

%.o : %.cpp
	g++ -c $(COMPILE_FLAGS) $< -o $@

clean:
	rm -f *.o
	rm -f tbb-leak-test

test: $(TEST_NAME)
	./$(TEST_NAME)
```

### tbb-leak-test.cpp
```
#include <cassert>
#include <cstdio>
#include <sys/resource.h>

#include <oneapi/tbb/global_control.h>
#include <oneapi/tbb/parallel_for.h>
#include <oneapi/tbb/partitioner.h>

void print_rss() {
    struct rusage ru;
    int result = getrusage(RUSAGE_SELF, &ru);
    assert(result == 0);

    printf("Max RSS = %ld kB\n", ru.ru_maxrss);
}

void busy_work(int x) {
    volatile int result = x * x;
    (void)result;
}

int main() {
    using tbb_partitioner = tbb::static_partitioner;

    constexpr int num_threads = 8;
[...]

PR #1404, opened 2024-06-13
Fix memory leak for static_partitioner

pavelkumbrasev:
### Description 
`global_control` we limit concurrency that all the internal arenas can share between each other. Therefore, each particular arena doesn't not know its actual concurrency limit but only one you explicitly set during construction or for `default arena` (one that will be used for simple `parallel_for` call for example) the concurrency will be a whole machine.
When you start `parallel_for` with `static_partitioner` it will create as many internal `proxy tasks` as `normal tasks` (`proxy tasks` are used to assign tasks to specific threads). `Proxy tasks` have a property they should be executed twice. When execute is called for the `proxy task` for the first time it will return actual task that it propagated. When execute is called for the second time `proxy task` can be deleted.
In test case each time we call `parallel_for` with `static_partitioner` it will create `hardware_concurrency` proxy tasks but because `global_control` is present the concurrency of the default arena will not be fully satisfied and some of the proxy tasks will be called only once so they never be destroyed.

This fix will not help if `global_control` is set concurrently with parallel algorithm execution. Perhaps, this scenario is less probable.

#1403 

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

@lennoxho

### DIFF
--- a/include/oneapi/tbb/partitioner.h
+++ b/include/oneapi/tbb/partitioner.h
@@ -46,6 +46,7 @@
 #include "cache_aligned_allocator.h"
 #include "task_group.h" // task_group_context
 #include "task_arena.h"
+#include "global_control.h"
 
 #include <algorithm>
 #include <atomic>
@@ -70,7 +71,8 @@ class affinity_partitioner_base;
 
 inline std::size_t get_initial_auto_partitioner_divisor() {
     const std::size_t factor = 4;
-    return factor * static_cast<std::size_t>(max_concurrency());
+    return factor * std::min(static_cast<std::size_t>(max_concurrency()),
+        tbb::global_control::active_value(tbb::global_control::max_allowed_parallelism));
 }
 
 //! Defines entry point for affinity partitioner into oneTBB run-time library.
@@ -90,7 +92,8 @@ class affinity_partitioner_base: no_copy {
     /** Retains values if resulting size is the same. */
     void resize(unsigned factor) {
         // Check factor to avoid asking for number of workers while there might be no arena.
-        unsigned max_threads_in_arena = static_cast<unsigned>(max_concurrency());
+        unsigned max_threads_in_arena = std::min(static_cast<std::size_t>(max_concurrency()),
+            tbb::global_control::active_value(tbb::global_control::max_allowed_parallelism));
         std::size_t new_size = factor ? factor * max_threads_in_arena : 0;
         if (new_size != my_size) {
             if (my_array) {
--- a/include/oneapi/tbb/partitioner.h
+++ b/include/oneapi/tbb/partitioner.h
@@ -1,5 +1,5 @@
 /*
-    Copyright (c) 2005-2023 Intel Corporation
+    Copyright (c) 2005-2024 Intel Corporation
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
--- a/test/tbb/test_task.cpp
+++ b/test/tbb/test_task.cpp
@@ -20,6 +20,7 @@
 #include "common/spin_barrier.h"
 #include "common/utils_concurrency_limit.h"
 #include "common/cpu_usertime.h"
+#include "common/memory_usage.h"
 
 #include "tbb/task.h"
 #include "tbb/task_group.h"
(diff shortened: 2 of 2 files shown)


include/oneapi/tbb/partitioner.h
```diff
@@ -90,7 +92,8 @@ class affinity_partitioner_base: no_copy {
     /** Retains values if resulting size is the same. */
     void resize(unsigned factor) {
         // Check factor to avoid asking for number of workers while there might be no arena.
-        unsigned max_threads_in_arena = static_cast<unsigned>(max_concurrency());
+        unsigned max_threads_in_arena = std::min(static_cast<std::size_t>(max_concurrency()),
+            tbb::global_control::active_value(tbb::global_control::max_allowed_parallelism));
```
pavelkumbrasev, line 96:
```suggestion
        unsigned max_threads_in_arena = unsigned(std::min(static_cast<std::size_t>(max_concurrency()),
            tbb::global_control::active_value(tbb::global_control::max_allowed_parallelism)));
```

pavelkumbrasev:
Potentially, idle workers can drain tasks from free slots.

include/oneapi/tbb/partitioner.h
```diff
@@ -90,7 +92,8 @@ class affinity_partitioner_base: no_copy {
     /** Retains values if resulting size is the same. */
     void resize(unsigned factor) {
         // Check factor to avoid asking for number of workers while there might be no arena.
-        unsigned max_threads_in_arena = static_cast<unsigned>(max_concurrency());
+        unsigned max_threads_in_arena = unsigned(std::min(static_cast<std::size_t>(max_concurrency()),
+            tbb::global_control::active_value(tbb::global_control::max_allowed_parallelism)));
```
isaevil, line 96: Looks a little bit clumsy, don't you think? Does it make sense to move calculation of `maximum_arena_concurrency` into some dedicated function that could be reused in `get_initial_auto_partitioner_divisor` as well?

pavelkumbrasev:
> Potentially, idle workers can drain tasks from free slots.

This will require additional synchronization on mailbox due to concurrent access to it. So while this solution will solve all potential problems it also brings potential performance problems

include/oneapi/tbb/partitioner.h
pavelkumbrasev: I'm not sure. if need to introduce new method for these 2 lines. (we check not arena concurrency `min(arena_concurrency, allowed_concurrency)` but rather available amount of workers).

test/tbb/test_task.cpp
```diff
@@ -840,3 +841,28 @@ TEST_CASE("Check correct arena destruction with enqueue") {
...
+        stability_counter = current_memory_usage==previous_memory_usage ? stability_counter + 1 : 0;
+        // If the amount of used memory has not changed during 5% of executions,
+        // then we can assume that the check was successful
+        if (stability_counter > num_iterations / 20) {
+            no_memory_leak = true;
+            break;
+        }
+        previous_memory_usage = current_memory_usage;
+    }
+    REQUIRE_MESSAGE(no_memory_leak, "Seems we get memory leak here.");
```
aleksei-fedotov, line 867: Wrap it into the loop with gradual decrease of the `global_control`'s limit? E.g.,
```cpp
std::size_t current_limit = std::size_t(tbb::this_task_arena::max_concurrency());
while (current_limit /= 2) {
    tbb::global_control gc{ tbb::global_control::max_allowed_parallelism, current_limit };

    // iterations loop goes here {
        // repetitions loop goes here {
        // }
    // }
}
```

include/oneapi/tbb/partitioner.h
aleksei-fedotov: I agree to move it into something like `get_num_possible_workers()` to highlight that the solution is based on the immediate value of workers as `std::min(...)` does not mean much to reader.

aleksei-fedotov (commented):
Questions:

1. How does the patch fix memory leaks in case `global_control` is instantiated in the middle of a parallel algorithm's work?
2. Write such a test.

include/oneapi/tbb/partitioner.h
pavelkumbrasev: Remove from affinity

PR #1410, opened 2024-06-19
Improve robustness of leaked observer warning

pavelkumbrasev:
### Description 
The Leaked proxy observers warning could appear during the shutdown process even if there was no memory leak (related structures were destroying concurrently).
This patch improves reliability of the  warning by removing dependency on global variable destructor and moving warning to the `release_resources` (all of the proxies should be destroyed 100%).

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/test/tbb/test_task.cpp
+++ b/test/tbb/test_task.cpp
@@ -857,7 +857,7 @@ TEST_CASE("Try to force Leaked proxy observers warning") {
 
     arena.enqueue([] {
         tbb::parallel_for(0, 100000, [] (int) {
-            for (volatile int i = 0; i < 1000; ++i);
+            utils::doDummyWork(1000);
         });
     });
 }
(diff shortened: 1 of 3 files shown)


test/tbb/test_task.cpp
```diff
@@ -840,3 +840,24 @@ TEST_CASE("Check correct arena destruction with enqueue") {
...
+    }
+
+    for (auto& observer : observers) {
+        observer.observe(true);
+    }
+
+    arena.enqueue([] {
+        tbb::parallel_for(0, 100000, [] (int) {
+            for (volatile int i = 0; i < 1000; ++i);
+        });
```
dnmokhov, line 861: `volatile` warning in C++20

dnmokhov (commented):
LGTM, but adjust the test to get rid of the warning.

test/tbb/test_task.cpp
pavelkumbrasev: Improved

PR #1411, opened 2024-06-20
Collect contribution ideas

pavelkumbrasev:
### Description 
This patch provides a list of tasks that the community can help us with.

### DIFF
--- a/CONTRIBUTION_IDEAS.MD
+++ b/CONTRIBUTION_IDEAS.MD
@@ -49,10 +49,23 @@ You have the option to contribute a new example that you deem significant.
 Alternatively, you can check library APIs that have not yet been covered and provide an example that addresses them.
 Any example you provide should adhere to the style of the existing benchmarks.
 
+### Stabilize examples performance.
+Several oneTBB examples serve as performance benchmarks for the library. It's recommended to check the relative error for these examples.
+If the error exceeds 5%, consider stabilizing the example to improve the reproducibility of results, while still maintaining a reasonable execution time.
+
 ## Intermediate
 
-### NUMA improvements
-In the current state 
+### C++ 20 Modules
+With the introduction of C++ 20 Modules, it is worth analyzing the applicability of this feature to the oneTBB library.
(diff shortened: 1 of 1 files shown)


pavelkumbrasev:
I did not cover each and every task that are available for contribution in this initial commits.
Please, @kboyarinov @aleksei-fedotov @isaevil @akukanov @vossmjp feel free to add more ideas as part of this PR.

CONTRIBUTION_IDEAS.MD
```diff
@@ -0,0 +1,77 @@
...
+
+- Expert: These tasks are challenging and require advanced knowledge and expertise.
+They often involve complex algorithms, performance optimizations, or deep understanding of the project's internals.
+Contributors with extensive experience in the project can tackle these tasks to push the boundaries of the project's capabilities.
+
+When choosing a task to work on, consider your skill level and the complexity bracket that best matches your expertise.
+This will help ensure that you can make meaningful contributions and continue to grow as a developer.
+
+## Beginner
+
```
isaevil, line 42: Сonsider the following "beginner" task:
```suggestion
### Align symbol exporting rules across oneTBB libraries/plugins
oneTBB uses .def files to list global symbols to be exported and also local symbols to hide. 
For example, TBBMalloc hides symbols from static libraries which is a result of compilation with Intel Compiler (see https://github.com/oneapi-src/oneTBB/blob/master/cmake/compilers/Intel.cmake#L34). 
oneTBB library hides all symbols which are not listed in global section.
But other libraries like TBBBind, for example, for some reason does neither of these things and it may result in some symbol conflicts in them environment where multiple libraries depends on the same symbols.
The goal of this task is to align approaches across all of these libraries:
- Define local section with the symbols not meant for the exporting for each library.
- Analyze if is there any unintendedly exported symbols that are not related to the oneTBB libraries themselves. Consider hiding these symbols not in .def files but in the corresponding compiler CMake files.
```

pavelkumbrasev:
Idea: improve proxy tasks clean-up mechanism

Issue #1417, opened 2024-06-24
tbb::global_control does not stop spinning threads


Sedeniono:
# Summary
Reducing the max. allowed parallelism via `tbb::global_control` does not "stop" threads anymore. The worker threads are spinning in `thread_dispatcher::process()` and are eating CPU power, meaning that the execution is slower. 

The problem gets triggered when executing first e.g. `tbb::parallel_for_each` (to "fire up" threads), then reducing the number of threads to 1 via `tbb::global_control` and then executing a second `tbb::parallel_for_each`. While the second `tbb::parallel_for_each` is executed, only a single core should be active. Using the debugger I also see that the business logic is executed only by a single thread. However, looking in the task manager, I see that all cores are busy. The unnecessary threads cause the code to run slower by ~50% (either because of oversubscription, or because the CPU frequency gets reduced, not sure).

# Version
Version 2021.10.0 was ok. Version 2021.11.0 is broken. The current master (55bf2b33417ebd7f8b9add37709259b5c7ec3a5c) is still broken.

I was able to bisect the problem to commit c45684495599d41ba10893effa0682eceb1a3169 (pull request: #758).

# Environment
* Windows 11
* Intel Core i9-13900
* Microsoft Compiler version 19.40 (`_MSC_FULL_VER=194033811`)
 
# Observed Behavior
100% CPU load (all cores are busy) even though `tbb::global_control` was used to reduce the max. allowed parallelism to 1.

# Expected Behavior
After setting the max. allowed parallelism to 1 via `tbb::global_control`, only a single core should be busy (corresponds to ~3% CPU load in the task manager because of 32 virtual cores of my Intel i9-13900).

# Steps To Reproduce
* Code:
```C++
#include <chrono>
#include <iostream>
#include <numeric>
#include <tbb/global_control.h>
#include <tbb/parallel_for_each.h>
#include <tbb/version.h>
#include <vector>

int main()
{
  std::cout << "Start. TBB: " << TBB_VERSION_STRING << ", TBB_runtime_version=" << TBB_runtime_version()
            << ", TBB_runtime_interface_version=" << TBB_runtime_interface_version() << ", MSVC: " << _MSC_FULL_VER
            << std::endl;

  //------------------------------------------------
  // First call of tbb::parallel_for_each() to 'create' threads
  static constexpr bool TRIGGER_BUG = true;
  if (TRIGGER_BUG) {
    std::cout << "Warmup to trigger bug" << std::endl;
[...]

PR #1418, opened 2024-06-25
Refactor set of soft_limit to fix a bug

pavelkumbrasev:
### Description 
In the current implementation zero soft_limit would be ignored because of errors in `mandatory concurrency` checks.
This patch refactors the function allows zero soft_limit to be set.

#1417

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/test/tbb/test_global_control.cpp
+++ b/test/tbb/test_global_control.cpp
@@ -275,7 +275,7 @@ TEST_CASE("test concurrent task_scheduler_handle destruction") {
 }
 
 //! \brief \ref regression
-TEST_CASE("Thread should sleep when soft_limit is zero") {
+TEST_CASE("Test worker threads remain inactive in enforced serial execution mode") {
     int num_threads = int(utils::get_platform_max_threads());
     std::atomic<int> barrier{num_threads};
 
@@ -295,7 +295,7 @@ TEST_CASE("Thread should sleep when soft_limit is zero") {
         });
     });
 
-    // Workers should sleep because of the soft_limit
+    // Workers should sleep because of global_control enforced serial execution of tasks
     TestCPUUserTime(utils::get_platform_max_threads() - 1);
 
     thr.join();
--- a/test/tbb/test_global_control.cpp
+++ b/test/tbb/test_global_control.cpp
@@ -22,6 +22,7 @@
 #include "common/utils.h"
 #include "common/spin_barrier.h"
 #include "common/utils_concurrency_limit.h"
+#include "common/cpu_usertime.h"
 
 #include "tbb/global_control.h"
 #include "tbb/parallel_for.h"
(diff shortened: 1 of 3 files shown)


test/tbb/test_global_control.cpp
```diff
@@ -272,3 +273,30 @@ TEST_CASE("test concurrent task_scheduler_handle destruction") {
     stop = true;
     thr1.join();
 }
+
+//! \brief \ref regression
+TEST_CASE("Thread should sleep with when soft_limit is zero") {
```
pavelkumbrasev, line 278:
```suggestion
TEST_CASE("Thread should sleep when soft_limit is zero") {
```

test/tbb/test_global_control.cpp
```diff
@@ -272,3 +273,30 @@ TEST_CASE("test concurrent task_scheduler_handle destruction") {
     stop = true;
     thr1.join();
 }
+
+//! \brief \ref regression
+TEST_CASE("Thread should sleep when soft_limit is zero") {
```
isaevil, line 278: Suggesting more blackbox approach for naming this test.
```suggestion
TEST_CASE("Test worker threads remain inactive in enforced serial execution mode") {
```

test/tbb/test_global_control.cpp
```diff
@@ -272,3 +273,30 @@ TEST_CASE("test concurrent task_scheduler_handle destruction") {
...
+    int num_threads = int(utils::get_platform_max_threads());
+    std::atomic<int> barrier{num_threads};
+
+    // Warm-up threads
+    tbb::parallel_for(0, num_threads, [&] (int) {
+        --barrier;
+        while (barrier > 0) {
+            std::this_thread::yield();
+        }
```
isaevil, line 287: Reuse utils::SpinBarrier?

test/tbb/test_global_control.cpp
```diff
@@ -272,3 +273,30 @@ TEST_CASE("test concurrent task_scheduler_handle destruction") {
...
+
+    tbb::global_control control(tbb::global_control::max_allowed_parallelism, 1);
+
+    std::thread thr([&] {
+        tbb::parallel_for(0, 100000, [&] (int) {
+            utils::doDummyWork(100);
+        });
+    });
+
+    // Workers should sleep because of the soft_limit
```
isaevil, line 298:
```suggestion
    // Workers should sleep because of global_control enforced serial execution of tasks
```

test/tbb/test_global_control.cpp
isaevil, line 278: Shouldn't this test be moved under `conformance_global_control.cpp`? It seems like this test verifies compliance with the specification:

> With max_allowed_parallelism set to 1, global_control enforces serial execution of all tasks by the application thread(s), that is, the task scheduler **does not allow** worker threads to run. There is one exception: if some work is submitted for execution via task_arena::enqueue, a single worker thread will still run ignoring the max_allowed_parallelism restriction.

isaevil (commented):
Bug fix looks good. Left comments about the test.

test/tbb/test_global_control.cpp
pavelkumbrasev: Done

test/tbb/test_global_control.cpp
pavelkumbrasev: Done

isaevil (approved):
LGTM!

PR #1446, opened 2024-07-12
Added a utility class for Relative Error compute.

sarathnandu:
### Description 
Added a utility class for Relative Error compute.
Added relative error compute for seismic example.

### Type of change

- [x ] new feature - _change that adds functionality_

### Tests

- [ x] not needed

### Documentation

- [ ] needs to be updated
- [x ] not needed

- [ ] Yes
- [x ] No
- [ ] Unknown


examples/common/utility/utility.hpp
```diff
@@ -20,6 +20,7 @@
 #include <cassert>
 #include <cstring>
 #include <cstdlib>
+#include <cmath>
```
dnmokhov, line 23: Probably need to also include `oneapi/tbb/tick_count.h`?

sarathnandu: The tbb header files are not supposed to be added to utility.hpp. Instead of using tbb_tick_counts I have used std::chrono::time_point

examples/common/utility/utility.hpp
```diff
@@ -356,6 +358,46 @@ class cli_argument_pack {
...
+        _secPerFrame.clear();
+    }
+    void start() {
+        _startTime = std::chrono::steady_clock::now();
+    }
+    void stop() {
+        _endTime = std::chrono::steady_clock::now();
+        std::chrono::duration<double> duration = _endTime - _startTime;
+        // store the duration in seconds
+        _secPerFrame.push_back(duration.count());
```
aleksei-fedotov, line 377: Usually, it makes sense to introduce as little as possible overhead during the measurements to avoid influence the the computations being measured. So, consider pushing the postprocessing step to a later stage. Here only store `_startTime` and `_endTime`, and post process them in the independent from the main computations part, e.g., `computeRelError`.

aleksei-fedotov (commented):
I believe we need to continue measurements until one of the following conditions are met:
1. Relative error is within the acceptable limit
2. Overall limit on running time is reached.

examples/common/utility/utility.hpp
sarathnandu: Thanks, Aleksei, for bringing this point. I wasn't sure which approach is less intrusive. 
Currently I have only one store with a subtract instruction. In the approach you suggested we need to 2 vector stores in the measurement loop. I was under the impression subtract + 1 store is less intrusive, please correct me if I am wrong.

sarathnandu: In any case I think I should reserve some memory for the vector stores based on the iteration count. I will update the constructor accordingly.

sarathnandu:
> I believe we need to continue measurements until one of the following conditions are met:
> 
> 1. Relative error is within the acceptable limit
> 2. Overall limit on running time is reached.

Yes, that's the intent. The relative error may also vary from platform to platform for the same iteration count. In my experiments, for certain benchmarks, I have seen Relative error more stable on client systems whereas on high core count NUMA systems it sometimes increases with iteration.

The Performance CI scripts will utilize the "Relative_Err" string to adjust the iteration count to minimize the relative error in respective systems.

examples/common/utility/utility.hpp
aleksei-fedotov: That's correct! We want to have the impact from the measurement system itself as little as possible. Therefore, we need to consider implementing at least the following:
* Having the memory pre-allocated. I guess just need to add a parameter to the constructor.
  * Are there cases when the number of iterations is not known beforehand?
* Do all the calculations only after the measurement loop is completed. This means inside the measurement loop can only be stores of the retrieved timings.

examples/common/utility/utility.hpp
```diff
@@ -356,6 +358,46 @@ class cli_argument_pack {
...
+        double sumOfSquareDiff = std::inner_product(diff.begin(), diff.end(), diff.begin(), 0.0);
+        double stdDev = std::sqrt(sumOfSquareDiff / _secPerFrame.size());
+        double relError = 100 * (stdDev / averageTimePerFrame);
+        return relError;
+    }
+
+private:
+    std::vector<double> _secPerFrame;
+    std::chrono::steady_clock::time_point _startTime;
+    std::chrono::steady_clock::time_point _endTime;
```
aleksei-fedotov, line 398: To avoid having two vectors, where one is for `_startTime` timings and another one is for `_endTime` timings, and hence two separate memory pre-allocations, we may have single vector consisting of corresponding pairs of time points, that is:
```suggestion
    using time_point = std::chrono::steady_clock::time_point;
    time_point _startTime;
    std::vector< std::pair<time_point, time_point> > _time_intervals;
```
`_startTime` variable is constantly re-used during the `start()` call, and the `_time_intervals` being pre-allocated in the constructor, is pushed new pair into it during the `stop()` method call.

What do you think?

examples/parallel_for/seismic/main.cpp
```diff
@@ -106,8 +120,13 @@ int main(int argc, char *argv[]) {
             else {
                 oneapi::tbb::global_control c(oneapi::tbb::global_control::max_allowed_parallelism,
                                               p);
-                for (int i = 0; i < numberOfFrames; ++i) {
-                    u.ParallelUpdateUniverse();
+                mu.clear();
```
aleksei-fedotov, line 123: We can move instantiation of the `utility::measurements mu` here and have instead `double rel_error` created above. That will allow us dropping `utility::measurements::clear()` method.
```suggestion
                utility::measurements mu;
```
Then, just after the measurement loop, assign the `mu.computeRelError()` to that global `rel_error`.

examples/common/utility/utility.hpp
```diff
@@ -356,6 +358,46 @@ class cli_argument_pack {
...
+        auto averageTimePerFrame =
+            std::accumulate(_secPerFrame.begin(), _secPerFrame.end(), 0.0) / _secPerFrame.size();
+        std::vector<double> diff(_secPerFrame.size());
+        std::transform(_secPerFrame.begin(),
+                       _secPerFrame.end(),
+                       diff.begin(),
+                       [averageTimePerFrame](double x) {
+                           return (x - averageTimePerFrame);
+                       });
+        double sumOfSquareDiff = std::inner_product(diff.begin(), diff.end(), diff.begin(), 0.0);
```
aleksei-fedotov, line 389: We can also transform the data in place. This will save us one pass through the whole vector.
```suggestion
        double sumOfSquareDiff = 0.0;
        std::for_each(_secPerFrame.begin(),
                      _secPerFrame.end(),
                       [averageTimePerFrame, &sumOfSquareDiff](double x) {
                           double diff = (x - averageTimePerFrame);
                           sumOfSquareDiff += diff * diff;
                       });
```

aleksei-fedotov:
> > I believe we need to continue measurements until one of the following conditions are met:
> > 
> > 1. Relative error is within the acceptable limit
> > 2. Overall limit on running time is reached.
> 
> Yes, that's the intent. The relative error may also vary from platform to platform for the same iteration count. In my experiments, for certain benchmarks, I have seen Relative error more stable on client systems whereas on high core count NUMA systems it sometimes increases with iteration.
> 
> The Performance CI scripts will utilize the "Relative_Err" string to adjust the iteration count to minimize the relative error in respective systems.

Do I understand correctly that CI scripts will keep re-starting the measurements each time specifying increased number of iterations to run? So that they work as if in the following steps:

1. Set `num_iterations` to previously saved value (see step 3).
1. Run the benchmark and get the relative error computed.
1. If it is within the per benchmark pre-defined limit, save that number of iterations. The system will start from that value in the next measurement session (see step 1).
1. If it is not within the per benchmark pre-defined limit, then:
   * If overall per benchmark measurement session time is acceptable, increase the `num_iterations` (for example, double it) and go to the step 2.
   * If overall time is not acceptable, stop and report error saying that the measurement session is unstable, results are not reliable and that further analysis or re-consideration of either system or benchmark is required.

We also, can add step 3.1 that before saving the `num_iterations` will try decreasing it a little bit and see if still produces acceptable value of relative error. This way, we could make the system that will automatically and dynamically adjusts to the value of `num_iterations` closest to the one that gives stable relative error within defined range.

examples/parallel_for/seismic/main.cpp
```diff
@@ -91,11 +99,17 @@ int main(int argc, char *argv[]) {
             std::cout << "Substituting 1000 for unlimited frames because not running interactively"
                       << "\n";
         }
+        if (options.numberofIterations == 0) {
+            options.numberofIterations = 10;
+            std::cout << "Setting the number of iterations = 10 default"
+                      << "\n";
+        }
```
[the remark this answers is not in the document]
sarathnandu: The utility::cli_argument_pack() doesn't have the ability to specify default. Added a TODO comment, since this change is not in the scope of this pull request.

examples/common/utility/utility.hpp
sarathnandu: Modified the constructor to accept a parameter for number of iterations and reserve the memory accordingly. All calculations are done after the measurement loop is completed.

sarathnandu:
> > > I believe we need to continue measurements until one of the following conditions are met:
> > > 
> > > 1. Relative error is within the acceptable limit
> > > 2. Overall limit on running time is reached.
> > 
> > 
> > Yes, that's the intent. The relative error may also vary from platform to platform for the same iteration count. In my experiments, for certain benchmarks, I have seen Relative error more stable on client systems whereas on high core count NUMA systems it sometimes increases with iteration.
> > The Performance CI scripts will utilize the "Relative_Err" string to adjust the iteration count to minimize the relative error in respective systems.
> 
> Do I understand correctly that CI scripts will keep re-starting the measurements each time specifying increased number of iterations to run? So that they work as if in the following steps:
> 
> 1. Set `num_iterations` to previously saved value (see step 3).
> 2. Run the benchmark and get the relative error computed.
> 3. If it is within the per benchmark pre-defined limit, save that number of iterations. The system will start from that value in the next measurement session (see step 1).
> 4. If it is not within the per benchmark pre-defined limit, then:
>    
>    * If overall per benchmark measurement session time is acceptable, increase the `num_iterations` (for example, double it) and go to the step 2.
>    * If overall time is not acceptable, stop and report error saying that the measurement session is unstable, results are not reliable and that further analysis or re-consideration of either system or benchmark is required.
> 
> We also, can add step 3.1 that before saving the `num_iterations` will try decreasing it a little bit and see if still produces acceptable value of relative error. This way, we could make the system that will automatically and dynamically adjusts to the value of `num_iterations` closest to the one that gives stable relative error within defined range.

Yes this workflow needs to be implemented in the CI perf runner script to adjust the number of iterations until the relative error is reached within acceptable limits for the example benchmark. This intent is to improve the stability of measurements from run to run so that perf report geomean results are more reliable.

examples/common/utility/utility.hpp
```diff
@@ -356,6 +358,62 @@ class cli_argument_pack {
...
+        }
+        auto total_duration = std::accumulate(
+            _time_intervals.begin(),
+            _time_intervals.end(),
+            0, // Start with 0 count
+            [](long long total, const std::pair<time_point, time_point>& interval) {
+                // Compute the difference and add it to the total
+                return total + std::chrono::duration_cast<std::chrono::microseconds>(
+                                   interval.second - interval.first)
+                                   .count();
```
aleksei-fedotov, line 392: I wonder if it would be better to optimize by computing the `std::chrono::duration_cast<std::chrono::microseconds>(interval.second - interval.first).count()` only once and reuse the result below. What do you think?

aleksei-fedotov (commented):
More remarks to consider.

aleksei-fedotov:
> this workflow needs to be implemented in the CI perf runner script to adjust the number of iterations until the relative error is reached within acceptable limits for the example benchmark.

Don't you think that this better be implemented here instead? At least, it won't require restarting the whole benchmark for each new trial of measurements, which might be relatively complex process for some of them.

sarathnandu:
> Don't you think that this better be implemented here instead? At least, it won't require restarting the whole benchmark for each new trial of measurements, which might be relatively complex process for some of them.

We would like this feature to be available as part of performance CI and extend it to all benchmarks that are part of the geomean calculation.  Integrating this feature into the performance CI enables benchmark modifications to minimal (specify iteration counts and output relative error directly). For benchmarks where source code access is unavailable, the script can leverage the application runtime to compute relative error.

examples/common/utility/utility.hpp
sarathnandu: Storing the duration at each time point in a vector and reading it back for accumulation and standard deviation calculations increases memory usage. In contrast, calculating the duration on the fly for these functions is more computationally complex but avoids additional memory overhead. Since this code is not on the critical path, I’m fine with the current approach. However, if you think storing durations in a vector for later access is preferable, I can switch to that.

examples/common/utility/utility.hpp
```diff
@@ -356,6 +358,62 @@ class cli_argument_pack {
...
+        auto _endTime = std::chrono::steady_clock::now();
+        // store the end time and start time
+        _time_intervals.push_back(std::make_pair(_startTime, _endTime));
+    }
+    double computeRelError() {
+        // Accumulate the total duration in microseconds using std::accumulate with a lambda function
+        if (0 == _time_intervals.size()) {
+            std::cout << "No time samples collected \n";
+            return 0;
+        }
```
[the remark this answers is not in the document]
sarathnandu: Make sense, modified to use assert that will fire in debug mode.

examples/common/utility/utility.hpp
```diff
@@ -356,6 +358,62 @@ class cli_argument_pack {
...
+            _time_intervals.end(),
+            0, // Start with 0 count
+            [](long long total, const std::pair<time_point, time_point>& interval) {
+                // Compute the difference and add it to the total
+                return total + std::chrono::duration_cast<std::chrono::microseconds>(
+                                   interval.second - interval.first)
+                                   .count();
+            });
+        long long averageTimePerFrame = total_duration / _time_intervals.size();
+        long long sumOfSquareDiff = 0;
```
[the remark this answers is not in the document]
sarathnandu: Changed to use the unsigned long long type.

examples/parallel_for/seismic/main.cpp
```diff
@@ -53,6 +56,7 @@ RunOptions ParseCommandLine(int argc, char *argv[]) {
         utility::get_default_num_threads, 0, utility::get_default_num_threads());
 
     int numberOfFrames = 0;
+    int numberofIterations = 0;
```
aleksei-fedotov, line 59: To be consistent with the `numberOfFrames`:
```suggestion
    int numberOfIterations = 0;
```

examples/parallel_for/seismic/main.cpp
```diff
@@ -91,11 +99,18 @@ int main(int argc, char *argv[]) {
             std::cout << "Substituting 1000 for unlimited frames because not running interactively"
                       << "\n";
         }
+        // TODO : Extend utility::cli_argument_pack() to allow specifying the default value.
+        if (options.numberofIterations == 0) {
```
aleksei-fedotov, line 103: Make it a bit more reliable?
```suggestion
        if (options.numberofIterations <= 0) {
```

examples/parallel_for/seismic/main.cpp
```diff
@@ -91,11 +99,18 @@ int main(int argc, char *argv[]) {
...
+            options.numberofIterations = 10;
+            std::cout << "Setting the number of iterations = 10 default"
+                      << "\n";
+        }
         for (int p = options.threads.first; p <= options.threads.last;
              p = options.threads.step(p)) {
             oneapi::tbb::tick_count xwayParallelismStartTime = oneapi::tbb::tick_count::now();
             u.InitializeUniverse(video);
             int numberOfFrames = options.numberOfFrames;
+            int numberOfIterations = options.numberofIterations;
```
aleksei-fedotov, line 113: Also, consider:
```suggestion
            assert(options.numberOfIterations > 0 && "Number of iterations cannot be <= 0");
            unsigned numberOfIterations = unsigned(options.numberOfIterations);
```

aleksei-fedotov (commented):
I think the patch is mostly done. Consider the last bunch of comments that are targeted to improve reliability of the patch.

examples/common/utility/utility.hpp
```diff
@@ -356,6 +358,59 @@ class cli_argument_pack {
     }
 }; // class cli_argument_pack
 
+// utility class to aid relative error measurement of samples
+class measurements {
+public:
+    measurements() = default;
```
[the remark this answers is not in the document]
sarathnandu: I would like to keep this class default constructible. This allows flexibility in cases where the number of iterations is unknown in advance, or when the user chooses not to specify it. While this may lead to multiple vector allocations, which isn’t optimal, I believe it's an acceptable trade-off.

aleksei-fedotov (dismissed):

aleksei-fedotov (approved):
Looks good to me!

PR #1448, opened 2024-07-13
doc: Improve cmake README and INSTALL docs

ChihweiLHBird:
### Description 
Making the README more clear for beginner C++ developers (like me).

### Type of change

### Tests

- [x] not needed

### Documentation

- [x] updated in # - _add PR number_
- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -12,7 +12,7 @@
 From the build directory, run:
 
 ```bash
-cmake [options] <repo_root>
+cmake [options] <onetbb_repo_directory>
 ```
 
 You may want to use some additional options for configuration:
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -9,9 +9,10 @@
 
 ## Configure oneTBB
 
-At the command prompt, type:
-```
-cmake <options> <repo_root>
+In your build directory, type:
+
+```bash
+cmake [options] <repo_root>
 ```
 
 You may want to use some additional options for configuration:
@@ -25,8 +26,9 @@ You may want to use some additional options for configuration:
 ## Build oneTBB
  
 To build the system, run:
-```
-cmake --build . <options>
+
+```bash
+cmake --build . [options]
 ```
 
 Some useful build options:
@@ -73,7 +75,7 @@ Simple packaging using CPack is supported.
 The following commands allow you to create a simple portable package that includes header files, libraries, and integration files for CMake:
 
 ```bash
-cmake <options> ..
+cmake [options] ..
 cpack
 ```
 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -9,7 +9,7 @@
 
 ## Configure oneTBB
 
-In your build directory, type:
+From the build directory, run:
 
 ```bash
 cmake [options] <repo_root>
--- a/cmake/README.md
+++ b/cmake/README.md
@@ -35,7 +35,7 @@ cd /tmp/my-build
 
 ### Configure
 
-From the build directory, to generate a build system for it, run:
+From the build directory, to generate a build system, run:
 
 ```bash
 cmake [options] <onetbb_repo_directory>
--- a/cmake/README.md
+++ b/cmake/README.md
@@ -26,7 +26,7 @@ TBB_BUILD_APPLE_FRAMEWORKS - Enable the Apple* frameworks instead of dylibs, onl
 
 ### Preparation
 
-To perform an out-of-source build, create a build directory and go there:
+To perform an out-of-source build, create a build directory in an arbitrary location outside of the oneTBB source directory and go there.
 
 ```bash
 mkdir /tmp/my-build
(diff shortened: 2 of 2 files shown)


cmake/README.md
```diff
@@ -26,17 +26,25 @@ TBB_BUILD_APPLE_FRAMEWORKS - Enable the Apple* frameworks instead of dylibs, onl
...
+mkdir ./build && cd ./build
+```
+
 ### Configure
 
+In the build directory, specify the oneTBB source directory to generate a build system for it.
+
 ```bash
-cmake <options> <repo_root>
+cmake <options> <source_dir>
```
omalyshe, line 47: It is not correct, because the main CMakeList.txt is in <repo_root> - not in /src.

cmake/README.md
```diff
@@ -26,17 +26,25 @@ TBB_BUILD_APPLE_FRAMEWORKS - Enable the Apple* frameworks instead of dylibs, onl
...
 
-To perform an out-of-source build, create a build directory and go there:
+To perform an out-of-source build, create a build directory outside of the oneTBB source directory and go there.
 
 ```bash
 mkdir /tmp/my-build
 cd /tmp/my-build
 ```
 
+Or you may also create a build directory within the oneTBB source directory.
```
omalyshe, line 36: Purpose of this addition is not clear. A temporary build directory can be created in an arbitrary location - this is what is stated in the original message above. 

Probably it is worth to add a better example instead of changing the general instructions.

cmake/README.md
ChihweiLHBird: I changed it to source because a repo dir is referred as `<path-to-source>` in `cmake` menu:
```
cmake [options] <path-to-source>
```

But I am happy to restore it back to repo to avoid confusion.

cmake/README.md
ChihweiLHBird: Yea. What do you think about changing [the examples here](https://github.com/oneapi-src/oneTBB/blob/master/INSTALL.md#example-of-installation) to be out of source build?

ChihweiLHBird:
I just updated it to address some concerns, let me know if you have other suggestions. Thanks!

INSTALL.md
```diff
@@ -9,9 +9,10 @@
 
 ## Configure oneTBB
 
-At the command prompt, type:
-```
-cmake <options> <repo_root>
+In your build directory, type:
```
isaevil, line 12:
```suggestion
From the build directory, run:
```

cmake/README.md
```diff
@@ -26,7 +26,7 @@ TBB_BUILD_APPLE_FRAMEWORKS - Enable the Apple* frameworks instead of dylibs, onl
 
 ### Preparation
 
-To perform an out-of-source build, create a build directory and go there:
+To perform an out-of-source build, create a build directory in an arbitrary location outside of the oneTBB source directory and go there.
```
isaevil, line 29:
```suggestion
To perform an out-of-source build, create a build directory in an arbitrary location outside of the oneTBB source directory and go there:
```

cmake/README.md
```diff
@@ -35,8 +35,10 @@ cd /tmp/my-build
 
 ### Configure
 
+In the build directory, specify the oneTBB source directory to generate a build system for it.
```
isaevil, line 38: I don't personally think that this comment adds useful information here. Flow of this README already implies that the user's current working directory to be the build directory after the steps from "Preparation" section.

cmake/README.md
aepanchi, line 29:
```suggestion
To perform an out-of-source build, create a build directory in an arbitrary location outside the oneTBB source directory and go there:
```

cmake/README.md
aepanchi: agree

ChihweiLHBird: It might be the case for users who are familiar with `cmake`, but for users who don't, it can be very confusing because they have no idea why running an unfamiliar command in an empty directory and what does it do.

ChihweiLHBird: But if you guys strongly against it, feel free to suggest a change and I will follow.

isaevil: https://github.com/oneapi-src/oneTBB/blob/master/cmake/README.md#configure-build-and-test is a step-by-step guide to configure, build and test oneTBB. So the document is assuming that the reader follows it from top to bottom.
I think you could probably replace <repo_root> by <onetbb_source_directory> and add `From the build directory, run:` as it has been done for INSTALL.md. Will it help?

ChihweiLHBird: And yes, it's fine to assume the developer is reading from top to bottom; however, in my opinion, it's also very import to descript what each step does in a README doc. Especially for `cmake` which has a very different build workflow than other common programming languages (e.g., Java, Python, Go).

PR #1449, opened 2024-07-13
Extension of Nd range

tobiasweinzierl80:
### Description 

We found that it is very useful for our code if we can construct the range from plain integers. For this, the underlying datatype storing the ranges has to be changed to vector. This should have no negative impact on the performance. However, it makes using the class easier.

The class could do with more assertions and more docu.

### Type of change

- [X ] new feature - _change that adds functionality_
- [X ] tests - _change in tests_

### Tests

- [ X] added - _required for new features and some bug fixes_
- [ ] not needed

### Documentation

- [ X] updated in https://github.com/uxlfoundation/oneAPI-spec/pull/555
- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ X] No
- [ ] Unknown


include/oneapi/tbb/blocked_rangeNd.h
```diff
@@ -17,20 +17,16 @@
...
-#endif
-
 #include <algorithm>    // std::any_of
-#include <array>
+#include <vector>
 #include <cstddef>
 #include <type_traits>  // std::is_same, std::enable_if
 
-#include "detail/_config.h"
-#include "detail/_template_helpers.h" // index_sequence, make_index_sequence
```
isaevil, line 30: Missed this header, that leads CI to fail. https://github.com/oneapi-src/oneTBB/actions/runs/9953702919/job/27513502922?pr=1449

include/oneapi/tbb/blocked_rangeNd.h
```diff
@@ -63,20 +60,35 @@ class blocked_rangeNd_impl<Value, N, detail::index_sequence<Is...>> {
 private:
     //! Helper type to construct range with N tbb::blocked_range<value_type> objects.
     template<std::size_t>
-    using dim_type_helper = tbb::blocked_range<value_type>;
+    using dim_range_type = tbb::blocked_range<value_type>;
```
vossmjp, line 63: Is public in the specification proposal.

tobiasweinzierl80: Indeed needs to be public, as it is used in the constructor.

include/oneapi/tbb/blocked_rangeNd.h
```diff
@@ -100,17 +110,23 @@ class blocked_rangeNd_impl<Value, N, detail::index_sequence<Is...>> {
...
     static_assert(N != 0, "zero dimensional blocked_rangeNd can't be constructed");
 
-    //! Ranges in each dimension.
-    std::array<tbb::blocked_range<value_type>, N> my_dims;
+    /*! Ranges in each dimension.
+     *
+     * This has to be a vector, as there are situations where we can't
+     * initialise the entries all in one go.
+     */
+    std::vector<tbb::blocked_range<value_type>> my_dims;
```
akukanov, line 129: The change to vector might not be necessary, as I think the constructor can be written so that the array is properly initialized - see another comment.

include/oneapi/tbb/blocked_rangeNd.h
```diff
@@ -60,23 +57,36 @@ class blocked_rangeNd_impl<Value, N, detail::index_sequence<Is...>> {
...
+    blocked_rangeNd_impl(const dim_range_type<Is>&... args) : my_dims{ {args...} } {
+      __TBB_ASSERT(my_dims.size()==N, "range is not properly initialised" );
+    }
+
+    blocked_rangeNd_impl(
+      Value begin[N],
+      Value end[N],
+      typename tbb::blocked_range<value_type>::size_type grainsize
+    ):
+      my_dims() {
```
akukanov, line 76: The following way to do the initialization of `my_dims` should I think work:
```suggestion
    blocked_rangeNd_impl(
      Value begin[N],
      Value end[N],
      typename tbb::blocked_range<value_type>::size_type grainsize
    ):
      my_dims{ tbb::blocked_range<value_type>{begin[Is], end[Is], grainsize}... } {
```
And then the body can be simplified as there is no need to initialize data elements.

sasalla23:
I have implemented the changes discussed in the spec as well as the changes concerning `std::array`. The initialization of the array in the constructor with parameter `size[N]` as described by @akukanov works from what I tested. As I said in the spec, I have removed the constructor that uses `begin[N]` and `end[N]` and renamed `ndims` to `dim_count`. Let me know if this poses a problem or if there is anything missing.

include/oneapi/tbb/blocked_rangeNd.h
```diff
@@ -60,23 +57,32 @@ class blocked_rangeNd_impl<Value, N, detail::index_sequence<Is...>> {
...
-    using dim_type_helper = tbb::blocked_range<value_type>;
+    using dim_range_type = tbb::blocked_range<value_type>;
 
-public:
     blocked_rangeNd_impl() = delete;
 
     //! Constructs N-dimensional range over N half-open intervals each represented as tbb::blocked_range<Value>.
-    blocked_rangeNd_impl(const dim_type_helper<Is>&... args) : my_dims{ {args...} } {}
+    blocked_rangeNd_impl(const dim_range_type<Is>&... args) : my_dims{ {args...} } {
+      __TBB_ASSERT(my_dims.size()==N, "range is not properly initialised" );
```
akukanov, line 68: With the change back to array, these assertions make no sense anymore, as they are always true.

include/tbb/blocked_rangeNd.h
```diff
@@ -1,5 +1,5 @@
 /*
-    Copyright (c) 2017-2021 Intel Corporation
+    Copyright (c) 2017-2024 Intel Corporation
```
akukanov, line 2: I am not sure if this change is needed, as the file content remains the same. @omalyshe can you comment?

include/oneapi/tbb/blocked_rangeNd.h
```diff
@@ -100,16 +106,22 @@ class blocked_rangeNd_impl<Value, N, detail::index_sequence<Is...>> {
...
 
 private:
     static_assert(N != 0, "zero dimensional blocked_rangeNd can't be constructed");
 
-    //! Ranges in each dimension.
+    /*! Ranges in each dimension.
+     *
+     * This has to be a vector, as there are situations where we can't
+     * initialise the entries all in one go.
+     */
```
akukanov, line 124: Also seems unnecessary after switching back to array.

include/oneapi/tbb/blocked_rangeNd.h
```diff
@@ -17,20 +17,17 @@
...
 #include <cstddef>
 #include <type_traits>  // std::is_same, std::enable_if
 
 #include "detail/_config.h"
 #include "detail/_template_helpers.h" // index_sequence, make_index_sequence
+#include "detail/_namespace_injection.h"
 #include "detail/_range_common.h"
 
-#include "blocked_range.h"
+#include <oneapi/tbb/blocked_range.h>
```
vossmjp, line 30:
```suggestion
#include "blocked_range.h"
```

vossmjp: I think there's no need to change this.

include/oneapi/tbb/blocked_rangeNd.h
```diff
@@ -110,20 +110,21 @@ class blocked_rangeNd_impl<Value, N, detail::index_sequence<Is...>> {
...
     void do_split(blocked_rangeNd_impl& r, split_type proportion) {
-        static_assert((std::is_same<split_type, split>::value || std::is_same<split_type, proportional_split>::value), "type of split object is incorrect");
+        static_assert((std::is_same<split_type, split>::value || std::is_same<split_type, proportional_split>::value),
+                      "type of split object is incorrect");
         __TBB_ASSERT(r.is_divisible(), "can't split not divisible range");
 
-        auto my_it = std::max_element(my_dims.begin(), my_dims.end(), [](const tbb::blocked_range<value_type>& first, const tbb::blocked_range<value_type>& second) {
-            return (first.size() * second.grainsize() < second.size() * first.grainsize());
+        auto my_it = std::max_element(my_dims.begin(), my_dims.end(), [](const dim_range_type& first, const dim_range_type& second) {
+            return (first.size() * double(second.grainsize()) < second.size() * double(first.grainsize()));
```
pavelkumbrasev, line 122: Why cast to double?

pavelkumbrasev (commented):
Other than that LGTM

include/oneapi/tbb/blocked_rangeNd.h
akukanov: That's what all other blocked_range implementations do :) I guess it is to avoid the risk of multiplication overflow.

pavelkumbrasev: Ohh, I see

test/conformance/conformance_blocked_nd_range.cpp
```diff
@@ -19,11 +19,10 @@
 #include "common/utils_assert.h"
 #include "common/utils_concurrency_limit.h"
 
-//! \file conformance_blocked_rangeNd.cpp
+//! \file conformance_blocked_nd_range.cpp
 //! \brief Test for [preview] functionality
```
pavelkumbrasev, line 23:
```suggestion
//! \brief Test for [algorithms.blocked_nd_range] functionality
```

test/tbb/test_blocked_range.cpp
```diff
@@ -25,7 +25,7 @@
 #include "tbb/blocked_range2d.h"
 #include "tbb/blocked_range3d.h"
 #define TBB_PREVIEW_BLOCKED_RANGE_ND 1
-#include "tbb/blocked_rangeNd.h"
+#include "tbb/blocked_nd_range.h"
 
 //! \file test_blocked_range.cpp
 //! \brief Test for [algorithms.blocked_range] specification
```
pavelkumbrasev, line 31:
```suggestion
//! \brief Test for [algorithms.blocked_range algorithms.blocked_nd_range] specification
```

test/conformance/conformance_blocked_nd_range.cpp
akukanov: Fixed.

test/tbb/test_blocked_range.cpp
akukanov: Fixed.

pavelkumbrasev (approved):
LGTM!

pavelkumbrasev (dismissed):

include/oneapi/tbb/blocked_nd_range.h
```diff
@@ -37,45 +34,48 @@ namespace detail {
...
+    //! Type for the size of a range.
+    using size_type = typename dim_range_type::size_type;
+
+    blocked_nd_range_impl() = delete;
 
     //! Constructs N-dimensional range over N half-open intervals each represented as tbb::blocked_range<Value>.
-    blocked_rangeNd_impl(const dim_type_helper<Is>&... args) : my_dims{ {args...} } {}
+    blocked_nd_range_impl(const indexed_t<dim_range_type, Is>&... args) : my_dims{ {args...} } {}
+
+    blocked_nd_range_impl(const value_type (&size)[N], size_type grainsize = 1) :
```
kboyarinov, line 71: It seems like one of our tests identified an issue with this new constructor in the following case:
`blocked_nd_range<int, 1> br{ {0, 10, 1} }`

This code generates a compile-time error because it better matches on the new constructor instead of converting it to the single `blocked_range` object. `{0, 10, 1}` better matches on C array and generates an error that the number of elements was exceeded (expected 1, provided 3). 

We can fix it by extending this constructor to accept a C array of any length M and add constraint that M should be equal to N:

```suggestion
    template <unsigned int M, typename = typename std::enable_if<M == N>::type>
    blocked_nd_range_impl(const value_type (&size)[M], size_type grainsize = 1) :
```

In this case the `{0, 10, 3}` would first match on the array constructor and then it would be dropped because of the constraint failure.

akukanov: Another idea how to resolve: can we add, only for the case of N==1, an extra constructor taking an initializer list and delegating to a constructor with blocked_range? That would be an implementation detail I think, no spec changes required.

tobiasweinzierl80: I would personally always favour the specialisation route, i.e. the second variant.

kboyarinov, line 71:
```suggestion
#if __TBB_CLANG_VERSION < 140000
    // On clang prior to version 14.0.0, passing a single braced init list to the constructor of blocked_nd_range<T, 1>
    // matches better on the C array constructor and generates compile-time error because of unexpected size
    // Adding constraints for this constructor to force the compiler to drop it from overload resolution if the size is unexpected
    template <unsigned int M, typename = typename std::enable_if<M == N>::type>
    blocked_nd_range_impl(const value_type (&size)[M], size_type grainsize = 1) :
#else
    blocked_nd_range_impl(const value_type (&size)[N], size_type grainsize = 1) :
#endif
```

I have investigated a bit deeper and it seems like this is a Clang issue and it only takes place on versions prior to 14.0.0. GCC and Microsoft compiler works correct without any changes. 

According to this, I guess we can workaround this only for affected versions of Clang and keep the specification as-is.

include/oneapi/tbb.h
```diff
@@ -29,7 +29,7 @@
 #include "oneapi/tbb/blocked_range2d.h"
 #include "oneapi/tbb/blocked_range3d.h"
 #if TBB_PREVIEW_BLOCKED_RANGE_ND
```
kboyarinov, line 31: It seems like this macro should be removed.

test/tbb/test_blocked_range.cpp
```diff
@@ -25,10 +25,10 @@
 #include "tbb/blocked_range2d.h"
 #include "tbb/blocked_range3d.h"
 #define TBB_PREVIEW_BLOCKED_RANGE_ND 1
```
kboyarinov, line 27:
```suggestion
```

include/oneapi/tbb.h
akukanov: The patch does not remove the preview status; a separate patch is needed for this.

akukanov: OK, now the patch does remove the preview status.

pavelkumbrasev (approved):

kboyarinov (dismissed):
LGTM!

pavelkumbrasev (approved):

PR #1451, opened 2024-07-19
[try_put_and_wait] Part 10: implementation for broadcast-push buffers

kboyarinov:
### Description 
Extend flow::write_once_node and flow::overwrite_node with try_put_and_wait API + tests

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown


test/tbb/test_overwrite_node.cpp
```diff
@@ -183,6 +184,163 @@ void test_deduction_guides() {
...
+            CHECK_MESSAGE(processed_items[check_index++] == new_message, "unexpected new_message processing");
+            CHECK(check_index == processed_items.size());
+        });
+    }
+    // Test explicit clear
+    {
+        tbb::flow::graph g;
+        tbb::flow::overwrite_node<int> buffer(g);
+
+        int processed_item = -42;
```
vossmjp, line 322: I'd suggest to still use a vector for processed_items. It is possible here that you could receive more than 1 copy of wait_message in the function_node and the test would still pass. Instead if you keep it as a vector and check the size, you are sure that you only received the value once.

include/oneapi/tbb/flow_graph.h
```diff
@@ -3329,20 +3329,41 @@ class overwrite_node : public graph_node, public receiver<T>, public sender<T> {
...
     }
 
+#if __TBB_PREVIEW_FLOW_GRAPH_TRY_PUT_AND_WAIT
+    bool try_get( input_type &v, message_metainfo& metainfo ) override {
+        spin_mutex::scoped_lock l( my_mutex );
+        if (my_buffer_is_valid) {
+            v = my_buffer;
+            metainfo = my_buffered_metainfo;
+
+            // Since the successor of the node will use move semantics while wrapping the matainfo
```
vossmjp, line 3339:
```suggestion
            // Since the successor of the node will use move semantics while wrapping the metainfo
```

include/oneapi/tbb/flow_graph.h
```diff
@@ -3329,20 +3329,41 @@ class overwrite_node : public graph_node, public receiver<T>, public sender<T> {
...
-        return try_reserve(v);
-    }
-
-    bool try_get( input_type& v, message_metainfo& ) override {
-        return try_get(v);
+    bool try_reserve(T& v, message_metainfo& metainfo) override {
+        spin_mutex::scoped_lock l( my_mutex );
+        if (my_buffer_is_valid) {
+            v = my_buffer;
+            metainfo = my_buffered_metainfo;
```
vossmjp, line 3363: I suppose there is no move of the metainfo when reserved/consumed and so there is no need to increment ref count here or in consume. Is that right? It is handled in the downstream node.

vossmjp (approved):
A minor suggestion and a typo fix.

test/tbb/test_overwrite_node.cpp
kboyarinov: Done

include/oneapi/tbb/flow_graph.h
kboyarinov: Yes, move semantics optimization is only applied for try_get.

pavelkumbrasev (approved):

PR #1461, opened 2024-07-30
WIP: numa_partitioner for parallel_for.

JhaShweta1:
- Added numa_partitioner for parallel_for. files "test1.cpp" & "test.pp" will be deleted, once I add tests.
- labelling it 'WIP' as this needs to be add for scan, sort and for_each too.

### DIFF
--- a/include/oneapi/tbb/blocked_range.h
+++ b/include/oneapi/tbb/blocked_range.h
@@ -108,6 +108,14 @@ public:
         // only comparison 'less than' is required from values of blocked_range objects
         __TBB_ASSERT( !(my_begin < r.my_end) && !(r.my_end < my_begin), "blocked_range has been split incorrectly" );
     }
+  
+    // fill elements with their index values
+    void first_touch(std::vector<Value>& container) const {
+      container.resize(this->end() - this->begin());
+      for (Value i = this->begin(); i != this->end(); ++i) {
+	container[i] = i; // Set each element to its index
+        }
+    }
 
 private:
     /** NOTE: my_end MUST be declared before my_begin, otherwise the splitting constructor will break. */
--- a/include/oneapi/tbb/blocked_range2d.h
+++ b/include/oneapi/tbb/blocked_range2d.h
@@ -89,6 +89,19 @@ public:
     //! The columns of the iteration space
     const col_range_type& cols() const { return my_cols; }
 
+    // First touch method
+    template <typename Container>
+    void first_touch(Container& container) const {
+      container.resize(my_rows.end() * my_cols.end());
+      for (RowValue r = my_rows.begin(); r != my_rows.end(); ++r) {
+	for (ColValue c = my_cols.begin(); c != my_cols.end(); ++c) {
+	  container[r * my_cols.end() + c] = r * my_cols.size() + c; // Example initialization
+	}
+      }
+    }
+  
+
+
 private:
     template <typename Split>
     void do_split( blocked_range2d& r, Split& split_obj ) {
--- a/include/oneapi/tbb/blocked_range3d.h
+++ b/include/oneapi/tbb/blocked_range3d.h
@@ -100,6 +100,19 @@ public:
     //! The columns of the iteration space
     const col_range_type& cols() const { return my_cols; }
 
+     // First touch method
+    template <typename Container>
+    void first_touch(Container& container) const {
+      container.resize(my_rows.end() * my_cols.end() * my_pages.end());
+      for (PageValue p = my_pages.begin(); p != my_pages.end(); ++p) {
+	for (RowValue r = my_rows.begin(); r != my_rows.end(); ++r) {
+	  for (ColValue c = my_cols.begin(); c != my_cols.end(); ++c) {
+	    container[(r * my_cols.size() + c) * my_pages.size() + p] = (r * my_cols.size() + c) * my_pages.size() + p; 
+	  }
+	}
+      }
+    }
+
 private:
     template <typename Split>
     void do_split( blocked_range3d& r, Split& split_obj) {
--- a/include/oneapi/tbb/parallel_for.h
+++ b/include/oneapi/tbb/parallel_for.h
@@ -261,6 +261,17 @@ void parallel_for( const Range& range, const Body& body, affinity_partitioner& p
     start_for<Range,Body,affinity_partitioner>::run(range,body,partitioner);
 }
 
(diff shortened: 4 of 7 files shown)


milubin:
@JhaShweta1, could you look into adding oneTBB support for overriding the [Distributed Ranges](https://github.com/oneapi-src/distributed-ranges) (DR) approach that introduces a concept of  Distributed Data Structures (DDS). With DDS, one large flat array will be broken into segments, each allocated on the separate NUMA domain.  The motivation for introducing oneTBB overrides is to have an option for developers that use DR to achieve the best performance on CPUs by using oneTBB.

include/oneapi/tbb/blocked_range.h
```diff
@@ -108,6 +108,22 @@ class blocked_range {
         // only comparison 'less than' is required from values of blocked_range objects
         __TBB_ASSERT( !(my_begin < r.my_end) && !(r.my_end < my_begin), "blocked_range has been split incorrectly" );
     }
+  
+    // fill elements with their index values
+  void first_touch(std::vector<Value>& container) const {
```
vossmjp, line 113: I think having a first_touch function as part of the range is not the right abstraction. This looks like an algorithm. Also, the first-touch code would be application-specific and likely achieved by combining a range, partitioner and algorithm (very likely parallel_for). So I would not expect an explicit first_touch algorithm either.

include/oneapi/tbb/blocked_range2d.h
```diff
@@ -89,6 +89,19 @@ class blocked_range2d {
     //! The columns of the iteration space
     const col_range_type& cols() const { return my_cols; }
 
+    // First touch method
+    template <typename Container>
+    void first_touch(Container& container) const {
```
vossmjp, line 94: Same as previous comment,.

include/oneapi/tbb/blocked_range3d.h
```diff
@@ -100,6 +100,19 @@ class blocked_range3d {
     //! The columns of the iteration space
     const col_range_type& cols() const { return my_cols; }
 
+     // First touch method
+    template <typename Container>
+    void first_touch(Container& container) const {
```
vossmjp, line 105: Same as previous comment,.

include/oneapi/tbb/parallel_reduce.h
```diff
@@ -402,7 +402,42 @@ class lambda_reduce_body {
     }
 };
 
+template<typename BasePartitioner>
+template<typename Range, typename Body>
+void numa_partitioner<BasePartitioner>::execute_reduce(const Range& range, Body& body) const{
```
vossmjp, line 407: For now let's focus on parallel_for. We may consider parallel_reduce later.

include/oneapi/tbb/parallel_scan.h
```diff
@@ -537,6 +538,40 @@ class lambda_scan_body {
     }
 };
 
+
+  template<typename BasePartitioner>
+  template<typename Range, typename Body>
+  void numa_partitioner<BasePartitioner>::execute_scan(const Range& range, Body& body) const{
```
vossmjp, line 544: Let's skip parallel_scan completely. Already there are only a limited set of partitioners that work with parallel_scan. Let's not worry about making this work.

vossmjp (commented):
Not a full review, but comments on first-touch and which algorithms to focus on.

include/oneapi/tbb/parallel_for.h
```diff
@@ -179,6 +179,33 @@ task* start_for<Range, Body, Partitioner>::cancel(execution_data& ed) {
     return nullptr;
 }
 
+template<typename BasePartitioner>
+template<typename Range, typename Body>
+void numa_partitioner<BasePartitioner>::execute_for(const Range& range, const Body& body) const{
```
vossmjp, line 184: Is it necessary to define a member function of numa_partitioner inside of the parallel_for header. That seems very unusual.

include/oneapi/tbb/parallel_for.h
```diff
@@ -179,6 +179,33 @@ task* start_for<Range, Body, Partitioner>::cancel(execution_data& ed) {
...
+    if (range.is_divisible() && num_numa_nodes > 1) {
+        std::vector<Range> subranges;
+	split_range(range, subranges, num_numa_nodes);
+	std::vector<oneapi::tbb::task_group> task_groups(num_numa_nodes);
+	initialize_arena();
```
vossmjp, line 189: Can the task_arenas be initialized once instead of during each parallel_for execution. I would expect that a partitioner like this could be created and then passed to a number of parallel_fors, amortizing the initialization cost.

include/oneapi/tbb/partitioner.h
```diff
@@ -642,18 +649,63 @@ class static_partitioner {
...
+    std::size_t num_numa_nodes;
+    BasePartitioner& base_partitioner;
+
+public:
+  numa_partitioner() : num_numa_nodes(get_number_of_numa_nodes()), base_partitioner() {}
+  numa_partitioner(BasePartitioner& bp) : num_numa_nodes(get_number_of_numa_nodes()), base_partitioner(bp) {}
+  
+  void initialize_arena() const {
+    for (std::size_t node = 0; node < num_numa_nodes; ++node) {
+      this->arenas.emplace_back(tbb::task_arena::constraints().set_numa_id(node));
```
vossmjp, line 676: If the same instance is used across multiple parallel_fors, won't the arenas vector keep growing? I think initialize would be repeatedly invoked.

test/tbb/test_parallel_for.cpp
```diff
@@ -462,3 +462,60 @@ TEST_CASE("parallel_for constraints") {
...
+        for (size_t i = r.begin(); i != r.end(); ++i) {
+            vec[i] = 1; // Set each element to 1
+        }
+    };
+
+    tbb::affinity_partitioner ap;
+    tbb::numa_partitioner<tbb::affinity_partitioner> n_partitioner(ap);
+    
+    // Test parallel_for with numa_partitioner and a lambda body
+    parallel_for(range, body, n_partitioner);
```
vossmjp, line 508: You need a test with more than one parallel_for invocation. I think that would uncover some of the design issues.

vossmjp (changes requested):

milubin:
I second the above comment, that arenas must be initialized only once before any calls to parallel_fors.

PR #1469, opened 2024-08-05
[RFC] task_group_dynamic_dependencies

vossmjp:
### Description 
A proposal to extend task_group:

1. Extend semantics and useful lifetime of `task_handle`. We propose `task_handle` to represent tasks for the purpose of adding dependencies. The useful lifetime and semantics of `task_handle` will need to be extended to include tasks that have been submitted, are currently executing, or have been completed.
2. Add functions to set task dependencies. In the current `task_group`, tasks can only be waited for as a group and there is no direct way to add any before-after relationships between individual tasks. We will discuss options for spelling.
3. Add a function to move successors from a currently executing task to a new task. This functionality is necessary for recursively generated task graphs. This case represents a situation where it is safe to modify dependencies for an already submitted task.

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown


rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,311 @@
...
+
+## Introduction
+
+Back in 2021, during the move from TBB 2020 to the first release of oneTBB, 
+the lowest level tasking interface changed significantly and was no longer 
+promoted as a user-facing feature. Instead, the guidance since then has been 
+to use the task_group or the flow graph APIs to express patterns that were 
+previously expressed using with the lowest level tasking API. And for most 
+cases, this has been sufficient. However, there are some use cases that create 
+dynamically expanding graphs of dependent tasks are awkward to express using
```
tobiasweinzierl80, line 11: I don't get this grammar. And I think awkward is likely not a good wording.

tobiasweinzierl80: Suggestion: However, there is one use case which is not straightforward to express by the revised API: Dynamic task graphs which are not trees.

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,311 @@
...
+previously expressed using with the lowest level tasking API. And for most 
+cases, this has been sufficient. However, there are some use cases that create 
+dynamically expanding graphs of dependent tasks are awkward to express using 
+the existing interfaces and so this proposal expands tbb::task_group to make 
+additional use cases easier to express.
+
+The current class definition in the latest oneTBB specification for 
+`tbb::task_group` is shown below. Note the `defer` function because this 
+function and its return type, `task_handle`, are the foundation of our proposed
+extensions:
```
tobiasweinzierl80, line 18: Does the defer already exist? This is not clear to me.

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,311 @@
...
+        task_group();
+        task_group(task_group_context& context);
+
+        ~task_group();
+
+        template<typename Func>
+        void run(Func&& f);
+
+        template<typename Func>
+        task_handle defer(Func&& f);
```
tobiasweinzierl80, line 31: I notice that task_handle is not extensively documented atm. The docu says what it does, but not why. Can we add more docu there?

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,311 @@
...
+Dynamic tasks graphs order the execution of tasks via dependencies on the 
+completion of other tasks. They are dynamic because the creation of tasks, 
+specification of dependencies between, submission of tasks for scheduling, and
+execution of the tasks may happen concurrently and in various orders. Different
+concrete use cases have different requirements on when new tasks are created 
+and when dependencies between tasks are specified. 
+
+For the sake of discussion, let’s label four points in a task’s lifetime: 
+
+1. pre-submitted
```
tobiasweinzierl80, line 65: Would it make sense to call this deferred?

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,311 @@
...
+2. submitted for dependence tracking
```
tobiasweinzierl80, line 66: The addition "for dependence tracking" is confusing

rfcs/proposed/task_group_dynamic_dependencies/README.md
tobiasweinzierl80: Alternatively, call it "created" in line with the text below

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,311 @@
...
+a `task_handle` to `task_group::run` or `task_group::run_and_wait` only makes
+it available for dependency tracking, and does not make it immediately legal to
+execute.
+
+### Add function(s) to set dependencies.
+
+The obvious next extension is to add a mechanism for specifying dependencies
+between tasks. In the most conservative view, it should only be legal to add 
+additional predecessors / in-dependencies to tasks in the pre-submitted state. 
+After a task starts executing or is completed, it doesn’t make sense to add
```
tobiasweinzierl80, line 102: I think this is not 100% correct: If a task is already running and we add a new predecessor, we can read this as a suspend/interrupt. Isn't that what you want to add? We don't need it in our code btw.

pavelkumbrasev:
I think this proposal is lacking the final definition of `class task_handle` including all the new methods.

rfcs/proposed/task_group_dynamic_dependencies/README.md
vossmjp: I like "deferred" since it matches the defer function name. I think it will accurately imply that it is not yet submitted. I had been concerned that a submitted task that has not-yet-ready-to-run because of dependencies could also be thought of as "deferred", but on second thought, I think deferred is different enough.

rfcs/proposed/task_group_dynamic_dependencies/README.md
vossmjp: I was trying, again, to differentiate with ready-to-run. Submitted to what? Submitted to the scheduler to be tracked before executing or submitted to the ready queue.  But again, I was probably over thinking this. I agree that states of 1. deferred, 2. submitted, 3. executing and 4. completed are clear enough.

rfcs/proposed/task_group_dynamic_dependencies/README.md
vossmjp: Hmm, I was not intending to support a suspend. But yes, the language "doesn't make sense" is inaccurate since, as you just pointed out, it can make sense. I'll fix this to match my intention and not be overly broad.

rfcs/proposed/task_group_dynamic_dependencies/README.md
tobiasweinzierl: I suggest that we are careful with the term "deferred". A deferred task has a specific meaning in OpenMP. While OpenMP is not the "bible", I think it is clever to stick to their semantics or to explicitly avoid their terms where we disagree with their usage. So in this sense, my idea with "deferred" was wrong. I should have thought about this more carefully.

rfcs/proposed/task_group_dynamic_dependencies/README.md
vossmjp: Fixed

rfcs/proposed/task_group_dynamic_dependencies/README.md
vossmjp: Fixed

rfcs/proposed/task_group_dynamic_dependencies/README.md
vossmjp: Done

rfcs/proposed/task_group_dynamic_dependencies/README.md
vossmjp: Already changed it. Do you have a better suggestion instead? So maybe "created" then?

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,365 @@
...
+
+The task tree for this example matches the one shown earlier for merge-sort. 
+
+## Open Questions in Design
+
+Some open questions that remain:
+
+- Are the suggested APIs sufficient?
+- Are there additional use cases that should be considered that we missed in our analysis?
+- Are there other parts of the pre-oneTBB tasking API that developers have struggled to find a good alternative for?
```
aleksei-fedotov, line 365: These are interesting questions.

While reading this RFC, I kind of rushed proposing additional syntax sugar that besides being more user-friendly, since it seems to represent popular use cases, can save some CPU cycles. So I am posting them here for a discussion.

1. Would it be useful to have a method that simultaneously joins (or even fuses?) the instantiation of a task and transferring of the current task successors?
Something like:
```cpp
    template <typename Func>
    task_handle transfer_successors_to(Func&& f);
```
For the recursive decomposition scenario, instantiating a new task within an executing task and transferring successors to that new task seems to be the main model of writing such algorithms. Although, it seems to be not saving much (only one call to the library and assign to a couple of pointers?), there is always(?) going to be such a sequence in the code. Otherwise, how else an execution of already submitted tasks can be postponed?

Shall we also consider a question of having that API _instead_ or _in addition_ to the one proposed?

2. As for `add_predecessor(s)` and `add_successor(s)` I have a couple of thoughts.
   **a.** It seems again that it might be useful to merge instantiation of the new task handles and adding them as successors/predecessors:
    ```cpp
        template <typename Func>
        void add_predecessor(Func&& f);
    
        template <typename Func>
        void add_successor(Func&& f);
    ```
   **b.** Also, I think having an API that would allow adding more than one predecessor/successor at once can be useful, since usually a number of successors/predecessors are instantiated. I only think that we don't need to limit ourselves to only two parameters as it was proposed optionally, but allow passing of an arbitrary size of task handles or even user lambdas. Of course, a pattern of having a single task producer might be viewed as a limiting one, but there actually might be the cases where tasks cannot be made available to the scheduler (i.e. spawned) until all of them are gathered together from different producers, which essentially represents a barrier in the execution pipeline. Not to mention that the spawning of a bunch of tasks all together are done faster than regular one by one spawnings. Spawning of a bunch of tasks at once was implemented in the old TBB, as far as I remember.
So here I suggest to have something like:
    ```cpp
        template <typename... Func>
        void add_predecessors(Func&& ...f);
    
        template <typename... Func>
        void add_successor(Func&& ...f);
    ```
    However, this also seems to ask for having the `task_group::run()` method to accept an arbitrary size of task handles and/or functors. So, perhaps, it is more related to another RFC/extension...

vossmjp: I think we should start with minimal API then extend with syntactic sugar based on use cases. I suspect this will start as an experimental feature to allow some feedback on API. 

Even so, I'm also open to including these additional APIs in the initial implementation, if others think they're likely needed.

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,365 @@
...
+            // existing functions
+            task_handle();
+            task_handle(task_handle&& src);
+            ~task_handle();
+            task_handle& operator=(task_handle&& th);
+            explicit operator bool() const noexcept;
+
+            // proposed additions
+            void add_predecessor(task_handle& th);
+            void add_successor(task_handle& th);
```
akukanov, line 227: If we speak about the minimally required API, I doubt we need two variants of the same function, just with different name and the order of arguments. I would only keep one.

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,365 @@
...
+        };
+
+        void transfer_successors_to(task_handle& th);
```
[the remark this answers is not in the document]
akukanov: I would use `tbb::task` namespace, since it is already used for `suspend` (which applies to the currently running task) and `resume` functions.

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,365 @@
...
+specification of `task_group`, accessing a `task_handle` after it is submitted
+via one of the run functions is undefined behavior. Currently, therefore, a 
+`task_handle` can only represent a created task. And currently, any task
+that is run can immediately be scheduled for execution since there is no notion
+of task dependencies for task_group.
+
+The first extension is to expand the semantics and usable lifetime of
+`task_handle` so that remains valid after it is passed to run and it can 
+represent tasks in any state, including submitted, executing, and completed 
+tasks. Similarly, a `task_handle` in the submitted state may represent a task
```
akukanov, line 99: I think it will be necessary to specify in more detail what that means for a task handle to be valid and represent a task in a certain state. As a C++ object, it is valid until it is destroyed, and nothing changes in this regard I guess. But what you can and cannot do with a task handle might depend on the current state of the associated task.

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,365 @@
...
+predecessors in one call.
+
+Given two `task_handle` objects `h1` and `h2`, some possible options 
+for adding `h1` as an in-dependence / predecessor of `h2` include:
+
+- `h2.add_predecessor(h1)`
+- `h2 = defer([]() { … }, h1)`
+- `make_edge(h1, h2)`
+
+We propose including the first option. Similarly, there could be
```
akukanov, line 147: I would prefer to not add methods to `task_handle` but use "external" functions, perhaps in the `task_group` class. This would be more consistent with the current approach (`defer`, `run`, run_and_wait`) as well as with `transfer_successors_to`.

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,365 @@
...
+            tg.run(sortright);
+            tg.run(merge);
+        }
+    }
+
+The task tree for this example matches the one shown earlier for merge-sort. 
+
+## Open Questions in Design
+
+Some open questions that remain:
```
akukanov, line 361: Should there be more implementation-level questions added, such as feasibility of implementing the proposed design, performance implications, etc.?

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,365 @@
...
+A key point about this recursive parallel algorithm is that we must change 
+the predecessors of the merge tasks. But the merge tasks are already 
+submitted at the time their predecessors are modified! In the previous 
+section, we noted that updating the predecessors of a submitted task is 
+risky, because there is a potential race.  However, in the example shown 
+here, we know it’s safe to add additional predecessors to the merge task 
+because it simply cannot start executing until all its current predecessors 
+complete, and its predecessors are the tasks modifying the predecessors! 
+
+We therefore propose a very limited extension that allows the transfer of
```
aepanchi, line 197:
```suggestion
Therefore, the proposal is a limited extension that allows transferring 
```

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,365 @@
...
+the predecessors of the merge tasks. But the merge tasks are already 
+submitted at the time their predecessors are modified! In the previous 
+section, we noted that updating the predecessors of a submitted task is 
+risky, because there is a potential race.  However, in the example shown 
+here, we know it’s safe to add additional predecessors to the merge task 
+because it simply cannot start executing until all its current predecessors 
+complete, and its predecessors are the tasks modifying the predecessors! 
+
+We therefore propose a very limited extension that allows the transfer of 
+all the successors of a currently executing task to become the successors
```
aepanchi, line 198:
```suggestion
all the successors of a currently executing task to become successors 
```

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,365 @@
...
+risky, because there is a potential race.  However, in the example shown 
+here, we know it’s safe to add additional predecessors to the merge task 
+because it simply cannot start executing until all its current predecessors 
+complete, and its predecessors are the tasks modifying the predecessors! 
+
+We therefore propose a very limited extension that allows the transfer of 
+all the successors of a currently executing task to become the successors 
+of a different created task. This function can only access the successors
+of the currently executing task, and those tasks are prevented from executing 
+by a dependence on the current task itself, so we can ensure that we can safely
```
aepanchi, line 201:
```suggestion
due to dependence on the current task itself. This ensures that it is safe to 
```

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,365 @@
...
+here, we know it’s safe to add additional predecessors to the merge task 
+because it simply cannot start executing until all its current predecessors 
+complete, and its predecessors are the tasks modifying the predecessors! 
+
+We therefore propose a very limited extension that allows the transfer of 
+all the successors of a currently executing task to become the successors 
+of a different created task. This function can only access the successors
+of the currently executing task, and those tasks are prevented from executing 
+by a dependence on the current task itself, so we can ensure that we can safely 
+update the incoming dependencies for those tasks without worrying about any
```
aepanchi, line 202:
```suggestion
update the incoming dependencies for those tasks without worrying about
```

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,365 @@
...
+because it simply cannot start executing until all its current predecessors 
+complete, and its predecessors are the tasks modifying the predecessors! 
+
+We therefore propose a very limited extension that allows the transfer of 
+all the successors of a currently executing task to become the successors 
+of a different created task. This function can only access the successors
+of the currently executing task, and those tasks are prevented from executing 
+by a dependence on the current task itself, so we can ensure that we can safely 
+update the incoming dependencies for those tasks without worrying about any 
+potential race.
```
aepanchi, line 203:
```suggestion
potential race conditions. 
```

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,365 @@
...
+
+One possible spelling for this function would be `transfer_successors_to(h)`,
```
aepanchi, line 205:
```suggestion
The possible spelling for this function is `transfer_successors_to(h)`. 
```

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,365 @@
...
+We therefore propose a very limited extension that allows the transfer of 
+all the successors of a currently executing task to become the successors 
+of a different created task. This function can only access the successors
+of the currently executing task, and those tasks are prevented from executing 
+by a dependence on the current task itself, so we can ensure that we can safely 
+update the incoming dependencies for those tasks without worrying about any 
+potential race.
+
+One possible spelling for this function would be `transfer_successors_to(h)`, 
+where `h` is a `task_handle` to a created task and the
```
aepanchi, line 206:
```suggestion
Where `h` is a `task_handle` to a created task, and the 
```

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,365 @@
...
+
+Adds `th` as a successor that cannot start executing until the task represented by 
+`*this` is complete.
+
+#### void transfer_successors_to(task_handle& th);
+
+Transfers all of the successors from the currently executing task to the task 
+represented by `th`.
+
+### Small examples
```
aepanchi, line 250:
```suggestion
### Examples
```

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,365 @@
...
+
+##### A simple three task graph
```
aepanchi, line 252:
```suggestion
##### Simple Three-Nodes Task Graph
```

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,365 @@
...
+
+The example below shows a very simple graph with three nodes. The
```
aepanchi, line 254:
```suggestion
The example below shows a simple graph with three nodes. 
```

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,365 @@
...
+
+Transfers all of the successors from the currently executing task to the task 
+represented by `th`.
+
+### Small examples
+
+##### A simple three task graph
+
+The example below shows a very simple graph with three nodes. The 
+`final_task` must wait for both the `first_task` and `second_task`
```
aepanchi, line 255:
```suggestion
`final_task` must wait for  `first_task` and `second_task` 
```

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,365 @@
...
+    tg.run(first_task);
+    tg.run(second_task);
+
+    tg.wait();
+
+The dependency graph for this example is:
+
+<img src="three_task_graph.png" width=400>
+
+#### Adding predecessors in unknown states
```
aepanchi, line 279:
```suggestion
#### Predecessors in Unknown States
```

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,365 @@
...
+
+The example below shows a graph where the dependencies are determined
```
aepanchi, line 281:
```suggestion
The example below shows a graph with the dynamically determined dependencies.
```

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,365 @@
...
+    tg.wait();
+
+The dependency graph for this example is:
+
+<img src="three_task_graph.png" width=400>
+
+#### Adding predecessors in unknown states
+
+The example below shows a graph where the dependencies are determined 
+dynamically. The state of the predecessors may be unknown – they may
```
aepanchi, line 282:
```suggestion
The state of the predecessors is unknown. 
```

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,365 @@
...
+
+The dependency graph for this example is:
+
+<img src="three_task_graph.png" width=400>
+
+#### Adding predecessors in unknown states
+
+The example below shows a graph where the dependencies are determined 
+dynamically. The state of the predecessors may be unknown – they may 
+be created, submitted, executing or completed. Although not shown,
```
aepanchi, line 283:
```suggestion
Assuming that
```

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,365 @@
...
+let's assume that the user's `users::find_predecessors` function
```
aepanchi, line 284:
```suggestion
the `users::find_predecessors` function 
```

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,365 @@
...
+        tbb::task_handle new_task = tg.defer([=] { do_work(work_id); });
+
+        for (tbb::task_handle& p : users::find_predecessors(work_id)) {
+            new_task.add_predecessor(p);
+        }
+
+        tg.run(new_task);
+    }
+
+Again, the graph, as shown below, is simple. However, now we do not
```
aepanchi, line 298:
```suggestion
While the graph, as shown below, is simple, 
```

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,365 @@
...
+
+        for (tbb::task_handle& p : users::find_predecessors(work_id)) {
+            new_task.add_predecessor(p);
+        }
+
+        tg.run(new_task);
+    }
+
+Again, the graph, as shown below, is simple. However, now we do not 
+know the completion status of the predecessors. Therefore,
```
aepanchi, line 299:
```suggestion
the completion status of the predecessors is unknown. Therefore, 
```

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,365 @@
...
+        for (tbb::task_handle& p : users::find_predecessors(work_id)) {
+            new_task.add_predecessor(p);
+        }
+
+        tg.run(new_task);
+    }
+
+Again, the graph, as shown below, is simple. However, now we do not 
+know the completion status of the predecessors. Therefore, 
+for ease-of-use, a `task_handle` should be usable as a dependency
```
aepanchi, line 300:
```suggestion
for ease of use, `task_handle` should be usable as a dependency 
```

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,365 @@
...
+            new_task.add_predecessor(p);
+        }
+
+        tg.run(new_task);
+    }
+
+Again, the graph, as shown below, is simple. However, now we do not 
+know the completion status of the predecessors. Therefore, 
+for ease-of-use, a `task_handle` should be usable as a dependency 
+regardless of state of the task it represents. Any predecessor
```
aepanchi, line 301:
```suggestion
regarding the task state. Any predecessor 
```

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,365 @@
...
+        tg.run(new_task);
+    }
+
+Again, the graph, as shown below, is simple. However, now we do not 
+know the completion status of the predecessors. Therefore, 
+for ease-of-use, a `task_handle` should be usable as a dependency 
+regardless of state of the task it represents. Any predecessor 
+that is already completed when it is added as a predecessor will 
+not delay the start of the dependent task. Otherwise, end-users 
+will need to track these states explicitly.
```
aepanchi, line 304:
```suggestion
do not need to track the states explicitly.
```

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,365 @@
...
+
+        tg.run(new_task);
+    }
+
+Again, the graph, as shown below, is simple. However, now we do not 
+know the completion status of the predecessors. Therefore, 
+for ease-of-use, a `task_handle` should be usable as a dependency 
+regardless of state of the task it represents. Any predecessor 
+that is already completed when it is added as a predecessor will 
+not delay the start of the dependent task. Otherwise, end-users
```
aepanchi, line 303:
```suggestion
not delay the start of the dependent task. Otherwise, the end users 
```

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,365 @@
...
+know the completion status of the predecessors. Therefore, 
+for ease-of-use, a `task_handle` should be usable as a dependency 
+regardless of state of the task it represents. Any predecessor 
+that is already completed when it is added as a predecessor will 
+not delay the start of the dependent task. Otherwise, end-users 
+will need to track these states explicitly.
+
+<img src="unknown_states.png" width=400>
+
+#### And example of recursive decomposition
```
aepanchi, line 308:
```suggestion
#### Recursive Decomposition
```

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,365 @@
...
+
+This example is a version of merge-sort (with many of the details left out).
```
aepanchi, line 310:
```suggestion
This example shows a version of merge sort, with mane details left out for simplicity. 
```

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,365 @@
...
+that is already completed when it is added as a predecessor will 
+not delay the start of the dependent task. Otherwise, end-users 
+will need to track these states explicitly.
+
+<img src="unknown_states.png" width=400>
+
+#### And example of recursive decomposition
+
+This example is a version of merge-sort (with many of the details left out). 
+Assume that there is an initial task that executes the function shown
```
aepanchi, line 311:
```suggestion
Assume an initial task executes the function shown
```

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,365 @@
...
+not delay the start of the dependent task. Otherwise, end-users 
+will need to track these states explicitly.
+
+<img src="unknown_states.png" width=400>
+
+#### And example of recursive decomposition
+
+This example is a version of merge-sort (with many of the details left out). 
+Assume that there is an initial task that executes the function shown 
+below as its body, and the function implements that task, and also serves
```
aepanchi, line 312:
```suggestion
This function implements that task and serves 
```

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,365 @@
...
+will need to track these states explicitly.
+
+<img src="unknown_states.png" width=400>
+
+#### And example of recursive decomposition
+
+This example is a version of merge-sort (with many of the details left out). 
+Assume that there is an initial task that executes the function shown 
+below as its body, and the function implements that task, and also serves 
+as the body for the recursively decomposed pieces. The beginning and
```
aepanchi, line 313:
```suggestion
as the body for the recursively decomposed pieces. 
```

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,365 @@
...
+end of the sequence are represented by `b` and `e`, and much of the
```
aepanchi, line 314:
```suggestion
The range of the sequence is defined by `b` (beginning) and `e` (end). 
```

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,365 @@
...
+(unimportant) details of the implementation of merge-sort is hidden
```
aepanchi, line 315:
```suggestion
Most of the implementation details of merge sort are abstracted into the following helper functions:
```

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,365 @@
...
+            // task and its successors
+            tbb::transfer_successors_to(merge);
+
+            tg.run(sortleft);
+            tg.run(sortright);
+            tg.run(merge);
+        }
+    }
+
+The task tree for this example matches the one shown earlier for merge-sort.
```
aepanchi, line 357:
```suggestion
This task tree matches the one shown earlier for merge-sort. 
```

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,365 @@
...
+the lowest level tasking interface changed significantly and was no longer 
+promoted as a user-facing feature. Instead, the guidance since then has been 
+to use the `task_group` or the flow graph APIs to express patterns that were 
+previously expressed using with the lowest level tasking API. And for most 
+cases, this has been sufficient. However, there is one use case which is not 
+straightforward to express by the revised API: Dynamic task graphs which are 
+not trees. This proposal expands `tbb::task_group` to make additional use cases 
+easier to express.
+
+The class definition from section **[scheduler.task_group]** in the oneAPI
```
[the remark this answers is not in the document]
vossmjp: done

rfcs/proposed/task_group_dynamic_dependencies/README.md
vossmjp: Done

rfcs/proposed/task_group_dynamic_dependencies/README.md
vossmjp: What would the signature of such a function look like? The nice part of it being in `task_handle` is that `h2.add_predecessor(h1)` is easy to understand -- h1 becomes a predecessor of h2.  As part of `task_group`, the order becomes less clear: `tg.add_predecessor(h1,h2)` might be read as adding two predecessors to `task_group` `tg`.

akukanov: > What would the signature of such a function look like?

Similar to `make_edge`, but perhaps with a different name: `connect`, `set_order`, `set_dependency`, etc., with the left-to-right ordering semantics.

Currently, a `task_handle` is what its name says - a handle, an object that might represent a created task or be empty. It cannot even be copied, only moved.

The proposal suggests that a `task_handle` should represent a task at any state, so should perhaps internally track the task state somehow - that's OK. We can even think of a method to query the task state; that would also be OK.

But making the handle also tracking task dependencies, and so serving as a part of the scheduling system, rather than just an object to be scheduled, does not sound right to me. Conceptually, I see scheduling and dependency management as functions of a task group.

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,378 @@
...
+
+## Open Questions in Design
+
+Some open questions that remain:
+
+- Are the suggested APIs sufficient?
+- Should we add a function to adds more than one predecessor as single call, such as `add_predecessors`?
+- Should we add functions that merge creation and definition of predecessor tasks, such as
+`template <typename Func> add_predecessor(Func&& f);`.
+- Are there additional use cases that should be considered that we missed in our analysis?
```
akukanov, line 375: I can suggest a few more potentially interesting cases:
- a wavefront, where tasks have multiple predecessors and multiple successors;
- a two-stage parallel scan (reduce-then-scan), with left-to-right propagation of the accumulated prefix sum;
- non-trivial divide-and-conquer patterns, such as Matteo Frigo's algorithm for N-Body problem (see e.g. https://dspace.mit.edu/bitstream/handle/1721.1/122680/6-172-fall-2010/contents/projects/MIT6_172F10_proj4_1.pdf)

These seem to be interesting examples of "dynamic task graphs that are not trees", as stated in the introduction. I do not suggest that we must support these patterns, but it is interesting to see if we could, and what would be needed for that.

akukanov (commented):
I support this proposal in principle, and would approve it up to but not including the proposed API changes.
The API changes seem both insufficient (see the comment about the run function semantics) and arguable (when it comes to the way to set dependencies). I would not approve it even for an experimental API, at least until alternative API semantics are considered and compared.

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,378 @@
...
+    tbb::task_handle second_task = tg.defer([&] { /* task body */ });
+    tbb::task_handle final_task = tg.defer([&] { /* task body */ });
+
+    final_task.add_predecessor(first_task);
+    final_task.add_predecessor(second_task);
+
+    // order of submission is not important
+    tg.run(final_task);
+    tg.run(first_task);
+    tg.run(second_task);
```
kboyarinov, line 283:
```suggestion
    tg.run(std::move(final_task));
    tg.run(std::move(first_task));
    tg.run(std::move(second_task));
```

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,378 @@
...
+logic, the tasks that must complete before the new work can start.
+
+    void add_another_task(tbb::task_group& tg, int work_id) {
+        tbb::task_handle new_task = tg.defer([=] { do_work(work_id); });
+
+        for (tbb::task_handle& p : users::find_predecessors(work_id)) {
+            new_task.add_predecessor(p);
+        }
+
+        tg.run(new_task);
```
kboyarinov, line 305:
```suggestion
        tg.run(std::move(new_task));
```

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,378 @@
...
+            merge.add_predecessor(sortleft);
+            merge.add_predecessor(sortright);
+
+            // insert new subgraph between currently executing
+            // task and its successors
+            tbb::task_group::transfer_successors_to(merge);
+
+            tg.run(sortleft);
+            tg.run(sortright);
+            tg.run(merge);
```
kboyarinov, line 361:
```suggestion
            tg.run(std::move(sortleft));
            tg.run(std::move(sortright));
            tg.run(std::move(merge));
```

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,378 @@
...
+3. **Executing:** has started executing its body but is not yet complete.
+4. **Completed:** fully executed to completion.
+
+In the current `task_group` specification, the `task_group::defer` function
+already allows separate task creation and submission.
+`task_group::defer` returns a `tbb::task_handle` that represents a created
+task. A created task remains created until it is submitted through
+`task_group::run` or `task_group::run_and_wait`. The current
+`task_group` specification treats accessing a `task_handle` after it is submitted
+via one of the run functions as undefined behavior. Therefore, a
```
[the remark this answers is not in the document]
vossmjp: Yes, needs to be added.

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,378 @@
...
+does not support task dependencies, any task that is run can be immediately
+scheduled for execution without considering dependencies.
+
+The first extension is to expand the semantics and usable lifetime of
+`task_handle` so that it can used as a predecessor to other tasks even after it is
+passed to run. The handle will track the state of its associated task and may represent
+a task that is created, submitted, executing, or completed. Similarly, a submitted `task_handle`
+may represent a task that depends on predecessors that must complete before it can execute.
+In that case, passing a `task_handle` to `task_group::run` or `task_group::run_and_wait` only makes
+it available for dependency tracking but does not make it immediately eligible for execution.
```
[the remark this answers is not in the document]
vossmjp: Yes, run and enqueue functions are not yet well covered.

rfcs/proposed/task_group_dynamic_dependencies/README.md
vossmjp: Yes, more example would be good. In particular, we need to see if the very limited ways to add and modified predecessors is sufficient. And how lifetimes of task handles will be managed.

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,378 @@
...
+of the currently executing task, and those tasks are prevented from executing 
+by a dependence on the current task itself, so we can ensure that we can safely 
+update the incoming dependencies for those tasks without worrying about
+potential race conditions. 
+
+One possible spelling for this function would be `transfer_successors_to(h)`, 
+Where `h` is a `task_handle` to a created task, and the 
+`transfer_successors_to` function must be called from within a task. Calling
+this function from outside a task or passing anything other than a `task_handle`
+representing a task in the created state is undefined behavior.
```
kboyarinov, line 203: I think we need to add information about possible limitations (or lack of limitations) for the task states represented by `task_handle` to which we are transferring the successors. Does it sufficient to allow only the handles representing created tasks, or other 3 states are also allowed (i.e. transferring successors to the task_handle that was already submitted to `task_group::run`).

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,378 @@
+# Extend ``task_group`` for Dynamic Task Dependencies
```
kboyarinov, line 1: I have noticed one thing that is not related to the dependencies between tasks, but relates into covering the migration from the old tasking API to another, so I decided to add it as a comment here.

Let's consider the recursive Fibonacci example rewritten using the API proposed in this RFC (the splitting stage):

```cpp
long* left = new long(0);
long* right = new long(0);

tbb::task_handle fib_left = tg.defer([&tg, num, left] {
    recursive_fib(tg, num - 2, *left);
});
tbb::task_handle fib_right = tg.defer([&tg, num, right] {
    recursive_fib(tg, num - 1, *right);
});
tbb::task_handle fib_sum = tg.defer([&result, left, right] {
    result = *left + *right;
    delete left;
    delete right;
});
```
|
The main difference between merge sort and this example is some data that is required for executing the task (`left` and `right` - the placeholders for partial results of Fibonacci calculations on leaft).
Since the lifetime of this data should be preserved until the sum task is executed, it cannot be placed on stack of current function and needs to be allocated dynamically. 
Back to old TBB, this data was placed inside of the corresponding task that provides the required lifetime guarantees.
The question is do we need to extend the `task_handle` API somehow to allow putting the additional data to the task.

vossmjp: I agree that this should be explored.  Maintaining the lifetime of referenced objects is not a new issue, since the same problem exists for task_group::run and task_arena::enqueue, and even arguable for objects pass through a flow graph. That said, the old pre-oneTBB tasking interface provided a more obvious way to do this for task graphs via task objects.

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,366 @@
...
+
+    void add_another_task(tbb::task_group& tg, int work_id) {
+        tbb::task_handle new_task = tg.defer([=] { do_work(work_id); });
+
+        for (tbb::task_handle& p : users::find_predecessors(work_id)) {
+            new_task.add_predecessor(p);
+        }
+
+        tg.run(std::move(new_task));
+    }
```
akukanov, line 280:
```suggestion
    tbb::task_handle add_another_task(tbb::task_group& tg, int work_id) {
        tbb::task_handle new_task = tg.defer([=] { do_work(work_id); });

        for (tbb::task_handle& p : users::find_predecessors(work_id)) {
            new_task.add_predecessor(p);
        }
        tg.run(new_task);
        return new_task;
    }
```
The suggested change does not move the created task handle to `run()`, instead returns it to the caller. I think that would be practically necessary for `users::find_predecessors` to work,

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,366 @@
...
+        }
+    }
+
+This task tree matches the one shown earlier for merge-sort. 
+
+## Open Questions in Design
+
+Questions that should be addressed before moving into `experimental` and providing a preview feature:
+- Are the suggested APIs sufficient?
+- Are there additional use cases that should be considered that we missed in our analysis?
```
akukanov, line 345: The APIs seem sufficient for the considered use cases, except for semantical changes that need to be done. Some additional use cases to analyze were proposed, and currently are targeted as "post-experimental" work. I think the first question can be closed, and the second one can be either also closed or moved to that later stage.

rfcs/proposed/task_group_dynamic_dependencies/README.md
kboyarinov: Am I correct that `new_task` can be useful for the caller in terms of `users::find_predecessors` because it allows to make `new_task` visible for other calls of `add_another_task` and makes it possible to use is as a predecessor for other tasks?

rfcs/proposed/task_group_dynamic_dependencies/README.md
kboyarinov: Applied

rfcs/proposed/task_group_dynamic_dependencies/README.md
akukanov: Right, that's the idea.

akukanov (approved):

akukanov:
Though I approved the PR, the RFC process says there should be two approvals from maintainers - so formally this approval does not even count :)

rfcs/proposed/task_group_dynamic_dependencies/README.md
vossmjp: We have decided to split RFC into umbrella describing goals and then specific API proposals to follow.

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,369 @@
...
+            tbb::task_handle sortleft =
+                tg.defer([lb, le, &tg] {
+                    merge_sort(tg, lb, le);
+                });
+            tbb::task_handle sortright =
+                tg.defer([rb, re, &tg] {
+                    merge_sort(tg, rb, re);
+                });
+            tbb::task_handle merge =
+            tg.defer([lb, le, rb, re, &tg] {
```
aleksei-fedotov, line 326:
```suggestion
                tg.defer([lb, le, rb, re, &tg] {
```

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,369 @@
...
+            merge.add_predecessor(sortleft);
+            merge.add_predecessor(sortright);
+
+            // insert new subgraph between currently executing
+            // task and its successors
+            tbb::task_group::transfer_successors_to(merge);
+
+            tg.run(std::move(sortleft));
+            tg.run(std::move(sortright));
+            tg.run(std::move(merge));
```
aleksei-fedotov, line 340: As far as I understand, the `sortleft` and `sortright` can start executing right away after they are "moved into" `task_group` because they don't have predecessors. I wonder if there could be a usability concern similar to one that might have been with the `transfer_successors_to` function if we introduced it differently. We are introducing that extension with exactly that semantics because it allows to avoid potential data races between newly created predecessors and a successor task which has not yet started executing. I mean that in this particular example there is nothing to race between currently executing task and `sortleft`/`sortright` tasks, but in real example there might be some shared data between all these tasks, and the existence of potential data race might not be that obvious to the user. I guess even the `merge` task, which was "moved into" `tg` the last, might have been completed (including its successors that have just been moved to it? and successors of successors, and so on ...) before this `tg.run(std::move(merge))` call returns to the current task.

We should choose: either defer execution of the newly created tasks until current one completes ("insert new subgraph in between" for real) and hence reduce potential parallelism in the system or allow (perhaps not that obvious) races.

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,369 @@
...
+task. A created task remains created until it is submitted through
+`task_group::run` or `task_group::run_and_wait`. The current
+`task_group` specification treats accessing a `task_handle` after it is submitted
+via one of the run functions as undefined behavior. Therefore, a
+`task_handle` represents a created task only. Furthermore, since `task_group`
+does not support task dependencies, any task that is run can be immediately
+scheduled for execution without considering dependencies.
+
+The first extension is to expand the semantics and usable lifetime of
+`task_handle` so that it can used as a predecessor to other tasks even after it is
```
aleksei-fedotov, line 115: A typo?
```suggestion
`task_handle` so that it can be used as a predecessor to other tasks even after it is
```

rfcs/proposed/task_group_dynamic_dependencies/README.md
```diff
@@ -0,0 +1,369 @@
...
+### Add Functions to Set Dependencies.
+
+The next logical extension is to add a mechanism for specifying dependencies
+between tasks. In the most conservative view, it should only be allowed to add
+predecessors (in-dependencies) to tasks in the created state.
+After a task starts, adding more predecessors is irrelevant,
+since it’s too late to delay the task execution.
+
+It might be reasonable to add new predecessors to a task that is
+currently executing if it is suspended until these additional dependencies complete.
```
aleksei-fedotov, line 142: Alternatively, the "suspension support" can be added via the `tg.wait_for(task_handle)` future extension.

aleksei-fedotov (approved):
Looks good to me! My comments above are for consideration in future RFCs. It is not necessary to address them in this one.

rfcs/proposed/task_group_dynamic_dependencies/README.md
akukanov: The possibility of data races you describe is not new. With or without the new APIs, any `task_group` task can get a reference to that very group and start new tasks via `run`, not even using task handles. If data are shared between these tasks, users are responsible to prevent data races.

PR #1477, opened 2024-08-13
Make thread registration in local_wait_for_all

isaevil:
### Description 
Current implementation of thread registration is only focused on worker threads to be registered in TCM. When external threads are participating in work, they do not register thus TCM won't have enough information on resource usage for better distribution. 

`local_wait_for_all` function seems to be a common spot for both external and worker threads to perform thread registration. So, basically whole implementation is to simply call `register_thread` only on first thread enter to `local_wait_for_all` and `unregister_thread` on last thread exit.

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

@pavelkumbrasev @vossmjp @aleksei-fedotov

### DIFF
--- a/src/tbb/tcm_adaptor.cpp
+++ b/src/tbb/tcm_adaptor.cpp
@@ -170,7 +170,7 @@ public:
         __TBB_ASSERT_EX(res == TCM_RESULT_SUCCESS, nullptr);
     }
 
-    void init(d1::constraints& constraints) {
+    void init(tcm_client_id_t client_id, d1::constraints& constraints) {
         __TBB_ASSERT(tcm_request_permit, nullptr);
         __TBB_ASSERT(tcm_deactivate_permit, nullptr);
 
@@ -190,6 +190,12 @@ public:
 
         my_permit_request.min_sw_threads = 0;
         my_permit_request.max_sw_threads = 0;
+        my_permit_request.flags.request_as_inactive = 1;
+
+        tcm_result_t res = tcm_request_permit(client_id, my_permit_request, this, &my_permit_handle, nullptr);
+        __TBB_ASSERT_EX(res == TCM_RESULT_SUCCESS, nullptr);
+
+        my_permit_request.flags.request_as_inactive = 0;
     }
 
     void register_thread() override {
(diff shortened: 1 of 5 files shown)


isaevil:
@vossmjp @pavelkumbrasev @aleksei-fedotov There are several reasons behind choosing this patch instead of #1363:
1. `on_scheduler_entry` for `tbb::task_scheduler_observer` is not always invoked for external threads under implicit arena (only when stealing tasks or getting critical tasks). In order to make the registration work all the time it should be called unconditionally which, as I understand, is not initially intended behavior.
2. Implementation in #1363 tries to put "the registration observer" in arena's observers list during `r1::arena` initialization (to be precise, during initialization of `pm_client` structure for `r1::arena`) and doing so requires [extra entry point](https://github.com/oneapi-src/oneTBB/pull/1363/files#diff-55e778c918de9de39ed84b8a5bbc3b4f291df6398e0f1083227ec30831991d26R261):
    1. Trying to avoid using extra entry point will result in completely restructuring existing internal interfaces (since for registration we need to support both implicit and explicit arenas)
    2.  I guess another way is to do a lazy initialization, but I am not sure if it is an optimal solution.

Current patch seems to not have those issues, but possible concerns are: is it OK to have `local_wait_for_all` as a place to do such things and overall quality of such change.

src/tbb/tcm_adaptor.cpp
```diff
@@ -190,6 +190,12 @@ class tcm_client : public pm_client {
 
         my_permit_request.min_sw_threads = 0;
         my_permit_request.max_sw_threads = 0;
+        my_permit_request.flags.request_as_inactive = 1;
```
aleksei-fedotov, line 193:
```suggestion
        my_permit_request.flags.request_as_inactive = true;
```

src/tbb/tcm_adaptor.cpp
```diff
@@ -190,6 +190,12 @@ class tcm_client : public pm_client {
...
+
+        tcm_result_t res = tcm_request_permit(client_id, my_permit_request, this, &my_permit_handle, nullptr);
+        __TBB_ASSERT_EX(res == TCM_RESULT_SUCCESS, nullptr);
+
+        my_permit_request.flags.request_as_inactive = 0;
```
aleksei-fedotov, line 198:
```suggestion
        my_permit_request.flags.request_as_inactive = false;
```

aleksei-fedotov (approved):
Looks good to me!

PR #1478, opened 2024-08-13
Fix tbb malloc proxy on 14.30.33807

lplewa:
In this version of the ucrtbase.dll, the assembly of the _msize function has changed.

The first instruction of this function is a relative jump (0xE9) followed by a 4-byte offset. This offset has changed, but we can safely assume that 0xE9 followed by any offset value is correct, especially since we have special handling for the 0xE9 instruction in the InsertTrampoline() function.

The 0xCC is an INT 3 instruction, so it's a free space to use.

### Description 

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

@pbalcer @Alexandr-Konovalov

### DIFF
--- a/src/tbbmalloc_proxy/proxy.cpp
+++ b/src/tbbmalloc_proxy/proxy.cpp
@@ -431,14 +431,12 @@ void __TBB_malloc__free_base(void *ptr)
 const char* known_bytecodes[] = {
 #if _WIN64
 //  "========================================================" - 56 symbols
+    "E9********CCCC",         // multiple - jmp(0xE9) with address followed by empty space (0xCC - INT 3)
     "4883EC284885C974",       // release free()
     "4883EC284885C975",       // release _msize()
     "4885C974375348",         // release free() 8.0.50727.42, 10.0
-    "E907000000CCCC",         // release _aligned_msize(), _aligned_free() ucrtbase.dll
     "C7442410000000008B",     // release free() ucrtbase.dll 10.0.14393.33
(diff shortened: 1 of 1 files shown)


src/tbbmalloc_proxy/proxy.cpp
```diff
@@ -434,11 +434,9 @@ const char* known_bytecodes[] = {
     "4883EC284885C974",       // release free()
     "4883EC284885C975",       // release _msize()
     "4885C974375348",         // release free() 8.0.50727.42, 10.0
-    "E907000000CCCC",         // release _aligned_msize(), _aligned_free() ucrtbase.dll
+    "E9********CCCC",         // release _aligned_msize(), _aligned_free() ucrtbase.dll and release _msize() ucrtbase.dll 10.0.14393.33
```
Alexandr-Konovalov, line 437: Probably the explanation from the commit message deserves to be added in the comment.

ldorau (approved):
LGTM when the https://github.com/oneapi-src/oneTBB/pull/1478#discussion_r1715556447 issue is fixed

src/tbbmalloc_proxy/proxy.cpp
lplewa: Done

Alexandr-Konovalov (approved):
LGTM.

PR #1488, opened 2024-08-22
fix coverity issue in tbbmalloc HugePagesStatus

lplewa:
### Description 
page_size variable was set to -1024 if hugepages was not used, while being unsigned type. It did not matter as it was not used, when hugepages are not used, but it is better to set it just to 0.

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/src/tbbmalloc/tbbmalloc_internal.h
+++ b/src/tbbmalloc/tbbmalloc_internal.h
@@ -496,7 +496,11 @@ private:
         MALLOC_ASSERT(!pageSize, "Huge page size can't be set twice. Double initialization.");
 
         // Initialize object variables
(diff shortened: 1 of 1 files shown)


Alexandr-Konovalov (approved):
LGTM, another possibility is to modify `pageSize` only when `hugePageSize` got some valid value.

lplewa:
> LGTM, another possibility is to modify `pageSize` only when `hugePageSize` got some valid value.

it is zero-initialized anyway - so doesn't matter anyway.

PR #1490, opened 2024-08-23
Fix coverity issue in BitMaskMax

lplewa:
### Description 
It should never happen but coverity do not know it. Assert should enough here

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/src/tbbmalloc/tbbmalloc_internal.h
+++ b/src/tbbmalloc/tbbmalloc_internal.h
@@ -232,9 +232,13 @@ template<unsigned NUM>
 class BitMaskMax : public BitMaskBasic<NUM> {
 public:
     void set(size_t idx, bool val) {
+        MALLOC_ASSERT(NUM >= idx + 1, ASSERT_TEXT);
+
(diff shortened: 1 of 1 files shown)


src/tbbmalloc/tbbmalloc_internal.h
```diff
@@ -235,6 +235,8 @@ class BitMaskMax : public BitMaskBasic<NUM> {
         BitMaskBasic<NUM>::set(NUM - 1 - idx, val);
```
Alexandr-Konovalov, line 235: For the reason of symmetry, is it making sense to have similar assert in the `set()` as well? Because now a random reader may ask "Why only in `getMaxTrue()`"?

lplewa: Done

Alexandr-Konovalov (approved):
LGTM

KFilipek (approved):

ldorau (approved):

PR #1491, opened 2024-08-23
Fix coverity issue in block::initEmptyBlock

lplewa:
### Description 
getObjectSize should not return values bigger then 2^16-1. We are assigning it's return value, which is 32 bit to 16 bit one, so it is good to assert it anyway.

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/src/tbbmalloc/frontend.cpp
+++ b/src/tbbmalloc/frontend.cpp
@@ -817,6 +817,7 @@ unsigned int getSmallObjectIndex(unsigned int size)
 /*
  * Depending on indexRequest, for a given size return either the index into the bin
  * for objects of this size, or the actual size of objects in this bin.
+ * TODO: Change return type to unsigned short.
  */
 template<bool indexRequest>
 static unsigned int getIndexOrObjectSize (unsigned int size)
@@ -1581,6 +1582,7 @@ void Block::initEmptyBlock(TLSData *tls, size_t size)
     unsigned int objSz = getObjectSize(size);
 
     cleanBlockHeader();
+    MALLOC_ASSERT(objSz <= USHRT_MAX, "objSz must not be less 2^16-1");
     objectSize = objSz;
     markOwned(tls);
     // bump pointer should be prepared for first allocation - thus mode it down to objectSize


Alexandr-Konovalov:
If result of `getObjectSize()` must fit to `uint16_t`, why not change it return type to `uint16_t`? And move the related `assert` inside.

lplewa:
> If result of `getObjectSize()` must fit to `uint16_t`, why not change it return type to `uint16_t`? And move the related `assert` inside.

Because it is scary change to do few days before deadline.  To do it correctly we should change it from unsigned int to unsigned short in few places. TBH i would rewrite this function anyway at it has other issues, but few days before deadline, i just want fix coverity and do not create new issues. 

I checked that this code is correct, and added assert just to make coverity also aware that it code is fine.

Alexandr-Konovalov:
> Because it is scary change to do few days before deadline.
To late in the release circle is a good point. Could the TODO be added, to keep the intention explicit?

I'm just curious, is moving the assert to `getObjectSize()` body can make coverity happy? Because for a human reader it may be unclear, why on another call path a result of `getObjectSize()` is not covered by assert.

lplewa:
> > Because it is scary change to do few days before deadline.
> 
> To late in the release circle is a good point. Could the TODO be added, to keep the intention explicit?
> 
> I'm just curious, is moving the assert to `getObjectSize()` body can make coverity happy? Because for a human reader it may be unclear, why on another call path a result of `getObjectSize()` is not covered by assert.

I added TODO.

IMHO readability depends how you read this code. If thru GH maybe. If you use IDE then you can check that block::objectSize is uint_16_t so it make sense to assert before assessment were you do "type cast".

Alexandr-Konovalov:
> > > Because it is scary change to do few days before deadline.
> > 
> > 
> > To late in the release circle is a good point. Could the TODO be added, to keep the intention explicit?
> > I'm just curious, is moving the assert to `getObjectSize()` body can make coverity happy? Because for a human reader it may be unclear, why on another call path a result of `getObjectSize()` is not covered by assert.
> 
> I added TODO.
> 
> IMHO readability depends how you read this code. If thru GH maybe. If you use IDE then you can check that block::objectSize is uint_16_t so it make sense to assert before assessment were you do "type cast".

Sorry, I'm not totally understand you position wrt moving the assert inside `getObjectSize()` body. Is it too late to run Coverity once again, so we don't want to risk not to resolve the Coverity diagnostics?

lplewa:
> > > > Because it is scary change to do few days before deadline.
> > > 
> > > 
> > > To late in the release circle is a good point. Could the TODO be added, to keep the intention explicit?
> > > I'm just curious, is moving the assert to `getObjectSize()` body can make coverity happy? Because for a human reader it may be unclear, why on another call path a result of `getObjectSize()` is not covered by assert.
> > 
> > 
> > I added TODO.
> > IMHO readability depends how you read this code. If thru GH maybe. If you use IDE then you can check that block::objectSize is uint_16_t so it make sense to assert before assessment were you do "type cast".
> 
> Sorry, I'm not totally understand you position wrt moving the assert inside `getObjectSize()` body. Is it too late to run Coverity once again, so we don't want to risk not to resolve the Coverity diagnostics?

This is one argument.
But main argument is that, in my option, we should assert it just before type conversion. So as we are assigning int to short, we are doing assertion that it is safe. 

Also sure if you seen `getObjectSize()` function but it is literally `return getIndexOrObjectSize<false>(size);` so assert will have no sense here. we could put it to `getIndexOrObjectSize()` then this function is a mess, with multiple returns, so there in no good place to put this assert.

Alexandr-Konovalov (approved):
LGTM.

KFilipek (approved):
Generally, it breaks nothing, but getIndexOrObjectSize() should get more error handling to resolve the issue:
image

ldorau (approved):

PR #1492, opened 2024-08-23
Fix coverity issue in the alignToBin

lplewa:
### Description 
Those values should be always >=0 but coverity do not know it. Assert should fix it.

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown


omalyshe:
Note that copyright check failed. Need to update the year.

src/tbbmalloc/large_objects.h
```diff
@@ -81,7 +81,13 @@ struct HugeBinStructureProps {
 
     static size_t alignToBin(size_t size) {
         MALLOC_ASSERT(size >= StepFactor, "Size must not be less than the StepFactor");
-        size_t minorStepExp = BitScanRev(size) - StepFactorExp;
+
+        int sizeExp = (int)BitScanRev(size);
+        MALLOC_ASSERT(sizeExp >= 0, "A shift amount (sizeExp) must not be negative");
+
+        int minorStepExp = sizeExp - StepFactorExp;
+        MALLOC_ASSERT(minorStepExp >= 0, "A shift amount (minorStepExp) must not be negative");
```
Alexandr-Konovalov, line 89: Can be comments re-worded a little? To me, the reason here is that log() is monotonic, so after `size >= StepFactor` we must have `log(size) >= log(StepFactor)`, while you just postulate that something is true.

lplewa: You are right - this code is correct. What coverity do not know that:
- BitScanRev(size) cannot return -1 as `size >=stemfactor >0`
- `StepFactorExp = log2(StepFactor)` as it is a separate variable, and coverity is not good at math to find it out.

So we need those assert just to show coverity that code is correct. 

Also this asserts are literal copied from function below (and was added probably for the same reason)

lplewa:
> Note that copyright check failed. Need to update the year.

Done

src/tbbmalloc/large_objects.h
lplewa: I changed this test. Not sure if it is exactly what you want - if not please give me an example what you prefer in those asserts.

src/tbbmalloc/large_objects.h
```diff
@@ -81,18 +81,25 @@ struct HugeBinStructureProps {
 
     static size_t alignToBin(size_t size) {
         MALLOC_ASSERT(size >= StepFactor, "Size must not be less than the StepFactor");
-        size_t minorStepExp = BitScanRev(size) - StepFactorExp;
+
+        int sizeExp = (int)BitScanRev(size);
+        MALLOC_ASSERT(sizeExp >= 0, "Size must be greater than 0");
```
Alexandr-Konovalov, line 86: Can we re-use your wording like "BitScanRev cannot return -1 as size >=stepfactor >0"?

src/tbbmalloc/large_objects.h
```diff
@@ -81,18 +81,25 @@ struct HugeBinStructureProps {
...
+        MALLOC_ASSERT(sizeExp >= StepFactorExp, "sizeExp must not be less than the StepFactorExp");
```
Alexandr-Konovalov, line 87: Can it be like "sizeExp >= StepFactorExp, because size >= StepFactor"

Alexandr-Konovalov (commented):
As we have an explanation why something must be so and so, why not add that explanation in the assert's text?

src/tbbmalloc/large_objects.h
lplewa: Done

src/tbbmalloc/large_objects.h
lplewa: Done

Alexandr-Konovalov (approved):
LGTM.

KFilipek:
This change would not fix the issue. I've tried to run Coverity on my own and the results are following:
Before:
image
With proposed change:
image

KFilipek:
@Alexandr-Konovalov I've compiled it as a RelWithDebInfo and Łukasz correct me, that it should be compiled in Debug mode. Generally, it doesn't solve the issue, only way to workaround is to return INT_MAX.
Otherwise, it will mark those MALLOC_ASSERT's and ignore them:
image

ldorau:
IMHO if Coverity still sees an issue here it should be just marked as a false-positive in Coverity.

PR #1499, opened 2024-08-29
Fix futex support on OpenBSD

brad0:
### Description 
OpenBSD no longer supports direct syscalls.

https://github.com/openbsd/src/commit/cafeb892b121ee89c39c2b940e8ccd6950f50009

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown

### DIFF
--- a/src/tbb/semaphore.h
+++ b/src/tbb/semaphore.h
@@ -1,5 +1,5 @@
 /*
-    Copyright (c) 2005-2022 Intel Corporation
+    Copyright (c) 2005-2024 Intel Corporation
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


pavelkumbrasev:
Please correct the copyright year

akukanov:
@pavelkumbrasev Are systems that do not have `futex` still in support? If not, it might make sense to keep only the new implementation. Or maybe replace the check for OpenBSD with a check for presence of `futex` if such can be made.

pavelkumbrasev:
> @pavelkumbrasev Are systems that do not have `futex` still in support?

That's a good question we have never discussed this specific requirement: `#if defined(SYS_futex)`

brad0:
> That's a good question we have never discussed this specific requirement: `#if defined(SYS_futex)`

I was looking at that and although that does apply for OpenBSD at the moment that isn't necessarily the case in the future.

brad0:
@pavelkumbrasev Let me know if updating the copyright alone is Ok or if you want to go in some other direction based off of this.

pavelkumbrasev:
Since futex is still present on the OS I think this approach is fine. The only question: will this approach affect older versions of OpenBSD or it will work just fine?
(Copyright year should be fixed)

brad0:
> Since futex is still present on the OS I think this approach is fine. The only question: will this approach affect older versions of OpenBSD or it will work just fine? (Copyright year should be fixed)

It is fine for any version of OpenBSD that has futex support.

brad0:
@pavelkumbrasev Ping.

pavelkumbrasev (approved):
LGTM!

PR #1500, opened 2024-08-29
Fix coverity issue in the alignToBin with return

KFilipek:
### Description 
Original desc: Those values should be always >=0 but coverity do not know it. Assert should fix it.

This PR adds returning INT_MAX.

Fixes #1492

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [ ] No
- [x] Unknown


omalyshe:
Fixed with #1492

PR #1513, opened 2024-09-13
[RFCs] Add Flow Graph try_put_and_wait RFC

kboyarinov:
### Description 

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown

### DIFF
--- a/rfcs/experimental/flow_graph_wait_for_one/README.md
+++ b/rfcs/experimental/flow_graph_wait_for_one/README.md
@@ -90,8 +90,257 @@ The idea of this proposal is to extend the existing Flow Graph API with the new 
 ``node.try_put_and_wait(msg)``. This function should submit the msg into the Flow Graph (similarly to ``try_put()``) and wait for its completion. The function should be exited only if all of the tasks corresponds to the ``msg`` and skip waiting for any other tasks to
 complete.
 
+Consider the following graph:
+
+```cpp
+
+using namespace oneapi::tbb;
+
+flow::graph g;
+
+flow::broadcast_node<int> start(g);
+
+flow::function_node<int, int> f1(g, unlimited, f1_body);
+flow::function_node<int, int> f2(g, unlimited, f2_body);
+flow::function_node<int, int> f3(g, unlimited, f3_body);
+
+flow::join_node<std::tuple<int, int, int>> join(g);
+
+flow::function_node<int, int> pf(g, serial, pf_body);
+
+flow::make_edge(start, f1);
+flow::make_edge(start, f2);
+flow::make_edge(start, f3);
+
+flow::make_edge(f1, flow::input_port<0>(join));
+flow::make_edge(f2, flow::input_port<1>(join));
+flow::make_edge(f3, flow::input_port<2>(join));
+
+flow::make_edge(join, pf);
+
+// Parallel submittion
+oneapi::tbb::parallel_for (0, 100, [](int input) {
+    start.try_put(input);
+});
+
+start.try_put_and_wait(444);
+// Post-processing 444
+
+g.wait_for_all();
+```
+
+<img src="try_put_and_wait_queues.png" width=400>
+
+Each message is broadcasted from ``start`` to three concurrent computational functions ``f1``, ``f2`` and ``f3``. The result is when joined into single tuple in ``join`` node and
+post-processed in a serial ``pf`` function node. The task queue corresponding to each node in the graph is exposed under the node in the picture. The tasks that corresponds
+to the parallel loop 0-100 are shown as blue tasks in the queue. Red tasks corresponds to the message submitted as an argument in ``try_put_and_wait``.
+The ``try_put_and_wait`` is expected to exit when all of the red tasks and the necessary amount of blue tasks would be completed. Completion of all blue tasks as in ``wait_for_all``
+is not guaranteed.
+
+From the implementation perspective, the feature is implemented currently by creating an instance of special class ``message_metainfo`` with the input message in ``try_put_and_wait``
+and then broadcast it through the graph together with the message. The actual value of message can be changed during the computation but the stored metainformation should be preserved.
+
+When the message is buffered in one of the buffering nodes or one of the internal buffers (such as ``queueing`` ``function_node`` or ``join_node``), the corresponding metainformation
+instance should be buffered as well.
+
+For reference counting on single messages, the dedicated ``wait_context`` is assigned to each message passed to ``try_put_and_wait``. It is possible to use ``wait_context`` itself
+instead of ``message_metainfo``, but it can be useful to pass something with each message through the graph, not only for single message waiting. The initial implementation of
+``message_metainfo`` just wrapping the ``wait_context``, but it can be extended to cover additional use-cases. Each task corresponding to the completion of the message
+associated with the awaited message holds the reference counting on the corresponding ``wait_context``. In case of buffering the message somewhere in the graph,
+the additional reference counter would be held and released when the item is removed from the buffer.
+
+From the implementation perspective, working with metainformation is exposed by adding the new internal virtual functions in the Flow Graph:
+
+| Base Template Class | Existing Function Signature   | New Function Signature                              | Information                                      |
+|---------------------|-------------------------------|-----------------------------------------------------|--------------------------------------------------|
+| receiver            | bool try_put_task(const T& t) | bool try_put_task(const T& t)                       | Performs an action required by the node logic.   |
+|                     |                               | bool try_put_task(const T& t,                       | May buffer both ``t`` and ``metainfo``.          |
+|                     |                               |                   const message_metainfo& metainfo) | May broadcast the result and ``metainfo`` to     |
+|                     |                               |                                                     | successors of the node.                          |
+|                     |                               |                                                     | The first function can reuse the second with     |
+|                     |                               |                                                     | the empty metainfo.                              |
+|---------------------|-------------------------------|-----------------------------------------------------|--------------------------------------------------|
+| sender              | bool try_get(T& t)            | bool try_get(T& t)                                  | For buffers, gets the element from the buffer.   |
+|                     |                               | bool try_get(T& t, message_metainfo&)               | The second function provides both placeholders   |
+|                     |                               |                                                     | for metainformation and the element.             |
+|---------------------|-------------------------------|-----------------------------------------------------|--------------------------------------------------|
+| sender              | bool try_reserve(T& t)        | bool try_reserve(T& t)                              | For buffers, reserves the element in the buffer. |
+|                     |                               | bool try_reserve(T& t, message_metainfo& metainfo)  | The second function provides both placeholders   |
+|                     |                               |                                                     | for metainformantion and the element.            |
+|---------------------|-------------------------------|-----------------------------------------------------|--------------------------------------------------|
+
+The ``message_metainfo`` class is described in details in the [separate section](#details-about-metainformation-class).
+The [Nodes behavior](#nodes-behavior) section describes the behavior of each particular node when the metainformation is received.
+
+## Nodes behavior
+
+This chapter describes detailed behavior of each Flow Graph node when the item and the metainformation is received. Similarly to the message itself, the metainformation
+can be received from the predecessor node (explicit ``try_put_task`` call) or initially from ``try_put_and_wait``.
+
+### ``function_node<input, output, queueing>``
+
+If the concurrency of the ``function_node`` is ``unlimited``, the node creates a task for executting the body of the node. The created task should hold the metainfo
+received by the function node and broadcast it to the node successors when the task is completed.
+
+Otherwise, similarly to the original ``function_node`` behavior, the node tries to occupy its concurrency. If the limit is not yet reached, creates a body task
+similarly to the ``unlimited`` case. If the concurrency limit is reached, both input message and the associated metainformation would be stored in the internal queue, associated
+with the node. When one of the other tasks, associated with the node would be completed, it will retrieve the postponed message together with the metainformation and spawn it as
+a task.
+
+Since the ``function_node`` guarantees that all of the elements would be retrieved from the internal queue at some time, [buffering issues](#buffering-the-metainfo) cannot take place.
+
+### ``function_node<input, output, rejecting>``
+
+For the ``unlimited`` use-case, behaves the same as ``queueing`` node.
+
+If the concurrency limit of the node is reached, both message and the associated metainfo would be rejected and it is a predecessor responsibility to buffer them.
+If the predecessor is not the buffering node, both message and the metainfo would be lost.
+When another task would be completed, it will try to get a buffered message together with the metainfo (by calling the ``try_get(msg, metainfo)`` method) from the predecessing node.
+
+Since the ``function_node`` guarantees that all of the elements would be retrieved from the internal queue at some time, [buffering issues](#buffering-the-metainfo) cannot take place
+for buffering nodes, preceding the ``function_node``.
+
+### ``function_node<input, output, lightweight>``
+
+In regard to the concurrency limit, the lightweight function node behaves as it described in the corresponding buffering policy section (``queueing`` or ``rejecting``).
+The only difference is that for such nodes the tasks would not be spawned and the associated function will be executted by the calling thread. And since we don't have tasks,
+the calling thread should broadcast the metainformation to the successors after completing the function.
+
+### ``continue_node``
+
+The ``continue_node`` has one of the most specific semantics in regard to the metainformation. Since the node only executes the associated body (and broadcasts the signal
+to the successors) if it receives ``N`` signals from it's predecessors (where ``N`` is the number of predecessors). It means that prior to executing the body,
+the node can receive several metainformation instances from different predecessors.
+
+To handle this, the ``continue_node`` initially stores an empty metainfo instance and on each ``try_put_task(continue_msg, metainfo)`` call, it [merges](#details-about-metainformation-class)
+the received metainformation with the stored instance. Under the hood the merged instance will contain the ``wait_context`` pointers from its previous state and all of the pointers from
+the received ``metainfo``.
+
+When the ``continue_node`` receives ``N`` signals from the predecessors, it wraps stored metainformation into the task for completion of the associated body. Once the task is ready, the stored
+metainformation instance switchs back to the empty state for further work. Once the function would be completed, the task is expected to broadcast the metainfo to the successors.
+
+The lightweight ``continue_node`` behaves the same as described above, but without spawning any tasks. Everything would be performed by the calling thread.
+
+### Multi-output functional nodes
+
+``multifunction_node`` and ``async_node`` classes are not currently supported by this feature because of issues described in [the separate section](#multi-output-nodes-support).
+
+Passing the metainformation to such a node by the predecessor would have no effect and no metainfo would be broadcasted to further successors.
+
+### Single-push buffering nodes
+
+This section describes the behavior for ``buffer_node``, ``queue_node``, ``priority_queue_node`` and ``sequencer_node`` classes. The only difference between them would be in
+ordering of retrieving the elements from the buffer.
+
+Once the buffering node receives a message and the metainformation, both of them should be stored into the buffer.
+
+Since bufferring nodes are commonly used as part of the Flow Graph push-pull protocol, e.g. before the rejecting ``function_node`` or reserving ``join_node``, it means that
+the waiting for the message should be prolonged once it stored in the buffer. In particular, once the metainformation is in the buffer, the buffer should call ``reserve(1)`` on each
+associated ``wait_context`` to prolongue the wait and call ``release(1)`` once the element is retrieved from the buffer (while calling ``try_get`` or ``try_consume``).
+
+Once the element and the metainfo are stored in the buffer, the node will try to push them to the successor. If one of the successors accepts the message and the metainfo,
+both of them are removed from the buffer. Otherwise, the push-pull protocol works and the successor should return for the item once it becomes available by calling
+``try_get(msg, metainfo)`` or ``try_reserve(msg, metainfo)``.
+
+Since placing the buffers before rejecting nodes is not the only use-case, there is a risk of issues relates to buffering. It is described in details in the [separate section](#buffering-the-metainfo).
+
+### Broadcast-push buffering nodes
+
+The issue with broadcast-push ``overwrite_node`` and ``write_once_node`` is these nodes stores the received item and even if this item is accepted by one successor, it would be broadcasted to others and
+kept in the buffer. Since the metainformation is kept in the buffer together with the message itself, even if the computation is completed, the ``try_put_and_wait`` would stuck because of the reference
+held by the buffer.
+
+Even the ``wait_for_all()`` call would be able to finish in this case since it counting only the tasks in progress and ``try_put_and_wait`` would still stuck.
+
+``try_put_and_wait`` feature for the graph containing these nodes should be used carefully because of this issue:
+
+* The ``overwrite_node`` should be explicitly resetted by calling ``node.reset()`` or the element with the stored metainfo should be overwritten with another element without metainfo.
+* The ``write_once_node`` should be explicitly resetted by calling ``node.reset()`` since the item cannot be overwritten.
+
+### ``broadcast_node``
+
+The behavior of ``broadcast_node` is preety obvious - the metainformation would just be broadcasted to each successor of the node.
+
+### ``limiter_node``
+
+### ``join_node<output_tuple, queueing>``
+
+### ``join_node<output_tuple, reserving>``
+
+### ``join_node<output_tuple, key_matching>``
+
+### ``split_node``
+
+### ``indexer_node``
+
+### ``composite_node``
+
+## Additional implementation details
+
+### Details about metainformation class
+
+``message_metainfo`` class synopsis:
+
+```cpp
+class message_metainfo
+{
+public:
+    using waiters_type = std::forward_list<d1::wait_context_vertex*>;
+
+    message_metainfo();
+
+    message_metainfo(const message_metainfo&);
+    message_metainfo(message_metainfo&&);
+
+    message_metainfo& operator=(const message_metainfo&);
+    message_metainfo& operator=(message_metainfo&&);
+
+    const waiters_type& waiters() const &;
+    waiters_type&& waiters() &&;
+
+    bool empty() const;
+
+    void merge(const message_metainfo&);
+};
+```
+
+The initial implementation of ``message_metainfo`` class wraps only the list of single message waiters. The class may be extended if necessary to cover additional use-cases.
+
+The metainfo is required to hold a list of message waiters instead of single waiter to cover the ``continue_node`` and ``join_node`` joining use-cases. Consider the example:
+
+```cpp
+using namespace oneapi::tbb;
+
+flow::function_node<int, int> start1(g, ...);
+flow::function_node<int, int> start2(g, ...);
+
+flow::join_node<std::tuple<int, int>> join(g);
+
+flow::function_node<std::tuple<int, int>, int> post_process(g, ...);
+
+flow::make_edge(start, flow::input_port<0>(join));
+flow::make_edge(start, flow::input_port<1>(join));
+flow::make_edge(join, post_process);
+
+std::thread t1([&]() {
+    start1.try_put_and_wait(1);
+});
+
+std::thread t2([&]() {
+    start2.try_put_and_wait(2);
+})
+```
 
 
-## Combined or separated
+
+### Combined or separated wait
+
+### Buffering the metainfo
+
+### Multi-output nodes support
+
+### Optimization while retrieving from buffer to task
 
 ## Process Specific Information
+
+Open questions:
--- a/rfcs/experimental/flow_graph_wait_for_one/README.md
+++ b/rfcs/experimental/flow_graph_wait_for_one/README.md
@@ -1,0 +1,97 @@
+# Waiting for single messages in the Flow Graph
+
+Extending the oneTBB Flow Graph interface with the new ``try_put_and_wait(msg)`` API which allows to wait
+for completion of the chain of tasks corresponding to the ``msg``.
+
+The feature should improve the Flow Graph performance in scenarios where multiple threads submitts the work into the
+Flow Graph simultaneously and each of them should
+
+## Introduction
+
+Current API of oneTBB Flow Graph allows doing two basic actions after building the graph:
+
+- Submitting messages in some nodes using the ``node.try_put(msg)`` API.
+- Waiting for completion of **all** messagges in ghe graph using the ``graph.wait_for_all()`` API.
+
+Since the only API currently available for waiting until the work would be completed waits for all tasks in the graph
+to complete, there can be negative performance impact in use cases when the thread submitting the work should be notified as soon as possible
+when the work is done. Having only ``graph.wait_for_all()`` forces the thread to wait until **all** of the tasks in the Flow Graph, no metter
+corresponds to the waited message or not.
+
+Consider the following example:
+
+```cpp
+
+struct ComputeInput;
+struct ComputeOutput;
+
+// High-level computation instance that performs some work
+// using the oneTBB Flow Graph under the hood
+// but the Flow Graph details are not expressed as part of public API
+class ComputeTool
+{
+private:
+    oneapi::tbb::flow::graph m_graph;
+    oneapi::tbb::flow::broadcast_node<ComputeInput> m_start_node;
+    // Other Flow Graph nodes
+public:
+    ComputeTool()
+    {
+        // Builds the Flow Graph
+    }
+
+    // Peforms the computation using the user-provided input
+    ComputeOutput compute(ComputeInput input)
+    {
+        m_start_node.try_put(input); // Submit work in the graph
+        m_graph.wait_for_all(); // Waiting for input to be processed
+    }
+};
+
+int main()
+{
+    ComputeTool shared_tool;
+
+    oneapi::tbb::parallel_for(0, 10,
+        [&](int i)
+        {
+            // Preparing the input for index i
+
+            ComputeOutput output = shared_tool.compute(input);
+
+            // Post process output for index i
+        });
+}
+
+```
+
+The ``ComputeTool`` is a user interface that performs some computations on top of oneTBB Flow Graph. The function ``compute(input)`` should
+perform the computations for the provided input. Since the Flow Graph is used, it should submit the message into the graph and wait for its
+completion. This function also can be called concurrently from several threads as it shown in the ``main``.
+
+Since the only available API in the Flow Graph is ``wait_for_all()``, each thread submitting the work to the graph would be required to wait
+until **all** of the tasks would be done, no metter if these tasks corresponds to the input processed by this thread or not. If some
+post-processing is required on each thread after receiving the computation result, it would be only possible to start it when the Flow Graph would be completed what can be inneficient if the post-processing of lightweight graph tasks would be blocked by processing the more mature input.
+
+To get rid of this negative performance effect, it would be useful to add some kind of new API to the Flow Graph that would wait for
+completion of only one message (instead of the full completion of the graph):
+
+```cpp
+ComputeOutput compute(ComputeInput input)
+{
+    m_start_node.try_put(input); // Submit work in the graph
+    g.wait_for_one(input); // Pseudo-code. Wait for completion of only messages for computations of input
+}
+```
+
+## Proposal
+
+The idea of this proposal is to extend the existing Flow Graph API with the new member function of each receiver nodes -
+``node.try_put_and_wait(msg)``. This function should submit the msg into the Flow Graph (similarly to ``try_put()``) and wait for its completion. The function should be exited only if all of the tasks corresponds to the ``msg`` and skip waiting for any other tasks to
+complete.
+
+
+
+## Combined or separated
+
+## Process Specific Information


rfcs/experimental/flow_graph_wait_for_one/README.md
```diff
@@ -0,0 +1,631 @@
+# Waiting for single messages in the Flow Graph
+
+Extending the oneTBB Flow Graph interface with the new ``try_put_and_wait(msg)`` API which allows to wait
+for completion of the chain of tasks corresponding to the ``msg``.
```
vossmjp, line 4:
```suggestion
This feature extends the oneTBB Flow Graph interface with a ``try_put_and_wait(msg)`` API that supports waiting
for completion of the chain of tasks related to ``msg``.
```

rfcs/experimental/flow_graph_wait_for_one/README.md
```diff
@@ -0,0 +1,631 @@
...
+
+The feature should improve the Flow Graph performance in scenarios where multiple threads submits the work into the
+Flow Graph simultaneously and each of them should
```
vossmjp, line 7:
```suggestion
The feature may improve Flow Graph performance in scenarios where multiple threads submit work into the same
Flow Graph and each of them need to wait for only the work associated with their message to complete.
```

rfcs/experimental/flow_graph_wait_for_one/README.md
```diff
@@ -0,0 +1,631 @@
...
+
+## Introduction
+
+Current API of oneTBB Flow Graph allows doing two basic actions after building the graph:
```
vossmjp, line 11:
```suggestion
Without this feature, the oneTBB Flow Graph supports two basic actions after building the graph:
```

rfcs/experimental/flow_graph_wait_for_one/README.md
```diff
@@ -0,0 +1,631 @@
...
+
+- Submitting messages in some nodes using the ``node.try_put(msg)`` API.
+- Waiting for completion of **all** messages in ghe graph using the ``graph.wait_for_all()`` API.
```
vossmjp, line 14:
```suggestion
- Waiting for completion of **all** messages in the graph using the ``graph.wait_for_all()`` API.
```

rfcs/experimental/flow_graph_wait_for_one/README.md
```diff
@@ -0,0 +1,631 @@
...
+
+Since the only API currently available for waiting until the work would be completed waits for all tasks in the graph
+to complete, there can be negative performance impact in use cases when the thread submitting the work should be notified as soon as possible
+when the work is done. Having only ``graph.wait_for_all()`` forces the thread to wait until **all** of the tasks in the Flow Graph, no meter
+corresponds to the waited message or not.
+
```
vossmjp, line 20:
```suggestion
when the work is done. Having only ``graph.wait_for_all()`` forces the thread to wait until **all** of the tasks in the Flow Graph complete,
even those that are unrelated to the message it submitted.

```

rfcs/experimental/flow_graph_wait_for_one/README.md
```diff
@@ -0,0 +1,631 @@
...
+}
+
+```
+
+The ``ComputeTool`` is a user interface that performs some computations on top of oneTBB Flow Graph. The function ``compute(input)`` should
+perform the computations for the provided input. Since the Flow Graph is used, it should submit the message into the graph and wait for its
+completion. This function also can be called concurrently from several threads as it shown in the ``main``.
+
+Since the only available API in the Flow Graph is ``wait_for_all()``, each thread submitting the work to the graph would be required to wait
+until **all** of the tasks would be done, no meter if these tasks corresponds to the input processed by this thread or not. If some
```
vossmjp, line 73:
```suggestion
until **all** of the tasks are done, no matter if these tasks correspond to the input processed by this thread or not. If some
```

rfcs/experimental/flow_graph_wait_for_one/README.md
```diff
@@ -0,0 +1,631 @@
...
+post-processing is required on each thread after receiving the computation result, it would be only possible to start it when the Flow Graph would be completed what can be inefficient if the post-processing of lightweight graph tasks would be blocked by processing the more mature input.
+
```
vossmjp, line 75:
```suggestion
post-processing is required on each thread after receiving the computation result, it would be only safe to start it when the Flow Graph is complete. 
This is inefficient since the post-processing of lightweight graph tasks are blocked by processing of unrelated inputs.

```

rfcs/experimental/flow_graph_wait_for_one/README.md
```diff
@@ -0,0 +1,631 @@
...
+To get rid of this negative performance effect, it would be useful to add some kind of new API to the Flow Graph that would wait for
+completion of only one message (instead of the full completion of the graph):
```
vossmjp, line 77:
```suggestion
To remove this negative performance effect, the ``wait_for_one`` extension waits for
completion of only one message (instead of the full completion of the graph):
```

rfcs/experimental/flow_graph_wait_for_one/README.md
```diff
@@ -0,0 +1,631 @@
...
+
+```cpp
+ComputeOutput compute(ComputeInput input)
+{
+    m_start_node.try_put(input); // Submit work in the graph
+    g.wait_for_one(input); // Pseudo-code. Wait for completion of only messages for computations of input
+}
+```
+
+## Proposal
```
vossmjp, line 87:
```suggestion
## Overview of experimental feature
```

rfcs/experimental/flow_graph_wait_for_one/README.md
```diff
@@ -0,0 +1,631 @@
...
+
+The idea of this proposal is to extend the existing Flow Graph API with the new member function of each receiver nodes -
```
vossmjp, line 89:
```suggestion
This feature extends the existing Flow Graph API with an additional member function for each of the receiver nodes -
```

rfcs/experimental/flow_graph_wait_for_one/README.md
```diff
@@ -0,0 +1,631 @@
...
+``node.try_put_and_wait(msg)``. This function should submit the msg into the Flow Graph (similarly to ``try_put()``) and wait for its completion. The function should be exited only if all of the tasks corresponds to the ``msg`` and skip waiting for any other tasks to
+complete.
+
```
vossmjp, line 92:
```suggestion
``node.try_put_and_wait(msg)``. This function submits the msg into the Flow Graph (similarly to ``try_put()``) and wait for its completion. 
The function blocks until all of the tasks related to processing ``msg`` complete and is allowed to skip waiting for any 
other tasks to complete.

```

rfcs/experimental/flow_graph_wait_for_one/README.md
```diff
@@ -0,0 +1,631 @@
...
+// Post-processing 444
+
+g.wait_for_all();
+```
+
+Each message is broadcasted from ``start`` to three concurrent computational functions ``f1``, ``f2`` and ``f3``. The result is when joined into single tuple in ``join`` node and
+post-processed in a serial ``pf`` function node. The task queue corresponding to each node in the graph is exposed under the node in the picture. The tasks that corresponds
+to the parallel loop 0-100 are shown as blue tasks in the queue. Red tasks corresponds to the message submitted as an argument in ``try_put_and_wait``.
+The ``try_put_and_wait`` is expected to exit when all of the red tasks and the necessary amount of blue tasks would be completed. Completion of all blue tasks as in ``wait_for_all``
+is not guaranteed.
```
vossmjp, line 136:
```suggestion
Each message is broadcasted from ``start`` to three concurrent computational functions ``f1``, ``f2`` and ``f3``. 
The result is joined into single tuple in ``join`` node and post-processed in a serial ``pf`` function node. 
The task queue corresponding to each node in the graph is exposed under the node in the picture. 
The tasks that corresponds to the parallel loop 0-100 are shown as blue tasks in the queue. Red tasks 
correspond to the message submitted as an argument in ``try_put_and_wait``. The ``try_put_and_wait`` is 
expected to exit when all of the red tasks and the necessary amount of blue tasks are completed. Completion 
of all blue tasks as in ``wait_for_all`` is not guaranteed.
```

rfcs/experimental/flow_graph_wait_for_one/README.md
```diff
@@ -0,0 +1,631 @@
...
+
+From the implementation perspective, the feature is implemented currently by creating an instance of special class ``message_metainfo`` with the input message in ``try_put_and_wait``
```
vossmjp, line 138:
```suggestion
## Feature Design

From the implementation perspective, the feature is implemented currently by creating an instance of special class ``message_metainfo`` with the input message in ``try_put_and_wait``
```

rfcs/experimental/flow_graph_wait_for_one/README.md
```diff
@@ -0,0 +1,481 @@
...
+Since the ``function_node`` guarantees that all of the elements would be retrieved from the internal queue at some time, [buffering issues](#buffering-the-metainfo) cannot take place.
+
+### Rejecting ``function_node``
+
+If the concurrency of the node is set to ``unlimited``, behaves the same as in the ``queueing`` case described above.
+
+Otherwise, if the concurrency limit of the node is reached, both message and the associated metainformations would be rejected and the predecessor that called the ``try_put_task``
+is responsible on buffering both of them.
+
+If the predecessor is not the buffering node, both message and the metainfo would be lost.
```
vossmjp, line 226: You should say what "lost" means in this context. How does it affect waiting.

rfcs/experimental/flow_graph_wait_for_one/README.md
```diff
@@ -0,0 +1,481 @@
...
+
+Since the metainformation is kept in the buffer together with the message itself, even if the computation is completed, the ``try_put_and_wait`` would stuck because of the reference
+held by the buffer.
+
+Even the ``wait_for_all()`` call would be able to finish in this case since it counting only the tasks in progress and ``try_put_and_wait`` would still be blocked.
+
+``try_put_and_wait`` feature for the graph containing these nodes should be used carefully because of this issue:
+
+* The ``overwrite_node`` should be explicitly reset by calling ``node.reset()`` or the element with the stored metainfo should be overwritten with another element.
+* The ``write_once_node`` should be explicitly reset by calling ``node.reset()`` since the item cannot be overwritten.
```
vossmjp, line 293:
```suggestion
* The ``overwrite_node`` should be explicitly reset by calling ``node.clear()`` or the element with the stored metainfo should be overwritten with another element.
* The ``write_once_node`` should be explicitly reset by calling ``node.clear()`` since the item cannot be overwritten.
```

rfcs/experimental/flow_graph_wait_for_one/README.md
```diff
@@ -0,0 +1,481 @@
...
+
+### ``limiter_node``
+
+If the threshold of the node was not reached, both value and the metainformation should be broadcasted to the successors. Otherwise- both should be rejected and buffered by another node. 
+
+Metainformation on the decrement port is ignored since this signal should not be considered part of working on the original message.
+
+### Queueing ``join_node``
+
+Each input port of the join_node should support the queue for both values and the associated metainformations. Once all of the input ports would contain the value, the values
```
vossmjp, line 307:
```suggestion
Each input port of the join_node should support the queue for both values and the associated metainformations. Once all of the input ports contain the value, the values
```

rfcs/experimental/flow_graph_wait_for_one/README.md
```diff
@@ -0,0 +1,481 @@
...
+should be combined into single tuple output and the metainformation objects should be combined into single metainfo using `metainfo1.merge(metainfo2)`, associated with the tuple
+and submitted to successors.
+
+If the item with the metainformation is stored in the internal queue of one of the input ports, but items never received by other ports, the item and the metainformation would be kept in the
+queue and block the corresponding ``try_put_and_wait`` call.
```
vossmjp, line 312:
```suggestion
If an item with the metainformation is stored in the internal queue of one of the input ports, but items are never received by other ports, the item and the metainformation will be kept in the
queue and block the corresponding ``try_put_and_wait`` call.
```

vossmjp: We may need to add a `clear` function to `join_node` to allow explicit clearing of the internal queues.

rfcs/experimental/flow_graph_wait_for_one/README.md
```diff
@@ -0,0 +1,481 @@
...
+* Introducing an extra unspecified type ``node_type::tag_type`` and require the user to accept it as a third argument of the body. 
+
+## Open Questions
+
+The following questions should be addressed before moving this feature to ``supported``:
+
+* Multi-output nodes support should be described and implemented
+* Feedback from the customers should be received
+* More multithreaded tests should be implemented for the existing functionality
+* The corresponding oneAPI specification update should be done
```
vossmjp, line 481: Should we add a `clear` function to all buffering nodes?

rfcs/experimental/flow_graph_wait_for_one/README.md
```diff
@@ -0,0 +1,481 @@
...
+
+// Other work
+// May submit more work into the Flow Graph
+// May create other descriptors that would be waited later
+
+g.wait(desc);
+
+```
+
+In that case it would be needed to extend the node with the new API returning some descriptor that can be used as a hit work waiting function `g.wait` that also should be added.
```
akukanov, line 421:
```suggestion
In that case it would be needed to extend the node with the new API returning some descriptor that can be used as the argument to the work waiting function `g.wait` that also should be added.
```

rfcs/experimental/flow_graph_wait_for_one/README.md
```diff
@@ -0,0 +1,481 @@
...
+is the number of predecessors. 
+
+It means that prior to executing the body, the node can receive several ``metainfo`` instances from different predecessors. To handle this, the node initially stores an
+empty metainfo instance inside itself and each call to ``try_put_task`` with non-empty metainfo, merges the received metainformation with the stored instance.
+
+Additional reference counter would be held on each input ``wait_context`` to make sure the corresponding ``try_put_and_wait`` will remain blocked until the item would leave the
+``continue_node``. 
+
+When the ``continue_node`` receives the last signal from the predecessors, it creates a task to complete the associated body. Once this task is executed, the metainfo instance stored in
+the ``continue_node`` switches back to the empty state for further work and the previously accumulated metainfo is broadcasted to the successors of the node.
```
akukanov, line 253: As described, there seems to be a race in case the `continue_node` can be triggered more than once during an execution of the graph. In such case, new messages might come while the body of the node is executed - so switching the state back to empty should seemingly happen before executing the body.

rfcs/experimental/flow_graph_wait_for_one/README.md
```diff
@@ -0,0 +1,481 @@
...
+``join`` will create a tuple from the message that relates to ``try_put_and_wait`` (red task) and the one that does not (blue task). 
+
+The ``try_put_and_wait`` is expected to exit when all of the red tasks and the necessary amount of blue tasks would be completed.
+
+Completion of all blue tasks as in ``wait_for_all`` is not guaranteed.
+
+## Feature Design
+
+This feature is implemented by creating an instance of special private class ``message_metainfo`` for each input message in ``try_put_and_wait``. This instance
+wraps the pointer to the dedicated ``wait_context`` object representing the reference counter for single input message.
```
akukanov, line 151: I did not find any description of how the reference counter is used - that is, at which point(s) it is incremented and decremented.

rfcs/experimental/flow_graph_wait_for_one/README.md
kboyarinov: Thanks, it is an incorrect description. The implementation resets the stored metainfo before spawning the task under the node's mutex. 
Updated the description.

rfcs/experimental/flow_graph_wait_for_one/README.md
```diff
@@ -0,0 +1,481 @@
...
+See [separate section](#details-about-metainformation-class) for more details.
+
+Implementation-wise, processing the metainformation is exposed by adding new internal virtual member functions to various Flow Graph instance:
+
+``` cpp
+template <typename T>
+class receiver {
+protected:
+    virtual bool try_put_task(const T& t) = 0; // Existing API
+    virtual bool try_put_task(const T& t, const message_metainfo& info); // New API
```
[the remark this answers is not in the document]
kboyarinov: Fixed

rfcs/experimental/flow_graph_wait_for_one/README.md
```diff
@@ -0,0 +1,481 @@
...
+
+<img src="try_put_and_wait_graph.png" width=400>
+
+Each message is broadcasted from ``start`` to two concurrent computational functions ``f1`` and ``f2``.
+
+The result is joined into single tuple in ``join`` node and post-processed in a serial ``pf`` function node.
+
+The task queue corresponding to each node in the graph is exposed under the node in the picture.
+
+Blue tasks in the queue relates to the parallel loop 0-100. Red tasks relates to the message submitted as an input to ``try_put_and_wait``.
```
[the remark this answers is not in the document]
kboyarinov: Updated both picture and the description

rfcs/experimental/flow_graph_wait_for_one/README.md
```diff
@@ -0,0 +1,481 @@
...
+submit the input provided to the computation graph and wait for completion of the input. 
+This function also can be executed concurrently from several threads as it shown in the ``main``.
+
+While using the ``wait_for_all()``, each thread submitting the work to the graph is required to wait
+until **all** of the tasks to be done, even those that are unrelated to the input submitted. If some
+post-processing is required on each thread after receiving the computation result, it would be only safe to start it when the 
+Flow Graph is completed. That is inefficient since the post-processing of lightweight graph tasks are blocked by processing of unrelated inputs.
+
+To remove this negative performance effect, the ``wait_for_one`` extension waits for completion of only one message
+(instead of the full completion of the graph):
```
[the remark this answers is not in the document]
kboyarinov: Renamed

rfcs/experimental/flow_graph_wait_for_one/README.md
kboyarinov: Added to the list of the questions, together with the safety guarantees for this method.

rfcs/experimental/flow_graph_wait_for_one/README.md
kboyarinov: Updated the description

rfcs/experimental/flow_graph_wait_try_put_and_wait/README.md
```diff
@@ -0,0 +1,510 @@
...
+
+### Buffering the metainfo
+
+As described in the [buffering nodes description section](#single-push-buffering-nodes), current proposal requires the nodes that stores the user values as part of buffers to store
+also the metainformation associated with these values in the same buffer for future use and also to hold the additional reference counter to force the `try_put_and_wait` to wait 
+until the rejecting receiver take the item from the buffer and process it. 
+
+Such a behavior may significantly affect the other common use-case for buffering nodes: when they are used to store the result of the computation at the end of the Flow Graph. 
+In such scenarios, the metainformation would be stored in the buffer and never taken from it since the buffering node is not used as part of push-pull protocol and hence there is
+no rejecting successor. The `try_put_and_wait` function associated with such a metainformation will hang forever since the reference counter on the stored metainformation would
```
akukanov, line 457: Is `will hang forever` the right description though? I thought that once the corresponding message is taken out of the buffer by some consumer (thread) outside of the graph, the last reference is removed and waiting finishes. So it should be possible to use an asynchronous consumer activity, and the deadlock would only happen if the waiting thread is the consumer, correct?

For such cases, should we consider adding some trait to buffer nodes to ignore the metainformation?

rfcs/experimental/flow_graph_wait_try_put_and_wait/README.md
```diff
@@ -0,0 +1,510 @@
...
+never be decreased.
+
+It is impossible to rely on the number of successors while making a decision to store the metainformation in the buffer since if the node is used as part of the push-pull protocol,
+the number of successors is also equal to `0` since the edge is reversed.
+
+Current proposal considers this scenario is a `try_put_and_wait` feature limitation and does not add any support for that.
+
+### Multi-output nodes support
+
+Multi-output nodes (`multifunction_node` and `async_node`) creates an extra issue for the wait-for-one feature. Consider the following example (all of the examples shown
```
akukanov, line 467: The last of wait-for-ones is spotted :)

rfcs/experimental/flow_graph_wait_try_put_and_wait/README.md
```diff
@@ -0,0 +1,510 @@
...
+Unlike the `function_node`, the computed output is not propagated implicitly to the successors. It is done explicitly by the user by calling
+`try_put(output)` on the corresponding output_port of the node. Since the metainformation associated with the particular input is hidden from the user
+inside of the Flow Graph implementation, it cannot be propagated to the successors as part of the explicit user try_put.
+
+An other interesting use-case is when the `multifunction_node` is used as a reduction for multiple input values with only one output. In that case the metainformation
+should not be automatically propagated at all and should be accumulated simultaneously with the outputs.
+
+Current implementation does not support ``multifunction_node`` and ``async_node``. 
+
+Possible approaches for implementing such support:
```
akukanov, line 492: Would all three approaches work for async node as well?

Are we not going to even consider an approach where the metainfo is passed and propagated with the message through the whole graph? Essentially the second option in the list, but applied to all node types?

rfcs/experimental/flow_graph_wait_try_put_and_wait/README.md
```diff
@@ -0,0 +1,510 @@
...
+
+The following questions should be addressed before moving this feature to ``supported``:
+
+* Multi-output nodes support should be described and implemented
+* Since the buffered item holds an additional reference counter on the associated metainfo, all elements should be retrieved from buffers to allow ``try_put_and_wait`` to exit.
+    * Should ``clear()`` member function be added to all of the buffering nodes in the graph (currently supported only in ``write_once_node`` and ``overwrite_node``).
+    * Should ``clear()`` member function be added to the non-rejecting ``join_node`` to handle the case when when we don't have the input present on each input port.
+    * Concurrent safety guarantees for ``clear()`` should be defined (e.g. is it safe to clear the buffer when other thread tries to insert the item).
+* Feedback from the customers should be received
+* More multithreaded tests should be implemented for the existing functionality
```
akukanov, line 509: An idea for an example/test: a version of dining philosophers where an external thread would use `try_put_and_wait()` for one of them (perhaps the fastest thinker :))

rfcs/experimental/flow_graph_wait_try_put_and_wait/README.md
```diff
@@ -0,0 +1,510 @@
...
+* Introducing an extra unspecified type ``node_type::tag_type`` and require the user to accept it as a third argument of the body. 
+
+## Open Questions
+
+The following questions should be addressed before moving this feature to ``supported``:
+
+* Multi-output nodes support should be described and implemented
+* Since the buffered item holds an additional reference counter on the associated metainfo, all elements should be retrieved from buffers to allow ``try_put_and_wait`` to exit.
+    * Should ``clear()`` member function be added to all of the buffering nodes in the graph (currently supported only in ``write_once_node`` and ``overwrite_node``).
+    * Should ``clear()`` member function be added to the non-rejecting ``join_node`` to handle the case when when we don't have the input present on each input port.
```
akukanov, line 506: I would instead consider something like `try_get_partial` which would return e.g. a tuple of optionals. It would allow draining the node without losing stored messages.

rfcs/experimental/flow_graph_wait_try_put_and_wait/README.md
```diff
@@ -0,0 +1,510 @@
...
+for multi-output nodes.
+* Introducing an extra unspecified type ``node_type::tag_type`` and require the user to accept it as a third argument of the body. 
+
+## Open Questions
+
+The following questions should be addressed before moving this feature to ``supported``:
+
+* Multi-output nodes support should be described and implemented
+* Since the buffered item holds an additional reference counter on the associated metainfo, all elements should be retrieved from buffers to allow ``try_put_and_wait`` to exit.
+    * Should ``clear()`` member function be added to all of the buffering nodes in the graph (currently supported only in ``write_once_node`` and ``overwrite_node``).
```
akukanov, line 505: Not sure it is needed, as you can drain the buffer by `try_get`, right? Just my 2 cents.

rfcs/experimental/flow_graph_wait_try_put_and_wait/README.md
```diff
@@ -0,0 +1,510 @@
...
+wraps the pointer to the dedicated ``wait_context`` object representing the reference counter for single input message.
+
+``message_metainfo`` is broadcasted through the graph together with the message itself. The actual value and the type of the message can be changed during the computation, but the stored 
+metainformation is preserved.
+
+The reference counter in the ``wait_context`` inside of the ``message_metainfo`` is increased in the following cases:
+* The task associated with the computations of the ``try_put_and_wait`` input message is created (*).
+* The item associated with the computations of the desired input is stored in the buffering node or in the internal buffer of the other node type.
+* When the ``continue_msg`` with the non-empty associated waiter is received by ``continue_node``. In this case, the node buffers the ``message_metainfo``s received from each
+  predecessor and increases the underlying reference counter to prolonge the wait until signals from each predecessor would be received.
```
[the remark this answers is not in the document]
kboyarinov: I have double-checked the current implementation and it seems like we can support reference counter increase and decrease designed around create/destroy and copy the `message_metainfo` objects. Enabling proper moving the metainfo without touching the reference counter would require introducing two virtual options of `try_put_task` in the receiver:

```
virtual bool try_put_task(const T& t, const message_metainfo& meta) = 0; // 1 -  preview, current implementation
virtual bool try_put_task(const T& t, message_metainfo&& meta) = 0; // 2 - overload for optimized reference counting
```

Let's consider the example of a `broadcast_node` with multiple successors. A call to `try_put_and_wait` would broadcast the metainfo as const lvalue to all successors except the last one - using the overload 1, and as rvalue to the last successor - using the overload 2. 
It seems safe since overload 1 would never save the reference to metainfo and hence the moved-from object cannot be accessed (e.g. by the task) after calling the overload 2 for the last successor.

I also thought about the possible FG improvements for better rvalues support for the message itself. According to my old investigations, such a support would also require extending the `try_put_task` with the new overloads for rvalues. If we assume that the message should be copied when the metainfo is copied and moved when the metainfo is moved, two overloads is enough. If it is not always true (I am not sure for now, deeper investigation needed), we would also need virtual functions for lvalue message and rvalue meta and wise versa (4 virtual functions total). 

O propose to mention all of the possible improvements in the dedicated section of the RFC and keep this description as-is since this is how current implementation works.

rfcs/experimental/flow_graph_wait_try_put_and_wait/README.md
kboyarinov: I think it is also a good example that shows the necessity of some trait for buffering nodes that should allow ignoring the metainformation. If we consider the implementation from our examples but without using the `multifunction_node`, the chopstick is always present in the graph. If we want to use `try_put_and_wait` for one of the philosophers, at the point when both chopsticks are acquired, we have the tuple of signal and two chopsticks, associated with the metainfo from the signal.
When the chopstick is returned back to its buffer, it would contain the metainfo from the signal, and the chopstick would be considered part of computations needed by `try_put_and_wait`. 

In this case, the buffer for chopsticks can be marked by the special trait to ignore the metainformation. 
I think the same problem takes place in each graph that uses token-passing.

Alternatively, it can be done by the `multifunction_node` by explicitly passing the metainfo only to the "keep thinking" port but not to the "return chopstick" ports.

rfcs/experimental/flow_graph_wait_try_put_and_wait/README.md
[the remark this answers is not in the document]
akukanov: > I propose to mention all of the possible improvements in the dedicated section of the RFC and keep this description as-is since this is how current implementation works.

Sure; I only wanted to bring the idea for further consideration.

rfcs/experimental/flow_graph_wait_try_put_and_wait/README.md
```diff
@@ -0,0 +1,510 @@
...
+* When the desired number of signals from the predecessors was received by the ``continue_node``. This case is equivalent to retrieving the set of buffered ``message_metainfo``s received previously
+  from each predecessor. 
+
+``message_metainfo`` class may be reused in the future to support additional use-cases when it is required to push additional information about the input message through the graph. E.g. supporting
+priorities for single messages - the corresponding priority tag can be assigned as part of the metainfo.
+
+Metainformation class supports containing multiple reference counters at the same time to support joining messages with different associated reference counters in the ``join_node``. 
+See [separate section](#details-about-metainformation-class) for more details.
+
+Implementation-wise, processing the metainformation is exposed by adding new internal virtual member functions to various Flow Graph instance:
```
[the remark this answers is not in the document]
akukanov: I would recommend to add the information about maintaining backward compatibility to the document.

rfcs/experimental/flow_graph_wait_try_put_and_wait/README.md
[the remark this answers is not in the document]
kboyarinov: I have added this as possible future improvement.

rfcs/experimental/flow_graph_wait_try_put_and_wait/README.md
```diff
@@ -0,0 +1,510 @@
...
+
+If the threshold of the node was not reached, both value and the metainformation should be broadcasted to the successors. Otherwise- both should be rejected and buffered by another node. 
+
+Metainformation on the decrement port is ignored since this signal should not be considered part of working on the original message.
+
+### Queueing ``join_node``
+
+Each input port of the join_node should support the queue for both values and the associated metainformations. Once all of the input ports contain the value, the values
+should be combined into single tuple output and the metainformation objects should be combined into single metainfo using `metainfo1.merge(metainfo2)`, associated with the tuple
+and submitted to successors.
```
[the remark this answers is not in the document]
kboyarinov: Yes, as far as I remember the risk of changing the tuple between join and split was the main reason why we did not make any splitting of the metainfo.

rfcs/experimental/flow_graph_wait_try_put_and_wait/README.md
kboyarinov: Removed

rfcs/experimental/flow_graph_wait_try_put_and_wait/README.md
kboyarinov: Added

rfcs/experimental/flow_graph_wait_try_put_and_wait/README.md
kboyarinov: I agree, removed for now.

rfcs/experimental/flow_graph_wait_try_put_and_wait/README.md
```diff
@@ -0,0 +1,510 @@
...
+should not be automatically propagated at all and should be accumulated simultaneously with the outputs.
+
+Current implementation does not support ``multifunction_node`` and ``async_node``. 
+
+Possible approaches for implementing such support:
+
+* Hiding metainfo inside of ``output_ports`` to preserve automatic metainfo propagation.
+* Merging the input and the ``message_metainfo`` together into some publicly available type ``tagged_input<T>`` and require the user to explicitly specify this type
+for multi-output nodes.
+* Introducing an extra unspecified type ``node_type::tag_type`` and require the user to accept it as a third argument of the body.
```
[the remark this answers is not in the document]
kboyarinov: Yes, the third option requires expressing `hint_type` as a public version of `message_metainfo` including the description of thread safety of each particular method. It would be considered as part of a separate RFC.

rfcs/experimental/flow_graph_wait_try_put_and_wait/README.md
```diff
@@ -0,0 +1,604 @@
...
+* When the desired number of signals from the predecessors was received by the ``continue_node``. This case is equivalent to retrieving the set of buffered ``message_metainfo``s received previously
+  from each predecessor. 
+
+``message_metainfo`` class may be reused in the future to support additional use-cases when it is required to push additional information about the input message through the graph. E.g. supporting
+priorities for single messages - the corresponding priority tag can be assigned as part of the metainfo.
+
+Metainformation class supports containing multiple reference counters at the same time to support joining messages with different associated reference counters in the ``join_node``. 
+See [separate section](#details-about-metainformation-class) for more details.
+
+Implementation-wise, processing the metainformation is exposed by adding new internal virtual member functions to various Flow Graph instance:
```
akukanov, line 185: I would add a note that the functions exposed below are details of implementation which might change over time and therefore should not be used directly.

rfcs/experimental/flow_graph_wait_try_put_and_wait/README.md
```diff
@@ -0,0 +1,604 @@
...
+    virtual bool try_put_task(const T& t) = 0; // Existing API
+    virtual bool try_put_task(const T& t, const message_metainfo& info) = 0; // New API
+};
+```
+
+For each particular implementation of ``receiver``, the ``try_put_task`` performs ab action that is required by the corresponding Flow Graph node. 
+
+It may buffer both ``t`` and ``metainfo`` or broadcast the result and the ``info`` to the successors of the node. 
+
+The existing API ``try_put_task(const T& t)`` can reuse the new one with the empty metainfo object.
```
akukanov, line 200: Would not that however increase overheads for programs that do not use `try_put_and_wait()`?

rfcs/experimental/flow_graph_wait_try_put_and_wait/README.md
```diff
@@ -0,0 +1,604 @@
...
+wraps the pointer to the dedicated ``wait_context`` object representing the reference counter for single input message.
+
+``message_metainfo`` is broadcasted through the graph together with the message itself. The actual value and the type of the message can be changed during the computation, but the stored 
+metainformation is preserved.
+
+The reference counter in the ``wait_context`` inside of the ``message_metainfo`` is increased in the following cases:
+* The task associated with the computations of the ``try_put_and_wait`` input message is created (*).
+* The item associated with the computations of the desired input is stored in the buffering node or in the internal buffer of the other node type.
+* When the ``continue_msg`` with the non-empty associated waiter is received by ``continue_node``. In this case, the node buffers the ``message_metainfo``s received from each
+  predecessor and increases the underlying reference counter to prolonge the wait until signals from each predecessor would be received.
```
akukanov, line 170: Here and in other cases below where you describe the design of the already provided experimental functionality, the use of `would [be]` is somewhat confusing. Please take a look at these places and change the wording to be more definite where needed.

rfcs/experimental/flow_graph_wait_try_put_and_wait/README.md
```diff
@@ -0,0 +1,604 @@
...
+std::thread t1([&]() {
+    start1.try_put_and_wait(1);
+});
+
+std::thread t2([&]() {
+    start2.try_put_and_wait(2);
+})
+```
+
+### Combined or separated wait
```
akukanov, line 429: It looks like this question is not added to the "Exit criteria" list; should it be?

rfcs/experimental/flow_graph_wait_try_put_and_wait/README.md
```diff
@@ -0,0 +1,604 @@
...
+Increment and decrement of the reference counter/s stored in the ``message_metainfo`` can be done in the RAII manner - increase while creating and copying the object and decreased
+while destroying the object.
+
+Extending this approach would allow implementing move semantics for ``message_metainfo`` and optimize passing it to the last successor of the buffering node or a ``continue_node`` without
+modifying the reference counter.
+
+## Exit criteria
+
+The following questions should be addressed before moving this feature to ``supported``:
+* Multi-output nodes support should be described and implemented.
```
akukanov, line 596: Add a link to the corresponding section of the document?

rfcs/experimental/flow_graph_wait_try_put_and_wait/README.md
```diff
@@ -217,7 +217,10 @@ For each particular implementation of ``sender``, ``try_get`` gets the element a
 the metainfo to ``info``. The reference counter/s, associated with the stored metainfo are decremented.
 
 ``try_reserve`` implementation reserves the element and the corresponding metainfo inside of the buffer and feels the placeholders provided. Since the elements are not
-removed from the buffer, the reference counter/s remains unchanged. They will be decremented when ``try_consume`` is called. 
+removed from the buffer, the reference counter/s remains unchanged. They will be decremented when ``try_consume`` is called.
+
+To handle the backward compatibility issues caused by adding new virtual interfaces to ``receiver`` and ``sender``, the ``detail::d`` namespace version number should be increased
+while moving ``try_put_and_wait`` out of preview.
```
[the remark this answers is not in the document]
kboyarinov: Only the APIs inheriting the ``sender`` and ``receiver`` are affected, but in the past, we have increased the namespace version number for the entire Flow Graph, even for the non-affected APIs (e.g. when we implemented the distributed reference counting)

rfcs/experimental/flow_graph_wait_try_put_and_wait/README.md
kboyarinov: I have made an initial analysis before the feature was released and did not see any performance degradations (at least on our examples and some simple benchmarks). Currently, the empty metainfo is represented as an empty `std::forward_list` that is almost identical to the single nullptr.

I have added a performance analysis section to the exit criteria that should cover both performance improvement while `try_put_and_wait` is used and absence of degradations when it is not used.

rfcs/experimental/flow_graph_wait_try_put_and_wait/README.md
kboyarinov: Done

rfcs/experimental/flow_graph_wait_try_put_and_wait/README.md
kboyarinov: Done

rfcs/experimental/flow_graph_wait_try_put_and_wait/README.md
```diff
@@ -0,0 +1,608 @@
...
+* The item associated with the computations of the desired input is stored in the buffering node or in the internal buffer of the other node type.
+* When the ``continue_msg`` with the non-empty associated waiter is received by ``continue_node``. In this case, the node buffers the ``message_metainfo``s received from each
+  predecessor and increases the underlying reference counter to prolonge the wait until signals from each predecessor is received. 
+
+The reference counter is decreased when:
+* The task associated with the computations of the ``try_put_and_wait`` input is finalized. If the output of the task should be propagated to the successors of the node, it is done
+  from the task body and hence the reference counter is decreased only after creating the task for each successor and increasing the reference counter for them.
+* The item associated with the computations is taken from the buffering node or from the internal buffer.
+* When the desired number of signals from the predecessors was received by the ``continue_node``. This case is equivalent to retrieving the set of buffered ``message_metainfo``s received previously
+  from each predecessor.
```
akukanov, line 177: For these cases, how it is ensured that the reference counters are not decreased prematurely?

rfcs/experimental/flow_graph_wait_try_put_and_wait/README.md
```diff
@@ -0,0 +1,608 @@
...
+    virtual bool try_put_task(const T& t) = 0; // Existing API
+    virtual bool try_put_task(const T& t, const message_metainfo& info) = 0; // New API
+};
+```
+
+For each particular implementation of ``receiver``, the ``try_put_task`` performs ab action that is required by the corresponding Flow Graph node. 
+
+It may buffer both ``t`` and ``metainfo`` or broadcast the result and the ``info`` to the successors of the node. 
+
+The existing API ``try_put_task(const T& t)`` can reuse the new one with the empty metainfo object if an empty metainfo is a preserved as a lightweight structure.
```
akukanov, line 201: I would instead say "if that does not incur a noticeable overhead for the normal use of flow graphs."

rfcs/experimental/flow_graph_wait_try_put_and_wait/README.md
```diff
@@ -0,0 +1,608 @@
...
+
+### Lightweight ``function_node``
+
+Calls to ``try_put_task`` in the lightweight node will operate on the concurrency limit of the node in the same manner as is defined by the message buffering policy -
+``queueing`` (default)  or ``rejecting``.
+
+The difference is that for lightweight nodes no tasks are created and spawned in most of the cases and the node body will be executed by the calling thread.
+Since there are no tasks, the calling thread broadcasts the output and the metainformation to the successors after completing the function.
+
+### ``continue_node``
```
akukanov, line 266: I am afraid that the overhead of metainfo for dependency graphs can be big with a straightforward implementation.

Imagine a 2D wavefront graph started with `try_put_and_wait`. Each node that receives the metainfo will send two copies of it, one for each successor. Each node in the middle of the graph will receive two piles of metainfo from its predecessors, merge those (doubling the size of the pile) and send to two its successors, This is an exponential growth pattern.

And the benefit of single-message wait for continue nodes is very doubtful, as dependency graphs are not supposed to handle several independent message flows.

I would reconsider implementation and/or applicability of this feature for continue nodes.

rfcs/experimental/flow_graph_wait_try_put_and_wait/README.md
kboyarinov: `continue_node` creates a copy of the stored metainfo merged from each of the predecessors and creates a task for executing the body that holds an additional reference on the counter (similarly to any task with the stored metainfo). 
And only after spawning the task, the reference counters on the copy are decreased.

I have added a note that the reference counter is decreased after spawning a task.

rfcs/experimental/flow_graph_wait_try_put_and_wait/README.md
kboyarinov: Applied

rfcs/experimental/flow_graph_wait_try_put_and_wait/README.md
kboyarinov: Added a separate section for this issue.

PR #1520, opened 2024-09-23
Stateful allocator support for concurrent_queue and concurrent_bounde…

YexuanXiao:
…d_queue

### Description 
Add support for [allocator propagation](https://en.cppreference.com/w/cpp/named_req/AllocatorAwareContainer) to concurrent_queue and concurrent_bounded_queue, and use the allocator’s construct function instead of placement new.

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/include/oneapi/tbb/concurrent_queue.h
+++ b/include/oneapi/tbb/concurrent_queue.h
@@ -137,25 +137,29 @@ public:
     }
 
     concurrent_queue& operator=( const concurrent_queue& other ) {
-        //TODO: implement support for std::allocator_traits::propagate_on_container_copy_assignment
-        if (my_queue_representation != other.my_queue_representation) {
-            clear();
+        if (my_queue_representation == other.my_queue_representation)
+            return *this;
+        clear();
+        if (queue_allocator_traits::propagate_on_container_move_assignment::value) {
             my_allocator = other.my_allocator;
-            my_queue_representation->assign(*other.my_queue_representation, my_allocator, copy_construct_item);
         }
+        my_queue_representation->assign(*other.my_queue_representation, my_allocator, copy_construct_item);
         return *this;
     }
 
     concurrent_queue& operator=( concurrent_queue&& other ) {
-        //TODO: implement support for std::allocator_traits::propagate_on_container_move_assignment
-        if (my_queue_representation != other.my_queue_representation) {
-            clear();
+        if (my_queue_representation == other.my_queue_representation)
+            return *this;
+        clear();
+        if (queue_allocator_traits::propagate_on_container_move_assignment::value) {
+            my_allocator = std::move(other.my_allocator);
+            internal_swap(other);
+        } else {
             if (my_allocator == other.my_allocator) {
                 internal_swap(other);
             } else {
-                my_queue_representation->assign(*other.my_queue_representation, other.my_allocator, move_construct_item);
+                my_queue_representation->assign(*other.my_queue_representation, my_allocator, move_construct_item);
                 other.clear();
-                my_allocator = std::move(other.my_allocator);
             }
         }
         return *this;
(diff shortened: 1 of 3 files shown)


include/oneapi/tbb/concurrent_queue.h
```diff
@@ -137,25 +137,29 @@ class concurrent_queue {
...
-            clear();
+        if (my_queue_representation == other.my_queue_representation)
+            return *this;
+        clear();
+        if (queue_allocator_traits::propagate_on_container_move_assignment::value) {
             my_allocator = other.my_allocator;
-            my_queue_representation->assign(*other.my_queue_representation, my_allocator, copy_construct_item);
         }
+        my_queue_representation->assign(*other.my_queue_representation, my_allocator, copy_construct_item);
         return *this;
```
kboyarinov, line 147: I propose to rewrite consistent with other containers and use TBB helper function for copy assigning the allocators (`tbb::detail::copy_assign_allocators)`:

```
if (this != &other) {
    clear();
    copy_assign_allocators(my_allocator, other.my_allocator);
    my_queue_representation->assign(*other.my_queue_representation, my_allocator, copy_construct_item);
}
return *this;
```

include/oneapi/tbb/concurrent_queue.h
```diff
@@ -137,25 +137,29 @@ class concurrent_queue {
...
     }
 
     concurrent_queue& operator=( concurrent_queue&& other ) {
-        //TODO: implement support for std::allocator_traits::propagate_on_container_move_assignment
-        if (my_queue_representation != other.my_queue_representation) {
-            clear();
+        if (my_queue_representation == other.my_queue_representation)
```
kboyarinov, line 151: Same as above

include/oneapi/tbb/concurrent_queue.h
```diff
@@ -137,25 +137,29 @@ class concurrent_queue {
...
+            return *this;
+        clear();
+        if (queue_allocator_traits::propagate_on_container_move_assignment::value) {
+            my_allocator = std::move(other.my_allocator);
+            internal_swap(other);
```
kboyarinov, line 156: I am afraid that with the full support for stateful allocators, propagation and uses-allocator construction, we cannot use `internal_swap` as an implementation for move semantics anymore:

Consider `this->my_allocator` to be stateful allocator. 
After `internal_swap`, the memory allocated by `*this` would be transferred to `other` but the allocator with the correct state would not be transfered to `*this` (and it cannot be since the standard requires us to move-construct allocators. 
Because of these, we will need a "fair" move semantics for both constructor and the assignment operator

include/oneapi/tbb/concurrent_queue.h
```diff
@@ -178,8 +182,12 @@ class concurrent_queue {
     }
 
     void swap ( concurrent_queue& other ) {
-        //TODO: implement support for std::allocator_traits::propagate_on_container_swap
-        __TBB_ASSERT(my_allocator == other.my_allocator, "unequal allocators");
+        if (queue_allocator_traits::propagate_on_container_swap::value) {
```
kboyarinov, line 185: Please use `tbb::detail::swap_allocators`

include/oneapi/tbb/concurrent_queue.h
```diff
@@ -178,8 +182,12 @@ class concurrent_queue {
...
+            using std::swap;
+            swap(my_allocator, other.my_allocator);
+        } else {
+            __TBB_ASSERT(my_allocator == other.my_allocator, "unequal allocators");
```
kboyarinov, line 189: Lets move it inside `internal_swap`

kboyarinov (commented):
@YexuanXiao, many thanks for working on this gap in our containers!

include/oneapi/tbb/concurrent_queue.h
YexuanXiao: my_queue_representation is allocated by r1::cache_aligned_allocate instead of my_allocator and all memory allocated by my_allocator has been deallocated at line 153, so I don’t think this is an issue.

include/oneapi/tbb/concurrent_queue.h
YexuanXiao: I reviewed all the containers’ `internal_swap`, which typically have two overloads. I think they can be combined into a single function using `if`:

```cpp
void internal_swap(concurrent_queue& src) {
   if (!allocator_traits_type::queue_allocator_traits::propagate_on_container)
        __TBB_ASSERT(my_allocator == src.my_allocator, "Swapping with unequal allocators is not allowed");
    using std::swap;
    swap(my_queue_representation, src.my_queue_representation);
}
```
The compiler will optimize it and eliminate the unreachable branch.
Allocator helpers have the same pattern. Do we need to perform these cleanups, or should we keep things as they are?

PR #1522, opened 2024-10-01
[RFC] Adding API for parallel block to task_arena to warm-up/retain/release worker threads

pavelkumbrasev:
Adding API for parallel block to task_arena to warm-up/retain/release worker threads


rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,108 @@
...
+the default block time will be restored.
+
+But what if user would like to disable default block time entirely?<br>
+Because the heuristic of extended block time is unsuitable for the task submitted
+in unpredictable pattern and duration. In this case, there should be an API to disable
+the default block time in the arena entirely.
+
+```cpp
+class task_arena {
+    void indicate_start_of_parallel_block(bool do_warmup = false);
```
vossmjp, line 74: How about

`void retain_threads();`
`void release_threads();`


`*_parallel_block` is misleading since even your example shows serial parts of the region.

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,108 @@
...
+    void indicate_end_of_parallel_block(bool disable_default_block_time = false);
+    void disable_default_block_time();
+    void enable_default_block_time();
+};
+
+namespace this_task_arena {
+    void indicate_start_of_parallel_block(bool do_warmup = false);
+    void indicate_end_of_parallel_block(bool disable_default_block_time = false);
+    void disable_default_block_time();
```
vossmjp, line 83: The end-user doesn't know what the default block time is, and it will be platform dependent. The first set of functions indicate of region of interest, these `default_block_time` functions change a property on the task_arena that is not tied to a region. That makes me think it is better as a constraint. Are there known cases where this needs to be disabled then reenabled dynamically?  

If the first two functions became something like `retain_threads` and `release_threads`, what would these be named? What about `set_sleep_policy( sleep_quickly | sleep_slowly )` or something like that.

pavelkumbrasev: Perhaps, it is a good idea to move it constraints. If you need different guarantees just use different arenas the same is  applicable to priorities.
If we include the property as part of the constraints which in turn represent HW Resources perhaps the name should represent how these resources will be used, like `greedy` or something.

@akukanov what do you think?

rfcs/proposed/parallel_block_for_task_arena/README.md
pavelkumbrasev: Hmmm, I think `retain_threads` and `release_threads` provides unnecessary guaranteed like the threads will be actually retained.   
Should it be something more relaxed?
Perhaps, `make_sticky` and `make_unsticky` suites better because we can set definition of `sticky`.

rfcs/proposed/parallel_block_for_task_arena/README.md
vossmjp: A couple questions:

- How would users think about "greedy" relative to per-arena priorities? It might imply some kind of priority between a greedy normal arena and non-greedy normal arena, even though that wouldn't be the case.
- In suggesting `sleep_quickly` and `sleep_slowly`, which I admit are not great names, I was trying to find something that indicated more about the wastefulness of holding onto resources once you have them and *while there's nothing better to do with them* in contrast to a greediness in acquiring resources, perhaps from some other competing arenas. I think this is the key point of `disable|enable_default_block_time` -- while it is a form of greediness, it is more about the amount of wastefulness tolerated to reduce startup latency on the next parallel algorithm.

rfcs/proposed/parallel_block_for_task_arena/README.md
vossmjp: Do think "sticky" could people think of thread-to-core affinity? Since constraints are used for affinity, I'm thinking the likelihood of confusion is low and so I'm ok with `make_sticky` and `make_unsticky`.

aleksei-fedotov: Should not all this be about work rather than threads? Threads are the execution resources, that should not be exposed to the user, should they? I mean that is the original idea of the TBB library. Therefore, I suggest something like `expect_[more/less_]parallel_work` or `assume_[more/less_]parallelism` as a good level of a loose terminology what library should tend to "_think_" about user's code when this API is used.

rfcs/proposed/parallel_block_for_task_arena/README.md
aleksei-fedotov: When I am thinking about this, thoughts about being nice/responsive to the demand from other arenas appear in my head. But I am not sure how to better combine these two sets of API as they kind of mutually exclusive to each other. Consider, _expect more parallel work to appear but be responsive to resources demand from other arenas_. Actually, this counterintuition applies to the proposed design as well. Perhaps, we need to express that mutual exclusiveness somehow in the API.

aleksei-fedotov (commented):
Overall, it looks as too certain about the things that will or will not happen when the new API is utilized. I think that the explanation should be written in a more vague terms using the more of "may", "might", etc. words. Essentially, conveying the idea that all this is up to the implementation and serve as a hint rather than a concrete behavior.

What do others think?

rfcs/proposed/parallel_block_for_task_arena/README.md
pavelkumbrasev: TBB exposes some level of "thread logic" with `observers`. I'm not sure whether this API should expose this logic too.
If we want to extend these functions with additional guarantees such as "warm-up" or "leave earlier"  perhaps we could not ignore threads completely.

rfcs/proposed/parallel_block_for_task_arena/README.md
pavelkumbrasev: > expect more parallel work to appear but be responsive to resources demand from other arenas.

What part of the proposal is stating this?

pavelkumbrasev:
> Overall, it looks as too certain about the things that will or will not happen when the new API is utilized. I think that the explanation should be written in a more vague terms using the more of "may", "might", etc. words. Essentially, conveying the idea that all this is up to the implementation and serve as a hint rather than a concrete behavior.
> 
> What do others think?

I tried to indicate that this set of APIs is a hint to the scheduler. But if you believe that we can relax this guarantees even more I think we should do this.

rfcs/proposed/parallel_block_for_task_arena/README.md
aleksei-fedotov: It is not stating explicitly, but sort of implying the question _What will it mean if I invoke `indicate_start_of_parallel_block` and call `disable_default_block_time` right after that?_

pavelkumbrasev:
Ping @aleksei-fedotov @vossmjp @akukanov

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,178 @@
...
+
+namespace this_task_arena {
+    void parallel_block_start();
+    void parallel_block_end(bool set_one_time_fast_leave = false);
+    scoped_parallel_block_type scoped_parallel_block(bool set_one_time_fast_leave = false);
+}
+```
+
+By the contract, users should call `parallel block start` for each
+subsequent call to `parallel block_end`.<br>
```
akukanov, line 144: I would say the other way around: users should indicate the end of block for each previous start of block.

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,178 @@
...
+Let's introduce RAII scoped object that will help to manage the contract.
+
+If the end of the parallel block is not indicated by the user, it will be done automatically when
+the last public reference is removed from the arena (i.e., task_arena is destroyed or a thread
+is joined for an implicit arena). This ensures correctness is
+preserved (threads will not stick forever).
```
akukanov, line 150: Yet another place where "sticking" threads should not be mentioned :)

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,178 @@
...
+  but rather, it brings performance problems can create an arena in “Fast leave” state.
+* And the opposite by default arena will be created in “Delayed leave” state because
+  the delayed leave behavior is a heuristic that benefit most of the workloads.
+
+<img src="completely_disable_new_behavior.png" width=800>
+
+There will be a question that we need to answer:
+* Do we see any value if arena potentially can transition from one to another state?
+  * What if different types of workloads are mixed in one application?
+  * Different types of arenas can be used for different types of workloads.
```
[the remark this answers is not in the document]
pavelkumbrasev: > Are these just random thoughts associated with the main bullet?

Not completely random. These are the questions/statements that were asked as follow ups for the main question.

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,178 @@
...
+
+Let's consider the semantics that an API for explicit parallel blocks can provide:
+* Start of a parallel block:  * Indicates the point from which the scheduler can use a hint and stick threads to the arena.
+  * Serve as a warm-up hint to the scheduler, making some worker threads immediately available
+    at the start of the real computation.
+* "Parallel block" itself:
+  * Scheduler can implement different policies to retain threads in the arena.
+* End of a parallel block:
+  * Indicates the point from which the scheduler can drop a hint and unstick threads from the arena.
+  * Indicates that arena should enter the “One-time Fast leave” thus workers can leave sooner.
```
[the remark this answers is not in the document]
isaevil: Integrated parts from the paragraphs into the list instead. What do you think?

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,178 @@
...
+thus, no real guarantee is provided. The scheduler can ignore the hint and
+move threads to another arena or to sleep if conditions are met.
+
+End of a parallel block:<br>
+Can indicate that arena should enter the “One-time Fast leave” thus workers can leave sooner.
+However, if work was submitted immediately after the end of the parallel block,
+the default arena "workers leave" state will be restored.
+If the default "workers leave" state was the "Fast leave" the result is NOP.
+
+```cpp
```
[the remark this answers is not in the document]
isaevil: Done.

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,178 @@
...
+        fast,
+        delayed
+    };
+
+    task_arena(const constraints& constraints_, unsigned reserved_for_masters = 1,
+               priority a_priority = priority::normal,
+               workers_leave a_workers_leave = workers_leave::delayed);
+
+    void parallel_block_start();
+    void parallel_block_end(bool set_one_time_fast_leave = false);
```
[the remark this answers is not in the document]
isaevil: Done.

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,178 @@
...
+    scoped_parallel_block_type scoped_parallel_block(bool set_one_time_fast_leave = false);
```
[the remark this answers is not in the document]
isaevil: Defined `scoped_parallel_block` with corresponding constructor.

rfcs/proposed/parallel_block_for_task_arena/README.md
isaevil: Done.

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,178 @@
...
+Can indicate that arena should enter the “One-time Fast leave” thus workers can leave sooner.
+However, if work was submitted immediately after the end of the parallel block,
+the default arena "workers leave" state will be restored.
+If the default "workers leave" state was the "Fast leave" the result is NOP.
+
+```cpp
+class task_arena {
+    enum class workers_leave {
+        fast,
+        delayed
```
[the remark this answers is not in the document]
isaevil: We have adjusted it with `task_arena::priority` from specification. How would you change the proposal so it would use `automatic`?

aleksei-fedotov (commented):
A bunch of comment from my side. Have not reviewed the new API yet.

rfcs/proposed/parallel_block_for_task_arena/README.md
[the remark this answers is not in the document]
akukanov: Fine with me, if you prefer so.

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,188 @@
...
+
+With this approach, the user not only releases threads when necessary but also specifies a
+programmable block where worker threads should expected new work coming regularly
+to the executing arena.
+
+Let’s add new state to the existing state machine. To represent "Parallel Block" state.
+
+> **_NOTE:_** The "Fast leave" state is colored Grey just for simplicity of the chart.
+              Assume, that arena was created with the "Delayed leave" and the same logic
+              is applicable to the "Fast leave".
```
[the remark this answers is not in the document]
isaevil: Split into two separate sentences. Rephrased the part about "the logic".

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,188 @@
...
+
+<img src="parallel_block_state_final.png" width=800>
+
+Let's consider the semantics that an API for explicit parallel blocks can provide:
+* Start of a parallel block:
+  * Indicates the point from which the scheduler can use a hint and keep threads in the arena
+    for longer.
+  * Serves as a warm-up hint to the scheduler:
+    * Makes some worker threads immediately available at the start of the real computation.
+    * Should have similar guarantees as `task_arena::enqueue` from a signal standpoint.
```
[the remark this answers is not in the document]
isaevil: Yeah, seems like they duplicate each other. Removed the second bullet.

rfcs/proposed/parallel_block_for_task_arena/parallel_block_state_final.png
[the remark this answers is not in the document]
isaevil: Make arrows be the same color, made arrows thicker.

rfcs/proposed/parallel_block_for_task_arena/README.md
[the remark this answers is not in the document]
akukanov: Sorry, I was not clear. I do not mean `task_arena::automatic`, I mean to rename `delayed` to `automatic` (or some analogous word which is not already in use) to indicate that this is the default task arena behavior.

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,189 @@
...
+
+
+## Open Questions in Design
+
+Some open questions that remain:
+* Are the suggested APIs sufficient?
+* Are there additional use cases that should be considered that we missed in our analysis?
+* Do we see any value if arena potentially can transition from one to another state?
+  * What if different types of workloads are mixed in one application?
+  * What if there concurrent calls to this API?
```
aleksei-fedotov, line 189: See my comment above about making the approach to be a bit more generic. Essentially, I think we can write something like "implementation-defined" in case of a concurrent calls to this API. However, it seems to me that the behavior should be kind of relaxed, so to say. Meaning that if there is at least one "delayed leave" request happening concurrently with possibly a number of "fast leave" requests, then it, i.e., "delayed leave" policy prevails.

Also, having the request stated up front allows scheduler to know the runtime situation earlier, hence making better decisions about optimality of the workers' behavior.

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,189 @@
...
+    void initialize(int max_concurrency, unsigned reserved_for_masters = 1,
+                    priority a_priority = priority::normal,
+                    workers_leave a_workers_leave = workers_leave::delayed);
+
+    void initialize(constraints a_constraints, unsigned reserved_for_masters = 1,
+                    priority a_priority = priority::normal,
+                    workers_leave a_workers_leave = workers_leave::delayed);
+
+    void start_parallel_block();
+    void end_parallel_block(bool set_one_time_fast_leave = false);
```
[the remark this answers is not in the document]
akukanov: There is no composability problem really, as all but the last end-of-block calls are simply ignored, and only the last one has the one-time impact on the leave policy. Also it does not affect the arena settings, according to the design.

Of course if the calls come from different threads, in general it is impossible to predict which one will be the last. However, even if the code is designed to create parallel blocks in the same arena by multiple threads, all these blocks might have the same leave policy so that it does not matter which one is the last to end.

Using the same enum for the end of block as for the construction of the arena seems more confusing than helpful to me, as it may be perceived as changing the arena state permanently.

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,189 @@
...
+  increment/decrement reference counter.
+
+<img src="parallel_block_state_final.png" width=800>
+
+Let's consider the semantics that an API for explicit parallel blocks can provide:
+* Start of a parallel block:
+  * Indicates the point from which the scheduler can use a hint and keep threads in the arena
+    for longer.
+  * Serves as a warm-up hint to the scheduler:
+    * Allows worker threads to be available by the time real computation starts.
```
akukanov, line 100: Do we want the thread warmup to always happen at the start of a block, or do we want to allow users having control over it?
Also, do we promise thread availability "by the time the real computation starts"? I do not think we do, because a) in case too little time has passed after the block start, threads might not yet come, and b) in case too much time has passed, threads might leave.
Maybe more accurate description is like "Allows reducing computation start delays by initiating the wake-up of worker threads in advance".

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,189 @@
...
+
+
+### Proposed API
+
+```cpp
+class task_arena {
+    enum class workers_leave : /* unspecified type */ {
+        fast = /* unspecifed */,
+        delayed = /* unspecifed */
+    };
```
akukanov, line 122: We need to find better names to the enum class and its values.
I am specifically concerned about the use of "delayed" in case the actual behavior might be platform specific, not always delayed. But also `workers_leave` is not a very good name.

rfcs/proposed/parallel_block_for_task_arena/README.md
akukanov, line 141: More on the API names.
- There might be alternative names to consider instead of "block" - e.g., "stage", "phase", "region". Personally, I like any of these three, and especially "phase", quite more than "block".
- `set_one_time_fast_leave` is too verbose and, due to "set", suggests that it might have impact beyond the call. Something like `with_fast_leave` could be better.

rfcs/proposed/parallel_block_for_task_arena/README.md
isaevil: I've been thinking... Combined with your previous comment about `automatic`, perhaps we could have 3 modes instead:
* automatic - platform specific default setting
* fast (or any other more appropriate name)
* delayed (or any other more appropriate name)

If we assume that we have these 3 modes now, `fast` and `delayed` modes would enforce behavior regardless of the platform. That would allow to use `start/end_parallel_block` on hybrid CPUs but if we set `automatic` instead, it would be translated to `fast` option for hybrid systems (to be aligned with current implementation of block time here https://github.com/uxlfoundation/oneTBB/blob/master/src/tbb/waiters.h#L61).
What do you think?

rfcs/proposed/parallel_block_for_task_arena/README.md
isaevil: Seems like "region" might be confused (or wrongly associated) with OpenMP's "parallel Region" term (we have different guarantees). Therefore, I think "stage" or "phase" are more preferred names.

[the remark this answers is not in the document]
aleksei-fedotov: > in general it is impossible to predict which one will be the last

My guess it is the last one that decreases the ref counter to zero. I don't see any issue with this. Later blocks use the arena's policy if not specified explicitly.

> Using the same enum for the end of block as for the construction of the arena seems more confusing than helpful to me, as it may be perceived as changing the arena state permanently.

I indicated the difference in the parameter naming _this_block_leave_, but if it is not enough, we can also indicate that more explicitly with an additional type: _arena_workers_leave_ and _phase_workers_leave_. Nevertheless, my opinion is that it would not be a problem if documentation/specification includes explanation of this.

rfcs/proposed/parallel_block_for_task_arena/README.md
akukanov: Yes, that's an option we can consider. We'd need to think though what the "enforced" delayed mode will guarantee.

For the "automatic" mode, we say that after work completion threads might be retained in the arena for unspecified time chosen by internal heuristics. How would the definition of "enforced" delayed mode differ, what additional guarantees it would provide to make sense for users to choose it?

rfcs/proposed/parallel_block_for_task_arena/README.md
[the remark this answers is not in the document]
akukanov: It seems to me that this discussion is not just about API semantics but really about a different architecture, where each parallel block/stage might have its own customizable retention policy. It differs significantly from what is proposed, so I think it needs deeper elaboration, perhaps with new state change diagrams etc.

rfcs/proposed/parallel_block_for_task_arena/README.md
isaevil: Seems like if we have a dedicated `automatic` policy, that means that `delayed` policy should guarantee at least some level of threads' retention relatively to `fast` policy.

pavelkumbrasev: I think the whole structure is just a hint to the scheduler no real guarantees provided therefore from the description we get:
1) threads will leave without delay with "Fast" 
2) threads might have a delay before leaving with "Delayed" 
3) automatic will decide what state to choose 

From the implementation stand point it has a lot of sense since we will have clear invariants for arena e.g., default arena on hybrid platform will have "Fast" leave state.
So it defiantly improves implementation logic while bringing some potential value to the users ("Delayed" will behave as thread retantion if user explicitly specified).

rfcs/proposed/parallel_block_for_task_arena/README.md
pavelkumbrasev: > Do we want the thread warmup to always happen at the start of a block, or do we want to allow users having control over it?

It seems we already discussed this before and agreed that it should be always on because logically start of parallel block should not only indicate the open brace for parallel computation but also start of these computations.

rfcs/proposed/parallel_block_for_task_arena/README.md
isaevil: @akukanov Regarding enumeration class name. Do you find `leave_policy` a better name than `workers_leave`? Seems more natural to me when using it during arena construction:
```cpp
tbb::task_arena ta{..., tbb::task_arena::leave_policy::fast};
```

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,189 @@
...
+    no longer retain threads in the arena.
+  * Indicates that arena should enter the “One-time Fast leave” thus workers can leave sooner.
+    * If work was submitted immediately after the end of the parallel block,
+      the default arena "workers leave" state will be restored.
+    * If the default "workers leave" state was the "Fast leave" the result is NOP.
+
+
+### Proposed API
+
+```cpp
```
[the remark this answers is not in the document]
isaevil: Done.

rfcs/proposed/parallel_block_for_task_arena/README.md
akukanov: @pavelkumbrasev please describe the semantics of the `delayed` hint in a way that is meaningful for users.

For example, the description I used above 
> after work completion threads might be retained in the arena for unspecified time chosen by internal heuristics.

as well as what you mentioned
> threads might have a delay before leaving with "Delayed"

are good for `automatic` but rather bad for `delayed` because all aspects of the decision are left to the implementation. Even if changed to "will be retained for unspecified time", it would still be rather weak because the time can be any, including arbitrarily close to 0 - that is, it's not really different from `automatic` and so there is no reason to prefer it.

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,200 @@
...
+      in advance.
+* "Parallel phase" itself:
+  * Scheduler can implement different policies to retain threads in the arena.
+  * The semantics for retaining threads is a hint to the scheduler;
+    thus, no real guarantee is provided. The scheduler can ignore the hint and
+    move threads to another arena or to sleep if conditions are met.
+* End of a parallel phase:
+  * Indicates the point from which the scheduler may drop a hint and
+    no longer retain threads in the arena.
+  * Indicates that arena should enter the “One-time Fast leave” thus workers can leave sooner.
```
akukanov, line 110: The wording "should enter..." might create/add to confusion. As already mentioned, from the usage standpoint "one time fast leave" is not a state to enter, but rather a command to execute when the parallel phase has ended and no other one is active.
So I would change to something like "Indicates that worker threads should avoid busy-waiting once there is no more work in the arena".

rfcs/proposed/parallel_block_for_task_arena/README.md
pavelkumbrasev: Sure. Perhaps we are not on the same page. I would like it to be:
1. delayed - after work completion threads might be retained in the arena for unspecified time chosen by internal heuristics.
2. automatic - implementation will choose between "fast" and "delayed"

Automatic is basically another heuristic on choosing "fast" or "leave" based on underlying HW. Perhaps, automatic not the best name.

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,200 @@
...
+    * If work was submitted immediately after the end of the parallel phase,
+      the default arena behavior with regard to "workers leave" policy is restored.
+    * If the default "workers leave" policy was the "Fast leave", the result is NOP.
```
akukanov, line 113: I would replace both these bullets with e.g. "Temporarily overrides the default arena policy/behavior, which will be restored when new work is submitted"

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -62,100 +62,111 @@ aspect of oneTBB behavior should be able to do it.
...
-    void end_parallel_block(bool set_one_time_fast_leave = false);
+    void start_parallel_phase();
+    void end_parallel_phase(bool with_fast_leave = false);
 }
 ```
 
-By the contract, users should indicate the end of _parallel block_ for each
-previous start of _parallel block_.<br>
+By the contract, users should indicate the end of _parallel phase_ for each
+previous start of _parallel phase_.<br>
```
akukanov, line 166: Maybe "The parallel phase continues until for each previous `start_parallel_phase` call a matching `end_parallel_phase` call has been made."

rfcs/proposed/parallel_block_for_task_arena/README.md
akukanov: With this definition, I see no difference between "automatic" and "delayed", because in both cases the decision of whether to delay or not to delay, and for how long, is left to the implementation. If that is the intended behavior, let's not complicate the API with a redundant enum value.

isaevil: Okay, let's have `fast` and `automatic` policies then since we're not sure right now whether we can provide some meaningful guarantees for `delayed` policy to user.

rfcs/proposed/parallel_block_for_task_arena/README.md
isaevil: What about "The _parallel phase_ continues until each previous `start_parallel_phase` call has a matching `end_parallel_phase` call"?

rfcs/proposed/parallel_block_for_task_arena/README.md
isaevil: Done.

rfcs/proposed/parallel_block_for_task_arena/README.md
isaevil: Replaced to the suggested one.

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,200 @@
...
+  * Indicates the point from which the scheduler can use a hint and keep threads in the arena
+    for longer.
+  * Serves as a warm-up hint to the scheduler:
+    * Allows reducing computation start delays by initationg the wake-up of worker threads
+      in advance.
+* "Parallel phase" itself:
+  * Scheduler can implement different policies to retain threads in the arena.
+  * The semantics for retaining threads is a hint to the scheduler;
+    thus, no real guarantee is provided. The scheduler can ignore the hint and
+    move threads to another arena or to sleep if conditions are met.
```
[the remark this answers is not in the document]
isaevil: I am not entirely sure how to better formulate that. I provided an example of how possible implementation of "parallel phase" can benefit its user.

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,241 @@
...
+Let's introduce RAII scoped object that will help to manage the contract.
+
+If the end of the parallel phase is not indicated by the user, it will be done automatically when
+the last public reference is removed from the arena (i.e., task_arena is destroyed or a thread
+is joined for an implicit arena). This ensures correctness is
+preserved (threads will not be retained forever).
+
+### Examples
+
+Following code snippets show how the new API can be used.
```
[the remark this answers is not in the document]
isaevil: Added snippet with `scoped_parallel_phase`.

akukanov (dismissed):
I support the proposal.

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -129,7 +130,6 @@ class task_arena {
     enum class leave_policy : /* unspecified type */ {
         automatic = /* unspecifed */,
         fast = /* unspecifed */,
-        delayed = /* unspecifed */
     };
```
dnmokhov: Wouldn't the user still need an explicit `delayed` to override an implementation choosing `fast` on a hybrid system?

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,267 @@
+# Adding API for parallel phase to task_arena to warm-up/retain/release worker threads
+
+## Introduction
+
+In oneTBB, there has never been an API that allows users to block worker threads within the arena.
+This design choice was made to preserve the composability of the application.<br>
```
vossmjp, line 6:
```suggestion
This design choice was made to preserve the composability of the application.
```

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,267 @@
...
+Before PR#1352, workers moved to the thread pool to sleep once there were no arenas with active
+demand. However, PR#1352 introduced a delayed leave behavior to the library that
+results in blocking threads for an _implementation-defined_ duration inside an arena
+if there is no active demand arcoss all arenas. This change significantly
+improved performance for various applications on high thread count systems.<br>
```
vossmjp, line 11:
```suggestion
improved performance for various applications on high thread count systems.
```

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,267 @@
...
+The main idea is that usually, after one parallel computation ends,
+another will start after some time. The delayed leave behavior is a heuristic to utilize this,
+covering most cases within _implementation-defined_ duration.
+
+However, the new behavior is not the perfect match for all the scenarios:
+* The heuristic of delayed leave is unsuitable for the tasks that are submitted
+  in unpredictable pattern and/or duration.
```
vossmjp, line 18:
```suggestion
  in an unpredictable pattern and/or durations.
```

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,267 @@
...
+* If oneTBB is used in composable scenarios it is not behaving as
+  a good citizen consuming CPU resources.
+  * For example, if an application builds a pipeline where oneTBB is used for one stage
```
vossmjp, line 21: Since oneTBB has parallel_pipeline, this might be confusing since a TBB pipeline would still be using TBB tasks during an filter with nested OpenMP.  So perhaps "if an applications runs a series of stages"

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,267 @@
...
+    and OpenMP is used for a subsequent stage, there is a chance that oneTBB workers will
+    interfere with OpenMP threads. This interference might result in slight oversubscription,
+    which in turn might lead to underperformance.
+
+So there are two related problems but with different resolutions:
+* Completely disable new behavior for scenarios where the heuristic of delayed leave is unsuitable.
+* Optimize library behavior so customers can benefit from the heuristic of delayed leave but
+  make it possible to indicate that "it is the time to release threads".
```
vossmjp, line 29:
```suggestion
  make it possible to indicate that "it is the time for the TBB arena to release threads".
```

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,267 @@
...
+
+### Completely disable new behavior
+
+Let’s consider both “Delayed leave” and “Fast leave” as 2 different states in state machine.<br>
+* The "Delayed leave" heuristic benefits most of the workloads. Therefore, this is the 
+  default behavior for arena. 
+* Workloads that has rather negative performance impact from the heuristic of delayed leave
+  can create an arena in “Fast leave” state.
+
+<img src="completely_disable_new_behavior.png" width=800>
```
vossmjp, line 43: The arrows and endpoint are very thin, so its hard to see the direction. Maybe make them thicker / larger?

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,267 @@
...
+
+There will be a question that we need to answer:
+* Do we see any value if arena potentially can transition from one to another state?
+
+To answer this question, the following scenarios should be considired:
```
vossmjp, line 48:
```suggestion
To answer this question, the following scenarios should be considered:
```

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,267 @@
...
+* What if different types of workloads are mixed in one application?
+* Different types of arenas can be used for different types of workloads.
+
+### When threads should leave?
+
+oneTBB itself can only guess when the ideal time to release threads from the arena is.
+Therefore, it does the best effort to preserve and enhance performance without completely
```
vossmjp, line 55:
```suggestion
Therefore, it does its best effort to preserve and enhance performance without completely
```

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,267 @@
...
+messing composability guarantees (that is how delayed leave is implemented).
```
vossmjp, line 56:
```suggestion
messing up composability guarantees (that is how delayed leave is implemented).
```

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,267 @@
...
+therefore customers that want to further optimize this
+aspect of oneTBB behavior should be able to do it.
+
+This problem can be considered from another angle. Essentially, if the user can indicate
+where parallel computation ends, they can also indicate where it starts.
+
+<img src="parallel_phase_introduction.png" width=800>
+
+With this approach, the user not only releases threads when necessary but also specifies a
+programmable block where worker threads should expected new work coming regularly
```
vossmjp, line 68:
```suggestion
programmable block where worker threads should expect new work coming regularly
```

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,267 @@
...
+to the executing arena.
+
+Let’s add new state to the existing state machine. To represent "Parallel Phase" state.
```
vossmjp, line 71:
```suggestion
Let’s add a new state to the existing state machine. To represent "Parallel Phase" state.
```

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,267 @@
...
+
+> **_NOTE:_** The "Fast leave" state is colored Grey just for simplicity of the chart.
+              Let's assume that arena was created with the "Delayed leave". 
+              The logic demonstrated below is applicable to the "Fast leave" as well.
+
+<img src="parallel_phase_state_initial.png" width=800>
+
+This state diagram leads to several questions. There are some of them:
```
vossmjp, line 79:
```suggestion
This state diagram leads to several questions:
```

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,267 @@
...
+* What if there are multiple Parallel Phases?
+* If “End of Parallel Phase” leads back to “Delayed leave” how soon threads
```
vossmjp, line 81:
```suggestion
* If “End of Parallel Phase” leads back to “Delayed leave” how soon will threads
```

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,267 @@
...
+  will be released from arena?
```
vossmjp, line 82:
```suggestion
  be released from arena?
```

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,267 @@
...
+  * What if we indicated that threads should leave arena after the "Parallel Phase"?
+  * What if we just indicated the end of the "Parallel Phase"?
+
+The extended state machine aims to answer these questions.
+* The first call to the “Start of Phase” will transition into the “Parallel Phase” state.
+* The last call to the “End of Phase” will transition back to the “Delayed leave” state
+  or into the "One-time Fast leave" if it is indicated that threads should leave sooner.
+* Concurrent or nested calls to the “Start of Phase” or the “End of Phase”
+  increment/decrement reference counter.
```
vossmjp, line 91:
```suggestion
  increment/decrement a reference counter.
```

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,267 @@
...
+      in advance.
+* "Parallel phase" itself:
+  * Scheduler can implement different policies to retain threads in the arena.
+    * For instance, more aggressive policy might be implemented for _parallel phase_.
+      It can be beneficial in cases when the default arena leave policy is not sufficient enough.
+  * The semantics for retaining threads is a hint to the scheduler;
+    thus, no real guarantee is provided. The scheduler can ignore the hint and
+    move threads to another arena or to sleep if conditions are met.
+* End of a parallel phase:
+  * Indicates the point from which the scheduler may drop a hint and
```
vossmjp, line 110:
```suggestion
  * Indicates the point from which the scheduler may drop the hint and
```

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,267 @@
...
+    };
+};
+
+namespace this_task_arena {
+    void start_parallel_phase();
+    void end_parallel_phase(bool with_fast_leave = false);
+}
+```
+The _parallel phase_ continues until each previous `start_parallel_phase` call
+has a matching `end_parallel_phase` call.<br>
```
vossmjp, line 165:
```suggestion
to the same arena has a matching `end_parallel_phase` call.<br>
```

rfcs/proposed/parallel_block_for_task_arena/README.md
dnmokhov: Or do we just expect the user to always create a parallel phase in such situation?

vossmjp: Yes, I think the expectation is to use a parallel phase. I suppose the danger with a "delayed" would be that is less targeted that "fast" or the parallel phase; since it overrides the default for the entire lifetime of the arena. Overriding with "fast" makes sense, since it alters the behavior for each execute call to the arena. Adding some kind of block around a single call would be strange to add "fast".  But "delayed" is about identifying a region that contains multiple execute calls and where there should be stickiness to the arena.

rfcs/proposed/parallel_block_for_task_arena/README.md
isaevil: Is it applicable only for this picture or would you like other pictures to have thicker arrows?

rfcs/proposed/parallel_block_for_task_arena/README.md
vossmjp: I think the last-block-to-end is a good way to break any conflicts, and this is the currently proposed behavior. If the last to end is "with_fast_leave", there's is some conflicting runtime use about to happen, so seems like the right thing to do is fast leave. If the last to end is not "with_fast_leave" on an arena that was created with "automatic" that means "automatic", so it still could opt to leave fast. As I read the state diagram, it doesn't really say it has to be a delayed leave if the final decrement was from a non fast leave end. So this looks like something to evaluate during implementation phase.

vossmjp (dismissed):
Proposal looks go to me. Maybe fix the size of lines/arrows in diagrams, but that is a superficial change. Seems like remaining questions are best addressed through by implementation experience.

rfcs/proposed/parallel_block_for_task_arena/README.md
isaevil: Made them thicker.

aleksei-fedotov (approved):
Looks good enough to me. Maybe, one question requires some sort of clarification.

rfcs/proposed/parallel_block_for_task_arena/README.md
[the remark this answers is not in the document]
aleksei-fedotov: OK, makes sense.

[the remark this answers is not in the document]
aleksei-fedotov: Definitely, thanks for the explanation! I tried to generalize the interface to a more versatile use cases, but it really seems to be about different architecture. Current proposal retains threads within an arena during a parallel phase with a possibility for future extension to have different retaining policies.

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,267 @@
...
+}
+```
+The _parallel phase_ continues until each previous `start_parallel_phase` call
+to the same arena has a matching `end_parallel_phase` call.<br>
+Let's introduce RAII scoped object that will help to manage the contract.
+
+If the end of the parallel phase is not indicated by the user, it will be done automatically when
+the last public reference is removed from the arena (i.e., task_arena is destroyed or a thread
+is joined for an implicit arena). This ensures correctness is
+preserved (threads will not be retained forever).
```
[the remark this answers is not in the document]
akukanov: "A thread is joined for an implicit arena" is not the best choice of words, but what it means is really "for an implicitly created arena, the thread that owns it has completed". "Joining a thread" is the term that means "waiting for a thread to complete".

rfcs/proposed/parallel_block_for_task_arena/README.md
```diff
@@ -0,0 +1,267 @@
...
+    void initialize(constraints a_constraints, unsigned reserved_for_masters = 1,
+                    priority a_priority = priority::normal,
+                    leave_policy a_leave_policy = leave_policy::automatic);
+
+    void start_parallel_phase();
+    void end_parallel_phase(bool with_fast_leave = false);
+
+    class scoped_parallel_phase {
+        scoped_parallel_phase(task_arena& ta, bool with_fast_leave = false);
+    };
```
akukanov, line 156: Looks like the scoped object can only be created for a known task_arena. How to use it with "this task arena"? Is it necessary to always "materialize" the arena with the help of the attaching constructor? Or should there be a special constructor for this class, either with no parameters or maybe with `tbb::attach` parameter?

aleksei-fedotov: `tbb::attach` is [listed as deprecated interface](https://oneapi-spec.uxlfoundation.org/specifications/oneapi/latest/elements/onetbb/source/deprecated/task_arena_attach_tag) in the spec. Not sure why.

isaevil: @aleksei-fedotov I believe it is `tbb::task_arena::attach` being deprecated since there is now a [dedicated ](https://oneapi-spec.uxlfoundation.org/specifications/oneapi/latest/elements/onetbb/source/task_scheduler/attach_tag_type)`tbb::attach` tag used for both `task_arena` and `task_scheduler_handle`.

isaevil: > Looks like the scoped object can only be created for a known task_arena. How to use it with "this task arena"? Is it necessary to always "materialize" the arena with the help of the attaching constructor? Or should there be a special constructor for this class, either with no parameters or maybe with a `tbb::attach` parameter?

With the current version of proposed API the only option is to create an arena with `tbb::attach` and then pass it to the `scoped_parallel_phase` constructor.
```cpp
tbb::task_arena ta{tbb::attach{}};
{
   tbb::task_arena::scoped_parallel_phase phase{ta};
   // parallel work
}
```
I guess it makes sense to improve usability and move this kind of logic under the hood. However, I am not sure about the way to approach it:
* I suppose using `tbb::attach` as a parameter makes sense when `scoped_parallel_phase` is within `task_arena` class scope.
* Having `tbb::attach` as a parameter when `scoped_parallel_phase` is within namespace `this_task_arena` scope seems excessive because `tbb::this_task_arena::scoped_parallel_phase` with no parameters is already self-explanatory.

Should we just introduce a new constructor with `tbb::attach` to `tbb::task_arena::scoped_parallel_phase` or should we have a separate `tbb::this_task_arena::scoped_parallel_phase` which doesn't accept `task_arena` at all?

rfcs/proposed/parallel_block_for_task_arena/README.md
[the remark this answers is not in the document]
isaevil: Changed wording to remove confusion with "joining the arena".

rfcs/proposed/parallel_block_for_task_arena/README.md
akukanov: I prefer not to introduce an almost identical class. Currently, neither the `this_task_arena` namespace nor the similar `std::this_thread` namespace contain classes, but only functions.

So I prefer a special constructor, and I slightly prefer it to have the attach parameter, though a no-parameter constructor would probably be OK too.

Another question is: what this new constructor would do when an object is created in a thread that does not yet have any associated task arena?

Keeping these questions open and addressing later would be fine for me as well.

isaevil: Documented these questions in "Open Questions in Design".

isaevil:
@aleksei-fedotov @akukanov I believe the remaining comments were addressed in some kind of way. Given the fact that you previously approved the PR (@vossmjp has already approved prior those comments), I suppose I just need at least one approve from you so we could merge it to the "proposed" folder.

aleksei-fedotov:
> @aleksei-fedotov @akukanov I believe the remaining comments were addressed in some kind of way. Given the fact that you previously approved the PR (@vossmjp has already approved prior those comments), I suppose I just need at least one approve from you so we could merge it to the "proposed" folder.

It seems we have not fully discussed [the example from this comment](https://github.com/uxlfoundation/oneTBB/pull/1522#discussion_r1871505223).

There is actually possibility to avoid such kind of mistakes. The idea is to constrain the API to only `scoped_parallel_phase` type, not introducing `start_parallel_phase` and `end_parallel_phase` functions whose number may not match.
By making the instances of `scoped` objects owning we effectively utilize the recent findings wrt better code quality and programming efficiency the community is nowadays constantly speaking about. Though, at the expense of a bit of inconvenience on the user side. These are:
- RAII style - users won't ever forget about releasing of resources. Note that the "Resource" word in the abbreviation is directly about CPU resources that the parallel phase is needed. Perhaps, not just a coincidence.
- Functional programming - user functions become more encapsulated => no side-effects wrt resource retaining. Consider my example to be wrapped in a function that is called dozens of times - and all this with only increments to the reference counting as there are no explicit phase ends - Definitely must be a unit-test, BTW. Though I admit that with programming braces API and reset-like implementation of reference counting the effect seems to be only in single-time prolonged retaining of CPU resources. Not sure about the exact moment when to reset it, especially for nested cases.

The inconvenience is that users need to create, copy, or move instances of `scoped`-objects whenever they want to prolong resource retaining.  

So, I just wanted to make sure we considered suggested approach and made more conscious design choice.

rfcs/proposed/parallel_block_for_task_arena/README.md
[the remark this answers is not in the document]
akukanov: > // some long enough period of time
// during which there is an expectation that
// threads are still inside 'ta'

It's a wrong expectation that should be ideally minimized by the proper documentation. Parallel phase is a hint, the time for threads to remain "inside" is defined by implementation heuristics.

> // As far as I understand, the retaining happens until task_arena is destroyed,
// which could be long.

No, it does not have to, and it's unlikely that we will want to do that.

akukanov:
@aleksei-fedotov, @isaevil - I suggest to add the question of RAII-only vs. RAII + functions API for parallel phase as another open question. I guess some practical usage experience is needed to judge; also, I think the downsides of forgetting to end a phase are not severe, as it should not cause endless busy-waiting (see https://github.com/uxlfoundation/oneTBB/pull/1522#discussion_r1874028341).

isaevil:
> @aleksei-fedotov, @isaevil - I suggest to add the question of RAII-only vs. RAII + functions API for parallel phase as another open question. I believe some practical usage experience is needed to judge; also, I think the downsides of forgetting to end a phase are not severe, as it should not cause endless busy-waiting (see [#1522 (comment)](https://github.com/uxlfoundation/oneTBB/pull/1522#discussion_r1874028341)).

@aleksei-fedotov @akukanov I have added this question to open questions section. I think that having "RAII + functions" during the "proposed/experimental" stage is fine until we have better understanding about usage to the moment this RFC will be reconsidered for "supported" stage. I agree with @akukanov on severity of forgetting to end parallel phase since _parallel phase_ 
 doesn't have strong guarantees of threads retention.

aleksei-fedotov (approved):
Looks good to me!

akukanov (approved):

PR #1524, opened 2024-10-07
Enable CTAD for blocked_rangeNd since C++17

rarutyun:
### Description 
Enable CTAD for blocked_rangeNd since C++17 by making the implementation through the inheritance rather than through alias template.

### Type of change

### Tests

- [X] not needed

### Documentation

- [ ] needs to be updated
- [X] not needed

- [x] Yes
- [x] No
- [ ] Unknown

Technically, the change breaks backward compatibility but this is a preview functionality, where breaking change is acceptable

@pavelkumbrasev, @vossmjp, @akukanov, @isaevil, @aleksei-fedotov, @kboyarinov 

### Other information
Of course, this PR should take into account the changes made in #1449. I cloned 1449 and checked my modifications there. Didn't find any broken tests (I might miss something, though).

Please provide feedback on my PR if anybody sees any issues with proposed changes

### DIFF
--- a/include/oneapi/tbb/blocked_nd_range.h
+++ b/include/oneapi/tbb/blocked_nd_range.h
@@ -47,11 +47,9 @@ namespace d1 {
 */
 
 template<typename Value, unsigned int N, typename = detail::make_index_sequence<N>>
-    __TBB_requires(blocked_range_value<Value>)
 class blocked_nd_range_impl;
 
 template<typename Value, unsigned int N, std::size_t... Is>
-    __TBB_requires(blocked_range_value<Value>)
 class blocked_nd_range_impl<Value, N, detail::index_sequence<Is...>> {
 public:
     //! Type of a value.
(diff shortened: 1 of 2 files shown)


kboyarinov:
I have tried to test the implementation locally and found that it enables the CTAD only for copy/move constructors of `blocked_rangeNd`. 
I had an idea on how can CTAD for the main constructor can be also enabled. I have opened a separate PR #1525 to add tests and this explicit deduction guide.

rarutyun:
> I have tried to test the implementation locally and found that it enables the CTAD only for copy/move constructors of `blocked_rangeNd`. I had an idea on how can CTAD for the main constructor can be also enabled. I have opened a separate PR #1525 to add tests and this explicit deduction guide.

@kboyarinov, I know that. The idea of my PR is to enable CTAD in principle, because it doesn't completely work with aliases prior C++20. I didn't write the deduction guides (which is just a "substory" of CTAD) intentionally. It's a different story and I propose to solve it differently. As far as I can tell, your PR doesn't enable everything that necessary (at least in my mind).

rarutyun:
> OK, no problem. In that case the patch looks fine for me, except for copyright years that should be actualized.

Thanks. And as you can guess, copyright issue is a 10 seconds fix ;)

vossmjp (approved):
Looks go to me.

PR #1525, opened 2024-10-08
Add explicit deduction guides for blocked_nd_range

kboyarinov:
### Description 

### Type of change

### Tests

- [ ] not needed

### Documentation

- [x] needs to be updated
- [ ] not needed

- [ ] Yes
- [x] No
- [ ] Unknown


include/oneapi/tbb/blocked_nd_range.h
```diff
@@ -147,6 +147,36 @@ class blocked_nd_range : public blocked_nd_range_impl<Value, N> {
...
+-> blocked_nd_range<Value, N>;
+
+template <typename Value, unsigned int N>
+blocked_nd_range(blocked_nd_range<Value, N>, oneapi::tbb::split)
+-> blocked_nd_range<Value, N>;
+
+template <typename Value, unsigned int N>
+blocked_nd_range(blocked_nd_range<Value, N>, oneapi::tbb::proportional_split)
+-> blocked_nd_range<Value, N>;
+
```
akukanov, line 177: Splitting constructors are not really for users but for algorithms. Do we need deduction guides for these?

include/oneapi/tbb/blocked_nd_range.h
```diff
@@ -147,6 +147,36 @@ class blocked_nd_range : public blocked_nd_range_impl<Value, N> {
...
+template <typename Value, typename... Values,
+          typename = std::enable_if_t<(... && std::is_same_v<Value, Values>)>>
+blocked_nd_range(blocked_range<Value>, blocked_range<Values>...)
+-> blocked_nd_range<Value, 1 + sizeof...(Values)>;
+
+template <typename Value, unsigned int N,
+          typename = std::enable_if_t<(N != 2 && N != 3)>>
+blocked_nd_range(const Value (&)[N])
+-> blocked_nd_range<Value, N>;
+
```
akukanov, line 165: What if the argument is explicitly an array of 2 or 3 elements, not a braced init list? Seems there is no deduction guide for it.

test/conformance/conformance_blocked_nd_range.cpp
```diff
@@ -291,3 +291,71 @@ TEST_CASE("blocked_nd_range proportional splitting") {
         utils::check_range_bounds_after_splitting(original.dim(0), first.dim(0), second.dim(0), expected_first_end);
     }
 }
+
+#if __TBB_CPP17_DEDUCTION_GUIDES_PRESENT
+//! Testing blocked_rangeNd deduction guides
+//! \brief \ref interface
+TEST_CASE("blocked_rangeNd deduction guides") {
```
akukanov, line 298: The test misses any checks for constructing explicitly from an array. It needs to check construction from arrays of different sizes (incl. 2 and 3), with and without a grainsize argument.

test/conformance/conformance_blocked_nd_range.cpp
```diff
@@ -291,3 +291,71 @@ TEST_CASE("blocked_nd_range proportional splitting") {
...
+    using oneapi::tbb::blocked_nd_range;
+
+    std::vector<const unsigned long*> v;
+    using iterator = typename decltype(v)::iterator;
+
+    oneapi::tbb::blocked_range<int> dim_range(0, 100);
+
+    blocked_nd_range<int, 2> source_range(dim_range, dim_range);
```
akukanov, line 306: Can we test with something more fancy than int for the value type?

akukanov (commented):
In addition to the notes below: it would be great to comment the guides explaining which ctor calls are covered by which.

include/oneapi/tbb/blocked_nd_range.h
kboyarinov: Passing single array of 2 or 3 elements should be considered as a valid use-case, but having an extra deduction guide for array of 2 or 3 will also cover the braced-init-list of size 2 and 3. 
As we discussed braced-init-list of size 2 and 3 should not be supported because of ambiguity between 1-D range constructed from single blocked_range and 2/3-D range constructed from the array.
This use-case is the main open question in the RFC #1607

test/conformance/conformance_blocked_nd_range.cpp
kboyarinov: Done

include/oneapi/tbb/blocked_nd_range.h
kboyarinov: As discussed in the RFC, added support for both single array and single braced-init-list with preference to multi-dimensional ranges

vossmjp (approved):
Looks good to me.

PR #1526, opened 2024-10-09
[RFC] Add fold_node RFC 

kboyarinov:
### Description 
Add RFC document for new `fold_node` in the Flow Graph.

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown

### DIFF
--- a/rfcs/proposed/flow_graph_fold_node/README.md
+++ b/rfcs/proposed/flow_graph_fold_node/README.md
@@ -1,0 +1,226 @@
+# Folding node in the Flow Graph
+
+## Introduction
+
+Current oneTBB Flow Graph API provides basic functional nodes for expressing basic user-scenarios:
+
+* `continue_node` that converts multiple input signals (usually from several different predecessors) to the single signal to each of successors.
+* `function_node` converting each input message from one or several predecessors to the single output message.
+* `multifunction_node` converting each input message into some amount of output messages (zero messages are also allowed) to one or several successors.
+
+But expressing the use-case converting multiple input messages (input stream) into single output is not straightforward to express using only the existing API.
+
+Such an API is extremely useful for expressing _reductions_ (or _folds_) using the oneTBB Flow Graph which is transformation of a sequence of
+objects (_fold input stream_) into a single object (_fold result_) starting from the user-defined initial value (_fold initializer_) by repeating the user defined binary operation (_fold operation_).
+Consider the basic example:
+
+*_S = sum(1, n) = 0 + 1 + 2 + 3 + 4 + ... + n_*
+
+_fold input stream_ is a sequence _(1, 2, 3, 4, ..., n)_, _fold result_ is _S_, _fold initializer_ is _0_ and the _fold operation_ is binary _+_.
+
+This paper proposes to add a new functional node to the oneTBB Flow Graph that would allow the user to specify _fold initializer_ and _fold operation_
+and would compute _fold_ on multiple _fold input stream_s in parallel (both several operations on a single stream and different streams can be
+calculated in parallel).
+
+## Proposal
+
+The new ``oneapi::tbb::flow::fold_node<InputType, OutputType>`` is a new oneTBB Flow Graph functional node that allows to calculate parallel _fold_
+on multiple input streams. The concurrency limit (similar to other functional nodes), the _fold initializer_ and a binary _fold operation_
+should be specified while constructing the node:
+
+```cpp
+tbb::flow::graph g;
+
+tbb::flow::fold_node<int, int> f_node(g, tbb::flow::unlimited,
+    /*fold initializer = */0,
+    /*fold operation = */std::plus<int>{});
+```
+
+Since the ``fold_node`` needs to operate multiple _fold input stream_s in parallel, some identifier of the stream should be provided
+as an input message together with the actual data to compute. It is proposed to use ``tbb::flow::tagged_msg`` class for that purpose
+as an input for the ``fold_node`` that would be sent by the predecessor or by manual ``try_put`` calls:
+
+```cpp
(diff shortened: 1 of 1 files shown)


rfcs/proposed/flow_graph_fold_node/README.md
```diff
@@ -0,0 +1,226 @@
...
+
+int result = 0;
+buf.try_get(result);
+
+// result is expected to be either 25 (result of the first stream fold) or
+// 30 (result of the second stream fold)
+```
+
+From the implementation perspective, there can be multiple approaches for implementing the ``fold_stream_end`` that affects the user API.
+Some of them were considered in a [separate section](#fold_stream_end-implementation-approaches).
```
akukanov, line 106: It seems that all the approaches are based on a message counter. The question then is, at which point the total number of messages to be folded (the stream "size") becomes known. If it is known a priori, or if we want to compute partial folds for each K messages, the counter might be set even at construction. If the stream size is dynamic but known before the first message, we may require that the stream is somehow "initialized" to set that counter.

The most flexible, but also the most complicated and risky approach is to get the stream size at any point, or to get the end of stream message. If the message does not carry any counter, there are problems with potential "still in the flight" messages that might be missed. If the message does carry a counter, what to do if more messages were already received and folded? Also, the message type is kind of problematic, as both the stream size and the end of stream message are likely of different type than the stream data.

A possible design alternative is to have a dedicated port for the "end of stream" signals.

rfcs/proposed/flow_graph_fold_node/README.md
```diff
@@ -0,0 +1,226 @@
...
+### ``fold_stream_end`` containing the number of elements in the stream
+
+The first implementation approach suggests ``fold_stream_end`` to contain the number of elements in the corresponding input stream.
+The ``fold_node`` should support an internal *signed* integer counter of elements for each input stream. Once the node is
+processing the element that is not a ``fold_stream_end``, the counter should be increased by 1 for each element. If the node is
+processing the ``fold_stream_end`` element, the counter should be decreased by _N_, where _N_ is a number of elements in the stream
+which is stored as part of the ``fold_stream_end``.
+
+In this case, having the internal counter not equal to _0_ indicates that there are still elements to process and once it is equal to _0_, the result
+of the fold is considered full and can be propagated to the ``fold_node`` successors.
```
akukanov, line 126: Making the single counter for both expected (negative count) and received (positive count) messages, which triggers at 0, has a downside - it loses the information about the number of messages and cannot pass it down. This value was computed as a byproduct of folding, and in theory it might not be known anywhere else, so it could possibly be useful in the subsequent processing.

rfcs/proposed/flow_graph_fold_node/README.md
```diff
@@ -0,0 +1,226 @@
...
+            // Fold logic for any other streams
+            return (lhs / 2) + (cast_to<int>(rhs) / 2);
+        }
+    });
+```
+
+The same flexible approach can be applied for solving this as well:
+* If the invocation of _fold operation_ with an object of type
+
+It is also an open question, should ``input_type`` be allowed as a first, second, or both operands of _fold operation_.
```
akukanov, line 188: Alternatively, there can be a separate optional argument for the stream index.

rfcs/proposed/flow_graph_fold_node/README.md
```diff
@@ -0,0 +1,226 @@
...
+
+Similar to the previous open question, the flexible approach can be applied to allow both use-cases:
+* If the invocation of the _fold operation_ associated with the ``fold_node`` with the argument of type ``input_type`` (i.e. ``tagged_msg``)
+is well-formed - propagate the index of the stream to the body.
+* Otherwise - the index of the input stream is not propagated to the _fold operation_.
+
+### Should ``InputType`` and ``OutputType`` be different?
```
akukanov, line 195: I think yes, generally these should be different types. If I remember correctly, one of our motivating scenarios is computing a histogram for a data sequence; it is obviously of a different type than the data elements.

rfcs/proposed/flow_graph_fold_node/README.md
akukanov, line 1: I think it is important to outline some use cases which motivate creation of the fold node. It will guide the design and help to prevent "suboptimal" decisions.

A few simplified use case ideas to consider:
- compute the word/token count in a text file read line by line.
- compute the sum of floating-point values, assuming that the sum might be many orders of magnitude bigger than some individual values.
- "reduce by key": for a key-value sequence, compute the value sums for all different keys.

akukanov (commented):
There are several other design questions which I think need to be discussed. I split those into separate comments to organize discussions.

rfcs/proposed/flow_graph_fold_node/README.md
```diff
@@ -0,0 +1,226 @@
...
+```cpp
+tbb::flow::graph g;
+
+tbb::flow::fold_node<int, int> f_node(g, tbb::flow::unlimited,
+    /*fold initializer = */0,
+    /*fold operation = */std::plus<int>{});
+```
+
+Since the ``fold_node`` needs to operate multiple _fold input stream_s in parallel, some identifier of the stream should be provided
+as an input message together with the actual data to compute. It is proposed to use ``tbb::flow::tagged_msg`` class for that purpose
```
vossmjp, line 40:
```suggestion
as an input message together with the actual data to compute. It is proposed to use [``tbb::flow::tagged_msg``](https://oneapi-spec.uxlfoundation.org/specifications/oneapi/latest/elements/onetbb/source/flow_graph/tagged_msg_cls) class for that purpose
```

PR #1532, opened 2024-10-22
Fix race in `concurrent_vector::grow_by()`

### LINKED ISSUE #1531: Deadlock in tbb::concurrent_vector
# Summary
Deadlock in tbb::concurrent_vector occurs in simplest example

# Version
2021.13

# Environment
The following information is important:
* Intel Xeon Gold 6248R CPU
* Windows 10 Enterprise LTSC
* Microsoft Visual Studio Version 17.9.7

# Observed Behavior
deadlock occurs in simplest example
deadlock

# Expected Behavior
absence of deadlock

# Steps To Reproduce
Code example

#include "random"
#include "oneapi/tbb/parallel_for.h"
#include "oneapi/tbb/concurrent_vector.h"

std::mt19937_64 gen;

int main()
{
    constexpr int max_grow_by = 32;
    constexpr int n_inserts = 1024;
    constexpr int n_repits = 1024*128;// *1024;
    auto rand = std::bind(std::uniform_int_distribution <int>{1, max_grow_by }, std::ref(gen));

    std::vector<int> n_grow_by(n_inserts);
    for (int iter = 0; iter < n_repits; ++iter) {
        std::generate(n_grow_by.begin(), n_grow_by.end(), rand);
        tbb::concurrent_vector<double> vec;
        tbb::parallel_for<int>(0, n_inserts, [&](int i) {
            vec.grow_by(n_grow_by[i]);
        });
    }

    return 0;
}

aleksei-fedotov:
### Description 
When thread tries to grow `concurrent_vector`, it first creates locally a new extended segmentation table, with which it then tries to update the global one. However, before this patch, the thread simply updated the global table not taking into account that there might be the other threads doing the same thing. This patch makes the thread _aware_ about its possibly concurrent environment. 

Fixes #1531

### Type of change

### Tests

- [X] updated
- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [X] No
- [ ] Unknown

@npotravkin

### DIFF
--- a/include/oneapi/tbb/concurrent_vector.h
+++ b/include/oneapi/tbb/concurrent_vector.h
@@ -565,14 +565,6 @@ private:
         return new_segment_table;
     }
 
-    void deallocate_long_table(segment_table_type table) {
-        auto& alloc = base_type::get_allocator();
-        for (size_type seg_idx = 0; seg_idx < this->pointers_per_long_table; ++seg_idx) {
-            segment_table_allocator_traits::destroy(alloc, &table[seg_idx]);
-        }
-        segment_table_allocator_traits::deallocate(alloc, table, this->pointers_per_long_table);
-    }
-
     // create_segment function is required by the segment_table base class
     segment_type create_segment( segment_table_type table, segment_index_type seg_index, size_type index ) {
         size_type first_block = this->my_first_block.load(std::memory_order_relaxed);
--- a/include/oneapi/tbb/concurrent_vector.h
+++ b/include/oneapi/tbb/concurrent_vector.h
@@ -1,5 +1,5 @@
 /*
-    Copyright (c) 2005-2022 Intel Corporation
+    Copyright (c) 2005-2024 Intel Corporation
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
@@ -565,6 +565,14 @@ private:
         return new_segment_table;
     }
 
+    void deallocate_long_table(segment_table_type table) {
+        auto& alloc = base_type::get_allocator();
+        for (size_type seg_idx = 0; seg_idx < this->pointers_per_long_table; ++seg_idx) {
+            segment_table_allocator_traits::destroy(alloc, &table[seg_idx]);
+        }
+        segment_table_allocator_traits::deallocate(alloc, table, this->pointers_per_long_table);
+    }
+
     // create_segment function is required by the segment_table base class
     segment_type create_segment( segment_table_type table, segment_index_type seg_index, size_type index ) {
         size_type first_block = this->my_first_block.load(std::memory_order_relaxed);
@@ -593,7 +601,7 @@ private:
 
             segment_type disabled_segment = nullptr;
             if (table[0].compare_exchange_strong(disabled_segment, new_segment)) {
-                this->extend_table_if_necessary(table, 0, first_block_size);
+                this->extend_table_if_necessary(table, /*start_index*/0, /*end_index*/first_block_size);
                 for (size_type i = 1; i < first_block; ++i) {
                     table[i].store(new_segment, std::memory_order_release);
                 }
@@ -826,8 +834,8 @@ private:
 
     template <typename... Args>
     iterator internal_grow( size_type start_idx, size_type end_idx, const Args&... args ) {
-        this->assign_first_block_if_necessary(this->segment_index_of(end_idx - 1) + 1);
         size_type seg_index = this->segment_index_of(end_idx - 1);
+        this->assign_first_block_if_necessary(seg_index + 1);
         segment_table_type table = this->get_table();
         this->extend_table_if_necessary(table, start_idx, end_idx);
 
--- a/include/oneapi/tbb/detail/_segment_table.h
+++ b/include/oneapi/tbb/detail/_segment_table.h
@@ -279,10 +279,14 @@ public:
     }
 
     void extend_table_if_necessary(segment_table_type& table, size_type start_index, size_type end_index) {
-        // extend_segment_table if an active table is an embedded table
-        // and the requested index is not in the embedded table
+        // Extend segment table if an active table is an embedded one and the requested index is
+        // outside it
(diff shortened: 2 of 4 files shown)


include/oneapi/tbb/concurrent_vector.h
```diff
@@ -565,6 +565,14 @@ class concurrent_vector
         return new_segment_table;
     }
 
+    void deallocate_long_table(segment_table_type table) {
+        auto& alloc = base_type::get_allocator();
+        for (size_type seg_idx = 0; seg_idx < this->pointers_per_long_table; ++seg_idx) {
+            segment_table_allocator_traits::destroy(alloc, &table[seg_idx]);
+        }
+        segment_table_allocator_traits::deallocate(alloc, table, this->pointers_per_long_table);
```
kboyarinov, line 573: Can we introduce this code as an extra method in `segment_table` and reuse in `segment_table::clear_table`?

test/tbb/test_concurrent_vector.cpp
```diff
@@ -692,14 +693,26 @@ TEST_CASE("swap with not always equal allocators"){
...
+                      [&gen, &uniform_dist, &expected_size]() {
+                          const unsigned random_value = uniform_dist(gen);
+                          expected_size += random_value;
+                          return random_value;
+                      });
+
+        tbb::concurrent_vector<double> test_vec;
+        tbb::parallel_for<unsigned>(0, num_inserts, [&] (unsigned j) {
+            test_vec.grow_by(grow_by_vals[j]);
+        });
```
kboyarinov, line 713: It seems like using `unsigned` is not important for the test and we can use `int` and improve readability.

include/oneapi/tbb/concurrent_vector.h
aleksei-fedotov: Good idea! Applied.

test/tbb/test_concurrent_vector.cpp
aleksei-fedotov: Well, this is subjective. Applied anyway though.

include/oneapi/tbb/detail/_segment_table.h
```diff
@@ -260,15 +260,19 @@ class segment_table {
         }
     }
 
+    void destroy_and_deallocate_table(segment_table_type table, size_type num_segments) {
```
kboyarinov, line 263: Is there a specific reason for having `num_segments` as a function parameter?
As far as I can see, `pointers_per_long_table` is used everywhere and it is a `static constexpr` value known by the class.

kboyarinov (commented):
Other than this, the patch looks good for me. 
I would still wait for feedback from @pavelkumbrasev to ensure we did not miss something.

include/oneapi/tbb/detail/_segment_table.h
aleksei-fedotov: This improves encapsulation of separate code pieces, allowing the function to deallocate segment tables of arbitrary size.

kboyarinov: OK, I don't mind.

include/oneapi/tbb/detail/_segment_table.h
```diff
@@ -260,34 +260,49 @@ class segment_table {
...
-
-                    if (table) {
-                        my_segment_table.store(table, std::memory_order_release);
-                    } else {
-                        table = my_segment_table.load(std::memory_order_acquire);
+                    segment_table_type new_table =
+                        self()->allocate_long_table(my_embedded_table, start_index);
+                    // It is possible that the table was extended by the thread that allocated first
+                    // block. In this case, the below CAS fails and re-reads the new table pointer.
+                    if (my_segment_table.compare_exchange_strong(
```
pavelkumbrasev, line 295:
```suggestion
                    if (new_table && my_segment_table.compare_exchange_strong(
```
Perhaps, we should avoid setting `nullptr` table

pavelkumbrasev: That's reasonable

include/oneapi/tbb/detail/_segment_table.h
```diff
@@ -260,34 +260,49 @@ class segment_table {
...
+                    {
+                        // CAS was successful, update the local table pointer with now actual
+                        table = new_table;
+                    } else if (new_table) {
+                        // Other thread was the first to replace the segment table. Current thread's
+                        // table is not needed anymore, so destroying it.
+                        destroy_and_deallocate_table(new_table, pointers_per_long_table);
                     }
                 }).on_exception([&] {
                     my_segment_table_allocation_failed.store(true, std::memory_order_relaxed);
```
kboyarinov, line 308: Would this logic work correct while having one thread which allocates the long table and successfully stores it into the `my_segment_table` and another which allocates the long table and receives a `bad_alloc` while trying to allocate?
It seems like we need to double check this here.

aleksei-fedotov: It depends on what logic we think is correct. Perhaps, not reporting allocation error in case it succeeded by some other thread makes sense, and double checking `my_segment_table` for updated state first can improve algorithm robustness in some cases. Although, I am not sure how frequent these cases can be. Also, it is not full cover anyway, as successful allocation can still sneak in after that double check. Full cover requires recording the failure state into single source of information such as `my_segment_table` or some kind of non-trivial synchronization between multiple information sources, but we don't have that kind of logic through the whole vector right now I guess. So, at most we can write something like the following instead:
```suggestion
                    // Last chance to overcome the failure, hoping that other thread has succeeded
                    // extending the table
                    table = get_table();
                    if (table == my_embedded_table) {
                        my_segment_table_allocation_failed.store(true, std::memory_order_relaxed);
                    }
```
Does it suffice though?

kboyarinov: I think we can keep it as-is for now since I believe this happens once in a blue moon and the ideal solution will require a lot of resources now.

kboyarinov (approved):
LGTM

PR #1535, opened 2024-10-23
[RFC] Added numa_support rfc

vossmjp:
### Description 
Adds RFC for simplified NUMA support

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown

### DIFF
--- a/rfcs/proposed/simplified_numa_support/README.md
+++ b/rfcs/proposed/simplified_numa_support/README.md
@@ -107,7 +107,7 @@ example and be unaware that a HWLOC installation error (or lack of HWLOC) has un
 can see in the preceding example, if we want to spread work across the NUMA nodes in 
 a system we need to query the topology using functions in the `tbb::info` namespace, create
 one `task_arena` per NUMA node, along with one `task_group` per NUMA node, and then add an
-extra loop that iterates overs these `task_arena` and `task_group` objects to execute the
+extra loop that iterates over these `task_arena` and `task_group` objects to execute the
 work on the desired NUMA nodes. We also need to handle all container allocations using OS-specific
 APIs (or behaviors, such as first-touch) to allocator or place them on the appropriate NUMA nodes.
 
--- a/rfcs/proposed/simplified_numa_support/README.md
+++ b/rfcs/proposed/simplified_numa_support/README.md
@@ -103,7 +103,7 @@ the failure to resolve HWLOC at runtime silently returns a default of `task_aren
 default does not pin threads to NUMA nodes. It is too easy to write code similar to the preceding 
 example and be unaware that a HWLOC installation error (or lack of HWLOC) has undone all your effort.
 
-**Getting good performance using these tools requres notable manual coding effort by users.** As we 
+**Getting good performance using these tools requires notable manual coding effort by users.** As we 
 can see in the preceding example, if we want to spread work across the NUMA nodes in 
 a system we need to query the topology using functions in the `tbb::info` namespace, create
 one `task_arena` per NUMA node, along with one `task_group` per NUMA node, and then add an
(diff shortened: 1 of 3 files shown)


rfcs/proposed/simplified_numa_support/README.md
```diff
@@ -0,0 +1,179 @@
...
+
+- Functions in the `tbb::info` namespace **[info_namespace]** 
+  - `std::vector<numa_node_id> numa_nodes()`
+  - `int default_concurrency(numa_node_id id = oneapi::tbb::task_arena::automatic)`
+- `tbb::task_arena::constraints` in **[scheduler.task_arena]**
+
+Below is the example that demonstrates the use of these APIs to pin threads to different 
+arenas to each of the NUMA nodes available on a system, submit work across those `task_arena` 
+objects and into associated `task_group`` objects, and then wait for work again using both 
+the `task_arena` and `task_group` objects.
```
akukanov, line 31: I think the code can be made simpler with `std::thread` per domain, instead of relying only TBB. On the one hand, it also signals that TBB lacks high-level NUMA APIs. On the other hand, TBB, and task arenas specifically, were designed to work well with application level threads where it makes sense. I think it is much better to assume/suggest each NUMA aware arena to be used by a special application thread than to add extra levels of complication with task groups.

vossmjp: This pattern of `task_arenas` and `task_groups` is what we show in our documentation: [for example here](https://oneapi-src.github.io/oneTBB/main/tbb_userguide/Guiding_Task_Scheduler_Execution.html). And, probably as a consequence, a pattern we see in applications that use NUMA constraints.

aleksei-fedotov (commented):
A few typos to fix.

rfcs/proposed/simplified_numa_support/README.md
```diff
@@ -0,0 +1,179 @@
+# Simplified NUMA support in oneTBB
```
akukanov, line 1: I would probably call it "Improved NUMA support".
Correspondingly, the RFC folder could be `numa_support_improvements`, meaning that NUMA support is a core feature and improvements are the gist of the proposal.

aleksei-fedotov: As far as I understand the whole `numa_support_improvements` or `simplified_numa_support` directory will be moved to `rfcs/supported` directory once these improvements are accepted. There may be another set of NUMA improvements in the future which could result in another `numa_support_improvements` directory be created in the same place. And then, when this new set is again accepted, it moves to the same directory. I see potential naming clash issue... It is not related to naming of this directory, but to the naming approach in general. Surely, we could use `numa_support_improvements2` as the name of the new directory, but I believe we can do better from the very beginning.

I propose having the directory with the name related to the feature itself, e.g., `numa_support`, without additionals such as `simplified` or `improvement`. This way we will convey the idea that the documents inside directly affect the support of a particular feature. For resolving naming clashes I propose having the file to be named as precise as possible to what the proposal changes avoiding general terms/adjectives such as `improved`, `increased`, etc.. For example, for sub-RFC that I wrote, I suggest naming the file to something like `introduce_tbbbind_static_library` or `introduce_tbbbind_statically_linked_with_hwloc`; for NUMA-aware allocators name something like `introduce_numa-aware_allocator`; for task_group dependencies name something like `introduce_dependencies_for_tasks_in_task_group`; and so on. This way we would avoid name clashing and still this allows grouping similar rfcs together into dedicated folder such as `numa_support`. Otherwise, I am afraid that the feature is not elaborated enough to be proposed since it sounds too generic in our mind.

vossmjp: I will move to a numa_support directory and change title to "NUMA support"

rfcs/proposed/simplified_numa_support/README.md
vossmjp: Modified to use defer.

akukanov (dismissed):
Looks good for its purpose.

aleksei-fedotov (approved):

PR #1559, opened 2024-11-12
Add RFC for creation and use of NUMA-constrained arenas

aleksei-fedotov:
### Description 
Add sub-RFC to https://github.com/oneapi-src/oneTBB/pull/1535 for creation and use of NUMA-constrained arenas.

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [x] No
- [ ] Unknown


rfcs/proposed/simplified_numa_support/numa-arenas-creation-and-use.org
```diff
@@ -0,0 +1,147 @@
...
+}
+
+for(unsigned j = 0; j < numa_indexes.size(); j++) {
+    arenas[j].execute([&task_groups, &j](){ task_groups[j].wait(); });   // [5]
+}
+#+end_src
+
+Usually the users of oneTBB employ this technique to tie oneTBB worker threads
+up within NUMA nodes and yet have all the parallelism of a platform utilized.
+The pattern allows to find out how many NUMA nodes are on the system. With that
```
vossmjp, line 29:
```suggestion
The pattern starts by finding the number of NUMA nodes on the system. With that
```

rfcs/proposed/simplified_numa_support/numa-arenas-creation-and-use.org
```diff
@@ -0,0 +1,147 @@
...
+number user creates that many ~tbb::task_arena~ objects, constraining each to a
```
vossmjp, line 30:
```suggestion
number, user creates that many ~tbb::task_arena~ objects, constraining each to a
```

rfcs/proposed/simplified_numa_support/numa-arenas-creation-and-use.org
```diff
@@ -0,0 +1,147 @@
...
+for(unsigned j = 0; j < numa_arenas.size(); j++) {
+    numa_arenas[j].wait();
+}
+#+end_src
+
+** New arena interface
+The example above requires new class named ~tbb::constrained_task_arena~. On one
+hand, it is a ~tbb::task_arena~ class that isolates the work execution from
+other parallel stuff executed by oneTBB. On the other hand, it is a constrained
+arena that represents an arena associated to a certain NUMA node and allows
```
vossmjp, line 98: Will there be additional functions added to create arenas constrained to core type too? Or will this be exclusively for NUMA?

rfcs/proposed/simplified_numa_support/numa-arenas-creation-and-use.org
```diff
@@ -0,0 +1,147 @@
...
+  and at most a segmentation fault, since the loop counter might not exist by
+  the time the functor starts its execution.
+- [5] - Synchronization point, where user needs to again make sure corresponding
+  indices are used. Otherwise, the waiting might be done in unrelated
+  ~tbb::task_arena~. It is also possible to mistakenly use
+  ~tbb::task_arena::enqueue~ method with the same consequences as were outlined
+  in the previous bullet, but since it is a synchronization point, usually the
+  blocking call is used.
+
+The proposal below addresses these issues.
```
akukanov, line 67: I'd prefer it not to address all these issues :)

Specifically, I believe that [4] and [5] are really orthogonal to NUMA, and need to be resolved for `task_arena` in general rather than only for its subclass. And even for the subclass I am not sure that "hiding" a task_group inside is the right thing to do.

Keeping task_group and task_arena separate and merely simplifying the combined use of those with some "syntax sugar" allows creating independent "work queues" in the same arena. Also we can think if a similar approach with a flow graph instead of task_group might be useful. Of course all that can be implemented with the hidden task_group, but likely at the expense of some overhead.

rfcs/proposed/simplified_numa_support/numa-arenas-creation-and-use.org
aleksei-fedotov: Applied.

rfcs/proposed/simplified_numa_support/numa-arenas-creation-and-use.org
aleksei-fedotov: Applied.

rfcs/proposed/simplified_numa_support/numa-arenas-creation-and-use.org
aleksei-fedotov: I guess we need to decide on this whether it is needed right away or can be added later in future RFCs. Initially, I wanted writing RFC that addresses the specific concern about verbose and error-prone API for creation of NUMA-constrained arenas.

rfcs/proposed/simplified_numa_support/numa-arenas-creation-and-use.org
akukanov: The code would be in between of the "verbose" and "concise" variants showed in the document, like this:
```cpp
std::vector<tbb::task_arena> numa_arenas =
    tbb::initialize_constrained_arenas(/*possibly some arguments*/);
std::vector<tbb::task_group> task_groups(numa_arenas.size());

for(unsigned j = 0; j < numa_arenas.size(); j++) {
    numa_arenas[j].enqueue( (){/*some parallel stuff*/}, task_groups[j] );
}

for(unsigned j = 0; j < numa_arenas.size(); j++) {
    numa_arenas[j].wait( task_groups[j] );
}
```
or, with modern C++, like this:
```cpp
std::vector<tbb::task_arena> numa_arenas =
    tbb::initialize_constrained_arenas(/*possibly some arguments*/);
std::vector<tbb::task_group> task_groups(numa_arenas.size());

for(auto& [arena, tg]: std::views::zip(numa_arenas, task_groups)) {
    arena].enqueue( (){/*some parallel stuff*/}, tg );
}

for(auto& [arena, tg]: std::views::zip(numa_arenas, task_groups)) {
    arena.wait( tg );
}
```

aleksei-fedotov: I agree that coupling of separate entities is rather a bad thing. Here we would like to improve usability of current interfaces without sacrificing their flexibility. Then the proposal boils down to:
- Introduce the interface that would simplify creation of arenas, each bind to its own NUMA node.
- Introduce the interface that would allow to avoid common mistakes related to loading with work of such arenas.

akukanov: No objections to these two goals.

But I believe the second goal can be achieved without introducing a new class, by extending the existing methods of `task_arena` to better integrate with `task_group`s. That would potentially be useful beyond NUMA scenarios and would give users more control with rather small increase in code complexity.

At the very least, it's an alternative to mention.

rfcs/proposed/numa_support/numa-arenas-creation-and-use.org
```diff
@@ -0,0 +1,147 @@
...
+  slot reserved for master thread, which in this particular example usually
+  results in undersubscription issue as the master thread can join only one
+  arena at a time to help with work processing.
+- [2] - Separate step for instantiation the same number of ~tbb::task_group~
+  objects, in which the actual work is going to be submitted. Note that user
+  also needs to make sure the size of ~arenas~ matches the size of
+  ~task_groups~.
+- [3] - Actual tying of ~tbb::task_arena~ instances with corresponding NUMA
+  nodes. Note that user needs to make sure the indices of ~tbb::task_arena~
+  objects match corresponding indices of NUMA nodes.
```
akukanov, line 52: The point about the need to match the indices is kind of strange. A single loop that works with several arrays/vectors is a typical pattern, you just use the loop index consistently. Moreover, with modern C++ you can rewrite the loop to not have any indices at all, e.g.

```
std::vector<tbb::numa_node_id> numa_indexes = tbb::info::numa_nodes();
std::vector<tbb::task_arena> arenas;      // note that the size is not set
std::vector<tbb::task_group> task_groups; // same for task groups

for (auto idx: numa_indexes) {
    arenas.emplace_back( tbb::task_arena::constraints(idx) );
    task_groups.emplace_back();
    arenas.back().execute([&tg = task_groups.back()]{
        tg.run([]{/*some parallel stuff*/});
    });
}
```

If you meant something else, perhaps try explaining it better.

rfcs/proposed/numa_support/numa-arenas-creation-and-use.org
```diff
@@ -0,0 +1,147 @@
...
+  necessity to know how many objects to initialize further.
+- [1] - Explicit step for creating the number of ~tbb::task_arena~ objects per
+  each NUMA node. Note that by default the arena objects are constructed with a
+  slot reserved for master thread, which in this particular example usually
+  results in undersubscription issue as the master thread can join only one
+  arena at a time to help with work processing.
+- [2] - Separate step for instantiation the same number of ~tbb::task_group~
+  objects, in which the actual work is going to be submitted. Note that user
+  also needs to make sure the size of ~arenas~ matches the size of
+  ~task_groups~.
```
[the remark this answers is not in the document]
vossmjp: I actually don't mind the repetition. Its easy to read over the "same number" without recognizing there's a potential error if the sizes of these two vectors don't match. So, in my opinion, the repetition highlights the potential danger here.

rfcs/proposed/numa_support/numa-arenas-creation-and-use.org
vossmjp: I think the point here might be not that there's a safe way to do it and a well written piece of code will do it the safe way. I think the point is that a not-so-well-written might mess it up. But this specific example (with everything in a loop body) doesn't provide much room for mismatched indices, so it doesn't seem at all likely in this case. In the more general case of task_arenas with a matching number of task_groups, it could be possible to mismatch them.

rfcs/proposed/numa_support/numa-arenas-creation-and-use.org
```diff
@@ -0,0 +1,147 @@
...
+- [3] - Actual tying of ~tbb::task_arena~ instances with corresponding NUMA
+  nodes. Note that user needs to make sure the indices of ~tbb::task_arena~
+  objects match corresponding indices of NUMA nodes.
+- [4] - Actual work submission point. It is relatively easy to make a mistake
+  here by using the ~tbb::task_arena::enqueue~ method instead. In this case not
+  only the work submission might be done after the synchronization point [5],
+  but also the loop counter ~j~ can be mistakenly captured by reference, which
+  at least results in submission of the work into incorrect ~tbb::task_group~,
+  and at most a segmentation fault, since the loop counter might not exist by
+  the time the functor starts its execution.
```
vossmjp, line 59: Since we're highlighting possible points of failure, I suppose we don't want to bring up the safer enqueue deferred tasks patten, right?

vossmjp (approved):
Some optional suggestions. Otherwise, looks good to me as a starting proposal.

rfcs/proposed/numa_support/numa-arenas-creation-and-use.org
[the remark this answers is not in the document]
akukanov: Fair enough, then I would rephrase as:
```suggestion
- [2] - The necessity to instantiate the same number of ~tbb::task_group~
  objects for the actual work to be submitted; that is, the size of ~task_groups~
  must match the size of ~arenas~.
```

rfcs/proposed/numa_support/numa-arenas-creation-and-use.org
akukanov: The sentence says "the user **needs to** make sure ...", and my point is - no, users do not necessarily need to. And we agree that this specific pattern "doesn't provide much room for mismatched indices".

My point is that we should not paint the usage worse than it really is.

rfcs/proposed/numa_support/numa-arenas-creation-and-use.org
akukanov: Well, I would really do want to bring it up.

rfcs/proposed/simplified_numa_support/numa-arenas-creation-and-use.org
aleksei-fedotov: Decided to split onto two RFCs. One will simplify constrained arenas instantiation. The other one will deal with submission of work into constrained arenas.

aleksei-fedotov:
Closed in favor of the less coupled alternative described in #1679.

PR #1570, opened 2024-11-26
Add reference implementation for parallel_phase feature

isaevil:
### Description 

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown


isaevil:
@akukanov @aleksei-fedotov @vossmjp Could you please take a loot at the PR in terms of implementation and ABI.

P.S. Tests are still WIP.

src/tbb/arena.h
```diff
@@ -179,6 +179,76 @@ class atomic_flag {
     }
 };
 
+#if __TBB_PREVIEW_PARALLEL_BLOCK
+class thread_leave_manager {
+    static const std::uint64_t FAST_LEAVE = 1;
+    static const std::uint64_t ONE_TIME_FAST_LEAVE = 1 << 1;
+    static const std::uint64_t DELAYED_LEAVE = 1 << 2;
+    static const std::uint64_t PARALLEL_BLOCK = 1 << 3;
+    static const std::uint64_t PARALLEL_BLOCK_MASK = ~((1LLU << 32) - 1) & ~(0x7);
```
aleksei-fedotov, line 188: Could you please describe in comments what bits are supposed to be set and unset after this operation?

Also, the following mask is equivalent:
```suggestion
    // The PARALLEL_BLOCK_MASK has the most significant 32 bits set
    static const std::uint64_t PARALLEL_BLOCK_MASK = ~((1LLU << 32) - 1);
```

include/oneapi/tbb/task_arena.h
```diff
@@ -95,6 +95,11 @@ TBB_EXPORT void __TBB_EXPORTED_FUNC isolate_within_arena(d1::delegate_base& d, s
 TBB_EXPORT void __TBB_EXPORTED_FUNC enqueue(d1::task&, d1::task_arena_base*);
 TBB_EXPORT void __TBB_EXPORTED_FUNC enqueue(d1::task&, d1::task_group_context&, d1::task_arena_base*);
 TBB_EXPORT void __TBB_EXPORTED_FUNC submit(d1::task&, d1::task_group_context&, arena*, std::uintptr_t);
+
+#if __TBB_PREVIEW_PARALLEL_BLOCK
+TBB_EXPORT void __TBB_EXPORTED_FUNC register_parallel_block(d1::task_arena_base&);
+TBB_EXPORT void __TBB_EXPORTED_FUNC unregister_parallel_block(d1::task_arena_base&, bool);
```
aleksei-fedotov, line 101: Do we anticipate any future extension? In other words, does it make sense to accept an untyped pointer for any future details?

```suggestion
TBB_EXPORT void __TBB_EXPORTED_FUNC unregister_parallel_block(d1::task_arena_base&, void* /*supplemental details*/);
```

Also, I don't think single bit can be passed. At least a word (or even more?) is used anyway. Then, it makes sense to accept at least that width and do not kill opportunity for future extensions:
```suggestion
TBB_EXPORT void __TBB_EXPORTED_FUNC unregister_parallel_block(d1::task_arena_base&, unsigned short/*supplemental details*/);
```
Please double check what size of information is passed.

aleksei-fedotov (commented):
Not a complete review, just a couple of starters to think about.

src/tbb/arena.h
pavelkumbrasev: ` & ~(0x7)` should be used to nullify state bits

include/oneapi/tbb/task_arena.h
isaevil: Yeah, I also have been thinking about that. Perhaps we should pass `std::uintptr_t` instead of `void*`?

src/tbb/arena.h
aleksei-fedotov: I believe in `~((1LLU << 32) - 1)` these bits are already nullified, hence applying `& ~(0x7)` does not change anything. What did I miss?

pavelkumbrasev: Yes. You are correct. However, the current mask is not 😄 
The mask should be `(1LLU << 32) - 1) & ~(0x7)`

isaevil: Yeah, it is incorrect mask. The correct one should be `((1LLU << 32) - 1) & ~(0x7)`.

include/oneapi/tbb/task_arena.h
```diff
@@ -154,6 +168,11 @@ class task_arena_base {
     //! Number of threads per core
     int my_max_threads_per_core;
 
+#if __TBB_PREVIEW_PARALLEL_BLOCK
+    //! Defines the initial behavior of the workers_leave state machine
+    workers_leave my_workers_leave;
+#endif
+
```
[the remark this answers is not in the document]
akukanov: Alternatively, you may not even introduce the new field but instead use a few bits in my_version_and_traits to pass the information about the default retention policy. The value of 0 for these bits should mean "automatic" to be consistent with the current behavior. Also note that the lowest bit is already use for core_type_support_flag, so the retention policy values should not change it when combined with bitwise OR.

include/oneapi/tbb/task_arena.h
akukanov: Following the logic of getting ready for future extensions, perhaps the "register" function should also take an extra argument.

src/tbb/waiters.h
```diff
@@ -54,11 +54,18 @@ class outermost_worker_waiter : public waiter_base {
...
 
         if (is_worker_should_leave(slot)) {
-            if (!governor::hybrid_cpu()) {
+            if (
+#if __TBB_PREVIEW_PARALLEL_BLOCK
+                !my_arena.my_thread_leave.should_leave()
+#else
+            !governor::hybrid_cpu()
+#endif
+               )
```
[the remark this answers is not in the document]
isaevil: I moved the logic to helper function called `can_worker_be_retained` and re-named `should_leave` to `is_retention_allowed`. What do you think?

include/oneapi/tbb/task_arena.h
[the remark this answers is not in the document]
isaevil: I've implemented proposed alternative approach which utilize `my_version_and_traits` variable instead (I think leave/retention policy falls into the **trait** part of this variable pretty naturally). Not sure about the mask though right now it is all remaining more significant bits, but I think we should reserve only some fixed part of it.

include/oneapi/tbb/task_arena.h
```diff
@@ -404,6 +500,33 @@ class task_arena : public task_arena_base {
...
+        __TBB_ASSERT(my_initialization_state.load(std::memory_order_relaxed) == do_once_state::initialized, nullptr);
+        r1::unregister_parallel_phase(this, with_fast_leave);
+    }
+
+    class scoped_parallel_phase {
+        task_arena& arena;
+        bool one_time_fast_leave;
+    public:
+        scoped_parallel_phase(task_arena& ta, bool with_fast_leave = true) 
+            : arena(ta), one_time_fast_leave(with_fast_leave)
```
dnmokhov, line 520: Shouldn't `with_fast_leave` default to `false` here?

aleksei-fedotov (commented):
A bunch of comments from me. Have not looked at the tests yet. Also, not finished reviewing the PHASE_* switching logic because of found issue.

src/tbb/arena.h
```diff
@@ -179,6 +179,75 @@ class atomic_flag {
...
 
+#if __TBB_PREVIEW_PARALLEL_PHASE
+class thread_leave_manager {
+    static const std::uint64_t FAST_LEAVE = 1;
+    static const std::uint64_t ONE_TIME_FAST_LEAVE = 1 << 1;
+    static const std::uint64_t DELAYED_LEAVE = 1 << 2;
+    static const std::uint64_t PARALLEL_PHASE = 1 << 3;
+    // Use 29 bits for the parallel block state + reference counter,
+    // reserve 32 most significant bits.
+    static const std::uint64_t PARALLEL_PHASE_MASK = ((1LLU << 32) - 1) & (PARALLEL_PHASE - 1);
```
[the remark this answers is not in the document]
isaevil: > Or these 32 bits are to indicate the reservation more explicitly?

Yes.

include/oneapi/tbb/task_arena.h
```diff
@@ -162,13 +182,36 @@ class task_arena_base {
...
+        my_version_and_traits = (my_version_and_traits & ~leave_policy_trait_mask) | leave_policy_to_traits(lp);
+    }
+#endif
+
     enum {
-        default_flags = 0
-        , core_type_support_flag = 1
+        default_flags               = 0,
+        core_type_support_flag      = 1,
     };
```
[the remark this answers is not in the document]
isaevil: Thanks, the closing comma is redundant here.

include/oneapi/tbb/task_arena.h
```diff
@@ -113,6 +120,11 @@ namespace d1 {
 static constexpr unsigned num_priority_levels = 3;
 static constexpr int priority_stride = INT_MAX / (num_priority_levels + 1);
 
+#if __TBB_PREVIEW_PARALLEL_PHASE
+static constexpr int leave_policy_trait_offset = 1;
+static constexpr int leave_policy_trait_mask = ~((1 << leave_policy_trait_offset) - 1);
```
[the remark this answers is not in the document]
isaevil: What is the problem here? First bit is reserved for `core_type_support_flag` thus we take (for now) remaining more significant bits for leave policies.

src/tbb/arena.h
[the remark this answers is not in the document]
isaevil: Sorry, I think I forgot to commit correct final mask:
`((1LLU << 32) - 1) & ~(PARALLEL_PHASE - 1);`
Thus `ONE_TIME_FAST_LEAVE` won't be present in the mask (referring to your other comment)

include/oneapi/tbb/task_arena.h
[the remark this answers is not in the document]
akukanov: I think you should just add `leave_policy_support_flag` next to `core_type_support_flag` with the value of 2 (or 1 << 1, if you like it more).

include/oneapi/tbb/task_arena.h
```diff
@@ -404,6 +498,33 @@ class task_arena : public task_arena_base {
         return execute_impl<decltype(f())>(f);
     }
 
+#if __TBB_PREVIEW_PARALLEL_PHASE
+    void start_parallel_phase() {
+        initialize();
+        r1::register_parallel_phase(this, /*reserved*/0);
+        // Trigger worker threads to join arena
+        enqueue([]{});
```
akukanov, line 506: Why cannot `register_parallel_phase` trigger thread demand internally?

include/oneapi/tbb/task_arena.h
```diff
@@ -404,6 +498,33 @@ class task_arena : public task_arena_base {
...
+    }
+    void end_parallel_phase(bool with_fast_leave = false) {
+        __TBB_ASSERT(my_initialization_state.load(std::memory_order_relaxed) == do_once_state::initialized, nullptr);
+        r1::unregister_parallel_phase(this, with_fast_leave);
```
akukanov, line 510: I would make the conversion to uintptr_t explicit (even though the implicit one works) and even would add a comment that the passed value is either 0 or 1 (it's guaranteed by the standard, but does everybody remember it?)

include/oneapi/tbb/task_arena.h
isaevil: Moved demand trigger into `register_parallel_phase`

include/oneapi/tbb/task_arena.h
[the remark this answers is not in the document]
isaevil: My idea here was to map values from enumeration class `leave_policy` directly into bits in the `my_version_and_traits` field and vice versa. User will pass values from enumeration class anyway to set default leave policy for the arena, so in order to not duplicate any non-automatic policies as flags in unnamed enum such way is used. If any new policies are going to be added- just single enum for `leave_policy` will need to be updated in `task_arena`.

@akukanov Do I understand it correctly that you are suggesting to add `fast_leave_policy_flag` and just check if it is set or not instead?

test/tbb/test_task_arena.cpp
aleksei-fedotov, line 1: In addition, I think need to cover in tests the following:
- The DELAYED_LEAVE is returned back once new task is submitted.
  - Here I think we can test that the return of a worker after one-time-fast-leave happens generally longer than the following returns made after new task(s) is/are submitted. 
- Nested parallel phases with combinations of DELAYED and (ONE TIME) FAST leaves.

src/tbb/waiters.h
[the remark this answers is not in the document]
aleksei-fedotov: Looks much better now. Thanks!

src/tbb/arena.cpp
```diff
@@ -875,6 +921,20 @@ int task_arena_impl::max_concurrency(const d1::task_arena_base *ta) {
     return int(governor::default_num_threads());
 }
 
+#if __TBB_PREVIEW_PARALLEL_PHASE
+void task_arena_impl::register_parallel_phase(d1::task_arena_base* ta) {
+    arena* a = ta ? ta->my_arena.load(std::memory_order_relaxed) : governor::get_thread_data()->my_arena;
```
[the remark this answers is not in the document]
isaevil: This is done to support starting and ending parallel phase using `tbb::this_task_arena` interfaces. You can call `tbb::this_task_arena::start_parallel_phase` which will invoke the same entry point with `nullptr` parameter, thus `arena` object will be taken from TLS.

src/tbb/arena.h
```diff
@@ -179,6 +179,86 @@ class atomic_flag {
...
+                         "Was the new value introduced for leave policy?");
+            my_state.store(FAST_LEAVE, std::memory_order_relaxed);
+        }
+    }
+
+    void restore_state_if_needed() {
+        std::uint64_t curr = ONE_TIME_FAST_LEAVE;
+        if (my_state.load(std::memory_order_relaxed) == curr) {
+            // Potentially can override decision of the parallel block from future epoch
+            // but it is not a problem because it does not violate the correctness
```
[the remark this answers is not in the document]
isaevil: Essentially, yes.

test/tbb/test_task_arena.cpp
```diff
@@ -41,6 +42,7 @@
 #include <stdexcept>
 #include <thread>
 #include <vector>
+#include <numeric>
```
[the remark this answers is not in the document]
isaevil: I was using `std::accumulate` in previous versions of test's code, but then replaced it. It is redundant now, yes.

src/tbb/arena.h
```diff
@@ -179,6 +179,86 @@ class atomic_flag {
...
+                // The parallel phase is already started, thus simply add a reference to it
+                desired = PARALLEL_PHASE + prev;
+            } else if (prev == ONE_TIME_FAST_LEAVE) {
+                // State was previously transitioned to "One-time Fast leave", thus
+                // with the start of new parallel phase, it should be then transitioned to "Delayed leave"
+                desired = PARALLEL_PHASE | DELAYED_LEAVE;
+            } else {
+                // Transition to "start of parallel phase" state and preserving the default state
+                desired = PARALLEL_PHASE | prev;
+            }
```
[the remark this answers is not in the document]
isaevil: Looks like this approach eliminates the need for `PARALLEL_PHASE_MASK`. Do we still want to use it though for something else?

test/tbb/test_task_arena.cpp
```diff
@@ -2068,3 +2070,251 @@ TEST_CASE("worker threads occupy slots in correct range") {
...
+        longest_start_times.push_back(get_longest_start(start_time));
+    };
+
+    for (std::size_t i = 0; i < num_runs; ++i) {
+        if (ta) {
+            ta->execute(work);
+        } else {
+            work();
+        }
+        utils::doDummyWork(i*500);
```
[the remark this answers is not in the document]
isaevil: I didn't found anything similar to what you are referring to. We did use sleep first, but on Win32 API `Sleep` function accepts only milliseconds.

src/tbb/arena.h
```diff
@@ -179,6 +179,86 @@ class atomic_flag {
...
+    }
+    
+    // Indicate start of parallel phase in the state machine
+    void register_parallel_phase() {
+        __TBB_ASSERT(my_state.load(std::memory_order_relaxed) != 0, "The initial state was not set");
+
+        std::uint64_t prev = my_state.load(std::memory_order_relaxed);
+        std::uint64_t desired{};
+        do {
+            if (prev & PARALLEL_PHASE_MASK) {
```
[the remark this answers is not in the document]
isaevil: I think the code after simplification you've suggested is already pretty much self-explanatory so I think there is no need in such helpers.

src/tbb/arena.h
```diff
@@ -179,6 +179,86 @@ class atomic_flag {
...
+        do {
+            if (((prev - PARALLEL_PHASE) & PARALLEL_PHASE_MASK) != 0) {
+                // There are still other parallel phases, thus just decrease the reference counter
+                desired = prev - PARALLEL_PHASE;
+            } else {
+                // We are the last parallel phase, thus transition to the default state
+                // or to the "One-time Fast leave" state if it was requested
+                desired = enable_fast_leave && (prev - PARALLEL_PHASE == DELAYED_LEAVE) ? ONE_TIME_FAST_LEAVE : prev - PARALLEL_PHASE;
+            }
+        } while (!my_state.compare_exchange_strong(prev, desired));
```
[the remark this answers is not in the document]
isaevil: Applied.

include/oneapi/tbb/task_arena.h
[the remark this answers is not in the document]
akukanov: > @akukanov Do I understand it correctly that you are suggesting to add fast_leave_policy_flag and just check if it is set or not instead?
@isaevil Yes, I believe it is the same functionally but easier to use than keeping the offset and the mask for a position in the traits. And also it would be consistent with `core_type_support_flag`

src/tbb/arena.cpp
```diff
@@ -500,6 +530,8 @@ struct task_arena_impl {
     static int max_concurrency(const d1::task_arena_base*);
     static void enqueue(d1::task&, d1::task_group_context*, d1::task_arena_base*);
     static d1::slot_id execution_slot(const d1::task_arena_base&);
+    static void register_parallel_phase(d1::task_arena_base*);
+    static void unregister_parallel_phase(d1::task_arena_base*, bool);
```
akukanov, line 534: It would be better for this internal methods to match the ABI functions, i.e. use `uintptr_t` parameters. This way you will limit the modifications related to these parameters strictly within the methods.

aleksei-fedotov (approved):
I think the PR looks good enough as a preview functionality. Consider my other suggestions to ponder on and insignificant remarks. Perhaps, we would be able to "close on" these before fully supporting this feature.

I approve it. Though you might also want implementing Alexey's comment about single `fast_leave_policy_flag` as well because suggested in-the-patch approach can be implemented in future if we would ever have more than two leave policies.

include/oneapi/tbb/task_arena.h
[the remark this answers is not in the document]
isaevil: @akukanov Added `fast_leave_policy_flag` instead of previous implementation with `leave_policy_trait_mask`.

test/tbb/test_task_arena.cpp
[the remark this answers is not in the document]
isaevil: From what I've seen, `sleep_for` translates into `Sleep` on Windows, which accepts only milliseconds hence the minimum amount of sleep is exact current blocking time in oneTBB. The test tries to include measurements with timings smaller than a millisecond so arena with delayed leave (or parallel phase) would in general start new work faster than with fast `leave_policy`.

src/tbb/arena.cpp
isaevil: Done.

src/tbb/waiters.h
```diff
@@ -70,7 +71,7 @@ class outermost_worker_waiter : public waiter_base {
                         return true;
                     }
 
-                    if (my_arena.my_threading_control->is_any_other_client_active()) {
+                    if (!my_arena.my_thread_leave.is_retention_allowed() || my_arena.my_threading_control->is_any_other_client_active()) {
```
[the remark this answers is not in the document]
isaevil: Done.

src/tbb/arena.h
```diff
@@ -281,13 +358,27 @@ class arena: public padded<arena_base>
     };
 
     //! Constructor
-    arena(threading_control* control, unsigned max_num_workers, unsigned num_reserved_slots, unsigned priority_level);
+    arena(threading_control* control, unsigned max_num_workers, unsigned num_reserved_slots, unsigned priority_level
+#if __TBB_PREVIEW_PARALLEL_PHASE
+          , tbb::task_arena::leave_policy wl
```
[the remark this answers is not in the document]
isaevil: Finally replaced all :)

src/tbb/waiters.h
```diff
@@ -54,13 +54,14 @@ class outermost_worker_waiter : public waiter_base {
...
+    bool continue_execution(arena_slot& slot, d1::task*& t) {
         __TBB_ASSERT(t == nullptr, nullptr);
 
         if (is_worker_should_leave(slot)) {
-            if (!governor::hybrid_cpu()) {
+            if (is_delayed_leave_enabled()) {
                 static constexpr std::chrono::microseconds worker_wait_leave_duration(1000);
-                static_assert(worker_wait_leave_duration > std::chrono::steady_clock::duration(1), "Clock resolution is not enough for measured interval.");
+                static_assert(worker_wait_leave_duration > std::chrono::steady_clock::duration(1),
+                        "Clock resolution is not enough for measured interval.");
```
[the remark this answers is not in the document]
isaevil: Done.

src/tbb/arena.h
```diff
@@ -179,6 +179,78 @@ class atomic_flag {
...
+            if (enable_fast_leave && /*it was the last parallel phase*/desired == DELAYED_LEAVE) {
+                desired = ONE_TIME_FAST_LEAVE;
+            }
+        } while (!my_state.compare_exchange_strong(prev, desired));
+    }
+
+    bool is_retention_allowed() {
+        __TBB_ASSERT(my_state.load(std::memory_order_relaxed) != 0, "The initial state was not set");
+
+        std::uint64_t curr = my_state.load(std::memory_order_relaxed);
```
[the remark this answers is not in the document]
isaevil: Done.

src/tbb/arena.h
```diff
@@ -179,6 +179,78 @@ class atomic_flag {
...
+    }
+
+    // Indicate the end of parallel phase in the state machine
+    void unregister_parallel_phase(bool enable_fast_leave) {
+        __TBB_ASSERT(my_state.load(std::memory_order_relaxed) != 0, "The initial state was not set");
+
+        std::uint64_t prev = my_state.load(std::memory_order_relaxed);
+        std::uint64_t desired{};
+        do {
+            desired = prev - PARALLEL_PHASE; // Mark the end of this phase in reference counter
```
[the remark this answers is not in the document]
isaevil: Good suggestion, done.

test/tbb/test_parallel_phase.cpp
```diff
@@ -0,0 +1,273 @@
...
+
+    for (std::size_t i = 0; i < num_runs; ++i) {
+        if (ta) {
+            ta->execute(work);
+        } else {
+            work();
+        }
+        utils::doDummyWork(i*500);
+    }
+    return utils::median(longest_start_times.begin(), longest_start_times.end());
```
dnmokhov, line 54: Doesn't "**_i_** * 500" mean that earlier runs will take less to restart than later runs? Would it make more sense to keep the amount of dummy work the same across each run?

If this increasing non-TBB work amount is intentional, wouldn't it eventually become larger than the current block time of 1ms?

isaevil: @dnmokhov Changed the constant.

dnmokhov (dismissed):
LGTM

src/tbb/arena.h
```diff
@@ -179,6 +179,81 @@ class atomic_flag {
...
 };
 
+#if __TBB_PREVIEW_PARALLEL_PHASE
+class thread_leave_manager {
+    static const std::uint64_t FAST_LEAVE = 1;
+    static const std::uint64_t ONE_TIME_FAST_LEAVE = 1 << 1;
+    static const std::uint64_t DELAYED_LEAVE = 1 << 2;
+    static const std::uint64_t PARALLEL_PHASE = 1 << 3;
+
+    std::atomic<std::uint64_t> my_state{0};
```
akukanov, line 189: Not sure if 64-bit atomic is really needed. Even if it's only 32 bit, 2^29 seems quite enough for the amount of simultaneous parallel phases. The benefit would be better support for 32 bit platforms, but maybe it's negligible.

src/tbb/arena.h
```diff
@@ -179,6 +179,81 @@ class atomic_flag {
...
+public:
+    void set_initial_state(tbb::task_arena::leave_policy lp) {
```
akukanov, line 191: Since this is essentially the constructor, please add a comment that this method is required to be called soon after construction, and is not thread-safe. Or maybe consider converting it into a real constructor.

src/tbb/arena.h
```diff
@@ -179,6 +179,81 @@ class atomic_flag {
...
+            std::uint64_t platform_policy = governor::hybrid_cpu() ? FAST_LEAVE : DELAYED_LEAVE;
+            my_state.store(platform_policy, std::memory_order_relaxed);
+        } else {
+            __TBB_ASSERT(lp == tbb::task_arena::leave_policy::fast,
+                         "Was the new value introduced for leave policy?");
+            my_state.store(FAST_LEAVE, std::memory_order_relaxed);
+        }
+    }
+
+    void restore_default_policy_if_needed() {
```
akukanov, line 202: I would just call it `reset` (or `reset_if_needed`) for better encapsulation.

akukanov (commented):
My comments below are not critical; feel free to commit as-is and address later.

test/tbb/test_parallel_phase.cpp
```diff
@@ -0,0 +1,288 @@
...
+    start_time_collection st_collector1{ta_automatic_leave, /*num_trials=*/10};
+    start_time_collection st_collector2{ta_fast_leave, /*num_trials=*/10};
+
+    auto times_automatic = st_collector1.measure();
+    auto times_fast = st_collector2.measure();
+
+    auto median_automatic = utils::median(times_automatic.begin(), times_automatic.end());
+    auto median_fast = utils::median(times_fast.begin(), times_fast.end());
+
+    WARN_MESSAGE(median_automatic < median_fast,
```
vossmjp, line 208: Are all of the tests guaranteed to work (i.e. not have warnings) for hybrid systems where automatic is not delayed leave?

include/oneapi/tbb/task_arena.h
```diff
@@ -95,6 +95,11 @@ TBB_EXPORT void __TBB_EXPORTED_FUNC isolate_within_arena(d1::delegate_base& d, s
 TBB_EXPORT void __TBB_EXPORTED_FUNC enqueue(d1::task&, d1::task_arena_base*);
 TBB_EXPORT void __TBB_EXPORTED_FUNC enqueue(d1::task&, d1::task_group_context&, d1::task_arena_base*);
 TBB_EXPORT void __TBB_EXPORTED_FUNC submit(d1::task&, d1::task_group_context&, arena*, std::uintptr_t);
+
+#if __TBB_PREVIEW_PARALLEL_PHASE
+TBB_EXPORT void __TBB_EXPORTED_FUNC register_parallel_phase(d1::task_arena_base*, std::uintptr_t);
```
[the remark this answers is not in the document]
isaevil: I think the main idea was to use more generic names, which do not depend on feature API. So, if API will change dramatically, entry point names would still make some sense.

test/tbb/test_parallel_phase.cpp
isaevil: That is a good question. I suppose that only one test case(Parallel Phase retains workers in task_arena) would produce expected results one hybrid systems. Other test don't make much sense on such configurations, but I think it is fine since these are warnings, not asserts.

include/oneapi/tbb/task_arena.h
```diff
@@ -162,13 +175,37 @@ class task_arena_base {
         return (my_version_and_traits & core_type_support_flag) == core_type_support_flag ? my_max_threads_per_core : automatic;
     }
 
+#if __TBB_PREVIEW_PARALLEL_PHASE
+    leave_policy get_leave_policy() const {
+        bool fast_policy_set = (my_version_and_traits & fast_leave_policy_flag) == fast_leave_policy_flag;
+        return fast_policy_set ? leave_policy::fast : leave_policy::automatic;
+    }
+
+    int leave_policy_to_traits(leave_policy lp) const {
```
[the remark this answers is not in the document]

include/oneapi/tbb/task_arena.h
```diff
@@ -259,31 +304,58 @@ class task_arena : public task_arena_base {
      *       Value of 1 is default and reflects behavior of implicit arenas.
      **/
     task_arena(int max_concurrency_ = automatic, unsigned reserved_for_masters = 1,
-               priority a_priority = priority::normal)
-        : task_arena_base(max_concurrency_, reserved_for_masters, a_priority)
+               priority a_priority = priority::normal
+#if __TBB_PREVIEW_PARALLEL_PHASE
+                    , leave_policy lp = leave_policy::automatic
+#endif
```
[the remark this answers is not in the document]
isaevil: Fixed.

src/tbb/arena.h
isaevil: Done.

src/tbb/arena.h
isaevil: Changed it to uintptr_t.

isaevil:
@aleksei-fedotov @akukanov @vossmjp @pavelkumbrasev @dnmokhov I have updated the RFC document: included some technical details and conditions to leave the experimental stage.

src/tbb/arena.h
isaevil: Added a comment.

include/oneapi/tbb/task_arena.h
```diff
@@ -491,6 +617,11 @@ using detail::d1::max_concurrency;
 using detail::d1::isolate;
 
 using detail::d1::enqueue;
+
+#if __TBB_PREVIEW_PARALLEL_PHASE
+using detail::d1::start_parallel_phase;
+using detail::d1::end_parallel_phase;
```
kboyarinov, line 623: Should not we introduce `scoped_parallel_phase` for `this_task_arena` as well?

src/tbb/arena.h
```diff
@@ -433,6 +528,9 @@ void arena::advertise_new_work() {
             workers_delta = 1;
         }
 
+#if __TBB_PREVIEW_PARALLEL_PHASE
+        my_thread_leave.reset_if_needed();
```
aleksei-fedotov, line 532: Will the following assert be useful?
```suggestion
        __TBB_ASSERT(mandatory_delta > 0 || workers_delta > 0, "Potentially delaying fast leave?");
        my_thread_leave.reset_if_needed();
```

include/oneapi/tbb/task_arena.h
kboyarinov: Oh, I see that it is already covered by the open questions in the RFC and I guess it should not be resolved in the initial implementation.
Feel free to resolve this conversation.

src/tbb/arena.h
isaevil: It is not a valid assert. In case of workerless arena and spawn of the task both of those values will be zeroes.

src/tbb/arena.h
```diff
@@ -1,5 +1,5 @@
-/*
-    Copyright (c) 2005-2024 Intel Corporation
+/*arena
```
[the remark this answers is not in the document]
isaevil: Must've been added by mistake while updating copyrights. Fixed.

aleksei-fedotov (approved):
Looks good to me.

src/tbb/arena.h
aleksei-fedotov: Okay... let it be then:
```suggestion
        __TBB_ASSERT((mandatory_delta > 0 || workers_delta > 0) ||
                     (is_arena_workerless() && work_type == work_spawned),
                     "Potentially delaying fast leave?");
        my_thread_leave.reset_if_needed();
```

include/oneapi/tbb/task_arena.h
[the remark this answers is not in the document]
akukanov: There might be alternative suitable generic verbs, such as "initiate", "enter", "commence", "launch", "open" for the beginning and "complete", "conclude", "close", "finish", "exit".

I will leave it to Mike to decide on the best choice of a verb. From the semantical point of view (i.e. speaking of a program execution phase), I would likely use "enter/exit" or "initiate/complete".

src/tbb/arena.h
akukanov: First of all, I do not readily see how this assert is related to the parallel phase.
And if you move the assertion next to the code in the lines 522-529, it seems it just duplicates almost exactly the conditions and assignments there.

src/tbb/arena.h
```diff
@@ -179,6 +179,82 @@ class atomic_flag {
...
+            if (enable_fast_leave && /*it was the last parallel phase*/desired == DELAYED_LEAVE) {
+                desired = ONE_TIME_FAST_LEAVE;
+            }
+        } while (!my_state.compare_exchange_strong(prev, desired));
+    }
+
+    bool is_retention_allowed() {
+        std::uintptr_t curr = my_state.load(std::memory_order_relaxed);
+        __TBB_ASSERT(curr != std::uintptr_t(-1), "The initial state was not set");
+        return curr != FAST_LEAVE && curr != ONE_TIME_FAST_LEAVE;
```
akukanov, line 253: Should not it be `curr & (FAST_LEAVE | ONE_TIME_FAST_LEAVE) != 0` instead, to account for the parallel phase counter?

src/tbb/arena.h
```diff
@@ -179,6 +179,82 @@ class atomic_flag {
...
+    }
+
+    void reset_if_needed() {
+        std::uintptr_t curr = ONE_TIME_FAST_LEAVE;
+        if (my_state.load(std::memory_order_relaxed) == curr) {
+            // Potentially can override decision of the parallel block from future epoch
+            // but it is not a problem because it does not violate the correctness
+            my_state.compare_exchange_strong(curr, DELAYED_LEAVE);
+        }
+    }
```
akukanov, line 211: In reset, you only need to drop the one-time flag, correct? Since now it does not intersect with other states, it should be enough to just `fetch_and(~ONE_TIME_FAST_LEAVE)`.

src/tbb/arena.h
```diff
@@ -179,6 +179,82 @@ class atomic_flag {
...
+            // previous phase.
+            desired = prev;
+            if (prev == ONE_TIME_FAST_LEAVE) {
+                // State was previously transitioned to "One-time Fast leave", thus with the start
+                // of new parallel phase, it should be transitioned to "Delayed leave"
+                desired = DELAYED_LEAVE;
+            }
+            __TBB_ASSERT(desired + PARALLEL_PHASE > desired, "Overflow detected");
+            desired += PARALLEL_PHASE; // Take into account this start of a parallel phase
+        } while (!my_state.compare_exchange_strong(prev, desired));
```
akukanov, line 231: I think it can be simplified as `fetch_add(PARALLEL_PHASE)` followed by `fetch_and(~ONE_TIME_FAST_LEAVE)` in case the bit is set in the returned value of `fetch_add`.

src/tbb/arena.h
```diff
@@ -179,6 +179,82 @@ class atomic_flag {
...
+
+        std::uintptr_t desired{};
+        do {
+            __TBB_ASSERT(prev - PARALLEL_PHASE < prev,
+                         "A call to unregister without its register complement");
+            desired = prev - PARALLEL_PHASE; // Mark the end of this phase in reference counter
+            if (enable_fast_leave && /*it was the last parallel phase*/desired == DELAYED_LEAVE) {
+                desired = ONE_TIME_FAST_LEAVE;
+            }
+        } while (!my_state.compare_exchange_strong(prev, desired));
```
akukanov, line 247: I think this can be simplified as `fetch_or(ONE_TIME_FAST_LEAVE)` if needed, and then `fetch_add(-PARALLEL_PHASE)`.

akukanov (commented):
> I think that might also simplify the implementation logic.

Below is how. Note that separation of significant bits for different leave states was a necessary prerequisite to operate with ONE_TIME_FAST_LEAVE as a single bit flag.

src/tbb/arena.h
isaevil: The idea of current implementation is that when parallel phase is active then threads are retained regardless of arena default leave policy (as it was said in RFC, scheduler can implement different retention policy). This allows to support this feature on hybrid platforms as well where worker threads are not retained by default, but if the phase is active, they are.

That is why the check is like that. It is used to verify if parallel phase isn't active anymore, meaning threads can fall back to default leave policy (fast or block time) or leave once fast if such option is chosen.

src/tbb/arena.h
akukanov: No, my suggestion above is incorrect, as the set of the flag should be only done if that was the last active parallel phase. Unlike with register, it looks like here the change of both the flag and the counter need to be transactional, i.e. a single atomic operation.

src/tbb/arena.h
aleksei-fedotov: I think the desire is to drop the one-time flag only if a new work is submitted in the empty arena. So, the flag cannot be blindly dropped.

src/tbb/arena.h
```diff
@@ -245,6 +325,11 @@ struct arena_base : padded<intrusive_list_node> {
     //! Waiting object for external threads that cannot join the arena.
     concurrent_monitor my_exit_monitors;
 
+#if __TBB_PREVIEW_PARALLEL_PHASE
+    //! Manages state of thread_leave state machine
+    thread_leave_manager my_thread_leave;
```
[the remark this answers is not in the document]
akukanov: > Multiple "ends" come first and only after that "starts" begin to appear;

According to the assertion at lines 241-242. this causes undefined behavior. Seems reasonable to me, as I have no better idea for handling "ends" coming before "starts".

> Number of "starts" does not equal to the number of "ends"

Missed "starts" would be also UB, as conceptually it's the same as "ends" before "starts".

For missed "ends", I do not see how we would catch those earlier than on arena destruction.

src/tbb/arena.h
isaevil: Well, at first glance it seems like @akukanov suggestion shouldn't contradict the desire. Since `unregister_parallel_phase` sets this bit only on the last parallel phase completion, when new work submission happens the state will be either still `ONE_TIME_FAST_LEAVE`, or already include some amount of parallel phase references, which means that this bit will be already dropped by some of the first parallel phases.

src/tbb/arena.h
aleksei-fedotov: Yeah, I know that it duplicates. But the stress here is on the words "almost exactly". With the assert I sort of wanted to make sure we do not fall into recently discovered issue of dropping the fast leave setting when threads are recalled from arena. I don't mind not having an assert here, since, as you said, the conditions duplicate it. However, I do think we need to tell reader about the desired behavior here, it feels like the code itself might not be good at telling that.

Another interesting consideration is to reduce cross-NUMA accesses as much as possible. If thread can base its decision whether it needs to read and/or modify global-to-arena variables on the data it has already read, then it makes sense to avoid doing more re-reads of other global-to-arena data. Of course, if we are sure that the set of "almost exactly" is empty here, then the thread already avoids unnecessary touches to the data that increases cross-NUMA traffic.

src/tbb/arena.h
akukanov: Right, if the arena is already not empty, then the flag should have been dropped already, right? And then `fetch_and` would just do nothing. We can still leave the load and the check if it gives a proven performance advantage, but certainly `compare_exchange_strong` is not needed.

src/tbb/arena.h
akukanov: Assertions are good to enforce preconditions for a function, and in some cases of sophisticated internal logic also good to check expected invariants in certain point. Having an assertion that checks the conditions explicitly set by the code right above is a pure duplication. The part that is not an exact duplication could be checked in a proper place, of course; also the `reset_if_needed` could internally assert on its expected preconditions.

For the NUMA part of the comment, I cannot make the connection to the assertion being discussed.

include/oneapi/tbb/task_arena.h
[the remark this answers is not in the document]
vossmjp: Of the suggestions, I like "enter/exit".

test/tbb/test_parallel_phase.cpp
vossmjp: Can we skip the warning on hybrid systems?  For example, don't output warning if there is more than 1 core type.

include/oneapi/tbb/task_arena.h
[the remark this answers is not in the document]
isaevil: @akukanov @vossmjp Renamed entry points to "enter/exit".

src/tbb/arena.h
```diff
@@ -179,6 +179,82 @@ class atomic_flag {
...
 };
 
+#if __TBB_PREVIEW_PARALLEL_PHASE
+class thread_leave_manager {
+    static const std::uintptr_t DELAYED_LEAVE       = 0;
+    static const std::uintptr_t FAST_LEAVE          = 1;
+    static const std::uintptr_t ONE_TIME_FAST_LEAVE = 1 << 1;
+    static const std::uintptr_t PARALLEL_PHASE      = 1 << 2;
+
+    std::atomic<std::uintptr_t> my_state{static_cast<std::uintptr_t>(-1)};
```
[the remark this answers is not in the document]
isaevil: Done.

src/tbb/arena.h
isaevil: Replaced CAS by `fetch_and`.

src/tbb/arena.h
isaevil: Done.

src/tbb/arena.h
```diff
@@ -179,6 +179,74 @@ class atomic_flag {
...
+    void unregister_parallel_phase(bool enable_fast_leave) {
+        std::uintptr_t prev = my_state.load(std::memory_order_relaxed);
+        __TBB_ASSERT(prev != UINTPTR_MAX, "The initial state was not set");
+
+        std::uintptr_t desired{};
+        do {
+            __TBB_ASSERT(prev - PARALLEL_PHASE < prev,
+                         "A call to unregister without its register complement");
+            desired = prev - PARALLEL_PHASE; // Mark the end of this phase in reference counter
+            if (enable_fast_leave && /*it was the last parallel phase*/desired == DELAYED_LEAVE) {
```
akukanov, line 236: A bit of nitpicking: the comment does not really match the condition tested. To check if that's the last parallel phase, you'd use `desired < PARALLEL_PHASE`. Here you also check that neither of the fast-leave flags is set. Perhaps can be improved like this:
```suggestion
            if (enable_fast_leave && /*it was the last parallel phase, and delayed leave is the default*/ desired == DELAYED_LEAVE) {
```

vossmjp (approved):
Looks good to me. At some point we need to improve tests though.

PR #1575, opened 2024-12-04
Fix CPU binding on HCC Windows systems

isaevil:
### Description 
Current implementation of thread pinning with HWLOC usage on Windows is not scalable and already does not work on HCC systems like SPR with multiple NUMA nodes. It doesn't work because `hwloc_set_cpubind` uses [SetThreadGroupAffinity](https://learn.microsoft.com/en-us/windows/win32/api/processtopologyapi/nf-processtopologyapi-setthreadgroupaffinity) on Windows and this API doesn't allow setting affinity across multiple processor groups.

This patch fixes binding by shrinking the mask to fit a particular processor group based on slot number in arena.

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/src/tbbbind/tbb_bind.cpp
+++ b/src/tbbbind/tbb_bind.cpp
@@ -37,6 +37,10 @@
 
 #define __TBBBIND_HWLOC_HYBRID_CPUS_INTERFACES_PRESENT (HWLOC_API_VERSION >= 0x20400)
 #define __TBBBIND_HWLOC_TOPOLOGY_FLAG_RESTRICT_TO_CPUBINDING_PRESENT (HWLOC_API_VERSION >= 0x20500)
+#define __TBBBIND_HWLOC_WINDOWS_API_AVAILABLE (_WIN32 && HWLOC_API_VERSION >= 0x20500)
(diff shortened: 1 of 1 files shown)


src/tbbbind/tbb_bind.cpp
```diff
@@ -369,6 +404,32 @@ class system_topology {
...
+        // Map slot number to a number within constraints mask if
+        // max concurrency is greater than weight of the mask.
+        slot_num %= constraints_mask_weight;
+        std::size_t total_weight = 0;
+        for (auto& processor_group : processor_groups_affinity_masks_list) {
+            if (hwloc_bitmap_intersects(constraints_mask, processor_group)) {
+                hwloc_bitmap_and(result_mask, processor_group, constraints_mask);
+                total_weight += hwloc_bitmap_weight(result_mask);
+                if (slot_num < total_weight) {
+                    return;     // Correponding processor group where to bind the thread is found
```
aleksei-fedotov, line 427:
```suggestion
                    return;     // Corresponding processor group where to bind the thread is found
```

aleksei-fedotov (approved):
LGTM!

Issue #843, opened 2022-06-10
build failure with gcc13: -Werror=stringop-overflow


nanoric:
GCC Version:
gcc (GCC) 13.0.0 20220529 (experimental)

steps to reproduce:
```
# build gcc with git@master
cd / && git clone --depth 1 --single-branch --branch master https://github.com/gcc-mirror/gcc.git
GMP_VERSION="6.2.1" MPFR_VERSION="4.1.0" MPC_VERSION="1.2.1" \
  && cd /tmp && wget https://ftp.gnu.org/gnu/gmp/gmp-${GMP_VERSION}.tar.xz\
  && cd /tmp && wget https://ftp.gnu.org/gnu/mpfr/mpfr-${MPFR_VERSION}.zip \
  && cd /tmp && wget https://ftp.gnu.org/gnu/mpc/mpc-$MPC_VERSION.tar.gz \
  && cd /tmp && tar -xf gmp-$GMP_VERSION.tar.xz && cp -r gmp-$GMP_VERSION/ /gcc/gmp \
  && cd /tmp && unzip mpfr-$MPFR_VERSION.zip && cp -r mpfr-$MPFR_VERSION/ /gcc/mpfr \
  && cd /tmp && tar -xf mpc-$MPC_VERSION.tar.gz && cp -r mpc-$MPC_VERSION/ /gcc/mpc \
  && mkdir /gcc12 && rm -rf /tmp/*

cd /gcc \
  &&  CXX_FLAGS="-O3 -march=native" ./configure --disable-multilib --prefix=/gcc12 --with-arch=native --disable-werror

cd /gcc \
  && CXX_FLAGS="-O3 -march=native" make -j16 \
  && make install \
  && rm -rf /gcc

export CC=/gcc12/bin/gcc
export LD_LIBRARY_PATH=/gcc12/lib64:$LD_LIBRARY_PATH
export PATH=/gcc12/bin:$PATH
export PKG_CONFIG_PATH=/gcc12/lib64/pkgconfig:$PKG_CONFIG_PATH

# build tbb with that gcc:
cd /tmp && wget https://github.com/oneapi-src/oneTBB/archive/refs/heads/master.zip && unzip master.zip \
 && cd oneTBB-master && cmake -B build . -DCMAKE_BUILD_TYPE=Release \
 && CFLAGS="-O3 -march=native" CXXFLAGS="-O3 -march-native" make -j16 && make install
```

error log :
```
[...]

PR #1583, opened 2024-12-09
Avoid `-Wstringop-overflow` in GCC 14

ashvardanian:
### Description 

Relates to the bug #843 - reappearing with GCC 14.

### Type of change

- [x] bug fix
- [ ] new feature
- [ ] tests
- [ ] infrastructure
- [ ] documentation

### Tests

- [ ] added
- [x] not needed

### Documentation

- [ ] updated
- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

@kboyarinov, @dnmokhov


phprus:
Please update this condition also in cmake config for gcc:

https://github.com/uxlfoundation/oneTBB/blob/bd689b5a7e8829786a47890cb54a117e3b45970a/cmake/compilers/GNU.cmake#L86-L87

ashvardanian:
Thanks for a fast response, @phprus! I've updated the condition. Let me know if something else is needed 🤗

dnmokhov:
Thanks for the patch, @ashvardanian! Could you please also update the copyright year in the header file?

ashvardanian:
@dnmokhov, done ;)

kboyarinov (approved):
LGTM

PR #1585, opened 2024-12-17
Fix deprecation warning: Compatibility with CMake < 3.10 will be removed from a future version of CMake

phprus:
### Description 
Fix CMake deprecation warning:

```
CMake Deprecation Warning at CMakeLists.txt:15 (cmake_minimum_required):
  Compatibility with CMake < 3.10 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

```

https://cmake.org/cmake/help/v3.31/command/cmake_minimum_required.html

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-# Copyright (c) 2020-2024 Intel Corporation
+# Copyright (c) 2020-2025 Intel Corporation
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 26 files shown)


CMakeLists.txt
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-cmake_minimum_required(VERSION 3.5)
+cmake_minimum_required(VERSION 3.5.0...3.31.3)
```
isaevil, line 15: From CMake docs:

> If the running version of CMake is older than 3.12, the extra ... dots will be seen as version component separators, resulting in the ...<max> part being ignored and preserving the pre-3.12 behavior of basing policies on <min>.

Shouldn't we add some comment (at least to the root CMake file) to reflect that?

isaevil (commented):
Looks good to me. One small comment though.

CMakeLists.txt
phprus: I added a comment with text from the documentation.

isaevil (approved):

PR #1587, opened 2024-12-18
Fallback to GetThreadGroupAffinity if GetNativeSystemInfo reports wrongly

isaevil:
### Description 
For some reason, `GetNativeSystemInfo` on Windows 11 (or WS2022) systems with multiple processor groups reports information from parent process (for example, when running tests using CTest) even though [documentation](https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/ns-sysinfoapi-system_info) states that `dwNumberOfProcessors` is "the number of logical processors in _the current group_". This patch fallbacks to using `GetActiveProcessorCount` on group from calling thread's affinity if possible. 

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/src/tbb/misc_ex.cpp
+++ b/src/tbb/misc_ex.cpp
@@ -297,11 +297,21 @@ static void initialize_hardware_concurrency_info () {
         if ( pam & m )
             ++nproc;
     }
-    __TBB_ASSERT( nproc <= (int)si.dwNumberOfProcessors, nullptr);
+    int number_of_processors = (int)si.dwNumberOfProcessors;
+    if (nproc > number_of_processors && TBB_GetThreadGroupAffinity) {
+        // Sometimes on systems with multiple processor groups GetNativeSystemInfo
+        // reports mask and processor count from the parent process
+        TBB_GROUP_AFFINITY ga;
+        if (TBB_GetThreadGroupAffinity( GetCurrentThread(), &ga)) {
+            number_of_processors = (int)TBB_GetActiveProcessorCount(ga.Group);
+        }
+    }
+
+    __TBB_ASSERT( nproc <= number_of_processors, nullptr);
     // By default setting up a number of processors for one processor group
     theProcessorGroups[0].numProcs = theProcessorGroups[0].numProcsRunningTotal = nproc;
     // Setting up processor groups in case the process does not restrict affinity mask and more than one processor group is present
-    if ( nproc == (int)si.dwNumberOfProcessors && TBB_GetActiveProcessorCount ) {
+    if ( nproc == number_of_processors && TBB_GetActiveProcessorCount ) {
         // The process does not have restricting affinity mask and multiple processor groups are possible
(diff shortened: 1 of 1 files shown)


src/tbb/misc_ex.cpp
```diff
@@ -297,11 +297,21 @@ static void initialize_hardware_concurrency_info () {
...
             ++nproc;
     }
-    __TBB_ASSERT( nproc <= (int)si.dwNumberOfProcessors, nullptr);
+    int number_of_processors = (int)si.dwNumberOfProcessors;
+    if (nproc > number_of_processors && TBB_GetThreadGroupAffinity) {
+        // Sometimes on systems with multiple processor groups GetNativeSystemInfo
+        // reports mask and processor count from the parent process
+        TBB_GROUP_AFFINITY ga;
+        if (TBB_GetThreadGroupAffinity(GetCurrentThread(), &ga)) {
+            number_of_processors = (int)TBB_GetActiveProcessorCount(ga.Group);
```
dnmokhov, line 306: 1. Should we just always be using `GetActiveProcessorCount` if it is more reliable?
2. Just curious why we are checking for `TBB_GetThreadGroupAffinity` here but not checking for `TBB_GetActiveProcessorCount`?

src/tbb/misc_ex.cpp
```diff
@@ -297,11 +297,21 @@ static void initialize_hardware_concurrency_info () {
...
+    }
+
+    __TBB_ASSERT( nproc <= number_of_processors, nullptr);
     // By default setting up a number of processors for one processor group
     theProcessorGroups[0].numProcs = theProcessorGroups[0].numProcsRunningTotal = nproc;
     // Setting up processor groups in case the process does not restrict affinity mask and more than one processor group is present
-    if ( nproc == (int)si.dwNumberOfProcessors && TBB_GetActiveProcessorCount ) {
+    if ( nproc == number_of_processors && TBB_GetActiveProcessorCount ) {
         // The process does not have restricting affinity mask and multiple processor groups are possible
         ProcessorGroupInfo::NumGroups = (int)TBB_GetActiveProcessorGroupCount();
```
dnmokhov, line 316: Similarly, why are we checking for `TBB_GetActiveProcessorCount` here but not checking for `TBB_GetActiveProcessorGroupCount` and `TBB_GetThreadGroupAffinity`?

src/tbb/misc_ex.cpp
isaevil: 1. I am not sure about that. The code seems to be designed in a way that these Group-related features may not be available. So I decided not to replace that functionality, but rather complement it.
2. It is actually enough to check just for one symbol. `dynamic_link` module in TBB works in a way that if some of the symbols in symbol table is not found, than the whole DLL won't be loaded and thus other symbols will be unavailable as well.

dnmokhov (approved):
LGTM

PR #1594, opened 2025-01-07
Allow loading dependencies by name

aleksei-fedotov:
### Description 

This PR allows loading dependencies by module name only as it is proposed in [this RFC](https://github.com/uxlfoundation/oneTBB/blob/77478a1fc3b56da65b11e82194ca2f5fd6e9062b/rfcs/proposed/loading-dependencies/loading-dependencies-by-module-name.org). To strengthen the security and avoid loading malicious code, the signature verification for modules being loaded is also introduced.
However, to avoid breaking compatibility with existing scenarios, the signature verification is disabled by default. To notify about that the following CMake warning appears on the console when `cmake` command is invoked for the first time:
```
CMake Warning at CMakeLists.txt:142 (message):
  Dependency signature verification during dynamic linking in run-time is
  disabled by default.  This may lead to security vulnerabilities.  See
  installation section in oneTBB documentation for more details.
```
, drawing user's attention to this change and allowing to explicitly choose whether user wants unsigned or having incorrect signature modules to be loaded or not.
To avoid loading of such modules, user needs specifying `-DTBB_VERIFY_DEPENDENCY_SIGNATURE=ON` flag in the invocation command of CMake. In this case the warning transforms in a regular status message of the form:
```
-- Dependency signature verification during dynamic linking in run-time is ENABLED
```
If, however, user ignores the warning, the following invocations of `cmake` command show that the signature verification setting remains disabled:
```
-- Dependency signature verification during dynamic linking in run-time is DISABLED
```

The patch also adds optional reporting of dynamic link issues. Examples of the output:
```
oneTBB dynamic link warning: The module "<path-to-dll>" is unsigned or has invalid signature.
oneTBB dynamic link warning: The module "<path-to-dll>" was not found. System error: 126
```
The issues reporting is disabled by default and can be enabled by setting `TBB_DYNAMIC_LINK_WARNING` macro during the build.


CMakeLists.txt
```diff
@@ -225,6 +226,13 @@ foreach(FILE_WITH_EXTRA_TARGETS ${FILES_WITH_EXTRA_TARGETS})
     include(${FILE_WITH_EXTRA_TARGETS})
 endforeach()
 
+if (NOT TBB_VERIFY_DEPENDENCY_SIGNATURE)
+    message(WARNING
+        "Dependency signature verification during dynamic run-time linking is disabled. This may "
+        "lead to security vulnerabilities. It is recommended to leave it enabled in production "
+        "environment. Unset TBB_VERIFY_DEPENDENCY_SIGNATURE or set it ON to enable it again.")
```
egfefey, line 233: Will unsetting TBB_VERIFY_DEPENDENCY_SIGNATURE not disable it instead? Did you mean "Set TBB_VERIFY_DEPENDENCY_SIGNATURE ..."

src/tbb/dynamic_link.cpp
```diff
@@ -110,9 +115,102 @@ namespace r1 {
...
+        std::va_list args;
+        va_start(args, code);
+
+        switch (code) {
+        case dl_lib_not_found:
+            str = va_arg(args, const char*);
+            error = va_arg(args, dlerr_t);
+            // TODO: Use fprintf_s once dynamic link functionality is extracted into a separate
+            // module (e.g., a static library)
+            std::fprintf(stdout, "%s The module \"%s\" was not loaded because it was not found. "
```
egfefey, line 141: Please look up and use the more secured version fprintf_s

src/tbb/dynamic_link.cpp
```diff
@@ -110,9 +115,102 @@ namespace r1 {
...
+                         "System error: " DLERROR_SPECIFIER "\n", prefix, str, error);
+            break;
+        case dl_sym_not_found:     // char const * sym, dlerr_t err:
+            // TODO: Print not found symbol once it is used by the implementation
+            break;
+        case dl_sys_fail:
+            str = va_arg(args, const char*);
+            error = va_arg(args, dlerr_t);
+            std::fprintf(stdout, "oneTBB: A call to \"%s\" failed with error " DLERROR_SPECIFIER
```
egfefey, line 150: Same here

egfefey: and all other locations of fprintf

CMakeLists.txt
aleksei-fedotov: I clarified the messaging.

src/tbb/dynamic_link.cpp
aleksei-fedotov: Done.

src/tbb/dynamic_link.cpp
aleksei-fedotov: Done.

src/tbb/dynamic_link.cpp
```diff
@@ -110,9 +125,100 @@ namespace r1 {
...
+        case dl_sig_other_error:
+            str = va_arg(args, const char*);
+            error = va_arg(args, dlerr_t);
+            TBB_FPRINTF(stdout, "%s The module \"%s\" was not loaded. System error code "
+                        DLERROR_SPECIFIER "\n", prefix, str, error);
+            break;
+        }
+
+        va_end(args);
+    } // library_warning
```
egfefey, line 215: Is there a case that covers a certificate that has expired?

src/tbb/dynamic_link.cpp
```diff
@@ -427,53 +544,204 @@ namespace r1 {
...
+            DYNAMIC_LINK_WARNING( dl_sig_security_settings, filepath );
+            break;
+        default:
+            DYNAMIC_LINK_WARNING( dl_sig_other_error, filepath, retval);
+            break;
+        }
+    }
+
+    bool has_valid_signature(const char* filepath, const std::size_t length) {
+        __TBB_ASSERT_EX(length <= PATH_MAX, "Too small buffer for path conversion");
```
egfefey, line 630: This is minor:
Is length too small when it is equal to PATH_MAX? (Maybe length < PATH_MAX)

src/tbb/CMakeLists.txt
```diff
@@ -71,6 +71,15 @@ if (NOT ("${CMAKE_SYSTEM_PROCESSOR}" MATCHES "(armv7-a|aarch64|mips|arm64|riscv)
     target_compile_definitions(tbb PRIVATE __TBB_USE_ITT_NOTIFY)
 endif()
 
+if (TBB_VERIFY_DEPENDENCY_SIGNATURE)
+    if (MSVC)
+        set(TBB_COMPILER_MACRO_SWITCH /D)
+    else()
+        set(TBB_COMPILER_MACRO_SWITCH -D)
+    endif()
+    set(TBB_LIB_COMPILE_FLAGS ${TBB_LIB_COMPILE_FLAGS} ${TBB_COMPILER_MACRO_SWITCH}__TBB_VERIFY_DEPENDENCY_SIGNATURE)
```
kboyarinov, line 80: Can `target_compile_definitions` be used instead?

src/tbb/dynamic_link.cpp
```diff
@@ -427,53 +544,212 @@ namespace r1 {
...
+        pWVTData.dwUIChoice          = WTD_UI_NONE;                    // Disable WVT UI
+        pWVTData.fdwRevocationChecks = WTD_REVOKE_WHOLECHAIN;          // Check the whole chain
+        pWVTData.dwUnionChoice       = WTD_CHOICE_FILE;                // Verify file signature
+        pWVTData.pFile               = &fdata;
+        pWVTData.dwStateAction       = WTD_STATEACTION_VERIFY;         // Verify action
+        // Perform revocation checking on the entire certificate chain but use only the local cache
+        pWVTData.dwProvFlags         = WTD_CACHE_ONLY_URL_RETRIEVAL | WTD_REVOCATION_CHECK_CHAIN;
+        pWVTData.dwUIContext         = WTD_UICONTEXT_EXECUTE;          // UI Context to run the file
+
+        const auto rc = WinVerifyTrust((HWND)INVALID_HANDLE_VALUE, &pgActionID, &pWVTData);
```
egfefey, line 669: Better to use LONG instead of auto to prevent ambiguity and unintended typecasting when doing comparisons later with the return code from WinVerifyTrust

egfefey (commented):
From the security standpoint, this is good for now.
A few notes:

- This is still up for discussion but anything beside ERROR_SUCCESS from WinVerifyTrust should result in an error instead of a warning sometimes. The user always has to option to turn off signature verification if they want, so once it's on, it's either success or failure.
- Might also need review from the TBB team.
- If there are some testing results that we can share, that will be great.

src/tbb/CMakeLists.txt
aleksei-fedotov: Of course! Thanks! Implemented.

aleksei-fedotov:
> Update to use SearchPathA is good from the security point of view.

Thanks, @egfefey! Then I proceed with productization of this patch.

CMakeLists.txt
```diff
@@ -233,6 +234,13 @@ foreach(FILE_WITH_EXTRA_TARGETS ${FILES_WITH_EXTRA_TARGETS})
     include(${FILE_WITH_EXTRA_TARGETS})
 endforeach()
 
+if (NOT TBB_VERIFY_DEPENDENCY_SIGNATURE)
+    message(WARNING
+        "Verification of a dependency signature during dynamic run-time linking is explicitly disabled."
+        "This may lead to security vulnerabilities. It is recommended to leave it enabled by default"
+        " or set TBB_VERIFY_DEPENDENCY_SIGNATURE ON to enable it explicitly.")
```
isaevil, line 241: What about making it shorter a little bit?
```suggestion
        "This may lead to security vulnerabilities. It is recommended to set "
        "TBB_VERIFY_DEPENDENCY_SIGNATURE to ON (default value)")
```

src/tbb/CMakeLists.txt
```diff
@@ -71,6 +71,10 @@ if (NOT ("${CMAKE_SYSTEM_PROCESSOR}" MATCHES "(armv7-a|aarch64|mips|arm64|riscv)
     target_compile_definitions(tbb PRIVATE __TBB_USE_ITT_NOTIFY)
 endif()
 
+if (TBB_VERIFY_DEPENDENCY_SIGNATURE)
+    target_compile_definitions(tbb PRIVATE __TBB_VERIFY_DEPENDENCY_SIGNATURE)
+endif()
```
isaevil, line 76: See target_compile_definitions call at line 58. You should be able to use the same pattern as for BUILD_SHARED_LIBS.

CMakeLists.txt
```diff
@@ -122,6 +122,7 @@ cmake_dependent_option(TBBMALLOC_PROXY_BUILD "Enable tbbmalloc_proxy build" ON "
 option(TBB_CPF "Enable preview features of the library" OFF)
 option(TBB_FIND_PACKAGE "Enable search for external oneTBB using find_package instead of build from sources" OFF)
 option(TBB_DISABLE_HWLOC_AUTOMATIC_SEARCH "Disable HWLOC automatic search by pkg-config tool" ${CMAKE_CROSSCOMPILING})
+option(TBB_VERIFY_DEPENDENCY_SIGNATURE "Verify signature of a dependency during its dynamic run-time linking" ON)
```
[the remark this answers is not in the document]
isaevil: I personally prefer the current name :)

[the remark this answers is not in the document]
aleksei-fedotov: Agreed to not change the user-facing option, but amend the internal.

CMakeLists.txt
aleksei-fedotov: Updated.

src/tbb/CMakeLists.txt
aleksei-fedotov: Applied.

CMakeLists.txt
```diff
@@ -233,6 +234,13 @@ foreach(FILE_WITH_EXTRA_TARGETS ${FILES_WITH_EXTRA_TARGETS})
     include(${FILE_WITH_EXTRA_TARGETS})
 endforeach()
 
+if (NOT TBB_VERIFY_DEPENDENCY_SIGNATURE AND WIN32)
+    message(WARNING "Verification of a dependency signature during dynamic "
+        "run-time linking is explicitly disabled. This may lead to security "
+        "vulnerabilities. It is recommended to set "
+        "TBB_VERIFY_DEPENDENCY_SIGNATURE to ON (default value)")
+endif()
```
aleksei-fedotov, line 242: Consider adding `else` branch to show that binary signatures are enabled so that users are aware about potential issues. Something like:
```cmake
message(STATUS "oneTBB is configured to verify signatures of dynamic dependencies loaded during run-time.")
```

INSTALL.md
```diff
@@ -16,11 +16,11 @@ cmake <options> <repo_root>
...
+| `-G <generator>`                           | Specify project generator                                                     | For more information, run cmake `–help`.                                                                                                                                                                                     |   |
+| `-DCMAKE_BUILD_TYPE=Debug`                 | Specify for Debug build                                                       | Not applicable for multi-configuration generators such as Visual Studio generator.                                                                                                                                           |   |
+| `-DTBB_VERIFY_DEPENDENCY_SIGNATURE=ON|OFF` | Controls signature verification of dynamic dependencies loaded during runtime | If set to ON, only the signed dynamic dependencies are loaded. This is the recommended behavior. By default, the value is unspecified. Therefore, the warning is printed. To suppress the warning, set the value explicitly. |   |
```
isaevil, line 23: Looks like it doesn't render correctly. Perhaps?
```suggestion
| `-DTBB_VERIFY_DEPENDENCY_SIGNATURE=ON\|OFF` | Controls signature verification of dynamic dependencies loaded during runtime | If set to ON, only the signed dynamic dependencies are loaded. This is the recommended behavior. By default, the value is unspecified. Therefore, the warning is printed. To suppress the warning, set the value explicitly. |   |
```

kboyarinov (dismissed):
LGTM overall but I didn't review warnings and error codes in details

omalyshe:
@aleksei-fedotov Please adjust the description because signature check is now OFF by default.

isaevil (approved):

isaevil (approved):

PR #1596, opened 2025-01-08
Fix the allocator for Flow Graph critical tasks creating

### LINKED ISSUE #1595: flow_graph: Assertion m_private_counter >= 0 failed
# Summary
When running TBB flow_graph I sporadically encounter  `Assertion m_private_counter >= 0 failed (located in the destroy function, line in file: 141)` on destruction. 
This is a similiar to [Issue-639](https://github.com/uxlfoundation/oneTBB/issues/639), only that I encounter it in succeeding TBB versions. 

One thing that I observed is that the failure does not occur when setting `tbb::flow::node_priority_t{0}`. 
This is not required in this minimal example, but needed in our implementation as we have other nodes with `tbb::flow::node_priority_t{0}`. 

# Version
Able to reproduce with `TBB 2022.0.0`
# Environment
* OS Ubuntu 22.04.4 LTS
* Compiler: gcc 11.4.0 | clang 17

# Steps To Reproduce
```cpp
#include <iostream>
#include <cassert>
#include <tbb/flow_graph.h>

int main() {
    struct Message {
        int idx;
        [[maybe_unused]] char pad[1] = {0};
    };
    using NODE = tbb::flow::function_node<Message, tbb::flow::continue_msg>;
    int num_iterations = 10000;
    int node_activations = 0;

    tbb::flow::graph graph;
    NODE node{graph, tbb::flow::serial, [&]([[maybe_unused]] const auto& msg){ node_activations++; },
              tbb::flow::queueing{}, tbb::flow::node_priority_t{1}};

    for (int i{0}; i < num_iterations; i++) {
        node.try_put(Message{i});
    }

    graph.wait_for_all();
    assert (num_iterations == node_activations);
    return 0;
}
```

kboyarinov:
### Description 
There is a bug in the current implementation of Flow Graph while using functional node concurrency limits together with node priorities.
Current algorithm for creating tasks in function node is the following:
* If the concurrency limit is not reached, two tasks are created - graph_task for executing the body and priority_task_selector task for prioritization. Both tasks are created by the same thread.
* If the concurrency limit is reached, the postponed items are stored in the queue.
* Once the resources of the node are available, the thread X tries to get the postponed item from the queue using the aggregator.
* Since the aggregator is used, another thread Y can take the "aggregated" work and hence, create the graph_task using its small object pool.
* Once the control goes back to thread X, it creates a priority_task_selector wrapper. The bug in the current implementation is that the small object pool, associated with the underlying task (created by thread Y) is used instead of small object pool of thread X.

Also this patch adds extra assert for small object pool to ensure that the pool, for which the allocation is requested is the same as the TLS pool.

Fixes #1595 

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/include/oneapi/tbb/detail/_flow_graph_impl.h
+++ b/include/oneapi/tbb/detail/_flow_graph_impl.h
@@ -1,5 +1,5 @@
 /*
-    Copyright (c) 2005-2024 Intel Corporation
+    Copyright (c) 2005-2025 Intel Corporation
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
--- a/include/oneapi/tbb/detail/_flow_graph_impl.h
+++ b/include/oneapi/tbb/detail/_flow_graph_impl.h
@@ -526,7 +526,7 @@ inline graph_task* prioritize_task(graph& g, graph_task& gt) {
     //! priority queue, and a new critical task is created to take and execute a work item with
     //! the highest known priority. The reference counting responsibility is transferred to
     //! the new task.
-    // The newly created small_object_allocator should be used to allocate the priority_task_selector
+    // A newly created small_object_allocator should be used to allocate the priority_task_selector
     // instead of the allocator, associated with gt since gt can be allocated by another thread
     d1::small_object_allocator allocator;
     d1::task* critical_task = allocator.new_object<priority_task_selector>(g.my_priority_queue, allocator);
--- a/include/oneapi/tbb/detail/_flow_graph_impl.h
+++ b/include/oneapi/tbb/detail/_flow_graph_impl.h
@@ -526,7 +526,8 @@ inline graph_task* prioritize_task(graph& g, graph_task& gt) {
     //! priority queue, and a new critical task is created to take and execute a work item with
     //! the highest known priority. The reference counting responsibility is transferred to
     //! the new task.
-    d1::task* critical_task = gt.my_allocator.new_object<priority_task_selector>(g.my_priority_queue, gt.my_allocator);
+    d1::small_object_allocator allocator;
+    d1::task* critical_task = allocator.new_object<priority_task_selector>(g.my_priority_queue, allocator);
     __TBB_ASSERT( critical_task, "bad_alloc?" );
     g.my_priority_queue.push(&gt);
     using tbb::detail::d1::submit;
--- a/include/oneapi/tbb/detail/_flow_graph_impl.h
+++ b/include/oneapi/tbb/detail/_flow_graph_impl.h
@@ -526,6 +526,8 @@ inline graph_task* prioritize_task(graph& g, graph_task& gt) {
     //! priority queue, and a new critical task is created to take and execute a work item with
     //! the highest known priority. The reference counting responsibility is transferred to
     //! the new task.
+    // The newly created small_object_allocator should be used to allocate the priority_task_selector
+    // instead of the allocator, associated with gt since gt can be allocated by another thread
     d1::small_object_allocator allocator;
     d1::task* critical_task = allocator.new_object<priority_task_selector>(g.my_priority_queue, allocator);
     __TBB_ASSERT( critical_task, "bad_alloc?" );
--- a/src/tbb/small_object_pool.cpp
+++ b/src/tbb/small_object_pool.cpp
@@ -40,7 +40,8 @@ void* __TBB_EXPORTED_FUNC allocate(d1::small_object_pool*& allocator, std::size_
     // TODO: optimize if the allocator contains a valid pool.
     auto tls = governor::get_thread_data();
     auto pool = tls->my_small_object_pool;
-    __TBB_ASSERT(allocator == nullptr || pool == allocator, "An attempt was made to allocate using another thread's small memory pool");
+    __TBB_ASSERT(allocator == nullptr || pool == allocator,
+                 "An attempt was made to allocate using another thread's small memory pool");
     return pool->allocate_impl(allocator, number_of_bytes);
 }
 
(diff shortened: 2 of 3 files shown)


include/oneapi/tbb/detail/_flow_graph_impl.h
```diff
@@ -526,7 +526,8 @@ inline graph_task* prioritize_task(graph& g, graph_task& gt) {
     //! priority queue, and a new critical task is created to take and execute a work item with
     //! the highest known priority. The reference counting responsibility is transferred to
     //! the new task.
-    d1::task* critical_task = gt.my_allocator.new_object<priority_task_selector>(g.my_priority_queue, gt.my_allocator);
+    d1::small_object_allocator allocator;
+    d1::task* critical_task = allocator.new_object<priority_task_selector>(g.my_priority_queue, allocator);
```
aleksei-fedotov, line 530: At the very least I would add a comment why it is necessary, explaining why graph_task's allocator cannot be used here. After all, its purpose is to cache allocations. BTW, if it cannot be used, do we need to have it in `graph_task` at all?

kboyarinov: Sure, I agree that the comment is required.
Regarding the necessity of the allocator inside of the `graph_task`, it is required not only for allocating the critical tasks, but also for deallocation of the `graph_task` itself after the execution (even in the case without critical tasks).

test/tbb/test_function_node.cpp
```diff
@@ -806,3 +806,25 @@ TEST_CASE("test function_node try_put_and_wait") {
...
+    tbb::flow::graph g;
+    node_type node(g, tbb::flow::serial, node_body, tbb::flow::node_priority_t{1});
+
+    for (int i = 0; i < num_iterations; ++i) {
+        node.try_put(i);
+    }
+
+    g.wait_for_all();
+    REQUIRE_MESSAGE(num_calls == num_iterations, "Incorrect number of body executions");
+}
```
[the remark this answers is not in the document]
kboyarinov: I will investigate if it is possible to change the test to make it more lightweight. 
As far as I understand, the root cause of the issue is the aggregator of `function_node` and we need to maximize the pressure on the aggregator to increase the chance that one thread will request creating of the task, another thread will create it under the aggregator and return the control to the first thread. So I am not sure that increasing or blocking the node body would help.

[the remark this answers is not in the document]
kboyarinov: As far as I understand, increasing the amount of work inside of the body will not help to reproduce the issue. It only takes place if we have one thread that enters the aggregator to add new tasks into the queue (as part of `try_put`) and another thread entering the aggregator to take the task from the queue (as part of post-processing the `body`) and wrapping it into the critical task. And we need the first thread to "win" on the aggregator and process the work from the second. I don't see how we can increase the probability of such a behavior using the Flow Graph public API. 

I think the current test is OK since it provides constant simultaneous adding items to the queue and taking items from it (because of the lightweight body). As I see, the test reproduces the issue in ~80% of runs.

include/oneapi/tbb/detail/_flow_graph_impl.h
kboyarinov: The comment was added

src/tbb/small_object_pool.cpp
```diff
@@ -40,6 +40,8 @@ void* __TBB_EXPORTED_FUNC allocate(d1::small_object_pool*& allocator, std::size_
     // TODO: optimize if the allocator contains a valid pool.
     auto tls = governor::get_thread_data();
     auto pool = tls->my_small_object_pool;
+    __TBB_ASSERT(allocator == nullptr || pool == allocator,
+                 "An attempt was made to allocate using another thread's small memory pool");
```
aleksei-fedotov, line 44: I suggest moving this assert into `allocate_impl` function below, to put it just before line 68, where an assignment is made so that they become:
```cpp
__TBB_ASSERT(allocator == nullptr || allocator == this,
             "An attempt was made to allocate using another thread's small memory pool");
allocator = this;
```
This way the `allocate(small_object_pool, number_of_bytes, execution_data)` overload above is covered as well.

kboyarinov: IMHO, it would be better to place it in `allocate_impl` but before doing something, in the very beginning of the function

test/tbb/test_function_node.cpp
```diff
@@ -806,3 +806,25 @@ TEST_CASE("test function_node try_put_and_wait") {
     test_try_put_and_wait();
 }
 #endif
+
+// It was an issue when the critical task wrapper was allocated using the small object pool
+// of the task being wrapped. Since the original task creates under the aggregator, there is no
+// guarantee that the thread that requested the task creating is the same as actually created the task
+// Mismatch between memory pull caused internal assertion failure while deallocating the task
```
vossmjp, line 813:
```suggestion
// Mismatch between memory pool caused internal assertion failure while deallocating the task
```

include/oneapi/tbb/detail/_flow_graph_impl.h
```diff
@@ -526,7 +526,10 @@ inline graph_task* prioritize_task(graph& g, graph_task& gt) {
     //! priority queue, and a new critical task is created to take and execute a work item with
     //! the highest known priority. The reference counting responsibility is transferred to
     //! the new task.
-    d1::task* critical_task = gt.my_allocator.new_object<priority_task_selector>(g.my_priority_queue, gt.my_allocator);
+    // The newly created small_object_allocator should be used to allocate the priority_task_selector
```
vossmjp, line 529:
```suggestion
    // A newly created small_object_allocator should be used to allocate the priority_task_selector
```

aleksei-fedotov (approved):
Looks good to me!

PR #1604, opened 2025-01-15
[try_put_and_wait] Add tag-based implementation for try_put_and_wait for multifunction and async nodes

kboyarinov:
### Description 
Add reference implementation of try_put_and_wait feature for multifunction and async nodes as described in the corresponding RFC. 

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown


test/tbb/test_try_put_and_wait.h
```diff
@@ -0,0 +1,468 @@
+/*
```
kboyarinov, line 1: Most part of this file was moved from test_buffering_try_put_and_wait without any changes

test/tbb/test_multifunction_node.cpp
```diff
@@ -541,11 +542,11 @@ TEST_CASE("Test ports retrurn references"){
...
-//! NativeParallelFor testing with various concurrency settings
-//! \brief \ref error_guessing
-TEST_CASE("Lightweight testing"){
-    lightweight_testing::test<tbb::flow::multifunction_node>(10);
-}
+// //! NativeParallelFor testing with various concurrency settings
+// //! \brief \ref error_guessing
+// TEST_CASE("Lightweight testing"){
+//     lightweight_testing::test<tbb::flow::multifunction_node>(10);
+// }
```
kboyarinov, line 549: I faced some issues with this test and will investigate them as part of this patch. 
But I guess the overall review can be started before making the fix

kboyarinov: Fixed

include/oneapi/tbb/detail/_flow_graph_node_impl.h
```diff
@@ -554,6 +554,70 @@ struct init_output_ports {
...
+
+    metainfo_tag_type(const message_metainfo& metainfo) : my_metainfo(metainfo) {
+        for (auto waiter : my_metainfo.waiters()) {
+            waiter->reserve();
+        }
+    }
+
+    metainfo_tag_type& operator=(const metainfo_tag_type&) = delete;
+    metainfo_tag_type& operator=(metainfo_tag_type&& other) {
+        // TODO: should this method be thread-safe?
```
vossmjp, line 576: Could someone want to use assignment in place of reset to start a next batch of accumulations -- starting with a different first tag. Then there could be a race between the assignment and merge..?

test/tbb/test_function_node.cpp
```diff
@@ -657,7 +657,7 @@ void test_try_put_and_wait_rejecting(size_t concurrency_limit) {
         std::vector<int> processed_items;
         std::vector<int> new_work_items;
 
-        int wait_message = 0;
+        int wait_message = 10;
```
vossmjp, line 660: You forgot to update copyright year.

kboyarinov: Fixed

include/oneapi/tbb/detail/_flow_graph_node_impl.h
kboyarinov: It is in theory really interesting, which thread-safety guarantees should be provided.
It is obvious, that:
* `merge` should be thread-safe with other `merge`
* `reset` should be thread-safe with `reset`

`merge` can be thread-safe with `reset` (as well as with `operator=`) but it is hard to imagine the real use case. If we have both `merge` and `reset` simultaneously, which result can the user expect - merged or reset object.
Currently, the first function that acquires the mutex will define the behavior.
But let's consider the 2 consecutive reductions that use the single accumulators. If in the edge between 2 reductions, there would be simultaneous `merge` from reduction 2 and `reset` from the reduction 1, some elements from the reduction 2 can be missed if the merge will go before the reset.

For me, it is unclear how we can manage this. I think the answer for this question is the answer also to thread-safety of `operator=`

kboyarinov: Other interesting question is should it be thread-safe to have the `left.merge(middle)` simultaneous with `middle.merge(right)`. I feel that the answer is no, but still want to highlight this.

PR #1606, opened 2025-01-15
Fix build with non-english locale

moritz-h:
### Description 

The build fails on non-english systems due to parsing localized gcc output in CMake.
This was partially already fixed in #1450, but next to the `LANG` env variable also `LC_ALL` needs to be explicitly set to `C` otherwise `LC_ALL` or `LANGUAGE` can take precedence over `LANG`.
See [gettext docs](https://www.gnu.org/software/gettext/manual/html_node/The-LANGUAGE-variable.html) and [GCC docs](https://gcc.gnu.org/onlinedocs/gcc/Environment-Variables.html).

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/cmake/compilers/GNU.cmake
+++ b/cmake/compilers/GNU.cmake
@@ -1,4 +1,4 @@
-# Copyright (c) 2020-2024 Intel Corporation
+# Copyright (c) 2020-2025 Intel Corporation
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


cmake/compilers/GNU.cmake
isaevil, line 1: Could you please bump the copyright year from 2024 to 2025?

isaevil (approved):

PR #1607, opened 2025-01-17
Add RFC for CTAD in blocked_nd_range

kboyarinov:
### Description 
Reference implementation is #1525 

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown


rfcs/proposed/blocked_nd_range_ctad/README.md
```diff
@@ -0,0 +1,203 @@
...
+tbb::blocked_nd_range range(array);
+#+end_src
+
+Since the `blocked_range` is not constructible from C-array and the braced-init-list is not used, the user expects the range to be deduced as
+`blocked_nd_range<int, 2>` and the constructor `[2]` to be used.
+
+If we add one more explicit deduction guide to support the code above, the single braced-init-list of size 2 or 3 would also match on this guide.
+
+There are the following options how this issue can be resolved:
+* Add a new deduction guide to support the code above. The downside of this approach is that it makes the ambiguity, discussed in the
```
kboyarinov, line 184: My vision is that it would be better to pick that option since the explanation of the downside would not be surprise for the user and will be understandable from the deduction guides definition. I also think that construction of multi-dimensional ranges are mostly common use-case for blocked_nd_range. 

The only thing that actually can be affected, is some generic code that can create both one and multi-dimensional ranges

rfcs/experimental/blocked_nd_range_ctad/README.md
```diff
@@ -0,0 +1,219 @@
...
+tbb::blocked_nd_range range(array);
+```
+
+Since the `blocked_range` is not constructible from C-array and the braced-init-list is not used, the user expects the range to be deduced as
+`blocked_nd_range<int, 2>` and the constructor `[2]` to be used.
+
+If we add one more explicit deduction guide to support the code above, the single braced-init-list of size 2 or 3 would also match on this guide.
+
+There are the following options how this issue can be resolved:
+* Add a new deduction guide to support the code above. The downside of this approach is that it makes the ambiguity, discussed in the
```
kboyarinov, line 199: My vision is that it would be better to pick that option since the explanation of the downside would not be surprise for the user and will be understandable from the deduction guides definition. I also think that construction of multi-dimensional ranges are mostly common use-case for blocked_nd_range.

The only thing that actually can be affected, is some generic code that can create both one and multi-dimensional ranges

rfcs/proposed/blocked_nd_range_ctad/README.md
kboyarinov: Reopened since this discussion is outdated after retargeting to experimented.

rfcs/experimental/blocked_nd_range_ctad/README.md
```diff
@@ -0,0 +1,219 @@
...
+and the braced-init-lists:
+
+```cpp
+tbb::blocked_range<int> dim_range(0, 100);
+tbb::blocked_nd_range nd_range(dim_range, {0, 200}, {0, 300}, dim_range);
+```
+
+These arguments would not match nether on the `[g1]` not `[g2]` and it is unclear how to define the deduction guide that covers this case.
+Current proposal is to keep this scenario a limitation for using the CTAD and always require using the consistent set of parameters - or 
+the set of braced-init-lists or the set of `tbb::blocked_range` objects.
```
vossmjp, line 219: Can you add exit criteria? Likely the exit conditions are that we get feedback on user experience and it confirms choices made about open questions. This could be a place to collect those choices. And another exit criteria is that we get specification updates accepted (likely backed by user experience).

kboyarinov: Done

vossmjp (approved):
Looks good to me.

Issue #1467, opened 2024-08-04
Test test_buffer_node sporadically hangs on x86_64


phprus:
# Summary
Test ``test_buffer_node`` sporadically hangs on x86_64
Previous discussion: https://github.com/oneapi-src/oneTBB/pull/1445#issuecomment-2260671869

@pavelkumbrasev , @kboyarinov

# Version
Commit: 306c75f5e906fc221e12f45dae45900f0d8db6ba

# Environment
OS: Debian 12
gcc version 12.2.0 (Debian 12.2.0-14)

# Steps To Reproduce
```
cmake -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_CXX_STANDARD=20 -DCMAKE_INTERPROCEDURAL_OPTIMIZATION=ON ../..
cmake --build . --verbose --config Release -j8
ctest --timeout 0 --build-config RelWithDebInfo -R test_buffer_node --repeat-until-fail 10000000
```

Backtrace:

```
phprus@srv2:~$ gdb /home/phprus/test/tbb/oneTBB-306c75f5e906fc221e12f45dae45900f0d8db6ba/build/cpp20rwd/gnu_12.2_cxx20_64_relwithdebinfo/test_buffer_node -p 2468289
GNU gdb (Debian 13.1-3) 13.1
Copyright (C) 2023 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<https://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from /home/phprus/test/tbb/oneTBB-306c75f5e906fc221e12f45dae45900f0d8db6ba/build/cpp20rwd/gnu_12.2_cxx20_64_relwithdebinfo/test_buffer_node...
[...]

PR #1609, opened 2025-01-20
Fix concurrent_bounding_queue capacity on copy, move and swap operations

### LINKED ISSUE #1598: A copy of a concurrent bounded queue doesn't preserve the capacity of the original queue object.
# Summary
A copy of a concurrent bounded queue doesn't preserve the capacity of the original queue.

# Version
2021.10.0

# Environment
AlmaLinux 9
gcc 13.1.0

# Observed Behavior
If I set the capacity of an instance of concurrent bounded queue to a specific value and then create a copy of this queue, the capacity of the new object is set to the default value (0x3fffffffffffffff)

# Expected Behavior
I would expect the copied queue to retain the same capacity as the original.

# Steps To Reproduce

The following code can be used to reproduce the issue:

```
#include <iostream>
#include <tbb/concurrent_queue.h>
int main(int ac, char**av) {
    tbb::concurrent_bounded_queue<int> q;
    q.set_capacity(100);
    std::cout << "q capacity = " << q.capacity() << std::endl;
    tbb::concurrent_bounded_queue<int> q2(q);
    std::cout << "q capacity = " << q.capacity() << " q2 capasity = " << q2.capacity() << std::endl;
}
```

It produces the following output:

```
q capacity = 100
q capacity = 100 q2 capasity = 4611686018427387903
```

kboyarinov:
### Description 
concurrent_bounded_queue capacity was not preserved while copy/move construction and assignment and on swap.
The target queue always used its previous capacity, or an unbounded capacity.

Fixes #1598 

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown


test/tbb/test_concurrent_queue.cpp
```diff
@@ -282,3 +282,45 @@ TEST_CASE("Test clear and dtor with TrackableItem") {
...
+TEST_CASE("test capacity on modifying operations") {
+    // Test that concurrent_bounded_queue capacity is preserved on copying, moving and swapping
+    using queue_type = oneapi::tbb::concurrent_bounded_queue<int>;
+    using capacity_type = typename queue_type::size_type;
+
+    queue_type q;
+    capacity_type desired_capacity = 64;
+
+    q.set_capacity(desired_capacity);
+    REQUIRE_MESSAGE(q.capacity()  == desired_capacity, "Capacity is not set correctly");
```
isaevil, line 296:
```suggestion
    REQUIRE_MESSAGE(q.capacity() == desired_capacity, "Capacity is not set correctly");
```

phprus:
``test_buffer_node`` timeout issue: #1467 (Aug 4, 2024)

aleksei-fedotov (approved):
LGTM!

PR #1611, opened 2025-01-21
Fix C++20 macro in parallel_for_each + fix concepts macro definition

### LINKED ISSUE #1552: tbb::parallel_for_each build fail on gcc 9.4.0 using C++20
# Summary

We are trying to use `tbb::parallel_for_each` within a project compiled with gcc 9.4.0 within an Ubuntu 20.04 environment. We have been using C++17 and are moving to C++20. tbb was compiling correctly with C++17. However with C++20 we see build errors relating to missing `std::random_access_iterator`. I think this is due to a mistake in dealing with collections in development versions of C++20.

Specifically I think the problem is with an if-def in `parallel_for_each.h`:
```
#if __TBB_CPP20_PRESENT
...
#endif // __TBB_CPP20_PRESENT
```
I believe this should be:
```
#if __TBB_CPP20_CONCEPTS_PRESENT
...
```
Other if-def in this file use the second form to cope with the absence of concepts in development versions of C++20 for gcc 9.4.0. The first form is sufficient for non-development versions of C++20 but not early versions. Making this change fixes the build for me.

# Version
#define TBB_VERSION_MAJOR 2021
#define TBB_VERSION_MINOR 13
#define TBB_VERSION_PATCH 0

# Environment
Ubuntu 20.04 (within WSL)
gcc 9.4.0
It's probably just gcc that's important.

# Observed Behavior
Trying to compile:
```
    std::vector<SampleStruct> tasks;
...
    tbb::parallel_for_each(tasks.begin(), tasks.end(), Invoker<SampleStruct>());
```

Build error:
```
.../vcpkg_installed/x64-ubuntu-20-04-release/include/oneapi/tbb/parallel_for_each.h:427:29: error: no member named 'random_access_iterator' in namespace 'std' [clang-diagnostic-error]
    std::conditional_t<std::random_access_iterator<It>,
                            ^
.../vcpkg_installed/x64-ubuntu-20-04-release/include/oneapi/tbb/parallel_for_each.h:427:52: error: 'It' does not refer to a value [clang-diagnostic-error]
    std::conditional_t<std::random_access_iterator<It>,
                                                   ^
.../vcpkg_installed/x64-ubuntu-20-04-release/include/oneapi/tbb/parallel_for_each.h:425:20: note: declared here
template <typename It>
                   ^
.../vcpkg_installed/x64-ubuntu-20-04-release/include/oneapi/tbb/parallel_for_each.h:427:55: error: expected ';' after alias declaration [clang-diagnostic-error]
    std::conditional_t<std::random_access_iterator<It>,
                                                      ^
.../vcpkg_installed/x64-ubuntu-20-04-release/include/oneapi/tbb/parallel_for_each.h:428:55: error: expected unqualified-id [clang-diagnostic-error]
                       std::random_access_iterator_tag,
                                                      ^
.../vcpkg_installed/x64-ubuntu-20-04-release/include/oneapi/tbb/parallel_for_each.h:435:5: error: no template named 'iterator_tag_dispatch_impl' [clang-diagnostic-error]
    iterator_tag_dispatch_impl<typename move_iterator_dispatch_helper<It>::type>;
    ^
.../vcpkg_installed/x64-ubuntu-20-04-release/include/oneapi/tbb/parallel_for_each.h:501:83: error: no template named 'iterator_tag_dispatch' [clang-diagnostic-error]
template <typename Iterator, typename Body, typename Item, typename IteratorTag = iterator_tag_dispatch<Iterator>>
                                                                                  ^
.../vcpkg_installed/x64-ubuntu-20-04-release/include/oneapi/tbb/parallel_for_each.h:629:9: error: too few template arguments for class template 'for_each_root_task' [clang-diagnostic-error]
        for_each_root_task<Iterator, Body, ItemType> root_task(first, last, body, w_context, context);
        ^
.../vcpkg_installed/x64-ubuntu-20-04-release/include/oneapi/tbb/parallel_for_each.h:502:7: note: template is declared here
class for_each_root_task : public for_each_root_task_base<Iterator, Body, Item>
      ^
```

# Expected Behavior
Clean build.

# Steps To Reproduce
Sorry. I haven't checked this as vcpkg isn't supplying the latest version. It should be visible with the correct compiler gcc 9.4.0 with a simple test program:
```
#include <tbb/parallel_for_each.h>

void ParallelApplyFooToList( const std::vector<Item>& list ) {
    parallel_for_each( list.begin(), list.end(), ApplyFoo() );
}
```

kboyarinov:
### Description 

Fix two issues:
* parallel_for_each failed to compile on gcc9 since __TBB_CPP20_PRESENT was while STL concepts were used actually. Changed to __TBB_CPP20_CONCEPTS_PRESENT
* __TBB_CPP20_CONCEPTS_PRESENT macro was always defined to 0 for Intel Compiler and Clang. Changed to the detection based on Standard feature-test macros (the same approach as previously used for three-way-comparison)

Fixes #1552 

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown

### DIFF
--- a/include/oneapi/tbb/detail/_config.h
+++ b/include/oneapi/tbb/detail/_config.h
@@ -263,8 +263,8 @@
 #define __TBB_CPP17_ALLOCATOR_IS_ALWAYS_EQUAL_PRESENT   (__TBB_LANG >= 201703L)
 #define __TBB_CPP17_IS_SWAPPABLE_PRESENT                (__TBB_LANG >= 201703L)
 
-// TODO: fix concepts on Apple Clang or define the broken versions
-#if !(__clang__ && __APPLE__) && defined(__cpp_concepts) && defined(__cpp_lib_concepts)
+// TODO: fix concepts on Clang or define the broken versions
+#if !(__clang__) && defined(__cpp_concepts) && defined(__cpp_lib_concepts)
     #define __TBB_CPP20_CONCEPTS_PRESENT ((__cpp_concepts >= 201907L) && (__cpp_lib_concepts >= 202002L))
 #else
     #define __TBB_CPP20_CONCEPTS_PRESENT 0
--- a/include/oneapi/tbb/detail/_config.h
+++ b/include/oneapi/tbb/detail/_config.h
@@ -1,5 +1,5 @@
 /*
-    Copyright (c) 2005-2024 Intel Corporation
+    Copyright (c) 2005-2025 Intel Corporation
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
--- a/include/oneapi/tbb/detail/_config.h
+++ b/include/oneapi/tbb/detail/_config.h
@@ -233,10 +233,8 @@
 #if __INTEL_COMPILER && (!_MSC_VER || __INTEL_CXX11_MOVE__)
     #define __TBB_CPP14_VARIABLE_TEMPLATES_PRESENT (__TBB_LANG >= 201402L)
     #define __TBB_CPP17_DEDUCTION_GUIDES_PRESENT   (__INTEL_COMPILER > 2021 && __TBB_LANG >= 201703L)
-    #define __TBB_CPP20_CONCEPTS_PRESENT           0 // TODO: add a mechanism for future addition
 #elif __clang__
     #define __TBB_CPP14_VARIABLE_TEMPLATES_PRESENT (__has_feature(cxx_variable_templates))
-    #define __TBB_CPP20_CONCEPTS_PRESENT           0 // TODO: add a mechanism for future addition
     #ifdef __cpp_deduction_guides
         #define __TBB_CPP17_DEDUCTION_GUIDES_PRESENT (__cpp_deduction_guides >= 201611L)
     #else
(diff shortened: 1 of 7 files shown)


mmichel11:
@kboyarinov @dmitriy-sobolev @sarathnandu Could this PR please be prioritized for review? The bug is significantly impacting oneDPL's pass rate here as it shows up in our oneTBB backend, and it needs to be fixed for our upcoming milestone.

aleksei-fedotov (approved):
Looks good to me, but some tests on macOS fail. Is it because of the compiler not supporting concept specifics we use?

kboyarinov:
@aleksei-fedotov @pavelkumbrasev @isaevil @sarathnandu @dnmokhov 
I faced some unexpected and strange concept issues with clang on MacOS.
I propose to submit all of the other fixes now and keep the concepts disabled on MacOS. I will open an issue to fix that in the future or define the versions of Apple Clang that are not working with all or part of our concepts.
What do you think?

sarathnandu:
> @aleksei-fedotov @pavelkumbrasev @isaevil @sarathnandu @dnmokhov I faced some unexpected and strange concept issues with clang on MacOS. I propose to submit all of the other fixes now and keep the concepts disabled on MacOS. I will open an issue to fix that in the future or define the versions of Apple Clang that are not working with all or part of our concepts. What do you think?

I agree, since this patch is affecting oneDPL's pass rate and they are waiting for this patch to be merged. We can investigate the MacOS build issue separately.

phprus:
I think this is a bug in Clang (not only AppleClang):
https://github.com/llvm/llvm-project/issues/61763

The bug is fixed in version LLVM 18.1.0.

AppleClang from Xcode 16.2 based on LLVM 17.

sarathnandu (approved):
LGTM!

sarathnandu (changes requested):
The icpx compiler is also failing. I have tested on Intel(R) oneAPI DPC++/C++ Compiler 2023.2.0 (2023.2.0.20230721)

I believe you need to disable it for all clang not just clang && APPLE.

dnmokhov:
> I think this is a bug in Clang (not only AppleClang): [llvm/llvm-project#61763](https://github.com/llvm/llvm-project/issues/61763)
> 
> The bug is fixed in version LLVM 18.1.0.
> 
> AppleClang from Xcode 16.2 based on LLVM 17.

I agree that we should just disable these concepts with any Clang for now.

My results of compiling `test_blocked_range.cpp` on Ubuntu:
* Clang 17.0.1 - same compilation error
* Clang 18.1.0 - compiles
* Clang 19.1.0 - internal compiler error

kboyarinov:
> > > I think this is a bug in Clang (not only AppleClang): [llvm/llvm-project#61763](https://github.com/llvm/llvm-project/issues/61763)
> > > The bug is fixed in version LLVM 18.1.0.
> > > AppleClang from Xcode 16.2 based on LLVM 17.
> > 
> > 
> > I agree that we should just disable these concepts with any Clang for now.
> > My results of compiling `test_blocked_range.cpp` on Ubuntu:
> > 
> > * Clang 14.0.0 - compiles
> > * Clang 15.0.0 - compiles
> > * Clang 17.0.1 - same compilation error as in MacOS CI above
> > * Clang 18.1.0 - compiles
> > * Clang 19.1.0 - internal compiler error
> 
> Should not we then check the clang's version and enable concepts only if the version is not 17 or not 19? We have `__TBB_CLANG_VERSION` macro.

I agree, we can enable back concepts on the clang versions not affected by the issue described above.
An other option we have is not to disable concepts completely since the issue only affects TBB classes that uses a constrained friend declaration inside (like `blocked_range` that friends `blocked_range2d`.

I still propose to keep concepts on clang totally disabled for this release and create a separate issue for enabling it to make sure other fixes introduced in this patch are landed. I still feel that we need to test the potential fix better on different versions of clang on both Linux and MacOS before enabling concepts.

aleksei-fedotov (approved):
Okay, let's have it disabled for Clang in this release.

sarathnandu (approved):
LGTM

mcaryl-digitalbarriers:
Thanks for this.

PR #1617, opened 2025-01-27
RFC for the task arena waiting functionality

akukanov:
### Description 
A proposal to extend the API of task arena with work completion waiting methods

### DIFF
--- a/rfcs/proposed/task_arena_waiting/readme.md
+++ b/rfcs/proposed/task_arena_waiting/readme.md
@@ -1,0 +1,135 @@
+# Wait for work completion in a task_arena
+
+## Motivation
+
+Task arenas in oneTBB are the places for threads to share and execute tasks.
+A `task_arena` instance represents a configurable execution context for parallel work.
+
+There are two primary ways to submit work to an arena: the `execute` and `enqueue` functions.
+Both take a callable object and run it in the context of the arena. The callable object
+might start more parallel work in the arena by invoking a oneTBB algorithm, running a flow graph,
+or submitting work into a task group.
+`execute` is a blocking call: the calling thread does not return until the callable object
+completes. `enqueue` is a “fire-and-forget” call: the calling thread submits the callable
+object as a task and returns immediately, providing no way to synchronize with the completion
+of the task.
+
+In [the oneDPL case study of asynchronous algorithms](https://github.com/uxlfoundation/oneDPL/tree/main/rfcs/archived/asynchronous_api_general#use-case-study)
+three main usage patterns were observed, each assuming subsequent synchronization either with
+the asynchronous job or with all jobs previously submitted into a work queue or device.
+A task arena can be considered analogous to a device or a work queue in these patterns,
+but it lacks synchronization capabilities and so cannot alone support these use cases -
+for that it has to be paired with something "waitable".
+
+### Current solution: combining with a task group
+
+In oneTBB, asynchronous execution is supported by `task_group` and the flow graph API; both allow
+submitting a job and waiting for its completion later. The flow graph is more suitable for building
+and executing graphs of dependent computations, while the `task_group`, as of now, allows starting
+and waiting for independent tasks. Notably, both require calling `wait`/`wait_for_all` to ensure that
+the work will be done. The `task_arena::enqueue`, on the other hand, being "fire-and-forget", enforces
+eventual availability of another thread in the arena to execute the task.
+
+So, a reasonable solution for the described use cases seems to combine a `task_arena` with a `task_group`.
+However, it is notoriously non-trivial to do right. For example, the following "naive" attempt is subtly
+incorrect:
+```cpp
+tbb::task_arena ta{/*args*/};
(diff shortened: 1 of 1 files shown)


rfcs/proposed/task_arena_waiting/readme.md
```diff
@@ -0,0 +1,135 @@
...
+}
+
+for(unsigned j = 0; j < numa_arenas.size(); j++) {
+    numa_arenas[j].wait_for( task_groups[j] );
+}
+```
+
+Additionally, a task group parameter can also be added to `execute` as well as to `this_task_arena::isolate`,
+which is similar to `execute` but provides additional work isolation guarantees. That would make
+the experimental `isolated_task_arena` class fully obsolete.
```
vossmjp, line 117:
```suggestion
the experimental `isolated_task_group` class fully obsolete.
```

aleksei-fedotov (dismissed):
I like it. Why is it in draft state? Do you want to elaborate something or write the other idea(s) regarding the waiting of work completion in arenas?

akukanov:
> I like it. Why is it in draft state? Do you want to elaborate something or write the other idea(s) regarding the waiting of work completion in arenas?

I would be happy to get it initially accepted :) The next thing is to describe the proposed APIs more formally, and perhaps elaborate on the design - for which I would really need some feedback from the development team.

tobiasweinzierl80:
I do appreciate this topic pops up again. I wonder however if the introduction of an unsafe wait in the tradition of a yield should not be considered, too? I.e. I do something, and then ask the arena to do at least one more task before it returns, i.e. not totally blocking but instructing the scheduler. This way, I can build in fairness into an algorithm.

akukanov:
@tobiasweinzierl80 Is the idea for an application thread to temporarily join arena and help making progress, or for it to "sleep" until some progress is made by other threads? And in any case, what would trigger the thread to return/wake up - completion of a specific task, completion of a certain amount of unspecified tasks, timeout, or an event external to the arena (or some combination of those)?

tobiasweinzierl80:
Very good question. In MPI, there is this context of callback points. I don't think that's require here. I think joining the other team without any further progress guarantees is the most useful thing for us. So it eventually will come back, although I don't know when.

vossmjp:
It seems that at this early stage, this RFC is intended to act as an umbrella for possible task arena waiting extensions. If so, should a third category be added to include the temporary-participation case? Perhaps, the API is generalization of the proposed task_arena::wait_for to include different wait conditions: empty-state of a task group, empty-state or the arena itself, a time or progress condition?

akukanov:
@tobiasweinzierl80, @vossmjp and others, I added my thoughts about temporary participation or progress delegation by application threads, please take a look.

rfcs/proposed/task_arena_waiting/readme.md
```diff
@@ -0,0 +1,167 @@
...
+Of the three options outlined above for application thread participation, the second one - exchanging
+an application thread for a TBB thread in the arena, or *progress delegation* - seems the least risky.
+At minimum, it would enforce mandatory concurrency similarly to `enqueue`; it might also increment
+the total demand for TBB worker threads, rebalance threads between active arenas, and in the most sophisticated
+case could even ensure that a certain arena gets a temporary boost during rebalancing.
+
+## Open Questions
+
+- API details need to be elaborated
+- Implementation feasibility for (2) and (3) need to be explored
```
isaevil, line 167:
```suggestion
- Implementation feasibility for (2) and (3) needs to be explored
```

isaevil (dismissed):
I like the proposal

tobiasweinzierl80:
To be honest, I like "try_wait_for_all" for all, as it is a drain of the task queues, but makes it clear that this is a try and hence there might still be tasks left-over at the end or currently running.

rfcs/proposed/task_arena_waiting/readme.md
```diff
@@ -0,0 +1,194 @@
...
+
+The implementation can likely be backward compatible, as no layout or function changes in `task_arena`
+seems necessary. A new library entry point would be added for the waiting function.
+
+The implementation should, if possible, assign the waiting thread to execute tasks; if not, the thread
+should *block with progress delegation*, as described below.
+
+### 3. Consider API for progress delegation
+
+Of the three options outlined above for application thread participation, the second one - exchanging
```
vossmjp, line 172: Either number the options in the earlier section, or don't refer to the selected option by number at all.  It took me a minute to figure our what the three things were.  I believe those options are described in the section, "Participation of application threads" but the use of a bullet list with two items there was confusing.

vossmjp (approved):
I have one suggestion for making it read better. But otherwise, it looks go to me as an initial proposal.

rfcs/proposed/task_arena_waiting/readme.md
akukanov: Thanks; improved.

kboyarinov (approved):
LGTM

PR #1622, opened 2025-01-30
Update build instructions for WASM

dnmokhov:
### Description 
Emscripten uses static linking, so we should explicitly request a static TBB library instead of

* explicitly asking for shared (current instructions)
* using the default, i.e., shared

Both of the above get overridden with multiple warnings.

### Type of change

### Tests
- [x] not needed

### Documentation
- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/WASM_Support.md
+++ b/WASM_Support.md
@@ -33,7 +33,7 @@ To build the system, run:
 
 ```
 mkdir build && cd build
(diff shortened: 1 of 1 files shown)


sarathnandu (approved):
LGTM!

PR #1623, opened 2025-01-31
Stabilize test_parallel_phase

isaevil:
### Description 
* Use `task_arena::enqueue` instead of `parallel_for` to reduce probability of missed wakeup

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/test/tbb/test_parallel_phase.cpp
+++ b/test/tbb/test_parallel_phase.cpp
@@ -27,7 +27,6 @@
 #include "common/spin_barrier.h"
 
 #include "tbb/task_arena.h"
-#include "tbb/parallel_for.h"
 
 void active_wait_for(std::chrono::microseconds duration) {
     for (auto t1 = std::chrono::steady_clock::now(), t2 = t1;
@@ -52,7 +51,7 @@ std::size_t measure_median_start_time(tbb::task_arena* ta, const F1& start = F1{
 
     std::vector<std::chrono::steady_clock::time_point> start_times(num_threads);
     utils::SpinBarrier barrier(num_threads);
-    auto measure_start_time = [&] (std::size_t) {
+    auto measure_start_time = [&] {
         start_times[tbb::this_task_arena::current_thread_index()] = std::chrono::steady_clock::now();
         barrier.wait();
     };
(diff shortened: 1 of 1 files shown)


src/tbb/waiters.h
```diff
@@ -68,6 +68,7 @@ class outermost_worker_waiter : public waiter_base {
                     t2 = std::chrono::steady_clock::now())
                 {
                     if (!my_arena.is_empty() && !my_arena.is_recall_requested()) {
+                        reset_wait();
```
kboyarinov, line 71: Can you please add a comment in the code as well, why the reset is required

aleksei-fedotov, line 71: I think such changes require performance and power consumption measurements at the very least. If this is done only for stabilizing the unit test results, and we do not have proofs the changes are acceptable I would not rush applying them.

isaevil: Ok, let's split the patch then? I suppose the test related changes are enough to stabilize the results.

aleksei-fedotov: Sure.

test/tbb/test_parallel_phase.cpp
```diff
@@ -69,7 +68,11 @@ std::size_t measure_median_start_time(tbb::task_arena* ta, const F1& start = F1{
     auto work = [&] {
         auto start_time = std::chrono::steady_clock::now();
         start();
-        tbb::parallel_for(std::size_t(0), num_threads, measure_start_time, tbb::static_partitioner{});
+        for(std::size_t thr = 0; thr < num_threads-1; ++thr) {
+            tbb::this_task_arena::enqueue(measure_start_time);
+        }
+        start_times[tbb::this_task_arena::current_thread_index()] = std::chrono::steady_clock::now();
+        barrier.wait();
```
aleksei-fedotov, line 75: Consider adding comment why enqueueing is preferrable here. Perhaps, something like _Enqueueing the work to increase probability all the threads could find the work_ might suffice.

src/tbb/waiters.h
isaevil: Agreed to move this change to a separate patch.

aleksei-fedotov (approved):

PR #1626, opened 2025-02-04
Suppress Intel Compiler deprecation warning for STL in test_allocators.cpp

kboyarinov:
### Description 
Intel LLVM compiler triggers a warning for deprecated `std::pmr::polymorphic_allocator::destroy` inside the implementation of `std::allocator_traits::destroy` method. Since this method is not used explicitly by the user, I think the warning should not be triggered and this can be considered a compiler issue. 
Suppressed the warning on the affected platform. 

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown

### DIFF
--- a/test/tbb/test_allocators.cpp
+++ b/test/tbb/test_allocators.cpp
@@ -14,6 +14,14 @@
     limitations under the License.
 */
 
+#include <tbb/version.h>
+#define TEST_BROKEN __INTEL_LLVM_COMPILER == 20250000 && __TBB_GLIBCXX_VERSION == 110000 && __TBB_CPP20_PRESENT
+
+#if TEST_BROKEN
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+#endif
+
 #include "tbb/cache_aligned_allocator.h"
 #include "tbb/tbb_allocator.h"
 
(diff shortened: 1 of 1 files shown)


aleksei-fedotov (commented):
I think we rather should re-write the test to avoid using deprecated stuff if possible instead of suppressing the warning? It is in the test code after all.

kboyarinov:
> I think we rather should re-write the test to avoid using deprecated stuff if possible instead of suppressing the warning? It is in the test code after all.

The issue is that nothing deprecated is used actually inside of the test, only the standard containers API and `std::pmr::polymorphic_allocator`. The real issue is that the Intel compiler thinks these standard library APIs uses the deprecated functionality internally and hence there is nothing we can do other that suppressing the warning.
I have already reported the issue to the Intel LLVM compiler.

test/tbb/test_allocators.cpp
```diff
@@ -14,6 +14,17 @@
     limitations under the License.
 */
 
+#include <tbb/version.h> // For __TBB_GLIBCXX_VERSION and __TBB_CPP20_PRESENT
+
+// Intel LLVM compiler triggers a deprecation warning in the implementation of std::allocator_traits::destroy
+// inside Standard Library while using STL PMR containers since std::polymorphic_allocator::destroy is deprecated since C++20
+#define TEST_LLVM_COMPILER_PMR_DESTROY_DEPRECATED_BROKEN __INTEL_LLVM_COMPILER == 20250000 && __TBB_GLIBCXX_VERSION == 110000 && __TBB_CPP20_PRESENT
```
dnmokhov, line 21: Should we be only testing for the major Intel LLVM compiler version instead of "20250000" (i.e., 2025.0.0)?

aleksei-fedotov: I think better to have workarounds as picky as possible. This way we won't miss other issues about deprecated stuff that might turn out as real issues in the library's code.

aleksei-fedotov (approved):
Looks good enough to me.

PR #1629, opened 2025-02-05
Add CI workflow for testing documentation examples

kboyarinov:
### Description 

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown


.github/workflows/ci.yml
```diff
@@ -366,3 +366,36 @@ jobs:
...
+            -DTBB_TEST=OFF ..
+          cmake --build . -v --target tbb --target tbbmalloc --target tbbmalloc_proxy
+      - name: Test doc examples
+        run: |
+          source `find ./build -name vars.sh`
+          cd doc/main/examples_testing
+          mkdir build && cd build
+          cmake -DCMAKE_CXX_STANDARD=${{ matrix.std }} -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} \
+            -DCMAKE_CXX_COMPILER=${{ matrix.cxx_compiler }} -DCMAKE_C_COMPILER=${{ matrix.c_compiler }} ..
+          cmake --build . -v --target build-doc-examples
```
aleksei-fedotov, line 400: If we require CMake of version 3.12 at least, we will be able to build in parallel. 
```suggestion
          cmake --build . --parallel -v --target build-doc-examples
```

I suppose `ubuntu-latest` is the latest available version of Ubuntu Linux. Even if it is the LTS release, which is currently 24.4, `cmake --version` reports "cmake version 3.28.3" on it. So, I think it is more than enough to be able to build in parallel.

aleksei-fedotov (commented):
Finally, we are going to have this step automated! Hooray!

.github/workflows/ci.yml
isaevil: Well, it is Linux. We can just call `VERBOSE=1 make -j$(nproc) build-doc-examples`

doc/main/examples_testing/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+
+macro(add_example example_path)
+    message(STATUS "Adding ${_example_path}")
+
+    get_filename_component(_example_name "${_example_path}" NAME_WE)
+    add_executable(${_example_name} EXCLUDE_FROM_ALL "${_example_path}")
+
+    target_include_directories(${_example_name} PRIVATE ${_tbb_include_path})
+    target_link_libraries(${_example_name} tbb)
+    target_link_libraries(${_example_name} tbbmalloc)
```
isaevil, line 33: ?
```suggestion
    target_link_libraries(${_example_name} tbb tbbmalloc)
```

.github/workflows/ci.yml
```diff
@@ -366,3 +366,36 @@ jobs:
...
+            job_name: linux_doc_examples_testing_ubuntu_latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Build library
+        run: |
+          mkdir build && cd build
+          cmake -DCMAKE_CXX_STANDARD=${{ matrix.std }} -DCMAKE_BUILD_TYPE=${{ matrix.build_type}} \
+            -DCMAKE_CXX_COMPILER=${{ matrix.cxx_compiler }} -DCMAKE_C_COMPILER=${{ matrix.c_compiler }} \
+            -DTBB_TEST=OFF ..
+          cmake --build . -v --target tbb --target tbbmalloc --target tbbmalloc_proxy
```
isaevil, line 392: You already excluded tests. Just build libraries that are lest :)
```suggestion
          VERBOSE=1 make -j$(nproc)
```

doc/main/examples_testing/CMakeLists.txt
isaevil, line 1: Do you find it useful to be able to build these examples right from root like tests or TBB examples?

doc/main/examples_testing/CMakeLists.txt
```diff
@@ -0,0 +1,50 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+cmake_minimum_required(VERSION 3.11)
+project(doc_examples_testing)
+
+set(_tbb_include_path "${CMAKE_CURRENT_SOURCE_DIR}/../../../include")
```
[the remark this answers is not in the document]
kboyarinov: I thought about this, but the idea was to explicitly specify that these examples should be built with the same TBB that was checked-out with the examples. As far as I remember, `find_package(tbb)` also considers other places, like libraries installed on the system, right? 
Or there is a way to prohibit that?

doc/main/examples_testing/CMakeLists.txt
kboyarinov: Yes, it is a good point to consider. I just did not remember where do we ship our CMakes and be sure that absence of the documentation sources would not break something. But I think we can consider adding it as a "private" optional testing.

.github/workflows/ci.yml
aleksei-fedotov: I would prefer writing the code that has fewer platform-specific dependencies. So, if it is possible, it is better to use `cmake` build command. Especially, if it can be then easily adjusted on the other platforms.

aleksei-fedotov: And why make it deliberately even more platform-specific? Especially, when likely there is a need running this in other platforms.

I guess we have expressed our viewpoints. Let's then just the author of the patch decide on this.

doc/main/examples_testing/CMakeLists.txt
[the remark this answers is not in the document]
kboyarinov: This line is removed since the CMake was moved under the top-level CMake

.github/workflows/ci.yml
kboyarinov: As discussed above, decided to use `cmake --build` for both Windows and Linux

doc/main/examples_testing/CMakeLists.txt
kboyarinov: Fixed

doc/main/examples_testing/CMakeLists.txt
kboyarinov: Moved

CMakeLists.txt
```diff
@@ -330,6 +330,17 @@ if (TBB_EXAMPLES)
     add_subdirectory(examples)
 endif()
 
+
+if (TBB_DOC_EXAMPLES)
+    set(TBB_DOC_EXAMPLES_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}/doc/main/examples_testing")
+    if (NOT EXISTS ${TBB_DOC_EXAMPLES_DIRECTORY})
+        message(FATAL_ERROR "Documentation examples are not found while testing was enabled")
```
isaevil, line 337: Do we really want to fail here? Maybe the warning will be enough?

doc/main/examples_testing/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
...
+
+set(_reference_examples_path "${CMAKE_CURRENT_SOURCE_DIR}/../reference/examples")
+set(_userguide_examples_path "${CMAKE_CURRENT_SOURCE_DIR}/../tbb_userguide/examples")
+
+add_custom_target(build-doc-examples
+    COMMENT "Build oneTBB documentation samples")
+set(doc_examples_test_label "doc-examples")
+
+macro(add_doc_example _doc_example_path)
+    message(STATUS "Adding ${_doc_example_path}")
```
isaevil, line 26: I would prefer not to output each added test during configuration. Maybe we should just output once that "Enabled testing of samples from documentation"?

doc/main/examples_testing/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
...
+file(GLOB_RECURSE REFERENCE_EXAMPLES_LIST "${_reference_examples_path}/*.cpp")
+file(GLOB_RECURSE USERGUIDE_EXAMPLES_LIST "${_userguide_examples_path}/*.cpp")
+
+foreach(_doc_example_path IN LISTS REFERENCE_EXAMPLES_LIST)
+    add_doc_example(${_doc_example_path})
+endforeach()
+
+foreach(_doc_example_path IN LISTS USERGUIDE_EXAMPLES_LIST)
+    add_doc_example(${_doc_example_path})
+endforeach()
```
isaevil, line 48: I think you should be able to specify multiple expressions in the same `file(...)` call. That will allow you to iterate only once instead of twice.

CMakeLists.txt
kboyarinov: The idea actually was to prevent some of our automatic workloads to silently fail if the `examples_testing` directory will be missed of moved somewhere. I believe in that case `ctest` will just run nothing and return the successful status, but no testing would be done.

doc/main/examples_testing/CMakeLists.txt
kboyarinov: Done

isaevil (approved):
Github Actions and CMake side looks good to me.

aepanchi (approved):

PR #1630, opened 2025-02-05
Suppress Intel Compiler debug iterator vectorization warning in test_openmp.cpp

dnmokhov:
### Description 
Intel LLVM compiler triggers a warning when using OpenMP in debug mode on Windows:
```
using /MTd or /MDd with '#pragma omp simd' may lead to unexpected fails due to the debugging version of iterators that cannot be vectorized correctly [-Wdebug-option-simd]
```
This happens even when vectorization is turned off with `/Qopenmp-simd-`. In fact, the only way not to trigger the warning seems to be explicitly disabling it. Since `test_openmp.cpp` does not use vectorization, the warning should not be triggered and this can be considered a compiler issue.

### Type of change

### Tests
- [x] not needed

### Documentation
- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -535,6 +535,8 @@ if (TARGET TBB::tbb)
 
         set_target_properties(test_openmp PROPERTIES COMPILE_FLAGS ${TBB_OPENMP_FLAG})
 
+        # Intel LLVM compiler triggers a warning when using OpenMP in debug mode on Windows.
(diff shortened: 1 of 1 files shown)


sarathnandu (approved):
LGTM

kboyarinov (approved):
LGTM

PR #1639, opened 2025-02-18
Reset stealing loop backoff counter when leaving block time back to stealing loop

isaevil:
### Description

Reset of stealing loop backoff is needed to lower a probability of a worker thread, leaving "block time" idle loop back to stealing loop (someone set arena's pool state to SET), to miss work coming as a result of work propagation. If not reset, it is possible that worker thread might not find any work and instead of trying to pause again it will immediately try to set arena pool state to UNSET.

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown


src/tbb/waiters.h
```diff
@@ -73,6 +73,9 @@ class outermost_worker_waiter : public waiter_base {
                     t2 = std::chrono::steady_clock::now())
                 {
                     if (!my_arena.is_empty() && !my_arena.is_recall_requested()) {
+                        // We are going back to stealing loop, thus in order to not set arena EMPTY instantly
+                        // in case when worker failed to find work we need to reset stealing loop backoff counter
```
aleksei-fedotov: Probably, rephrase to sound clearer:

```suggestion
                        // Before thread returns to a stealing loop, we need to reset its 
                        // waiting counters so that it starts work search from the very
                        // beginning. Otherwise, thread might fast reset arena's state
                        // to EMPTY again.
```

PR #1644, opened 2025-02-20
Add LTO visibility attribute for wait_context_vertex

kboyarinov:
### Description 

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown

### DIFF
--- a/include/oneapi/tbb/detail/_config.h
+++ b/include/oneapi/tbb/detail/_config.h
@@ -332,6 +332,12 @@
     #define __TBB_nodiscard
 #endif
 
+#if __clang__
+    #define __TBB_lto_visibility_public [[clang::lto_visibility_public]]
+#else
+    #define __TBB_lto_visibility_public
+#endif
+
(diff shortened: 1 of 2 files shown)


solbjorn:
Pinging just in case.

I've been using changes from this PR for 1.5 months already with no issues observed (with `-fwhole-program-vtables` enabled for both oneTBB and my app).

include/oneapi/tbb/detail/_config.h
```diff
@@ -332,6 +332,12 @@
     #define __TBB_nodiscard
 #endif
 
+#if __clang__
+    #define __TBB_lto_visibility_public [[clang::lto_visibility_public]]
+#else
+    #define __TBB_lto_visibility_public
+#endif
+
```
akukanov, line 340: I think we need to give the macro a better, semantical name, something similar to `__TBB_EXPORTED_FUNC`. The property of a type that we want to indicate and apply as needed is that it's a "callback interface" that is declared in the library headers but used in the binaries. We have other such types, for example `delegate_base`, and so it's likely only the matter of time for similar issues with whole program optimization options to be seen for these types as well.

PR #1648, opened 2025-02-21
Added RFC for Flow Graph Helper Functions

vossmjp:
### Description 
This adds the description of an already existing experimental feature to the rfcs/experimental directory. The content should be reviewed for accuracy in describing the feature and for the proposed exit criteria.

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [X] No
- [ ] Unknown


rfcs/experimental/flow_graph_node_sets/README.md
```diff
@@ -0,0 +1,256 @@
...
+  broadcast_node<int> input(precedes(n1, n2, n3));
+```
+
+## Exit Criteria
+
+The following conditions need to be met to move the feature from experimental to fully supported:
+* Collecting feedback on user experience confirming the choices made on the open questions and limitations:
+  * Limitation that constructors can be used to set predecessors or successors but not both.
+  * The multiport node rule that makes edges from each node in the set to the corresponding port.
+* The corresponding oneTBB specification update should be done backed by the user feedback provided.
```
akukanov, line 256: With regard to the specification update, I would explicitly ask the following questions:
- Should we keep the "node set" type unspecified or define it properly?
- Should the ordering (preceding/following) become the property of a node set?

rfcs/experimental/flow_graph_node_sets/README.md
```diff
@@ -0,0 +1,256 @@
...
+
+When run, the above code results in:
+
+```
+  1:100
+  3:400
+  2:200
+```
+
+The multi-port nodes include `multifunction_node`, `join_node`, `split_node` and `indexer_node`.
```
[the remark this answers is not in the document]
kboyarinov: Actially, the `composite_node` is an interesting aspect of this proposal. If I remember correct, currently it does not support `follows` and `precedes` API, but it may be useful to have such an ability for user-provided nodes.
I think we need to investigate on how it can be expressed on the node implementor's side - I think it may result in making the node set public.

rfcs/experimental/flow_graph_node_sets/README.md
```diff
@@ -0,0 +1,256 @@
...
+The rule for creating edges from sets is the same for each of these types.
+
+### `follows` and `precedes` Helper Functions
```
kboyarinov, line 214: One important thing is missed from the RFC in regard to helper functions - deduction guides. 
Current preview feature provides additional extensions for Flow Graph deduction guides while `follows` and `precedes` APIs are used. Actually, only them allows to use deduction guides for nodes without the body.
E.g. `join_node j(follows(f1, f2, f2));` deduces the node as `join_node<std::tuple<o1, o2, o3>>` where `oN` is the output type for the corresponding `fN` node. 

But as far as I remember, one of the issues was with the `multifunction_node` and `async_node` - as far as I see they does not support CTAD now even if the simplified API is used. I don't remember much details, but one of the reasons was the fact that to create a multifunction_node you need to define a body that takes `node_type::output_ports_type` as an argument that requires knowledge about the `node_type` at it was not yet deducted.
May be we can define some deduction guides for the case when the generic lambda and `precedes` are provided.
Anyway, we need to investigate it somehow.

rfcs/experimental/flow_graph_node_sets/README.md
```diff
@@ -0,0 +1,256 @@
...
+
+### Additional Constructors for Flow Graph nodes
+
+Each flow graph node has been extended with two new constructors, one that receives a *follows* 
+set of nodes `[1]` and another that receives a *precedes* set of nodes `[2]`. These node sets are 
+created using the `precedes` and `follows` helper functions described in the next section.
+
+The object returned by follows or precedes replaces the graph argument, which in other constructors is passed
+as the first argument. The graph argument for the node being constructed is obtained either from the specified
+node set or the sequence of nodes passed to follows or precedes.
```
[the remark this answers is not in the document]
kboyarinov: As far as I remember, we agreed that only the nodes that belongs to the same graph can be wrapped into the node set. I think it should be states as undefined behavior somewhere (and in the future specification for sure).

rfcs/experimental/flow_graph_node_sets/README.md
```diff
@@ -0,0 +1,256 @@
...
+In this proposal, `node_set` is an exposition-only name for the type returned from the
+`make_node_set` function. The `make_node_set` function template creates a set of nodes that
+can be passed as arguments to the `make_edges`, `follows` and `precedes` functions that are
+described in later sections.
+
+```cpp
+    // node_set is an exposition-only name for the type returned from make_node_set function
+
+    template <typename Node, typename... Nodes>
+    /*unspecified*/ make_node_set( Node& node, Nodes&... nodes );
```
kboyarinov, line 63: As far as I see, current `make_edge` function in the general case only allows setting edges between sender and receiver of the same type. Because of this, it seems useless to allow `make_node_set` to accept nodes with different `input_type` or `output_type`. I think we can think about adding reasonable constraints APIs mentioned in this paper.

rfcs/experimental/flow_graph_node_sets/README.md
```diff
@@ -0,0 +1,256 @@
+# Helper Functions for Expressing Flow Graphs
+
+## Introduction
+
+The flow graph API that lacks an simple way to create multiple edges with a single call. The result of this
+limitation is verbose, less-readable code.  For example, the follow pciture shows a flow graph where the
+node `input` has three successors.
+
+<img src="graph_with_many_edges.png" width=500>
```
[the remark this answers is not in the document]
vossmjp: I will remove explicit widths.

rfcs/experimental/flow_graph_node_sets/README.md
vossmjp: Will add.

rfcs/experimental/flow_graph_node_sets/README.md
```diff
@@ -0,0 +1,256 @@
...
+    /*unspecified*/ make_node_set( Node& node, Nodes&... nodes );
+```
+
+### The `make_edges` function template
+
+The make_edges function template creates edges between a single node and each node in a set of nodes.
+
+There are two ways to connect nodes in a set and a single node using make_edges:
+
+<img src="make_edges.png" width=600>
```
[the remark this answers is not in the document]
vossmjp: Will remove

rfcs/experimental/flow_graph_node_sets/README.md
```diff
@@ -0,0 +1,256 @@
...
+[Developer Guide](https://www.intel.com/content/www/us/en/docs/onetbb/developer-guide-api-reference/2022-0/helper-functions-for-expressing-graphs.html)
+
+There are four main parts to the extension:
+
+- make_node_set function template
+- make_edges function template
+- two additional constructors for each flow graph node
+- follows and precedes function templates
+
+### Node Sets
```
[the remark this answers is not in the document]
vossmjp: Will add as an open question

rfcs/experimental/flow_graph_node_sets/README.md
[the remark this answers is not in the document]
vossmjp: I will add a comment about undefined behavior.

rfcs/experimental/flow_graph_node_sets/README.md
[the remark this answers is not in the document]
vossmjp: Will note in Exit Criteria

rfcs/experimental/flow_graph_node_sets/README.md
vossmjp: Will add to exit criteria

rfcs/experimental/flow_graph_node_sets/README.md
vossmjp: For multioutput or multi port nodes, the set might contain different types at each index, as long as they match

rfcs/experimental/flow_graph_node_sets/README.md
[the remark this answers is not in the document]
vossmjp: Done

rfcs/experimental/flow_graph_node_sets/README.md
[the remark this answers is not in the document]
vossmjp: Done

rfcs/experimental/flow_graph_node_sets/README.md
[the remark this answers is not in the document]
vossmjp: Added

rfcs/experimental/flow_graph_node_sets/README.md
vossmjp: Added

rfcs/experimental/flow_graph_node_sets/README.md
vossmjp: Added

rfcs/experimental/flow_graph_node_sets/README.md
```diff
@@ -0,0 +1,262 @@
...
+```cpp
+    // node_set is an exposition-only name for the type returned from make_node_set function
+
+    template <typename Node, typename... Nodes>
+    /*unspecified*/ make_node_set( Node& node, Nodes&... nodes );
+```
+
+### The `make_edges` function template
+
+The make_edges function template creates edges between a single node and each node in a set of nodes.
```
akukanov, line 69: Unlike other APIs in this proposal, for `make_edges` no function signature is provided.

rfcs/experimental/flow_graph_node_sets/README.md
```diff
@@ -0,0 +1,262 @@
...
+
+    broadcast_node input(precedes(doubler, squarer, cuber));
+```
+
+## Proposal
+
+Several helper functions and additional constructors have been added as experimental features.
+These additions simplify expressing dependencies between nodes when building oneTBB flow graphs.
+This experimental feature is described in more detail in the
+[Developer Guide](https://www.intel.com/content/www/us/en/docs/onetbb/developer-guide-api-reference/2022-0/helper-functions-for-expressing-graphs.html)
```
akukanov, line 43: Technically, it's in the API Reference document, not in the Developer Guide.
Also, would it be better to refer to github.io pages?https://uxlfoundation.github.io/oneTBB/main/reference/helpers_for_expressing_graphs.html

rfcs/experimental/flow_graph_node_sets/README.md
```diff
@@ -0,0 +1,262 @@
...
+
+The order of the arguments determines the predecessor / successor relationship. The node or nodes represented
+by the left argument becomes the predecessor or predecessors of the node or nodes represented by the right
+argument.
+
+### Additional Constructors for Flow Graph nodes
+
+Each flow graph node has been extended with two new constructors, one that receives a *follows* 
+set of nodes `[1]` and another that receives a *precedes* set of nodes `[2]`. These node sets are 
+created using the `precedes` and `follows` helper functions described in the next section.
```
akukanov, line 83: There is some semantical/naming nuance that may need to be explained. The names `follows` and `precedes` do not describe the node set of their arguments, referring instead to a to-be-constructed node and describing its relation to the node set. In other words, `precedes` creates a set of "following" nodes (successors). and `follows` creates a set of "preceding" nodes (predecessors).

It might be yet another area where some feedback would be helpful. If the names do not make the usage semantics clear, the goal of simplifying graph construction will not be achieved in full.

rfcs/experimental/flow_graph_node_sets/README.md
```diff
@@ -0,0 +1,262 @@
...
+
+The object returned by follows or precedes replaces the graph argument, which in other constructors is passed
+as the first argument. The graph argument for the node being constructed is obtained either from the specified
+node set or the sequence of nodes passed to follows or precedes.
```
akukanov, line 87:
```suggestion
The object returned by `follows` or `precedes` replaces the graph argument, which in other constructors is passed
as the first argument. The graph argument for the node being constructed is obtained either from the specified
node set or the sequence of nodes passed to `follows` or `precedes`.
```

rfcs/experimental/flow_graph_node_sets/README.md
vossmjp: Will fix.

rfcs/experimental/flow_graph_node_sets/README.md
vossmjp: Will add.

rfcs/experimental/flow_graph_node_sets/README.md
vossmjp: Will fix as part of rewrite.

rfcs/experimental/flow_graph_node_sets/README.md
```diff
@@ -0,0 +1,564 @@
...
+}
+```
+
+The arguments to the `precedes` functions (eventually) become successors to a new node, and therefore the arguments
+ to `precedes` are named `successors` and the `node_set` is constructed with an `order::preceding` order type.
+ 
+ Is it confusing that a `preceding` node set contains nodes that are (eventually) successors, not predecessors?
+ And that a `following` node set contains nodes that are (eventually) predecessors, not successors?
+ If we define the `node_set` type in the specification instead of leaving it as an exposition-only name, does that
+ increase potential confusion since this allows the details of `node_set` to be viewed in isolation.
```
aleksei-fedotov, line 179: I think the power of `follows` and `precedes` helpers is in the way how they express relationships between graph nodes. However, this is only achieved when they are used right inside the node constructor.

```cpp
my_pipeline_link = make_node_set(pedestrian_detector, lane_departure_warner, obstacle_detector);
input_node<frame> input(precedes(my_pipeline_link)); // input node precedes a node set 
join_node jn(follows(my_pipeline_link));  // join node follows a node set
```

While the following constructions are less expressive to me:
```cpp
// It is unclear what set of nodes precedes
my_node_set1 = precedes(pedestrian_detector, lane_departure_warner, obstacle_detector);
// ...
input_node<frame> source(my_node_set1); // makes little sense without 'precedes'

my_node_set2 = follows(pedestrian_detector, lane_departure_warner, obstacle_detector);
// ...
// Okay, consuming `my_node_set2` somewhat indicates that `fn` accepts the output from various sources,
// basing its compute on those. However, it is not as clear as with 'follows'.
function_node fn(my_node_set2, [](frame const& f) {/*...*/});
```
I wonder if the usage of `follows` and `precedes` could be restricted to usages only within constructors of nodes. Is there such a flexibility in standard C++?

vossmjp: The function is not called in the constructor itself, but is used to create a set that is passed as an argument to the constructor.

aleksei-fedotov: That's the question - do we want to have such an API that when it is used outside the node's constructor call (see my second example) confuses readers wrt what it follows or precedes? Is there a way to disallow such usage with possible renaming, changing semantics, etc.?

akukanov: Whatever is passed into a constructor is an expression that has a certain type and evaluates to a certain value. So I cannot see how we can make `auto x = flow::precedes(a, b, c);` invalid.

vossmjp: I think this is captured as an open question in the exit criteria already. The question being, "Should `node_set` be defined or left as an exposition-only type? If `node_set` is defined, are the semantics clear or is the naming of the order type potentially confusing?".  Since this RFC is for an already release experimental feature, we can discuss design changes but that should not gate acceptance of this description. Should I change the exit criteria question?  It seems ok to me.

kboyarinov (approved):
LGTM

PR #1662, opened 2025-02-27
Consider task_arena and task_group combination in detail

akukanov:
### Description 
Propose specific API for co-usage of `task_arena` and `task_group`, the "part 1" of possible `task_arena` waiting extensions outlined in https://github.com/uxlfoundation/oneTBB/tree/master/rfcs/proposed/task_arena_waiting

### DIFF
--- a/rfcs/proposed/task_arena_waiting/readme.md
+++ b/rfcs/proposed/task_arena_waiting/readme.md
@@ -106,18 +106,12 @@ The proposal consists of ideas that are not mutually exclusive and can be implem
 
 ### 1. Simplify the use of a task group
 
-To address the existing shortcomings of the `task_arena` and `task_group` combination, we could add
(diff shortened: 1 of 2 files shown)


akukanov:
While the question of work isolation is not specific to the proposed functionality and can be considered independently, I would like at least to get opinions and preferences on the proposed approaches, because these might impact the API design.

akukanov:
> Although, we need to make clear separation between this isolated tag and `task_group_context::isolated` as the term becomes sort of overloaded.

Or maybe we should add the work isolation semantics to this existing kind of the `task_group_context`.

Thanks for the opinion Aleksei. Hopefully others will tell theirs as well.

kboyarinov (approved):

aleksei-fedotov (approved):
Looks good to me.

vossmjp (approved):
lgtm

PR #1664, opened 2025-03-03
R2 RFC for task_group dynamic dependencies

kboyarinov:
### Description 
Add RFC describing the semantics for concrete task_group dynamic dependencies APIs

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown


rfcs/proposed/task_group_dynamic_dependencies/task_handle_semantics.md
```diff
@@ -0,0 +1,364 @@
...
+};
+
+class task_arena {
+    void enqueue(task_handle&& h); // existing overload
+    void enqueue(task_handle& h);  // new overload
+};
+
+namespace this_task_arena {
+    void enqueue(task_handle&& h); // existing overload
+    void enqueue(task_handle& h);  // new overload
```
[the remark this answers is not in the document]
akukanov: With the other proposed extension, `enqueue` will support not only "fire-and-forget" but the more general asynchronous execution semantics as well. Therefore I do not see much contradiction here.

rfcs/proposed/task_group_dynamic_dependencies/task_handle_semantics.md
```diff
@@ -0,0 +1,364 @@
...
+
+The proposal is to
+* Extend possible task states that can be handled by non-empty ``task_handle`` to be any tasks in any state
+* Allow to use non-empty ``task_handle`` representing a task in any state for setting and transferring the dependencies between tasks
+
+Since current API allows submitting the ``task_handle`` for execution only as rvalues, having any usage of ``task_handle`` object after submitting for execution 
+(e.g. using ``task_group::run(std::move(task_handle))``) looks misleading even if some guarantees are provided for the referred handle object. It also creates an error-prone
+for TBB developers - any move-construct or move-assign from the accepted handle will break the guarantee. Code analyzers?
+
+To handle this, the proposal is to extend all functions that take the task handled by the ``task_handle`` with the new overload taking an non-const lvalue reference and provide the following guarantees:
```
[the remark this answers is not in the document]
kboyarinov: It seems like the required semantics for ownership is somewhere in the middle. It seems like we need "unique" ownership to submit the task and "weak" ownership to track the task progress and set the dependencies. And the option may be to define 2 handlers of the task:
* Current `task_handle` that owns the task and can be used to make any action on it - submit for execution or add dependencies. 
* New handle (let's say `task_view`) that does not own the task, but can be used to track progress or add dependencies. 

`task_handle` would be almost unchanged. `task_view` would be copyable and any copy would be able make actions on the underlying task that is owned either by the `task_handle` or the TBB scheduler. 

`task_view` can be constructible from `task_handle` but not the opposite. 

Submitting functions (`task_group::run` and others) would also remain unchanged and allow only the `task_handle` argument. 

Functions for adding dependencies would provide overloads taking both `task_handle` and `task_view` where the task in any state can be taken, and only the `task_handle` where only the task in created state in allowed (like the successor side of `make_edge`).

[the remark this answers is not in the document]
akukanov: I have come to essentially the same conclusion. Even though a new class is an API complication, it seems that trying to extend `task_handle` and adjust all the related APIs would complicate everything even more.

I would not use `task_view` for the name, as it is not really a view to a task. Rather, it's a `weak_task_handle`.

rfcs/proposed/task_group_dynamic_dependencies/task_handle_semantics.md
```diff
@@ -0,0 +1,364 @@
...
+
+<img src="transferring_between_two_separate_tracking.png" width=800>
+
+After the transfer, the successors of ``current`` and ``target`` are tracked separately and adding new successors to one of them would only
+affect the successors list of one task:
+
+<img src="transferring_between_two_separate_tracking_new_successors.png" width=800>
+
+Such an approach can be beneficial if ``current`` task is kind of generator task that collects the set of successors on each iteration of the loop
+and then transfers it to the newly created task.
```
vossmjp, line 191: Can you provide a sketch of such a generator example that shows where the loop would be, and how additional successors would be added to the current task from an outside task or thread. I the more common use case would be recursive decomposition, such as the merge sort example included in the umbrella RFC, where a task body refines its work into a subgraph. So an initial node N becomes N -> N1. N1, when executed, becomes N1 -> N2, etc. When viewed from outside the task, the original N represents a complete piece of work and so outside tasks or threads would want to make successors of the complete work represented by the N's new subgraph. So an outside task, might want to add N->M.  Or, if N is in some state of refinement N->N1->N2->M.  In those cases, I can't think of a reason why an outside task or thread would want to insert an additional successor of N ignoring its refinement into a subgraph, such as N->{M,N1->N2}.  Maybe your generator case is such as counter-example, but I'm not sure I can immediately imagine practical generator scenarios. Do you have some in mind?

rfcs/proposed/task_group_dynamic_dependencies/task_handle_semantics.md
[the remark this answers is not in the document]
kboyarinov: With the latest approach, no changes in the existing `enqueue` are required, but the progress can be tracked using the `task_tracker` tracking the task.

rfcs/proposed/task_group_dynamic_dependencies/task_handle_semantics.md
```diff
@@ -0,0 +1,364 @@
...
+## Introduction
+
+This document contains a concrete API and semantics proposal for ``task_group`` extensions defined in the parent RFC. 
+The following cases should be covered and be described:
+* ``task_handle`` semantic extension to allow handling the ``task_group`` tasks in various states: created, submitted, executing and completed. 
+  Existing API only allows having a non-empty ``task_handle`` to handle the created task and an empty one that does not handle any task.
+    * Semantics for ``task_group::run``, ``task_group::run_and_wait``, ``task_arena::enqueue`` and ``this_task_arena::enqueue`` should be defined
+      when the input ``task_handle`` is handling a task in various states.
+    * Semantics for returning a ``task_handle`` handling a task in various states from the body of the task should be described (while using the existing
+      preview extensions for ``task_group``).
```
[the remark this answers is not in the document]
kboyarinov: Addressed by introducing the `task_tracker`.

rfcs/proposed/task_group_dynamic_dependencies/task_handle_semantics.md
```diff
@@ -0,0 +1,364 @@
...
+
+Once the thread finishes executing the task body, the state of the task is changed to `completed`.
+
+The proposal is to
+* Extend possible task states that can be handled by non-empty ``task_handle`` to be any tasks in any state
+* Allow to use non-empty ``task_handle`` representing a task in any state for setting and transferring the dependencies between tasks
+
+Since current API allows submitting the ``task_handle`` for execution only as rvalues, having any usage of ``task_handle`` object after submitting for execution 
+(e.g. using ``task_group::run(std::move(task_handle))``) looks misleading even if some guarantees are provided for the referred handle object. It also creates an error-prone
+for TBB developers - any move-construct or move-assign from the accepted handle will break the guarantee. Code analyzers?
```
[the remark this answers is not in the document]
kboyarinov: I meant it was easy to make a mistake and break the guarantees while having the non-const rvalue reference in the argument to `run`.  This  issue should be addressed by introducing the `task_tracker`.

rfcs/proposed/task_group_dynamic_dependencies/task_handle_semantics.md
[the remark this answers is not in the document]
kboyarinov: `task_tracker` with the semantics stated above was introduced.

rfcs/proposed/task_group_dynamic_dependencies/task_handle_semantics.md
```diff
@@ -0,0 +1,364 @@
...
+```cpp
+class task_handle {
+    bool was_submitted() const;
+    bool is_completed() const;
+};
+```
+
+Having ``th.was_submitted()`` equal to ``true``, means the task state was changed from `created` to `submitted` in the past. The task can be either submitted, executing or completed.
+Extra care must be taken while working with this method anyway since even if it returns ``false``, it may be unsafe to submit the ``task_handle`` since the state 
+can be changed by the other thread.
```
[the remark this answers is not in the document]
kboyarinov: Addressed by introducing the `task_tracker`

rfcs/proposed/task_group_dynamic_dependencies/task_handle_semantics.md
kboyarinov: I have thought once more about that and seems like the generator example I had in mind is not realistic. Initially it was something like:

```
tbb::task_group tg;
tbb::task_tracker tracker;

tbb::task_handle handle = tg.defer([] {
    while (exit_condition) {
        tbb::task_handle new_task = tg.defer(...);
        tbb::task_group::current_task::transfer_successors_to(new_task);
        tg.run(new_task);
    }
});

tracker = handle;

tg.run(std::move(handle));

// Adding successors to tracker for each iteration of the loop
```

But the loop inside of the task should wait somehow for the new successors to be added which seems inefficient. 
I think now it looks more like finding a problem for the solution:) 
I have removed this paragraph from the RFC for now.

rfcs/proposed/task_group_dynamic_dependencies/task_handle_semantics.md
```diff
@@ -0,0 +1,445 @@
...
+The second limitation is handled by limiting the successor argument to be only ``task_handle``.
+
+Lets consider the different states of the task tracked or handled by ``predecessor``. 
+
+If the predecessor task is in any state except the completed one (created/scheduled/running), the API registers the successor task
+in the list of successors on the predecessor side and increase the corresponding reference counter on the successor side to ensure it
+would not be executed before the predecessor task. The successor task can only start executing once the associated reference counter is equal to 0.
+
+If the predecessor task is in `completed` state, the API has no effect in terms of modifying the list of successors and reference counters since no additional
+dependencies required and the successor task can be executed if all other dependent tasks are executed as well.
```
vossmjp, line 91:
```suggestion
dependencies are required and the successor task can be executed if all other dependent tasks are executed as well.
```

rfcs/proposed/task_group_dynamic_dependencies/task_handle_semantics.md
```diff
@@ -0,0 +1,445 @@
...
+if the body of the predecessor task did not return other task that should be bypassed.
+
+This implementation approach is illustrated in the picture below:
+
+<img src="predecessor_successor_implementation.png" width=800>
+
+### Adding successors to the current task
+
+Consider use-case of parallel wavefront pattern on the 2-d grid. Each cell is computed as part of a separate task in ``task_group``. Each cell task computes
+itself and creates more tasks to process the cell below and the cell on the right.
```
vossmjp, line 116: Usually the executing task does not create the other tasks in a wavefront, since there are typically multiple predecessors. For example, the task to the right of a task A, let's call it C, may also be the task below B. More typically, wavefront tasks are pre-allocated and dependencies set from outside of the tasks themselves. Otherwise, who is responsible for creating C? A or B?  What if A creates C, adds itself as a predecessor, and completes before B executes? Would there be an edge from B to C?

rfcs/proposed/task_group_dynamic_dependencies/task_handle_semantics.md
```diff
@@ -0,0 +1,445 @@
...
+In this case, after the transfer, the task ``current`` will have an empty successors list, and ``target`` will have a merged successors list:
+
+<img src="transferring_between_two_separate_tracking.png" width=800>
+
+After the transfer, the successors of ``current`` and ``target`` are tracked separately and adding new successors to one of them would only
+affect the successors list of one task:
+
+<img src="transferring_between_two_separate_tracking_new_successors.png" width=800>
+
+Alternative approach is to keep tracking ``current`` and ``target`` together after transferring. This requires introducing the new state of task - a `proxy` state.
```
vossmjp, line 164: I can think of examples where combining tracking is useful. I think we need a counter example, where separating is useful.  Otherwise, the choice is obvious.

rfcs/proposed/task_group_dynamic_dependencies/task_handle_semantics.md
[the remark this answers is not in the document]
vossmjp: I also don't like having another class, but I agree that it simplifies thinking about the problem, so while complicating the API is makes it easier to reason about.

rfcs/proposed/task_group_dynamic_dependencies/task_handle_semantics.md
```diff
@@ -0,0 +1,445 @@
...
+It is safe to add multiple predecessors to the same successor and add the same predecessor for multiple successor tasks.
+
+It is safe to add successors to the task that currently transfers it's successors to another task and
+to the task to which the successors are transferred.
+
+### Member functions of ``task_group::current_task``
+
+``static void add_successor(task_handle& succ)``
+
+Resisters currently executing task to be a predecessor that must complete before the task handled by ``succ`` can start executing.
```
aleksei-fedotov, line 385: A typo:
```suggestion
Registers currently executing task to be a predecessor that must complete before the task handled by ``succ`` can start executing.
```

rfcs/proposed/task_group_dynamic_dependencies/task_handle_semantics.md
```diff
@@ -0,0 +1,445 @@
...
+
+The exact wording for the semantics of this method should be defined after making a decision about merged or separate tracking of tasks,
+that was described above.
+
+## Alternative approaches 
+
+The alternative approaches are to keep only the ``task_handle`` as the only was to track the task, set the
+dependencies and submit the task for execution.
+
+### ``task_handle`` as a unique owner
```
aleksei-fedotov, line 406: It seems that having a `task_handle` as oneTBB has it now should be enough to cover all mentioned use cases with a few extensions to it that will allow adding dependencies between valid `task_handle`s. I am not sure we need to answer right now the question about setting already moved `task_handle` as a predecessor. We can just prohibit this behavior in the first iteration of this extension for task dynamic dependencies until we find the use cases and users who will benefit from it.
And even if we prove the necessity in such support we may just relax the wording about moved `task_handle` by saying that the move operation only affects that the task cannot be submitted multiple times, but still allow setting it as a predecessor for another tasks. I believe this will not be confusing to users because "the edges" between the tasks represent the weak thing by itself. Meaning that moving them with the body of the task or leaving them with the `task_handle` after the move does not make any influence on the execution of the tasks.

aleksei-fedotov (commented):
I think we are overcomplicating things without proofs we actually need to support all the described behaviors.

rfcs/proposed/task_group_dynamic_dependencies/task_handle_semantics.md
```diff
@@ -0,0 +1,445 @@
...
+``is_completed`` method returns ``true`` if the tracked task is completed. This state cannot be changed in the
+future.
+
+## Semantics for setting dependencies between tasks
+
+Lets consider creating a predecessor-successor dependency between the task tracked or owned by ``predecessor``
+and the task handled by ``successor_task_handle`` -
+``task_group::make_edge(predecessor_task_handle, successor_task_handle)``. 
+
+As it was stated in the parent RFC document, we would like to allow adding predecessors in any state described above and to limit the successor to be a task in created state since it can be too late to add predecessors to
```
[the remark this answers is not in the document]
vossmjp: But if the user doesn't know that there are other predecessors in the created state then adding a predecessor to a submitted task results in a race, where it may or may not become a predecessor quickly enough. Is such a race desirable?  I understand the case you've described, but I'm not sure we should allow it. Whichever way we decide to go, this should be a an open question that we want answered if its released as a preview.

rfcs/proposed/task_group_dynamic_dependencies/task_handle_semantics.md
vossmjp: We have found an example of a divide-and-conquer wavefront where edges are made from within an executing task and also from external tasks, so my previous comment is moot.

rfcs/proposed/task_group_dynamic_dependencies/task_handle_semantics.md
```diff
@@ -0,0 +1,445 @@
...
+```cpp
+class task_tracker {
+public:
+    bool was_submitted() const;
+    bool is_completed() const;
+};
+```
+
+Having ``task_tracker.was_submitted()`` equal to ``true`` means the tracked task state was changed from created
+to submitted in the past. The task can be either in submitted, executing or completed state.
```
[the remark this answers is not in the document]
aleksei-fedotov: Oh, okay. Seems like I just got distracted in the middle of the paragraph.

rfcs/proposed/task_group_dynamic_dependencies/task_handle_semantics.md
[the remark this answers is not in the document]
aleksei-fedotov: Yeah. That's the point. If we want building an API that does not allow all these unsafe peculiarities, then I would just start with existing `task_handle`, even not allowing to specify successors for already submitted tasks, but only for created ones. Let's then see if it won't be enough for the users.

rfcs/proposed/task_group_dynamic_dependencies/task_handle_semantics.md
kboyarinov: The `add_successor` API was removed from the proposal. I have added the separate section describing alternatives.
I have also added the "Advanced examples" section describing all wavefront patterns we found with the possible implementations.

rfcs/proposed/task_group_dynamic_dependencies/task_handle_semantics.md
kboyarinov: The latest version proposes the combined tracking. Added the "Eager + classic" wavefront in the advanced examples section as a motivation.

rfcs/proposed/task_group_dynamic_dependencies/task_handle_semantics.md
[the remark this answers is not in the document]
kboyarinov: Added this as an open question.

rfcs/proposed/task_group_dynamic_dependencies/task_handle_semantics.md
```diff
@@ -0,0 +1,445 @@
...
+// Existing API
+class task_handle;
+
+// New APIs
+class task_tracker {
+public:
+    task_tracker();
+
+    task_tracker(const task_tracker& other);
+    task_tracker(task_tracker&& other);
```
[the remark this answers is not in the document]
kboyarinov: I think it can be useful for storing `task_tracker` in the containers. I have added a section with advanced examples and possible implementations. Recursive eager wavefront shows the necessity of copy semantics as well (while storing the trackers to the parent subtasks in the child task).

rfcs/proposed/task_group_dynamic_dependencies/task_handle_semantics.md
kboyarinov: This line was removed.

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1453 @@
...
+The first option is to consider ``current`` and ``target`` a separate tasks even after the transferring the successors from one to another.
+
+In this case, after the transfer, the task ``current`` will have an empty successors list, and ``target`` will have a merged successors list:
+
+<img src="assets/transferring_between_two_separate_tracking.png" width=600>
+
+After the transfer, the successors of ``current`` and ``target`` are tracked separately and adding new successors to one of them would only
+affect the successors list of one task:
+
+<img src="transferring_between_two_separate_tracking_new_successors.png" width=600>
```
isaevil:
```suggestion
<img src="assets/transferring_between_two_separate_tracking_new_successors.png" width=600>
```

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1453 @@
...
+
+## Introduction
+
+This document contains a concrete API and semantics proposal for ``task_group`` extensions defined in the parent RFC.
+The following cases are covered:
+* ``task_group`` extensions allowing handling the tasks in various states: created, submitted, executing and completed.
+  Existing API only allows having a non-empty ``task_handle`` object handling the created task and an empty one that does not
+  handle any tasks.
+* API for setting the dependencies between tasks in various states.
+* API for transferring the dependencies from the executing task to the task in various states.
```
aleksei-fedotov: Perhaps, rephrase to avoid misinterpretation:
```suggestion
* API for transferring the dependent tasks (i.e. successors) from the executing task to the task in various states.
```

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1453 @@
...
+tbb::task_tracker tracker = handle; // create the tracker for the task owned by the handle
+
+tg.run(std::move(handle)); // task is in submitted state, handle is left in an empty state
+
+// tracker is non empty and can be used to set dependencies
+```
+
+In this case, no semantic changes are needed for the submission methods, because the ``task_handle`` semantics is not changed.
+
+Alternative approaches for handling tasks in different states are described in the [separate section](#alternative-approaches).
```
aleksei-fedotov, line 95: The link points to non-existing target.

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1453 @@
...
+
+## Semantics for setting dependencies between tasks
+
+Lets consider creating a predecessor-successor dependency between the task tracked or owned by ``predecessor``
+and the task handled by ``successor_task_handle`` -
+``task_group::make_edge(predecessor_task_handle, successor_task_handle)``. 
+
+As it was stated in the parent RFC document, we would like to allow adding predecessors in any state described above and to limit the successor to be a task in created state since it can be too late to add predecessors to
```
aleksei-fedotov: Shall we allow adding predecessors to a successor in submitted state when the user knows that the successor cannot start because its other dependencies are not completed yet?

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1453 @@
...
+is a pointer to the vertex instance of the successor task.
+
+The vertex instance is created once the first task is registered as a predecessor and is reused by any other predecessors. 
+
+Once the predecessor task is completed, it should go through the list of successor vertices and decrement the reference counter. Once the successor's
+reference counter is equal to 0, the successor task can be scheduled for execution.
+
+API-wise, the function that decreases the reference counter may also return the pointer to the task. If the reference counter is not equal to 0, the
+returned pointer is ``nullptr``. Otherwise, the successor task pointer is returned. It is required to allow bypassing one of the successor tasks
+if the body of the predecessor task did not return other task that should be bypassed.
```
aleksei-fedotov: I think _requiring_ to allow bypassing is somewhat overstated. It would be good to but not required, isn't it? I believe you instead wanted to write something like: _If the body of the predecessor task does not bypass another task, the reference counting mechanism that decreases counters of successor tasks may take advantage and bypass one of these successor tasks if it becomes eligible for execution._

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1453 @@
...
+## Semantics for transferring successors from the currently executing task to the other task
+
+Lets consider the use-case where the successors of the task ``current`` are transferred to the task ``target`` owned or tracked by the ``target_handler``. 
+In this case, the API ``tbb::task_group::current_task::transfer_successors_to(target_handler)`` should be called from the body of ``current``.
+
+As it was mentioned in the parent RFC, if ``transfer_successors_to`` is called outside of task belonging to the same ``task_group``, the behavior is
+undefined.
+
+Current proposal limits the applicability of this API and allows only transferring from the executing task to the created task.
+Transferring to tasks in `scheduled`, `executing` or `completed` states are described in a [separate section](#transferring-successors-to-the-task-in-any-state)
```
aleksei-fedotov: The link is broken as it seems there is no `transferring-successors-to-the-task-in-any-state` target for it.

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
aleksei-fedotov:
```suggestion
<img src="assets/transferring_between_two_separate_tracking_new_successors.png" width=600>
```

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1453 @@
...
+After the transfer, the successors of ``current`` and ``target`` are tracked separately and adding new successors to one of them would only
+affect the successors list of one task:
+
+<img src="transferring_between_two_separate_tracking_new_successors.png" width=600>
+
+Alternative approach is to keep tracking ``current`` and ``target`` together after transferring. This requires introducing a special state
+of ``task_tracker` - a kind of `proxy` state. If the executing task calls `transfer_successors_to` from the body, all trackers, initially created
+to track this task are changing the state to `proxy`.
+
+Tracker to the ``current`` and ``target`` are sharing the single merged list of successors if ``current`` tracker is in `proxy` state.
```
aleksei-fedotov: 1.
```suggestion
Alternative approach is to keep tracking ``current`` and ``target`` together after transferring. This requires introducing a special state
of ``task_tracker`` - a kind of ``proxy`` state. If the executing task calls ``transfer_successors_to`` from the body, all trackers, initially created
to track this task are changing the state to ``proxy``.

Tracker to the ``current`` and ``target`` are sharing the single merged list of successors if ``current`` tracker is in ``proxy`` state.
```
2. Also, still requires management of multiple copies but now of the `task_tracker`, which was the primary concern against having `task_handle` to encapsulate all the logic related to task group's task dependencies. Shall we still consider not adding new entity for something that tracks the tasks?

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1453 @@
...
+would be transferred to ``B`` together with the entire list.
+
+If the transferring was done before adding the successor - the new successor would be added to ``A`` only and would not appear in the successors list of ``B``.
+
+Hence, the deterministic predecessor-successor relationships cannot be guaranteed in this case, because of the logical race between ``make_edge`` and ``transfer_successors_to``.
+
+If the successors of ``A`` and ``B`` are tracked together (the second option described) in both linearization case the newly added successor will appear in both
+successors list of ``B`` ("real" task) and ``A`` (task in `proxy` state).
+
+As it is stated in [one of the advanced examples topic](#combination-of-eager-and-classic-approaches), there are real-life scenarios where the combined tracking
```
aleksei-fedotov: The link seems to be broken.

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1453 @@
...
+should be implemented. Generally, these are examples where one thread is setting a predecessor-successor dependency between a task that can be in the running state
+(using a `task_tracker` to such a task). Other thread can execute the tracked task and can replace it in the task graph with the subtree by calling `transfer_successors_to`.
```
aleksei-fedotov: Perhaps,
```suggestion
should be implemented. Generally, these are the examples where one thread sets
a predecessor-successor dependency for a task in running state (using its
``task_tracker``) while the other thread is replacing this tracked task 
in the task graph with the subtree by calling the ``transfer_successors_to``.
```

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1453 @@
...
+* Unclear semantics for these function when the tracked task is executing ``task_group::current_task::transfer_successors_to``. Consider a task ``A`` that transfers
+  it's successors to the task ``B`` and exits the body. On the one hand, task ``A`` was already submitted and completed and both described functions should return ``true``.
+  On the other hand, the task ``A`` replaces itself in the task graph with the task ``B`` and it may be expected that the tracker would now track ``B`` (similarly to how
+  `make_edge` is done - new edges added to ``A`` after transferring are added to ``B``).
+
+## Potential future enhancements
+
+### ``empty_task`` API
+
+Since it is only allowed to transfer successors to a single task only, for some use-cases (see [N-body example](#n-bodies-problem) as a reference) it
```
aleksei-fedotov: Broken link. Probably, it should be:
```suggestion
Since it is only allowed to transfer successors to a single task only, for some use-cases (see [N-body example](#N-bodies problem) as a reference) it
```

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1453 @@
...
+These APIs were removed from the proposal because of the following reasons:
+* Lack of use-cases for these functions. All of the considered examples could be implemented without using these functions.
+* Unclear semantics for these function when the tracked task is executing ``task_group::current_task::transfer_successors_to``. Consider a task ``A`` that transfers
+  it's successors to the task ``B`` and exits the body. On the one hand, task ``A`` was already submitted and completed and both described functions should return ``true``.
+  On the other hand, the task ``A`` replaces itself in the task graph with the task ``B`` and it may be expected that the tracker would now track ``B`` (similarly to how
+  `make_edge` is done - new edges added to ``A`` after transferring are added to ``B``).
+
+## Potential future enhancements
+
+### ``empty_task`` API
```
aleksei-fedotov, line 548: Can you please elaborate on why that `empty_task` cannot include subsequent statements that need to wait until these gathered predecessor tasks execute?

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1453 @@
...
+    static void make_edge(task_tracker& pred, task_handle& succ);
+};
+} // namespace tbb
+} // namespace oneapi
+```
+
+Two enhancements for this API can be considered:
+* Allow ``task_tracker`` tracking the task in created state to be used as a successor.
+* Allow ``task_tracker`` tracking the task `T` in submitted state to be used as a successor if other non-submitted
+  predecessors of ``T`` prevent it from being scheduled for execution.
```
aleksei-fedotov: I think this applies not only to non-submitted predecessors but also to submitted ones, and even to being executed predecessors if the user is confident that these predecessors won't be finished (hence the successor is picked up by scheduler for execution) before the edge is made.

So, I suggest re-phrase to something like:
```suggestion
* Allow ``task_tracker`` tracking the task `T` in submitted state to be used as a successor if other
predecessors of ``T`` (non-submitted or still being executed) prevent the task ``T`` from being scheduled for execution.
```

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1453 @@
...
+    });
+}
+```
+
+### Wavefront
+
+Wavefront is a scientific programming pattern in which data elements are distributed on multidimensional grids. Elements should be computed in order
+because of dependencies between them.
+
+Let's consider a 2 dimensional wavefront example where the computation starts on the upper left corner of the grid:
```
aleksei-fedotov: Seems like previously there was another picture, where the computation indeed started at the upper left corner, but now it is a different one, is not it?
```suggestion
Let's consider a 2 dimensional wavefront example where the computation starts at the top corner of the grid:
```

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1453 @@
...
+
+<img src="assets/wavefront.png" width=300>
+
+Computations of each element *(i,j)* in the grid can start when two dependent sub-elements are completed:
+* North dependency - computation of the element *(i, j - 1)* if present
+* West dependency - computation of the element *(i - 1, j)* if present
```
aleksei-fedotov: Also, it is not north and west dependency anymore. It is more like a _upper right_ and _upper left_. Although, I might be seen as if I am nitpicking here but it seems that the picture was indeed rotated on 45 degrees clockwise without updating the text. So, drawing your attention to these discrepancies.

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1453 @@
...
+
+    cell_tasks.reserve(in);
+
+    for (int i = 0; i < in; ++i) {
+        cell_tasks[i].reserve(jn);
+
+        for (int j = 0; j < jn; ++j) {
+            cell_tasks[i].emplace_back(tg.defer([=] { compute_cell(i, j); }));
+
+            // north dependency
```
aleksei-fedotov:
```suggestion
            // upper right dependency
```
?

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1453 @@
...
+            if (j != 0) { tbb::task_group::make_edge(cell_tasks[i, j - 1], cell_tasks[i][j]); }
+
+            // west dependency
```
aleksei-fedotov:
```suggestion
            // upper left dependency
```
?

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1453 @@
...
+        }
+    }
+
+    // Run the graph
+    for (int i = 0; i < in; ++i) {
+        for (int j = 0; j < jn; ++j) {
+            tg.run(std::move(cell_tasks[i][j]));
+        }
+    }
+}
```
aleksei-fedotov, line 1026: Forgot to call the `wait` method?

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1453 @@
...
+
+bool operator==(const task_tracker& t, const task_tracker& rhs) noexcept;
+bool operator!=(const task_tracker& t, const task_tracker& rhs) noexcept;
+
+class task_group {
+    static void make_edge(task_handle& pred, task_handle& succ);
+    static void make_edge(task_tracker& pred, task_handle& succ);
+
+    struct current_task {
+        static void transfer_successors_to(task_handle& h);
```
isaevil: I think it would be great if the name of this function better reflected the behavior where `current_task` becomes a sort of proxy to `h`. I am not a native English speaker but `transfer` sounds like a one-time move of something somewhere. I don't think the following names are 100% better but I think they are worth considering:
* redirect/forward/delegate_successors_to
* transform_as/become_proxy_for

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1453 @@
...
+
+<img src="assets/wavefront_recursive_eager_second_division_west.png">
+
+Additionally, ``W`` should add dependencies between it's subtasks ``WN`` and ``WE`` and the subtasks of ``N`` created
+on the previous step (``NW`` and ``NS``):
+
+<img src="assets/wavefront_recursive_eager_second_division_west_new_dependencies.png">
+
+It is important to mention that at the moment of making additional edges, ``NW`` and ``NS`` are at least
+in `submitted` state and even can be completed.
```
aleksei-fedotov: Could you please explain why `NW` and `NS` are at least in `submitted` state? It is not obvious to me.

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1453 @@
...
+    std::size_t m_yn;
+
+    // Division number
+    std::size_t m_n_division;
+}; // class eager_wavefront_task
+
+void eager_wavefront(std::size_t n) {
+    tbb::task_group tg;
+
+    std::size_t num_divisions = log2(n / eager_wavefront_task::wavefront_grainsize) + 1;
```
aleksei-fedotov, line 1368: `eager_wavefront_task::wavefront_grainsize` is private.

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1453 @@
...
+            // Publish subtasks for successors
+            publish_predecessor(m_n_division + 1, north_x_index, north_y_index, std::move(north_subtask_tracker));
+            publish_predecessor(m_n_division + 1, west_x_index, west_y_index, std::move(west_subtask_tracker));
+            publish_predecessor(m_n_division + 1, east_x_index, east_y_index, std::move(east_subtask_tracker));
+            publish_predecessor(m_n_division + 1, south_x_index, south_y_index, std::move(south_subtask_tracker));
+        }
+    }
+
+private:
+    static constexpr std::size_t wavefront_grainsize = 5;
```
aleksei-fedotov:
```suggestion
    static constexpr std::size_t wavefront_grainsize = 5;
private:
```

aleksei-fedotov (commented):
Still not a full review.

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1453 @@
...
+            predecessors_container[i].reserve(num_subtasks);
+            num_subtasks *= 4;
+        }
+    }
+
+    static tbb::task_tracker find_predecessor(std::size_t n_division, std::size_t x_index, std::size_t y_index) {
+        // Number of elements in the grid on each level of division
+        // On the first level - 4 subtasks total (2 on each dimension)
+        // on the second level - 16 subtasks total (4 on each dimension), etc.
+        std::size_t n_grid = (2 << n_division);
```
aleksei-fedotov: It seems from the description, depending on the value of `n_division`, which I assume could be any value from a natural number set, the `n_grid` variable should been having values equal to the power of two, i.e. 2, 4, 8, 16, 32, etc. However, the description says they are 4, 16, 64, 256, and so on.
So what are the values that `n_grid` variable can store?

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1453 @@
...
+
+        auto it = predecessors[n_division].find(x_index * n_grid + y_index);
+        return it != predecessors[n_division].end() ? it->second : tbb::task_tracker{};
+    }
+
+    static void publish_predecessor(std::size_t n_division, std::size_t x_index, std::size_t y_index,
+                                    tbb::task_tracker&& pred)
+    {
+        std::size_t n_grid = (2 << n_division>>);
```
aleksei-fedotov: This line seems incorrect.

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1453 @@
...
+
+        [[maybe_unused]] auto it = predecessors[n_division].emplace(x_index * n_grid + y_index, std::move(pred));
+        assert(it != predecessors[n_division].end());
```
aleksei-fedotov: I don't think that getting the past the end iterator is possible for the map. I believe the expression in the assert statement always equals to `true`.

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1453 @@
...
+    {}
+
+    void operator()() const {
+        std::size_t x_size = m_xn - m_x0;
+        std::size_t y_size = m_yn - m_y0;
+
+        // Do serial wavefront if the grainsize reached
+        if (i_size <= wavefront_grainsize) {
+            assert(j_size <= wavefront_grainsize);
+            serial_wavefront(m_i0, m_in, m_j0, m_jn);
```
aleksei-fedotov: `i_size` and `j_size` do not exist as well as `m_i0`, `m_in`, `m_j0`, and `m_jn`. Did you mean `x_size`, `y_size`, `m_x0`, `m_xn`, `m_y0`, `m_yn`?
```suggestion
        if (x_size <= wavefront_grainsize) {
            assert(y_size <= wavefront_grainsize);
            serial_wavefront(m_x0, m_xn, m_y0, m_yn);
```

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
akukanov: On (2) - no, I do not think multiple copies of `task_handle` were the primary concern.

The discussion started at https://github.com/uxlfoundation/oneTBB/pull/1469#discussion_r1938290313, where I noted that the way submission functions currently take a ``task_handle`` (only as an r-value) is kind of incompatible with the idea of "extended task_handle lifetime". The original RFC therefore states that 
> Semantics for task_handle objects usages by task_group and task_arena should be defined for all possible states of the handle.

Then, we considered allowing submission functions to take a handle as an l-value, but that is even more problematic, as shown in https://github.com/uxlfoundation/oneTBB/pull/1664#discussion_r1987471753. Quoting myself,
> We want either a shared ownership, which can be represented by copies of the task handle, or some form of weak reference/observation.

And then, there was a consensus between Konstantin, Mike, and myself that supporting shared ownership semantics for task handles would lead to more complicated design and reasoning of the behavior. compared to the "weak reference" alternative. As far as I recall, my concerns were about task submission API semantics & behavior and possible error diagnostics, and not about multiple copies of a handle.

We can of course reconsider this, but so far nobody has provided strong arguments why we should.

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1453 @@
...
+    }
+
+    static tbb::task_tracker find_predecessor(std::size_t n_division, std::size_t x_index, std::size_t y_index) {
+        // Number of elements in the grid on each level of division
+        // On the first level - 4 subtasks total (2 on each dimension)
+        // on the second level - 16 subtasks total (4 on each dimension), etc.
+        std::size_t n_grid = (2 << n_division);
+
+        auto it = predecessors[n_division].find(x_index * n_grid + y_index);
+        return it != predecessors[n_division].end() ? it->second : tbb::task_tracker{};
```
aleksei-fedotov: Not sure I understand how predecessor tasks cannot be found. They are added unconditionally and the dependencies are there by the algorithm design.

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1453 @@
...
+
+Each `FileInfo` contains additional reference to `task_group`, a `task_handle` owning the parsing task and `task_tracker` initially set
+to track parsing task.
+
+`process()` function submits the task owned by `m_parse_task` for execution.
+
+`parse()` function reads the file, creates a task for finalization, transfers the list of dependencies and creates an edge between the
+corresponding `task_tracker` and the finalization task.
+
+After creating all of the edges, the parsing task replaces itself in the task graph with finalize_task by
```
aleksei-fedotov:
```suggestion
After creating all of the edges, the parsing task replaces itself in the task graph with `finalize_task` by
```

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1453 @@
...
+
+Replaces the task tracked by ``*this`` to be a task owned by ``handle``.
+
+Returns a reference to ``*this``.
+
+#### Observers
+
+``explicit operator bool() const noexcept``
+
+Checks if `this`` tracks any task object.
```
aleksei-fedotov:
```suggestion
Checks if ``this`` tracks any task object.
```

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1453 @@
...
+``task_handle`` as an argument to one of the submission functions would invalidate all copies or set them
+in the "weak" state that allows only to set dependencies between tasks.
+
+## Exit criteria & open questions:
+* Are concrete names of APIs good enough and reflects the purpose of the methods?
+* The performance targets for this feature should be defined
+* API improvements and enhancements should be considered (may be a criteria for moving the feature to `supported`):
+  * Should comparison functions between ``task_tracker`` and ``task_handle`` be defined?
+  * Should ``task_tracker`` tracking the task in created or submitted state (not always) be allowed as a
+    successor in ``make_edge``? See [separate section](#using-a-task_tracker-as-a-successor) for more details.
```
aleksei-fedotov: Broken link - it refers to non-existing paragraph.

Perhaps, the following will work:
```suggestion
    successor in ``make_edge``? See [separate section](#Using a ``task_tracker`` as a successor) for more details.
```

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1453 @@
...
+  * Should ``empty_task`` helper API be provided to optimize creating and spawning the empty sync-only tasks. See
+    [separate section](#empty_task-api) for more details.
```
aleksei-fedotov: Broken link - it refers to non-existing paragraph.

Perhaps, the following will work:
```suggestion
[separate section](#``empty_task`` API) for more details.
```

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
aleksei-fedotov: From my point of view, introducing another entity for essentially the same thing from the user-perspective complicates the design. The user will need to keep them separate in mind and remember what they are good for and what they are not.

As far as I understand, implementation-wise `task_tracker` represents this kind of shared-ownership, as it still needs to have a pointer to representation of at least the task state and the list of successors shared among `task_tracker` instances pointing to the same task. In this sense, it differs from the `task_handle` only in impossibility to be submitted for execution.
To avoid possible confusion about acting on moved `task_handle`, can we just deprecate the move semantics and add only lvalue signatures? After all, the same function object can be submitted several times into `task_group`.

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1453 @@
...
+bool operator!=(const task_tracker& t, const task_tracker& rhs) noexcept;
+
+class task_group {
+    static void make_edge(task_handle& pred, task_handle& succ);
+    static void make_edge(task_tracker& pred, task_handle& succ);
+
+    struct current_task {
+        static void transfer_successors_to(task_handle& h);
+    };
+}; // class task_group
```
aleksei-fedotov: I think we should consider introducing the return of a `[[maybe_unused]]` `task_handle`/`task_tracker` by the `task_group::run(Func&& f)`?

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
akukanov: I think we disagree on "essentially the same thing".

I also think that you might misinterpret the semantics of `task_handle`. It does not represent a function object, it represents a created task. If you submit the same task for execution several times, you will certainly get some problems.

Anyway, without going into further argue, my point in the above comment was that multiple copies of a handle/tracker to the same task were not the primary concern.

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
kboyarinov: Applied

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
kboyarinov: The link works on my side in both GitHub and VSCode Markdown viewers. Which viewer do you use?

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
kboyarinov: This case is described in "Using a task_tracker as a successor" section and is included in the list of enhancements in the end of the document.

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
kboyarinov: Rephrased

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
kboyarinov: The link also works correct on my side. It seems like the common issue for all links to parts of the same document.

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
kboyarinov: Fixed

kboyarinov: Fixed

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
kboyarinov: 1 is Fixed

Regarding 2, my understanding and the idea I was trying to express through the document is that `task_tracker` and `task_handle` do not represent the same thing. `task_handle` represents the task itself that holds the reference counter in the `task_group`'s `wait_context`. Hence, it can be used to submit the task for execution, but only once (move-semantics perfectly implements the required semantics). 

`task_tracker` represents an entity that (at least currently) may be used only for adding successors to tasks in any state. It is possible to have one or several trackers to the task even after the task itself was submitted, scheduled or executed. This instance can be copyable and any instance can be used to add new dependencies. 

From the implementation perspective, `task_tracker` implements a shared-ownership over one of the internal structures, but from the design perspective, I would say it represents a weak-ownership over a task (does not own the task object but can be used to modify it if the task is still alive). 

For me, mixing these two semantics into the single object would create more error-prone places in the user's code and understanding

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
kboyarinov: Fixed

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
kboyarinov: I am not sure I have completely understood the question. 
If the user application implies any subsequent statements that should be executed after predecessors, they can be represented as a regular successor task (`task_handle` object). Usage of `empty_task` does not make sense in this case.

But for some cases (like the N-body problem in the examples section), there are no other actions that should be done after all of the predecessors are completed , we only need to wait until their completion to consider subtree finished.

Current implementation forces using the empty task represented by `task_handle` in this case, but since it is a regular task from TBB perspective, it would be scheduled and executed that can create additional overhead.

The idea of `empty_task` is to omit scheduling if the successor task is an object of a special type.

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
kboyarinov: As for other links, this one works fine on my side. Let's consider all of the links part of the same issue and try to fix them all together:)

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
kboyarinov: Applied

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
kboyarinov: Applied

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
kboyarinov: You are right, thanks. Fixed

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
kboyarinov: Applied

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
kboyarinov: Applied.

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
kboyarinov: Fixed.

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
kboyarinov: You are correct, thanks. The predecessor should be found since we starting the search only if the index allows. So the predecessor is guaranteed to be found. 

Rewrote the code and added the assertion here.

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
kboyarinov: Oops, VS Code thought it is clever:)

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
kboyarinov: Correct. Fixed

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
kboyarinov: Applied

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
kboyarinov: Fixed

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
kboyarinov: Applied

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
kboyarinov: Highlighted explicitly in the open question section

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
kboyarinov: Added to the list of future enhancements.

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
kboyarinov: Fixed

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1515 @@
...
+
+Once the thread finishes executing the task body, the task transitions to the `completed` state.
+
+The parent RFC proposes extending the ``task_handle`` to support tasks in any of the states defined above. This change
+would require modifying the semantics of submission methods (such as ``task_group::run``) to accommodate ``task_handle`` objects representing tasks
+in various states.
+
+In contrast, this document proposes keeping ``task_handle`` as a unique-owning handle that either owns a task in the `created` state or owns no task.
+
+To handle tasks in other states, this proposal introduces a new weak-owning handle, ``task_tracker``. It can track a task in any state.
```
akukanov: Let me suggest alternative names: `task_connector`, `task_link`, `task_junction`. There might be other suitable synonyms found.

Also we might consider making it a type within `task_handle`, e.g., `task_handle::connector`. Or even make it an unspecified type returned by certain member functions of `task_handle` or/and `task_group` (not my preference though).

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1515 @@
...
+The limitation is enforced by requiring the successor argument to be a ``task_handle``.
+
+Let's consider the different states of the task tracked or handled by ``predecessor``. 
+
+If the predecessor task is in any state except `completed` (i.e. `created`/`scheduled`/`executing`), the API registers the successor task
+in the predecessor's list of successors and increments the corresponding reference counter on the successor side to ensure it
+is not executed before the predecessor completes. The successor task can only begin executing once its reference counter reaches zero.
+
+If the predecessor task is in the `completed` state, the API has no effect: no changes are made to the list of successors or reference counters,
+since no additional dependencies are needed. The successor task can proceed if all other dependencies are also satisfied.
```
akukanov: However, what about adding "late" successors after transfer? The original task might be completed, but it transferred its successors to another one.

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1515 @@
...
+
+If the predecessor's state changes during registration, the API must ensure that dependencies are not added and reference counters are not incremented for tasks that have already completed.
+
+From an implementation perspective, this API requires each predecessor task to maintain a list of successors, with each successor represented by
+a vertex instance. Each vertex contains a reference counter and a pointer to the successor task.
```
akukanov: Does the implementation differ substantially from the task graphs of parallel algorithms? I guess, one difference are multiple successors kept in a list; anything else?

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1515 @@
...
+
+Each element in the successor list is a pointer to a vertex instance representing a successor task.
```
akukanov:
```suggestion
From the implementation perspective, this API requires each predecessor task to maintain a list of successors.
The list holds pointers to vertex instances, each representing a successor task.
A vertex contains the reference counter for a successor task and the pointer to it.
```

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
aleksei-fedotov: GitHub viewer. I am clicking "Display the rich diff" button. It seems it does not work with the links. Viewing the whole file through the context menu works with the hyperlinks correctly.

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -272,10 +310,9 @@ tbb::task_group::make_edge(right_lower_rectangle_task, empty_sync_task);
 tbb::task_group::current_task::transfer_successors_to(empty_sync_task);
 ```
 
-In the current TBB implementation, ``empty_sync_task`` would be spawned (or bypassed) and executed anyway that can add some overhead for scheduling.
+In the current TBB implementation, ``empty_sync_task`` is still spawned or bypassed and executed, which may introduce unnecessary scheduling overhead.
 
-Possible extension is to add some special API for `task_group` to provide an `empty_task` of a special type that would not be scheduled for execution
-and skipped when the dependent tasks are completed:
+A possible extension is to add a special API to `task_group` that provides an `empty_task` — a task type that is never scheduled for execution and is skipped once its dependencies are resolved:
```
aleksei-fedotov: What if instead of introducing a special `empty_task` and an interface to acquire it from the `task_group`, we simply allow the `...::current_task::transfer_successors_to` method to accept multiple parameters? You also mention current limitation of having only one parameter in this method somewhere below.

For this particular example, its usage will look like `tbb::task_group::current_task::transfer_successors_to(left_upper_rectangle_task, right_lower_rectangle_task)` which will just add these two passed tasks as predecessors to the successors of the current task (by updating the corresponding lists, increasing reference counters, etc. - whatever is necessary).

Internally, the runtime may create the same synchronization task but the implementation details of it will not be exposed to the user and they will use already familiar interface for transferring successors.

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1515 @@
...
+a vertex instance. Each vertex contains a reference counter and a pointer to the successor task.
+
+Each element in the successor list is a pointer to a vertex instance representing a successor task.
+
+A vertex instance is created when the first predecessor is registered and is reused by subsequent predecessors. 
+
+When a predecessor task completes, it iterates through its list of successor vertices and decrements each reference counter.
+When a successor's reference counter reaches zero, the task can be scheduled for execution.
+
+From the API perspective, the function that decrements the reference counter may return a pointer to the task. If the reference counter is not zero, the
```
akukanov: In other places within the section, "API" refers to the function to set a dependency, but here it seems to have a different meaning as you speak about implementation. I would perhaps not mention API here to avoid confusion.

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -348,33 +385,66 @@ class task_group {
 } // namespace oneapi
 ```
 
-If the ``task_tracker`` tracking the task that is scheduled for execution, executing or completed, is used
-as a second argument, the behavior of APIs above is undefined.
+If a ``task_tracker`` that tracks a task in the `submitted`, `executing`, or `completed` state is used
+as the second argument, the behavior of the above APIs is undefined.
```
aleksei-fedotov: I believe you wanted to exclude the submitted state since you are writing above the code snippet that it might be allowed if something prevents it to be scheduled.

Do you mean `scheduled`?
```suggestion
If a ``task_tracker`` that tracks a task in the `scheduled`, `executing`, or `completed` state is used
as the second argument, the behavior of the above APIs is undefined.
```

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1515 @@
...
+
+In this approach, no semantic changes are needed for submission methods, since the semantics of ``task_handle`` remain unchanged.
+
+Alternative approaches for handling tasks in different states are described in the [separate section](#alternative-approaches).
+
+## Semantics for setting dependencies between tasks
+
+Consider creating a predecessor-successor dependency between the task tracked or owned by ``predecessor``
+and the task handled by ``successor_task_handle``—
+``task_group::make_edge(predecessor_task_handle, successor_task_handle)``.
```
akukanov: I'd still like to consider other names for the function. There are pros and cons for `make_edge`. My concerns about that name are:
- "Edge" does not seem a good term for a short-living task dependency; at least we have not used it before in this context.
- It's not `tbb::flow::make_edge` but `tbb::task_group::make_edge`. I am not sure if reusing the name in such manner helps or confuses more. It might be fine if ADL works properly, but does it?
- in the flow graph, there is also `remove_edge` - but here it would make no sense.

On the other hand, I agree that semantically the operations are very close. But maybe we should express that differently, e.g. by adding a free function (or a "niebloid") in `namespace tbb` that works with both graph nodes and task handles? In that case, my preferred name is `tbb::connect`.

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1515 @@
+# API and semantics details for task group dynamic dependencies
+
+*Note:* This document is a sub-RFC of the [umbrella RFC for task group dynamic dependencies](README.md). 
+
+## Table of contents
```
akukanov, line 5: I suggest to structure the document a bit different - that way it would be easier (for me :)) to read and review.

- Introduction
- Make 2,3,4 subsections of a "Design" section
- Alternative approaches & "removed functionality"
- API proposal
- Potential enhancements
- Exit criteria & open questions
- Advanced examples

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
kboyarinov: Mentioned this as an alternative approach for this enhancement.

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
kboyarinov: Since no `scheduled` state was introduced, only the `executing` and `completed` were kept.

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
kboyarinov: Thanks, it was missed from this paragraph. Rewritten

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
kboyarinov: Rewritten

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
kboyarinov: Done

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
kboyarinov: `n_grid` stores the number of elements in each dimension of the grid on the particular level of division. 
So, the previous version of the implementation is not correct:) Thanks for catching that!

During the first division, we have 2x2 grid of subtasks (`n_grid` = 2). On the second - 4x4 (`n_grid` = 4).
I have rewritten this part and corrected the comment

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
kboyarinov: I would say the biggest differences are:
* The list of successors vs the only parent for parallel algorithm tasks (as you mentioned),
* Successors in `task_group` are real tasks, while for parallel algorithms, it is tree nodes that can execute the body in case of reductions (that behaves like bypassing the successor). It is also related to the first bullet, but since tasks in `task_group` may have multiple successors, some tasks should be spawned if ref counter reaches 0.
* Lazy initialization of everything related to the dependencies in the `task_group` (list of successors, transfer tracking, etc.) to preserve `task_group` tasks without dependencies without any additional overheads.
* "Late" successors that are not supported by parallel algorithms that have pre-defined tree structure when the task body is executed

akukanov: Thanks, that makes sense. I think lazy initialization could be mentioned in the document as one of the design goals.

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
kboyarinov: Added naming considerations section. Agreed to apply `set_task_order` as a replacement for `make_edge`

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
kboyarinov: Added naming considerations section. Agreed to apply `task_completion_handle` as a replacement for `task_tracker`.

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1599 @@
...
+
+Constructs a ``task_completion_handle`` that refers to the task owned by ``handle``.
+
+If ``handle`` is empty, the behavior is undefined.
+
+``~task_completion_handle()``
+
+Destroys the ``task_completion_handle``.
+
+If the task is a predecessor or a successor, the behavior is undefined.
```
akukanov: I think that destroying handles to a completed task should be fine, should not it?
Otherwise, since there is no way to remove task dependencies once added. this sounds like undefined behavior becomes inevitable.

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1599 @@
...
+class task_group {
+    static void set_task_order(task_handle& pred, task_completion_handle& succ);
+    static void set_task_order(task_completion_handle& pred, task_completion_handle& succ);
+};
+} // namespace tbb
+} // namespace oneapi
+```
+
+If a ``task_completion_handle`` referring to a task in the `executing`, or `completed` state is used
+as the second argument, the behavior is undefined.
```
akukanov, line 656: Alternatively, we can consider a function that "converts" a `task_completion_handle` to a valid `task_handle`, if the task has not been started yet. That `task_handle` would have to be resubmitted after adding dependencies.

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -0,0 +1,1599 @@
...
+```cpp
+tbb::task_group tg;
+tbb::task_completion_handle ch = tg.run([] { /*task body*/ });
+// ...
+
+tbb::task_handle successor = tg.defer([] { /*successor body*/ });
+tbb::task_group::set_task_order(ch, successor);
+```
+
+This could be achieved by either changing the return type of ``task_group::run`` or introducing a new submission function:
```
akukanov: Would not changing the return type of `task_group::run` introduce extra overhead even if the return value is not needed and ignored? I think it would, and therefore a new submission function (e.g., a new overload of `run` with an extra argument) is preferable.

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
kboyarinov: It was a mistake, this note should be added to the destructor of `task_handle`. I have added the small subsection to summary.

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
kboyarinov: Added a mention.

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
kboyarinov: Agree, changed the wording.

akukanov: Looks like one of the comments there was not updated.

kboyarinov: Thanks, updated

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -331,179 +332,222 @@ namespace tbb {
...
-Registers the task tracked by ``pred`` as a predecessor that must complete before the task owned
+Registers the task completion of which is handled by ``pred`` as a predecessor that must complete before the task owned
 by ``succ`` can begin execution.
 
-The behavior is undefined in the following cases:
-* ``pred`` is an empty ``task_tracker``, or ``succ`` is an empty ``task_handle``,
-* If the task tracked by ``pred`` and the task handled by ``succ`` belong to different ``task_group``s.
+It is thread-safe to concurrently add multiple predecessors to a single successor, and to register the same predecessor for multiple successors.
+
+It is thread-safe to concurrently add successors to both the task currently transferring its successors and the task receiving them.
```
aleksei-fedotov: Should we use the words "completion of the task" rather than its successors?
```suggestion
It is thread-safe to concurrently add successors to both the task currently transferring its completion and the task receiving it.
```

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -515,86 +559,84 @@ tbb::task_handle right_lower_rectangle_task = tg.defer(...);
...
+    static void transfer_this_task_completion_to(task_completion_handle& task);
 };
 
 } // namespace tbb
 } // namespace oneapi
 ```
 
 From an implementation perspective, this would require merged successor tracking to support not just two tasks (the sender and the recipient), but an entire set of
-tasks — since a task may transfer its successors to another task that is also performing `transfer_successors_to`, and so on.
+tasks — since a task may transfer its successors to another task that is also performing `transfer_this_task_completion_to`, and so on.
```
aleksei-fedotov:
```suggestion
tasks — since a task may transfer its completions to another task that is also performing `transfer_this_task_completion_to`, and so on.
```
?

rfcs/proposed/task_group_dynamic_dependencies/extended_semantics.md
```diff
@@ -1484,45 +1552,46 @@ int main() {
...
 
-`f4p` and `f3p` are in at least the `submitted` state when the edge to `f5f` is created.
+`f4p` and `f3p` are in at least the `submitted` state when the dependency to `f5f` is established.
 
 Let's assume that the task `f3p` is executed next. It creates the finalize task `f3f` and two tasks for the included files - `f2p` and `f1p`:
 
 <img src="assets/parser_tasks2.png" width=400>
 
-After the edges between the included files and `f3f` are added, `f3p` transfers its successors to `f3f` and is destroyed:
+After the dependencies between the included files and `f3f` are established, `f3p` transfers its successors to `f3f` and is destroyed:
```
aleksei-fedotov: Perhaps, change here as well?
```suggestion
After the dependencies between the included files and `f3f` are established, `f3p` transfers its completions to `f3f` and is destroyed:
```

aleksei-fedotov (approved):
LGTM. Found a couple of possible discrepancies due to recent renamings.

vossmjp (approved):
Looks good to me. 

That said, I think there should be a link to the sub-RFC from the main RFC. README.md is automatically previewed in github and so it may be easy to overlook the extended_semantics.md if a link does not appear in README.md. I'd suggest putting it immediately before the Open Questions section.

kboyarinov:
> Looks good to me.
> 
> That said, I think there should be a link to the sub-RFC from the main RFC. README.md is automatically previewed in github and so it may be easy to overlook the extended_semantics.md if a link does not appear in README.md. I'd suggest putting it immediately before the Open Questions section.

Thanks for catching that. Applied.

I think we will need to think on how to merge `README.md` and `extended_semantics.md` while moving to `experimental` since most of API from README.md were renamed, the semantic has changed and most of the open questions are resolved in extended_semantics.md or not relevant at all any more.

PR #1666, opened 2025-03-04
Add info about TBBBind and custom allocators to known limitations

dnmokhov:
### Description 

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown

### DIFF
--- a/doc/main/intro/limitations.rst
+++ b/doc/main/intro/limitations.rst
@@ -48,6 +48,6 @@ Incorrect Installation Location
 Dynamic Malloc Replacement and Topology API Incompatibilities
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-**Limitation:** On Linux* OS, using dynamic malloc replacement with ``tbb::info`` and ``tbb::task_arena::constraints`` APIs may result in undefined behavior.
+**Limitation:** On Linux* OS, using dynamic malloc replacement with ``tbb::info`` and ``tbb::task_arena::constraints`` APIs may result in runtime failures.
 
 **Solution:** Set ``TBB_ENABLE_SANITIZERS=1`` in the environment. This informs that dynamic malloc replacement is used.
--- a/doc/main/intro/limitations.rst
+++ b/doc/main/intro/limitations.rst
@@ -44,3 +44,10 @@ Incorrect Installation Location
 **Limitation:** oneTBB does not support ``fork()``. 
 
 **Solution:** To work-around the issue, consider using ``task_scheduler_handle`` to join oneTBB worker threads before using ``fork()``.
+
+Custom Allocator and Topology API Incompatibilities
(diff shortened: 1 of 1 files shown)


doc/main/intro/limitations.rst
```diff
@@ -44,3 +44,10 @@ Incorrect Installation Location
 **Limitation:** oneTBB does not support ``fork()``. 
 
 **Solution:** To work-around the issue, consider using ``task_scheduler_handle`` to join oneTBB worker threads before using ``fork()``.
+
+Custom Allocator and Topology API Incompatibilities
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+**Limitation:** On Linux* OS, using custom memory allocators together with ``tbb::info`` and ``tbb::task_arena::constraints`` APIs may result in undefined behavior.
```
aepanchi, line 51:
```suggestion
**Limitation:** On Linux* OS, using custom memory allocators with ``tbb::info`` and ``tbb::task_arena::constraints`` APIs may result in undefined behavior.
```

doc/main/intro/limitations.rst
```diff
@@ -44,3 +44,10 @@ Incorrect Installation Location
...
+
+**Solution:** Set ``TBB_ENABLE_SANITIZERS=1`` in the environment to hint that custom allocators are used.
```
aepanchi, line 53:
```suggestion
**Solution:** Set ``TBB_ENABLE_SANITIZERS=1`` in the environment. This informs that custom allocators are used.
```

doc/main/intro/limitations.rst
kboyarinov, line 51: I don't like the formulation `custom memory allocators` here since it can be interpreted as usage of custom allocator types. Since this issue only relates to the malloc replacement libraries like tbbmalloc, jemalloc, etc., I propose to use "dynamic malloc replacement" or "dynamic allocating functions replacement" instead.

kboyarinov: Also, do we use "undefined behavior" when defining the known limitations? It seems like we can be more concrete and say it is runtime failures.

doc/main/intro/limitations.rst
kboyarinov, line 53: It does not seem like the `TBB_ENABLE_SANITIZERS` variable is about usage of custom dynamic malloc replacements and also since the deepbind would be dropped, can we guarantee that the tbbbind will work correct in all use-cases? 
Does it makes sense to provide `LD_PRELOAD`-based workaround instead?

dnmokhov: I agree that the environment variable name seems unconnected. However, the `LD_PRELOAD`-based workaround might cause more issues due to incorrect/incompatible TBBBind/HWLOC versions.

When deepbind is dropped, I believe we would remove the known limitation and the environment variable, because the globally-defined malloc replacement would always be used.

aepanchi (approved):
LGTM!

kboyarinov (approved):

PR #1677, opened 2025-03-13
[rfc] Resource-limited Nodes

marcpaterno:
### Description 

This PR introduces an RFC describing two new node types, a *resource-limited function node* and a *resource-limiter node*. These support node bodies that require sole access to limited resources, such as access to a thread-unsafe library or to a database handle.

### Type of change

### Tests

- [x] added - as part of demonstrator code linked to in the RFC
- [ ] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

@knoepfel

### Other information

This RFC is motivated by discussions held between Fermilab and the Flow Graph development team.


rfcs/proposed/flow_graph_serializers/README.md
```diff
@@ -0,0 +1,596 @@
...
+
+``` cpp
+using namespace tbb;
+flow::graph g;
+
+resource_limiter_node<GPU> gpus{g, {GPU{ /*gpu 1*/ }, GPU{ /*gpu 2*/}}; // Permit access to two GPUs
+resource_limiter_node<ROOT> root{g, {ROOT{...}}};  // Only 1 ROOT handle available
+
+rl_function_node fn{g,
+                    std::tie(gpus, root), // Edges implicitly created to the required resource limiters
```
kboyarinov, line 139: It seems a bit strange to me to define `gpus` and `root` as nodes. As a user, I would expect a node in the graph as something that can send and/or receive some kind of value and may be connected with other nodes that can send/receive the same value type. 
It is a bit unclear what kind of values can be sent by `gpus` and `root`. They seem more like special instances that manages the resource and the `function_node`s just using the managed resources.
I propose not to look on things like `gpus` and `root` as on the graph nodes and choose a better public name for that. 

In theory, they can be implemented as a graph nodes sending/receiving the token that is used to manage the resource, but I don't think such details should be expressed in the API names.

rfcs/proposed/flow_graph_serializers/README.md
```diff
@@ -0,0 +1,596 @@
...
+at the same time, if the nature of `db_resource` were to allow two
+handles to be available, and as long as each activation was given a
+token to a different handle.
+
+# Proposal
+
+Our proposal is an addition to what already exists and does not break
+API backwards compatibility. The proposal consists of: 1. Introducing
+the equivalent of a `resource_limiter_node` class template that, when
+connected with another node, ensures limited access to the resource it
```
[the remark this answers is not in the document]
knoepfel: We will clarify the wording.

rfcs/proposed/flow_graph_serializers/README.md
```diff
@@ -0,0 +1,596 @@
...
+> user-provided node body.
+
+With the proposal here, in addition to what is written above we imagine
+the following could be done:
+
+``` cpp
+using namespace tbb;
+flow::graph g;
+
+resource_limiter_node<GPU> gpus{g, {GPU{ /*gpu 1*/ }, GPU{ /*gpu 2*/}}; // Permit access to two GPUs
```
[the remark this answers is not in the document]
knoepfel: We did not consider the case where (e.g.) one GPU should be prioritized over another.  We imagine such prioritization could be helpful, but they are unusual in our domain.

rfcs/proposed/flow_graph_serializers/README.md
knoepfel: We agree with your point, @kboyarinov.  We will rename `resource_limiter_node` to just `resource_limiter`, even though our implementation expressed it as a Flow Graph node.

knoepfel: It turns we also had a footnote in our analysis that anticipated such a change.  That footnote is now unnecessary.

rfcs/proposed/flow_graph_serializers/README.md
```diff
@@ -0,0 +1,596 @@
...
+We imagine the following constructors could exist
+
+``` cpp
+using namespace tbb;
+flow::graph g;
+
+// 1. Already discussed above
+rl_function_node fn1{g,
+                     std::tie(gpus, root),
+                     [](Hits const&, GPU const*, ROOT const*) -> Tracks { ... }};
```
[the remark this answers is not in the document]
knoepfel: We will add definitions for *resource*, *handle*, and *token* that clarify what we mean:

- *resource*: an entity external to the program for which access may need to be limited (e.g. a file on the file system, a database, a thread-unsafe library)
- *handle*: the program entity within the program that represents and provides access to the resource (the handle is not copyable as the resource itself cannot be copied)
- *token*: a small, copyable object passed into a node to grant access to the handle

With these definitions, the body receives tokens that happen to be implemented as pointers to the resource handles.  To emphasize this, we will introduce a type alias `token<T>` that is just implemented as a pointer.

rfcs/proposed/flow_graph_serializers/README.md
```diff
@@ -0,0 +1,613 @@
...
+                     std::tie(gpus, root),
+                     [](Hits const&, token<GPU>, token<ROOT>) -> Tracks { ... }};
+
+// 2. No access required to a limited resource (equivalent to flow::function_node)
+rl_function_node fn2{g, flow::unlimited, [](Hits const&) -> Tracks { ... }};
+
+// 3. Access required to a limited resource, and the user body must be serialized
+rl_function_node fn3{g,
+                     std::make_tuple(flow::serial, std::ref(db_limiter)),
+                     [](Hit const&, token<DB>) -> Tracks{ ... }};
```
dnmokhov, line 272: Instead of adding new `rl_xxx_node` class templates, would it make sense to add corresponding constructors directly to the relevant flow-graph nodes? I.e., in each node type that supports this functionality, add  constructors with `size_t concurrency` replaced by a tuple of `resource_limiter`s (1 and 3 above). Constructor 2 is already available in the current node type.

akukanov: That is my question/desire as well. But I think it should probably be the job of the oneTBB team to evolve the proposal into this direction. For the purpose of initial acceptance, it should be fine to add this as a design option or an open question.

knoepfel: We agree with both of you.  Please let us know how we can help.

aleksei-fedotov (commented):
Good proposal! Thanks!

rfcs/proposed/flow_graph_serializers/README.md
akukanov: @marcpaterno @knoepfel
> Please let us know how we can help.

For now could you please just add some note in the doc that there are two possible approaches: adding new "resource limited" types of functional nodes, like proposed, and trying to extend the existing functional nodes with the necessary support.

A deeper discussion of pros and cons etc., like in Aleksei's comment above, should likely be driven by someone from the team; of course you will be welcome to participate.

knoepfel: @akukanov, we've included the alternative proposal (i.e. just adding constructors to existing nodes) in the proposal description above.

rfcs/proposed/flow_graph_serializers/README.md
```diff
@@ -0,0 +1,613 @@
...
+  concurrency but *Calibration\[C\]* is serial. This means that, at the
+  same time, we can have:
+  - two *Calibration\[A\]* tasks running, or
+  - two *Calibration\[B\]* tasks running, or
+  - one *Calibration\[A\]* task running and one *Calibration\[B\]* task
+    running, or
+  - one *Calibration\[A\]* task running and one *Calibration\[C\]* task
+    running, or
+  - one *Calibration\[B\]* task running and one *Calibration\[C\]* task
+    running.
```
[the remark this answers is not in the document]
knoepfel: @aleksei-fedotov, I changed my mind—I added the edges between the resource limiters and the other nodes as well.

rfcs/proposed/flow_graph_serializers/README.md
```diff
@@ -0,0 +1,613 @@
...
+  /// \brief Activate the resource_limiter by placing one token to each handle
+  /// it manages into its buffer.
+  ///
+  /// This function must be called before the flow graph is started.  It
+  /// is not automatically called by the constructor because the
+  /// resource_limiter may be placed into a container (e.g. a vector) that
+  /// can grow after the resource_limiter is constructed.  If the
+  /// resource_limiter is activated before it is placed into its final
+  /// location, then the locations of the tokens in the buffer can become
+  /// invalid, resulting in memory errors when the flow graph is executed.
```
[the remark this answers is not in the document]
akukanov: @aleksei-fedotov I think this is because a token is a pointer to a handle, and so becomes invalid if the handle is reallocated for whatever reason.

akukanov:
I have opened a design discussion #1683 where we can collect ideas and opinions on the development directions for this proposal after its initial acceptance.

@aleksei-fedotov. I would suggest you to add your detailed comment there as well.

rfcs/proposed/flow_graph_serializers/README.md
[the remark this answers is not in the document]
knoepfel: @akukanov and @aleksei-fedotov, we actually added the `activate()` function in a much earlier implementation of the limiter that supported only 1 token, which was just an integer that was passed around the graph.  We noticed that if the limiter was placed into a container, then the token might be lost if the container reallocated after the token was placed into the buffer.

Of course, another option could have been to (a) never place the limiter in a container, or (b) if a container of limiters is required, make it a container of (smart) pointers to limiters.

akukanov (approved):
I think this describes very well the overall idea, motivation, and even some design considerations.

vossmjp (approved):
I also think this has enough detail to move to implementation.

Issue #1656, opened 2025-02-26
[Hackathon] Stabilize oneTBB Example Benchmarks 2


dnmokhov:
# Summary
Several oneTBB examples serve as performance benchmarks for the library. It's recommended to check the relative error for these examples. If the relative error exceeds 5%, consider stabilizing the example to improve the reproducibility of results, while still maintaining a reasonable execution time.

Additional task
Conduct scalability analysis and see if the examples are scalable in the given HW. See modifying the input parameters like dataSize improve scalability. 

**Target Examples**

seismic : https://github.com/uxlfoundation/oneTBB/tree/master/examples/parallel_for/seismic
can be considered as reference for the usage of a utility class for Relative Error compute.
https://github.com/uxlfoundation/oneTBB/blob/af1a128a933a2ea0c84e4b647e389585463b7db4/examples/common/utility/utility.hpp#L362

primes
sudoku
tachyon

golowanow:
oneTBB_team2 is working on this (Intel OSS Hackathon 2025)

golowanow:
#1669 is ready for review with instrumentations added to 'primes', 'sudoku', and 'tachyon' examples. 
Next step is to analyze performance statistics from these.

PR #1669, opened 2025-03-06
Example Benchmarks: calculate relative error on repeats

golowanow:
### Description 

Extend the `Tachyon` example with additional command line option `n-of-iterations` to execute ray tracing (rendering) several times and collect its performance metrics to estimate relative error (for rendering duration at the moment).

Extend the `Primes` example's `n-of-repeats` execution mode to calculate relative error for each of the parallel threads
execution block (`n-of-threads`), so it is now possible to measure their performance reliability.

Extend the `Sudoku` example with additional option to repeat solution several times and calculate relative error to estimate performance reliability.

The relative error allows to take informed decision on how reliable and reproduceable these examples are as oneTBB benchmarks, or some additional stabilization is required to achieve some KPI (e.g. 5% max).

Resolves: #1656 (part of)

### Type of change

- [ ] bug fix - _change that fixes an issue_
- [x] new feature - _change that adds functionality_
- [ ] tests - _change in tests_
- [ ] infrastructure - _change in infrastructure and CI_
- [x] documentation - _documentation update_

### Tests

- [ ] added - _required for new features and some bug fixes_
- [x] not needed

### Documentation

- [x] updated in README.md
- [ ] needs to be updated
- [ ] not needed

### Breaks backward compatibility
- [ ] Yes
- [x] No
- [ ] Unknown

### Notify the following users
@anchita20
@mehulBhandari20

### Other information
`class measurements` introduced with #1446 is moved to its own header file to make its use easier deep in example applications without conflicts with other utility classes.

Intel OSS Hackathon 2025

PR #1678, opened 2025-03-13
examples:  build targets for benchmarking 

golowanow:
### Description 

Add two new build targets to `Tachyon`, `Primes`, and `Sudoku` examples to execute them for benchmarking purposes with parameters obtained from experiments on 3 different platforms. 
Ideally, these parameters should allow to run the benchmarks in reasonable time with relative error 5% maximum.
The results can be optionally post-processed into .csv files to ease further analysis.

Resolves: #1656 (part of)

### Type of change

### Tests

- [x] not needed

### Documentation

- [x] updated in README.md files
- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### Other information
follow up for #1669

### DIFF
--- a/examples/common/cmake/common.cmake
+++ b/examples/common/cmake/common.cmake
@@ -1,4 +1,4 @@
-# Copyright (c) 2020-2021 Intel Corporation
+# Copyright (c) 2020-2025 Intel Corporation
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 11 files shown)


examples/common/cmake/common.cmake
```diff
@@ -46,3 +46,22 @@ macro(add_execution_target TARGET_NAME TARGET_DEPENDENCIES EXECUTABLE ARGS)
 
     add_dependencies(${TARGET_NAME} ${TARGET_DEPENDENCIES})
 endmacro()
+
+macro(add_benchmark_target TARGET_NAME TARGET_DEPENDENCIES EXECUTABLE ARGS)
+    find_program(AWK awk REQUIRED NO_CACHE)
```
isaevil, line 51: As far as I know, awk is not available on Windows by default. I don't think it is a good idea to make it a REQUIRED program then.
P.S. As an alternative implementation of report generation benchmarks could accept some argument `-r <path-to-csv>` and the report generation logic would implemented in the benchmarks themselves.

golowanow: > As far as I know, awk is not available on Windows by default. I don't think it is a good idea to make it a REQUIRED program then. then 

A check for gawk presence is added, so if it is not found a warning is logged and only benchmark run target is prepared without .csv compose.

> P.S. As an alternative implementation of report generation benchmarks could accept some argument `-r <path-to-csv>` and the report generation logic would implemented in the benchmarks themselves.

report compose target is intentionally proposed as an external transformation: this way it is more general following shell pipeline design and has no impact on any of the benchmarks with their different implementations. The .awk script can be easily customized/extended, e.g. to produce .json instead of .csv,  etc.

isaevil (approved):
Didn't review *.awk files in detail.

dnmokhov (approved):
LGTM

PR #1679, opened 2025-03-14
[RFC] API to simplify creation of task arenas constrained to NUMA nodes

akukanov:
This PR proposes a function to simplify creating a set of NUMA-constrained task arenas. It suggests an approach alternative to that proposed in #1559.

### DIFF
--- a/rfcs/proposed/numa_support/create-numa-arenas.md
+++ b/rfcs/proposed/numa_support/create-numa-arenas.md
@@ -87,7 +87,19 @@ std::vector<tbb::task_arena> create_numa_task_arenas(
 }
 ```
 
+### Shortcomings and downsides
+
+The following critics was provided in the RFC discussion:
+
+- It might be confusing that a single `constraints` object is used to generate multiple arenas,
+  especially with part of it (`numa_id`) being ignored.
+- The proposed API addresses just one, albeit important, use case for creating a set of arenas.
+
+See the "universal function" alternative below for related considerations.
+
 ## Considered alternatives
+
+### Sub-classing `task_arena`
 
 The earlier proposal [PR #1559](https://github.com/uxlfoundation/oneTBB/pull/1559) suggested to add
 a new class derived from `task_arena` which would also provide a method to wait for work completion.
@@ -107,9 +119,43 @@ use of a task arena and a task group to submit and wait for work. Combined, thes
 provide only a slightly more verbose solution for the NUMA use case, while being more flexible
 and having greater potential for other useful extensions and applications.
 
+### Universal function to create an arena set
+
+In the discussion of this RFC, it was suggested to generalize the function for creating a set of arenas
+with various prescribed characteristics, beyond only NUMA-bound arenas. It would take a "constraint
+generator" type which would represent multiple arena constraint objects according to given options,
+and each of these constraints would be used to create an arena.
+
+Generally, such generalized API needs some way to describe which arena parameters should vary
+across the arena set and how, and which should remain uniform. Possible ways for that are
+to use some descriptive "language" and/or to generate parameter sets by a function.
+
+In the suggestion, construction arguments of a constraint generator (including value sets
+and named patterns) serve as the description language.
+Another way could be to add named patterns as possible data values for `task_arena::constraints`;
+for example, `tbb::task_arena::constraints c{ .numa_id = tbb::task_arena::iterate }`
+would serve as a pattern for generating a set of constraints bound to NUMA domains.
+
+Using a function to generate arena parameter sets seems even more flexible comparing to a description
+language, as that function would be not limited in which parameters to alter, and how.
+
+Compared to the proposal, this approach would necessarily be more verbose because of the need to
+describe or generate a set of constraints. That could however be mitigated for common use cases
+with presets provided by the library allowing for a reasonably concise code. For example:
+```c++
+  std::vector<tbb::task_arena> arenas = tbb::create_task_arenas(tbb::iterate_over_numa_ids);
+```
+where `iterate_over_numa_ids` is a predefined variable of a type accepted by the function.
+
+Yet it is questionable if such flexibility is useful at the moment, and so if it justifies
+extra complexity. As of now, we know just a few arena set patterns with potential practical usage:
+besides NUMA arenas, we can think of splitting CPU resources by core type and of creating a set of
+arenas with different priorities. We do not however recall any requests to simplify these use cases.
+
 ## Open questions
 - Instead of a free-standing function in namespace `tbb`, should we consider
   a static member function in class `task_arena`?
 - The proposal does not consider arena priority, simply keeping the default `priority::normal`.
   Are there use cases for pre-setting priorities? Similarly for the experimental thread leave policy.
+- Are there more practical use cases which could justify the universal function approach?
 - Are there any reasons for the API to first go out as an experimental feature?
--- a/rfcs/proposed/numa_support/create-numa-arenas.md
+++ b/rfcs/proposed/numa_support/create-numa-arenas.md
@@ -61,6 +61,11 @@ It optionally takes a `constraints` argument to change default arena settings su
 is ignored. The second optional argument allows to override the number of reserved slots, which by default
 is 0 (unlike the `task_arena` construction default of 1) for the reasons described in the introduction.
 
+These arena parameters were selected for pre-setting because there appear to be practical use cases to modify
+it uniformly for the whole arena set - e.g., to suppress the use of hyper-threading or to reserve a slot
+for a dedicated application thread. For other arena parameters, such as priorities and thread leave policy,
+no obvious use cases are seen for uniformly changing default values; it can be addressed on demand.
+
 The function returns a `std::vector` of created arenas. The arenas should not be initialized,
 in order to allow changing certain arena settings before the use.
 
@@ -106,5 +111,5 @@ and having greater potential for other useful extensions and applications.
 - Instead of a free-standing function in namespace `tbb`, should we consider
   a static member function in class `task_arena`?
 - The proposal does not consider arena priority, simply keeping the default `priority::normal`.
-  Should a possibility to change priorities be considered?
+  Are there use cases for pre-setting priorities? Similarly for the experimental thread leave policy.
 - Are there any reasons for the API to first go out as an experimental feature?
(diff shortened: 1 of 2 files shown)


rfcs/proposed/numa_support/create-numa-arenas.md
```diff
@@ -0,0 +1,110 @@
...
+
+The function has the following signature:
+
+```c++
+// Defined in tbb/task_arena.h
+
+namespace tbb {
+    std::vector<tbb::task_arena> create_numa_task_arenas(
+        task_arena::constraints other_constraints = {},
+        unsigned reserved_slots = 0
```
vossmjp, line 54: You mention priorities later, but why _not_ have a default normal priority as an argument? Is there a reason to exclude it? It seems like it would naturally fall after reserved_slots as is usual for task_arena constructors?

rfcs/proposed/numa_support/create-numa-arenas.md
```diff
@@ -0,0 +1,110 @@
...
+    };
+}
+```
+
+It optionally takes a `constraints` argument to change default arena settings such as maximal concurrency
+(the upper limit on the number of threads), core type etc.; the `numa_id` value in `other_constraints`
```
vossmjp, line 60: If we elevate NUMA constraints to a named function, should we expect that core types will be next?

rfcs/proposed/numa_support/create-numa-arenas.md
```diff
@@ -0,0 +1,110 @@
...
+- Adding a special "flavour" of `task_arena` potentially increases the library learning curve and
+  might create confusion about which class to use in which conditions, and how these interoperate.
+- It seems very specialized, capable to only address specific and quite narrow set of use cases.
+
+Our proposal, instead, aims at a single usability aspect with incremental improvements/extensions
+to the existing oneTBB classes and usage patterns, leaving other aspects to complementary proposals.
+Specifically, the [Waiting in a task arena](../task_arena_waiting/readme.md) RFC improves the joint
+use of a task arena and a task group to submit and wait for work. Combined, these extensions will
+provide only a slightly more verbose solution for the NUMA use case, while being more flexible
+and having greater potential for other useful extensions and applications.
```
vossmjp, line 103: I would suggest showing a combined example here, since one of the tradeoffs between [PR #1559](https://github.com/uxlfoundation/oneTBB/pull/1559) and this proposal is verbosity vs flexibility. You have already reproduced the motivating example. PR 1559 shows its implementation of that complete motivating example. Showing the combined solutions with this proposal and [Waiting in a task arena](../task_arena_waiting/readme.md) will let us directly see how much it is simplified compared to the motivating example.

rfcs/proposed/numa_support/create-numa-arenas.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Open questions
+- Instead of a free-standing function in namespace `tbb`, should we consider
+  a static member function in class `task_arena`?
+- The proposal does not consider arena priority, simply keeping the default `priority::normal`.
```
vossmjp, line 108: See above. I'm not sure what is gained by excluding it.

rfcs/proposed/numa_support/create-numa-arenas.md
```diff
@@ -0,0 +1,110 @@
...
+  Should a possibility to change priorities be considered?
+- Are there any reasons for the API to first go out as an experimental feature?
```
vossmjp, line 110: I don't see much to gain from experimental feature first.

rfcs/proposed/numa_support/create-numa-arenas.md
aleksei-fedotov: I think we should not confuse single instance of `tbb::task_arena::constraints` with a set of them.
To be able to create a set of arenas we need to pass this function a set of constraints that would describe the hardware subset for which task arenas need to be created. At the beginning it may look something like (names are not thoroughly considered):
```cpp
    // Create arenas for every NUMA node
    auto numas = tbb::info::numa_nodes();
    tbb::constraints_generator cg1( /*numa nodes*/tbb::info::numa_nodes() );
    std::vector<task_arena> constrained_arena = tbb::create_constrained_task_arenas(cg1);
    
    // Create arenas for specific NUMA nodes
    tbb::constraints_generator cg2( {numas[0], numas[numas.size()-1]} );
    std::vector<task_arena> constrained_arena = tbb::create_constrained_task_arenas(cg2);
```
Later on we might consider extending the constraints descriptor allowing more strict filtering of hardware resources by intersecting them. For example (again, names are not thoroughly considered):
```cpp
    // Create arenas for specific NUMA nodes and efficient core types
    auto cts = tbb::info::core_types();
    tbb::constraints_generator cg3( tbb::info::numa_nodes(), {cts[0]} );
    std::vector<task_arena> constrained_arena = tbb::create_constrained_task_arenas(cg3);
    
    // Create arenas for specific core types that have L2 cache shared
    std::vector<thread_id> tids = tbb:info::shared_cache_threads(/*level=*/2);
    tbb::constraints_generator cg3( /*numa_nodes*/{no_constraint}, {cts[0]}, tids );
    std::vector<task_arena> constrained_arena = tbb::create_constrained_task_arenas(cg3);
```
I believe this is more general and hence scalable interface, while the one proposed represents its particular use.

akukanov: Yes, I thought that `create_core_type_task_arenas` and even `create_priority_task_arenas` could be added as new factory functions, if there is demand.

rfcs/proposed/numa_support/create-numa-arenas.md
```diff
@@ -0,0 +1,110 @@
...
+Specifically, the [Waiting in a task arena](../task_arena_waiting/readme.md) RFC improves the joint
+use of a task arena and a task group to submit and wait for work. Combined, these extensions will
+provide only a slightly more verbose solution for the NUMA use case, while being more flexible
+and having greater potential for other useful extensions and applications.
+
+## Open questions
+- Instead of a free-standing function in namespace `tbb`, should we consider
+  a static member function in class `task_arena`?
+- The proposal does not consider arena priority, simply keeping the default `priority::normal`.
+  Should a possibility to change priorities be considered?
```
aleksei-fedotov, line 109: We also have `leave_policy` setting as a preview feature now.

rfcs/proposed/numa_support/create-numa-arenas.md
akukanov: For now, the example can be found in the discussion of https://github.com/uxlfoundation/oneTBB/pull/1559#discussion_r1842415581 as well as in the arena waiting proposal https://github.com/uxlfoundation/oneTBB/blob/rfc/create-numa-task-arenas/rfcs/proposed/task_arena_waiting/readme.md#1-simplify-the-use-of-a-task-group.

Eventually we need to "merge" #1559 and this PR, describing one approach as the proposed solution and the other as a considered but declined alternative. I have done some initial steps for that here, but before further "merging" I would like to see some indication that this approach is generally preferred. I hope there is sufficient information available for reviewers to judge on that.

rfcs/proposed/numa_support/create-numa-arenas.md
akukanov: At this point, I do not want to complicate the API with more "syntax sugar" than practically necessary. I can see how setting other constraints (max concurrency and threads-per-core) can be useful, and I can see a scenario where NUMA arenas are served by different application threads to provide an option for reservation. However I do not easily see the case for changing the priority of all NUMA arenas; so why providing an option for that?

There is also the experimental API that adds thread leave policy as another arena construction option; should it also be a parameter to the arena set factory? And maybe we later decide to combine these options into a structure similar to arena constraints. So I would better wait now and add more overloads later, if there is demand.

rfcs/proposed/numa_support/create-numa-arenas.md
akukanov: Thanks for reminding me of it :)
To me, this is a reason not to extend the list of parameters beyond what is seen practically useful, at least for now.

rfcs/proposed/numa_support/create-numa-arenas.md
akukanov: Discussed at https://github.com/uxlfoundation/oneTBB/pull/1679#discussion_r1998920359

rfcs/proposed/numa_support/create-numa-arenas.md
akukanov: Expanded the question to mention the leave policy as well.

rfcs/proposed/numa_support/create-numa-arenas.md
akukanov: > To be able to create a set of arenas we need to pass this function a set of constraints that would describe the hardware subset for which task arenas need to be created

I have attempted to integrate Aleksei's feedback and the alternative idea into the document.

rfcs/proposed/numa_support/create-numa-arenas.md
akukanov: Since @aleksei-fedotov has supported this proposal, I added more details from #1559 as well as the code examples for comparison.

aleksei-fedotov (dismissed):

rfcs/proposed/numa_support/create-numa-arenas.md
```diff
@@ -0,0 +1,224 @@
...
+    std::vector<tbb::task_arena> create_numa_task_arenas(
+        task_arena::constraints other_constraints = {},
+        unsigned reserved_slots = 0
+    };
+}
+```
+
+It optionally takes a `constraints` argument to change default arena settings such as maximal concurrency
+(the upper limit on the number of threads), core type etc.; the `numa_id` value in `other_constraints`
+is ignored. The second optional argument allows to override the number of reserved slots, which by default
```
kboyarinov, line 82: Would it be more practical to raise the error if the user provided the `constraints` with some `numa_id` specified instead of ignoring it? For me, it seems like the user is doing something wrong in this case and it would be good to warn the user somehow.

akukanov: Then there should be some "valid" value. What would that be? `task_arena::automatic`?

If not just ignoring other values, then I would go for undefined behavior, which would mean "ignore in release builds, assert in debug builds".

vossmjp (approved):
Looks good to me as a proposal. After PoC, we will need to decide if it is shipped initially as experimental or not.

rfcs/proposed/numa_support/create-numa-arenas.md
akukanov: I have added an open for that.,

vossmjp (approved):
lgtm

aleksei-fedotov (approved):

kboyarinov (approved):

PR #1680, opened 2025-03-18
Add more hardening flags to library build

omalyshe:
### Description 
Add flags to library build with Clang and GCC:

- -fstack-clash-protection - 
- -fcf-protection=full flags - not supported in Emscripten and on aarch64.
Both are not supported in AppleClang and Clang for Android.

### Type of change

### Tests

- [X] not needed

### Documentation

- [ ] needs to be updated
- [X] not needed

- [ ] Yes
- [ ] No
- [X] Unknown


cmake/compilers/Clang.cmake
```diff
@@ -65,6 +65,10 @@ endif()
 # Clang flags to prevent compiler from optimizing out security checks
 set(TBB_COMMON_COMPILE_FLAGS ${TBB_COMMON_COMPILE_FLAGS} -Wformat -Wformat-security -Werror=format-security -fPIC $<$<NOT:$<BOOL:${EMSCRIPTEN}>>:-fstack-protector-strong>)
 
+if (NOT APPLE AND NOT ANDROID_PLATFORM)
+    set(TBB_LIB_COMPILE_FLAGS ${TBB_LIB_COMPILE_FLAGS} -fstack-clash-protection $<$<NOT:$<BOOL:${EMSCRIPTEN}>>:-fcf-protection=full>)
+endif()
```
dnmokhov, line 70: > * -fcf-protection=full flags - not supported in Emscripten and on aarch64.

Won't this be adding `-fcf-protection=full` on aarch64 on Windows? Or is that a GCC-only limitation?

cmake/compilers/GNU.cmake
```diff
@@ -105,6 +105,11 @@ endif ()
 set(TBB_COMMON_COMPILE_FLAGS ${TBB_COMMON_COMPILE_FLAGS} -fno-strict-overflow -fno-delete-null-pointer-checks -fwrapv)
 set(TBB_COMMON_COMPILE_FLAGS ${TBB_COMMON_COMPILE_FLAGS} -Wformat -Wformat-security -Werror=format-security
     -fstack-protector-strong )
+if (NOT CMAKE_SYSTEM_PROCESSOR MATCHES "aarch64")
+    set(TBB_LIB_COMPILE_FLAGS ${TBB_LIB_COMPILE_FLAGS} $<$<NOT:$<VERSION_LESS:${CMAKE_CXX_COMPILER_VERSION},8.0>>:-fcf-protection=full>)
+endif ()
```
dnmokhov, line 110: > * -fcf-protection=full flags - not supported in Emscripten and on aarch64.

Isn't a no-Emscripten check needed?

omalyshe: As I understand em++ is LLVM-Clang-based compiler, so GCC is not used for Emscripten. Let me know if it's not the case.

dnmokhov: You are right, but it seems it that it can be mixed with GCC somehow, as mentioned in #1391. Or was that just extra carefulness? @isaevil, do you know the details?

isaevil: Unfortunately, I don't know particular details about this GNU+Emscripten scenario. But as an extra carefulness, as @dnmokhov said, I would prefer to add the check.

cmake/compilers/Clang.cmake
omalyshe: > > * -fcf-protection=full flags - not supported in Emscripten and on aarch64.
> 
> Won't this be adding `-fcf-protection=full` on aarch64 on Windows? Or is that a GCC-only limitation?

Unfortunately, I can't verify it with Clang on Windows arm64.

cmake/compilers/GNU.cmake
omalyshe: Added the check.

dnmokhov (approved):
LGTM

PR #1682, opened 2025-03-21
POC implementation for task_group dynamic dependencies - part 1 - task_tracker

kboyarinov:
### Description 
First part of POC implementation for task_group dynamic dependencies. 
Design described in the RFC #1664
This PR extends the `task_group` tasks with the dynamic state, that currently allows tracking the progress of the task using the new object `task_tracker`.
In the future parts, dynamic state of the task would be reused for setting and transferring dependencies between tasks.

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown

### DIFF
--- a/include/oneapi/tbb/detail/_task_handle.h
+++ b/include/oneapi/tbb/detail/_task_handle.h
@@ -32,7 +32,81 @@ namespace d2 {
 
 class task_handle;
 
-class task_handle_task : public d1::task {
+#if __TBB_PREVIEW_TASK_GROUP_EXTENSIONS
+
+class task_with_dynamic_state;
+class task_dynamic_state {
+    enum class task_status {
+        no_status = 0,
+        submitted = 1,
+        completed = 2
+    };
+public:
+    task_dynamic_state(d1::small_object_allocator& alloc)
+        : m_task_status(task_status::no_status)
+        , m_num_references(0)
+        , m_allocator(alloc)
+    {}
+
+    void reserve() { ++m_num_references; }
+
+    void release() {
+        if (--m_num_references == 0) {
+            m_allocator.delete_object(this);
+        }
+    }
+
+    void mark_submitted() {
+        m_task_status.store(task_status::submitted, std::memory_order_release);
+    }
+
+    void complete_task() {
+        m_task_status.store(task_status::completed, std::memory_order_release);   
+    }
+
+    bool was_submitted() const {
+        return m_task_status.load(std::memory_order_acquire) >= task_status::submitted;
+    }
+
+    bool is_completed() const {
+        return m_task_status.load(std::memory_order_acquire) == task_status::completed;
+    }
+private:
+    std::atomic<task_status> m_task_status;
+    std::atomic<std::size_t> m_num_references;
+    d1::small_object_allocator m_allocator;
+};
+
+class task_with_dynamic_state : public d1::task {
+public:
+    task_with_dynamic_state(d1::small_object_allocator& alloc)
+        : m_state(nullptr)
+        , m_allocator(alloc)
+    {}
+
+    task_dynamic_state* get_dynamic_state() {
+        if (m_state == nullptr) {
+            m_state = m_allocator.new_object<task_dynamic_state>(m_allocator);
+        }
+        return m_state;
+    }
+
+    task_dynamic_state* get_dynamic_state_if_created() {
+        return m_state;
+    }
+private:
+    task_dynamic_state* m_state;
+    d1::small_object_allocator& m_allocator;
+};
+#endif // __TBB_PREVIEW_TASK_GROUP_EXTENSIONS
+
+class task_handle_task
+#if __TBB_PREVIEW_TASK_GROUP_EXTENSIONS
+: public task_with_dynamic_state
+#else
+: public d1::task
+#endif
+{
     std::uint64_t m_version_and_traits{};
     d1::wait_tree_vertex_interface* m_wait_tree_vertex;
     d1::task_group_context& m_ctx;
@@ -47,9 +121,14 @@ public:
     }
 
     task_handle_task(d1::wait_tree_vertex_interface* vertex, d1::task_group_context& ctx, d1::small_object_allocator& alloc)
-        : m_wait_tree_vertex(vertex)
+        : 
+#if __TBB_PREVIEW_TASK_GROUP_EXTENSIONS
+          task_with_dynamic_state(alloc),
+#endif
+          m_wait_tree_vertex(vertex)
         , m_ctx(ctx)
-        , m_allocator(alloc) {
+        , m_allocator(alloc)
+    {
         suppress_unused_warning(m_version_and_traits);
         m_wait_tree_vertex->reserve();
     }
@@ -64,7 +143,12 @@ public:
 
 class task_handle {
     struct task_handle_task_finalizer_t{
-        void operator()(task_handle_task* p){ p->finalize(); }
+        void operator()(task_handle_task* p){
+            p->finalize();
+#if __TBB_PREVIEW_TASK_GROUP_EXTENSIONS
+
+#endif
+        }
     };
     using handle_impl_t = std::unique_ptr<task_handle_task, task_handle_task_finalizer_t>;
 
@@ -84,8 +168,16 @@ public:
 
 private:
     friend struct task_handle_accessor;
+#if __TBB_PREVIEW_TASK_GROUP_EXTENSIONS
+    friend class task_tracker;
+#endif
 
-    task_handle(task_handle_task* t) : m_handle {t}{};
+    task_handle(task_handle_task* t) : m_handle {t} {
+#if __TBB_PREVIEW_TASK_GROUP_EXTENSIONS
+        task_dynamic_state* state = m_handle->get_dynamic_state();
+        state->reserve();
+#endif
+    }
 
     d1::task* release() {
        return m_handle.release();
@@ -93,12 +185,22 @@ private:
 };
 
 struct task_handle_accessor {
-static task_handle              construct(task_handle_task* t)  { return {t}; }
-static d1::task*                release(task_handle& th)        { return th.release(); }
-static d1::task_group_context&  ctx_of(task_handle& th)         {
-    __TBB_ASSERT(th.m_handle, "ctx_of does not expect empty task_handle.");
-    return th.m_handle->ctx();
-}
+    static task_handle construct(task_handle_task* t) { return {t}; }
+
+    static d1::task* release(task_handle& th) { return th.release(); }
+
+    static d1::task_group_context& ctx_of(task_handle& th) {
+        __TBB_ASSERT(th.m_handle, "ctx_of does not expect empty task_handle.");
+        return th.m_handle->ctx();
+    }
+
+#if __TBB_PREVIEW_TASK_GROUP_EXTENSIONS
+    static void mark_task_submitted(task_handle& th) {
+        __TBB_ASSERT(th.m_handle, "mark_task_submitted does not expect empty task_handle");
+        task_dynamic_state* state = th.m_handle->get_dynamic_state_if_created();
+        if (state) state->mark_submitted();
+    }
+#endif
 };
 
 inline bool operator==(task_handle const& th, std::nullptr_t) noexcept {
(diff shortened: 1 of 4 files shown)


include/oneapi/tbb/detail/_task_handle.h
```diff
@@ -1,5 +1,5 @@
 /*
-    Copyright (c) 2020-2024 Intel Corporation
+    Copyright (c) 2020-2025 Intel Corporation
```
vossmjp, line 2:
```suggestion
    Copyright (c) 2020-2024 Intel Corporation
    Copyright (c) 2025 UXL Foundation contributors
```

include/oneapi/tbb/detail/_task_handle.h
```diff
@@ -32,7 +32,82 @@ namespace d2 {
...
+        task_dynamic_state* current_state = m_state.load(std::memory_order_acquire);
+
+        if (current_state == nullptr) {
+            d1::small_object_allocator alloc;
+
+            task_dynamic_state* new_state = alloc.new_object<task_dynamic_state>(alloc);
+
+            if (m_state.compare_exchange_strong(current_state, new_state)) {
+                // Reserve a task co-ownership for dynamic_state
+                new_state->reserve();
```
vossmjp: Is it possible that two trackers T1 and T2 to the same handle H are constructor concurrently?  And if so, what if T2 is destroyed before T1 finishes its call to get_dynamic_state. Is it possible for the destruction of T2 to delete the task_dynamic_state before T1 has reserved it? I think T1 could succeed at exchanging a newly create state, T2 could see it and reserve and release, before T1 reserves.

test/tbb/test_task_group.cpp
```diff
@@ -1222,3 +1222,124 @@ TEST_CASE("task_handle cannot be scheduled into other task_group of the same con
 }
 
 #endif // TBB_USE_EXCEPTIONS
+
+#if __TBB_PREVIEW_TASK_GROUP_EXTENSIONS
+TEST_CASE("test task_tracker") {
```
vossmjp: Case that I think are missing:
-  is concurrent creation / destruction of trackers to same handle
- creation of tracker to deferred task but handle is destroyed without being run

include/oneapi/tbb/detail/_task_handle.h
omalyshe: > 

OK - removed my previous comment. Let's keep both copyright notices until have the guideline published. But Intel copyright year still should be 2020-202**5**

include/oneapi/tbb/detail/_task_handle.h
kboyarinov: Yes, it is possible, but when the `task_dynamic_state` creating is triggered by the `task_tracker` constructor (or by `make_edge`/ `transfer_successors_to` in parts 2 and 3), an additional reference is reserved for the task, that is represented by `H`. 
In particular, for creating a tracker, two references will be reserved - one for task and one for each tracker. A reference for a task will only be released when the task body is completed. Therefore, since we can create a new `task_tracker` only when having a `task_handle` or an other `task_tracker`, an early destructor of `task_tracker` cannot trigger the deallocation of dynamic state. 

Anyway, I will add a test case for concurrent creating of the trackers.

akukanov: So, which of the references is added here - the one for the task_handle?

And overall, if this method is the only way to create a `task_dynamic_state`, and the thread that succeeded in creation has to increment the ref count right after, why not just construct the state with the ref count of 1? Or, if other use cases are anticipated, why not add the initial refcount value as a construction argument?

kboyarinov: The reference added here is a reference for a task. I would not say it is for `task_handle` since this reference is released after completing the task, not while submitting the handle.

Regarding the constructing instead of explicit reserve - I think it can be done, I will update the PR. Thanks for catching this!

include/oneapi/tbb/detail/_task_handle.h
kboyarinov: Applied

include/oneapi/tbb/detail/_task_handle.h
kboyarinov: I have replaced the explicit `reserve` here with the construction with ref count of 1.

include/oneapi/tbb/detail/_task_handle.h
```diff
@@ -116,6 +193,104 @@ inline bool operator!=(std::nullptr_t, task_handle const& th) noexcept {
...
+        other.m_task_state = nullptr;
+    }
+
+    task_completion_handle(const task_handle& th)
+        : m_task_state(th.m_handle->get_dynamic_state())
+    {
+        __TBB_ASSERT(th, "Construction of task_completion_handle from an empty task_handle");
+        // Register new co-owner of the dynamic state
+        m_task_state->reserve();
+    }
```
akukanov:
```suggestion
    task_completion_handle(const task_handle& th)
        : m_task_state(nullptr)
    {
        __TBB_ASSERT(th, "Construction of task_completion_handle from an empty task_handle");
        m_task_state = th.m_handle->get_dynamic_state();
        // Register one more co-owner of the dynamic state
        m_task_state->reserve();
    }
```
Otherwise, a null pointer dereference may happen before the assertion.

include/oneapi/tbb/detail/_task_handle.h
```diff
@@ -116,6 +193,104 @@ inline bool operator!=(std::nullptr_t, task_handle const& th) noexcept {
...
+
+    ~task_completion_handle() {
+        if (m_task_state) m_task_state->release();
+    }
+
+    task_completion_handle& operator=(const task_completion_handle& other) {
+        if (this != &other) {
```
akukanov: Perhaps `*this != other` can be used instead, or maybe direct pointer comparison of `m_task_state` with that of `other`.

include/oneapi/tbb/detail/_task_handle.h
```diff
@@ -49,16 +78,59 @@ class task_handle_task : public d1::task {
...
     }
 
     d1::task_group_context& ctx() const { return m_ctx; }
+
+#if __TBB_PREVIEW_TASK_GROUP_EXTENSIONS
+    // Initializes the dynamic state if:
+    // * the task_completion_handle object was created
+    // * first dependency was added to a task_handle
+    // * Successors were transferred to the current task
+    task_dynamic_state* get_dynamic_state() {
```
akukanov: The comment seemingly describes the "expected" call sites for the function; that does not seem to add much value. A short description of what the function does would likely be better.

include/oneapi/tbb/detail/_task_handle.h
```diff
@@ -49,16 +78,59 @@ class task_handle_task : public d1::task {
...
+        task_dynamic_state* current_state = m_dynamic_state.load(std::memory_order_acquire);
+
+        if (current_state == nullptr) {
+            d1::small_object_allocator alloc;
+
+            task_dynamic_state* new_state = alloc.new_object<task_dynamic_state>(alloc);
```
aleksei-fedotov: Should not `task_handle_task`'s `m_allocator` be reused?

include/oneapi/tbb/detail/_task_handle.h
```diff
@@ -32,11 +33,39 @@ namespace d2 {
...
+    void complete_task() {
+    }
+private:
+    std::atomic<std::size_t> m_num_references;
+    d1::small_object_allocator m_allocator;
+};
+#endif // __TBB_PREVIEW_TASK_GROUP_EXTENSIONS
+
 class task_handle_task : public d1::task {
     std::uint64_t m_version_and_traits{};
```
isaevil, line 140: Should we set `m_version_and_traits` to some different value (for instance, 1), that would indicate, that this task supports dynamic dependencies?

include/oneapi/tbb/detail/_task_handle.h
kboyarinov: Changed to direct pointers comparison.

include/oneapi/tbb/detail/_task_handle.h
kboyarinov: Fixed

include/oneapi/tbb/detail/_task_handle.h
kboyarinov: No, we cannot reuse it since the `task_handle_task` and the associated `task_dynamic_state` can be allocated using different threads. And since small object pools are thread-local, we cannot use the pool from different thread to allocate the object.
Same applies for allocating the successor list nodes in the second part of the implementation.

include/oneapi/tbb/detail/_task_handle.h
kboyarinov: As far as I understand, `m_version_and_traits` is needed to allow the binary to distinguish tasks from the applications built with "old" and "new" versions of oneTBB. In the current implementation, all dynamic dependencies are managed on the header side. The scheduler only uses `task->execute` and `task->cancel` functions.
Hence, I do not think we should update `m_version_and_traits` for this implementation.

include/oneapi/tbb/detail/_task_handle.h
```diff
@@ -49,16 +78,56 @@ class task_handle_task : public d1::task {
...
+
+        if (current_state == nullptr) {
+            d1::small_object_allocator alloc;
+
+            task_dynamic_state* new_state = alloc.new_object<task_dynamic_state>(alloc);
+
+            if (m_dynamic_state.compare_exchange_strong(current_state, new_state)) {
+                current_state = new_state;
+            } else {
+                // Other thread created the dynamic state
```
akukanov:
```suggestion
                // CAS failed, current_state points to the dynamic state created by another thread
```
Some people, like me, tend to forget that `compare_exchange`, if failed, updates its first argument with the new value.

include/oneapi/tbb/detail/_task_handle.h
```diff
@@ -116,6 +190,108 @@ inline bool operator!=(std::nullptr_t, task_handle const& th) noexcept {
...
+        __TBB_ASSERT(th, "Assignment of task_completion_state from an empty task_handle");
+        task_dynamic_state* th_state = th.m_handle->get_dynamic_state();
+        if (m_task_state != th_state) {
+            // Release co-ownership on the previously tracked dynamic state
+            if (m_task_state) m_task_state->release();
+
+            m_task_state = th_state;
+
+            // Reserve co-ownership on the new dynamic state
+            __TBB_ASSERT(m_task_state != nullptr, "No state in the non-empty task_handle");
```
akukanov: Seems this assertion can be moved to the beginning of the function, checking `th_state` once that is obtained.

include/oneapi/tbb/detail/_task_handle.h
akukanov: Then `m_version_and_traits` is essentially a reserved field in `task_handle_task`, as it will not ever be necessary if the class is not used within the TBB binaries. Do not we need some space in the class for `task_dynamic_state`? :)

kboyarinov: I have just checked and `m_version_and_traits` is currently used in the scheduler to check if the task is proxy ((`m_version_and_traits & 1) != 0`) or resume task (`m_version_and_traits & 2) != 0`). 
I think `task::m_reserved` is a better place to store `task_dynamic_state`. Otherwise, the scheduler can think `task_handle_task` is a proxy or resume task if the address of the state is stored in the traits.

include/oneapi/tbb/detail/_task_handle.h

include/oneapi/tbb/detail/_task_handle.h
akukanov: In the scheduler a different one is used, I believe:
```
class task_traits {
    std::uint64_t m_version_and_traits{};
    friend struct r1::task_accessor;
};
...
class alignas(task_alignment) task : public task_traits { ... };
```

vossmjp: There are two m_version_and_traits, one in task_traits, accessible via task_accessor, and another one in task_handle_task.

kboyarinov: Yes, I was wrong about the `m_version_and_traits`, thank you. 

But what if we would like to use `m_version_and_traits` as intended in the future? I mean if we would like to access `task_handle_task` directly from scheduler and disambiguate between different versions of the task. If we put the dynamic state address here, we would still need an extra field for the version.
I still believe `task::m_reserved` is the better place if we don't want to increase the size of `task_handle_task`.

akukanov: Is `task_handle_task` different from other task classes in that regard? Do we have reasons to expect that in the future this class will be accessed directly in the binaries and will need versioning? With adding task dependencies, we will support most of the task management functionality of the old tasking API, yet the scheduler is agnostic of that. What else could we need that for, performance optimizations?

I am not against using `m_reserved` for a task specific payload either. But also if you say no runtime versioning is needed for the new functionality that we are adding now, I doubt it will ever be necessary, and so this space in the class layout seems to be wasted.

isaevil:
LGTM!

include/oneapi/tbb/detail/_task_handle.h
vossmjp: Decisions around reuse of other fields and versioning can be deferred until after the preview feature stage. For now, seems its ok to not reuse space and just add the additional field.

vossmjp (approved):
Looks good to me.

include/oneapi/tbb/detail/_task_handle.h
akukanov: I think with minimal code changes by reusing the traits field we can both safely test if it is a viable variant for production and avoid the class layout change for the preview feature. It looks a very good tradeoff to me. But I do not insist on that.

kboyarinov: I thought today that may be it make sense to keep `m_version_and_traits` and introduce special value indicating dependencies are supported, as Ilya initially proposed.
Consider having dependencies as a fully supported feature.
Let's say the user creates a `task_handle` in the translation unit compiled with the old version of the library (without dependencies, atomic is not present) and passes it to another translation unit compiled with the new version of the library (where dependencies are supported). The code in the second translation unit passes this `task_handle` to `set_task_order`.
Since `set_task_order` is reading the atomic that is not present, it leads to undefined behavior.
We can of course protect from this by increasing the `d` namespace version, but as we already know, it does not help if the user wrapper on top of `task_handle` is used.
Having `m_version_and_traits` indicating the task dependencies are supported will give us a bit more flexibility and allow to do something in `set_task_order` if the input handle represents the task with dependencies. For instance, add an assertion in debug without accessing the inexistent atomic.

Alternatively, we can use `task::m_version_and_traits` for this purpose. But it will be a bit more tricky to do since this field is private and is already used in the scheduler. 

What do you think?

akukanov: In the scenario you consider, there are two variations possible, depending on the implementation we choose.

With the current implementation, your analysis is correct - `set_task_order` will read a supposed-to-be atomic out of the object bounds. We can try preventing that with runtime versioning; more analysis is necessary what the API should do in that case though.

It is different if the atomic pointer replaces the version-and-traits field, which is value-initialized (https://en.cppreference.com/w/cpp/language/value_initialization.html) but unused. I think it can be treated as a null pointer, which practically makes old `task_handle_task` instances bitwise equivalent to new ones, even though formally there is no atomic pointer there. We can explore in the standard if it is still treated as UB, but at least it is predictable and testable.

include/oneapi/tbb/task_group.h
```diff
@@ -582,7 +614,14 @@ class task_group : public task_group_base {
         using acs = d2::task_handle_accessor;
         __TBB_ASSERT(&acs::ctx_of(h) == &context(), "Attempt to schedule task_handle into different task_group");
 
-        d1::spawn(*acs::release(h), context());
+        task_handle_task* task_ptr = acs::release(h);
+#if __TBB_PREVIEW_TASK_GROUP_EXTENSIONS
+        // Owned task has no dependencies or the task handle is the last dependency
```
vossmjp:
```suggestion
        // If the task has dependencies and the task_handle is not the last dependency
```

vossmjp (dismissed):
Looks good to me. Suggested one small change to a comment to make it match the other case where the same pattern appears.

PR #1685, opened 2025-03-24
POC implementation for task_group dynamic dependencies - part 2 - make_edge

kboyarinov:
### Description 
Second part of POC implementation for task_group dynamic dependencies. 
Design described in the RFC #1664 
First part is implemented in the PR #1682 

This PR extends the `task_group` API with the new method `make_edge` that allows to specify predecessor-successor dependency between two tasks. 
`task_dynamic_state` class was extended to contain current continuation vertex and the successors list. 

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown

### DIFF
--- a/include/oneapi/tbb/detail/_task_handle.h
+++ b/include/oneapi/tbb/detail/_task_handle.h
@@ -35,6 +35,10 @@ class task_handle;
 #if __TBB_PREVIEW_TASK_GROUP_EXTENSIONS
 
 class task_with_dynamic_state;
+class task_handle_task;
+class successor_vertex;
+
+inline task_with_dynamic_state* release_successors_list(successor_vertex* list);
 class task_dynamic_state {
     enum class task_status {
         no_status = 0,
@@ -42,8 +46,11 @@ class task_dynamic_state {
         completed = 2
     };
 public:
-    task_dynamic_state(d1::small_object_allocator& alloc)
-        : m_task_status(task_status::no_status)
+    task_dynamic_state(task_with_dynamic_state* task, d1::small_object_allocator& alloc)
+        : m_task(task)
+        , m_task_status(task_status::no_status)
+        , m_successors_list_head(nullptr)
+        , m_continuation_vertex(nullptr)
         , m_num_references(0)
         , m_allocator(alloc)
     {}
@@ -60,8 +67,15 @@ public:
         m_task_status.store(task_status::submitted, std::memory_order_release);
     }
 
-    void complete_task() {
+    task_with_dynamic_state* complete_task() {
         m_task_status.store(task_status::completed, std::memory_order_release);   
+
+        task_with_dynamic_state* next_task = release_successors_list(m_successors_list_head.exchange(nullptr));
+
+        // Releasing the task ownership of the dynamic_state
+        release();
+
+        return next_task;
     }
 
     bool was_submitted() const {
@@ -71,11 +85,50 @@ public:
     bool is_completed() const {
         return m_task_status.load(std::memory_order_acquire) == task_status::completed;
     }
+
+    bool has_dependencies() const {
+        return m_continuation_vertex.load(std::memory_order_acquire) != nullptr;
+    }
+    
+    void unset_dependency() {
+        m_continuation_vertex.store(nullptr, std::memory_order_release);
+    }
+
+    void add_successor(successor_vertex* successor);
+    void release_continuation();
+    
+    successor_vertex* get_continuation_vertex() {
+        successor_vertex* current_continuation_vertex = m_continuation_vertex.load(std::memory_order_acquire);
+
+        if (current_continuation_vertex == nullptr) {
+            d1::small_object_allocator alloc;
+            successor_vertex* new_continuation_vertex = alloc.new_object<successor_vertex>(m_task, alloc);
+
+            if (!m_continuation_vertex.compare_exchange_strong(current_continuation_vertex, new_continuation_vertex)) {
+                // Other thread already created a continuation vertex
+                alloc.delete_object(new_continuation_vertex);
+            } else {
+                // This thread has updated the continuation vertex
+                current_continuation_vertex = new_continuation_vertex;
+            }
+        }
+        __TBB_ASSERT(current_continuation_vertex != nullptr, "Failed to get continuation vertex");
+        return current_continuation_vertex;
+    }
+
 private:
+    task_with_dynamic_state* m_task;
     std::atomic<task_status> m_task_status;
+    std::atomic<successor_vertex*> m_successors_list_head;
+    std::atomic<successor_vertex*> m_continuation_vertex;
     std::atomic<std::size_t> m_num_references;
     d1::small_object_allocator m_allocator;
 };
+
+inline void internal_make_edge(task_dynamic_state* pred, task_dynamic_state* succ) {
+    __TBB_ASSERT(pred != nullptr && succ != nullptr , nullptr);
+    pred->add_successor(succ->get_continuation_vertex());
+}
 
 class task_with_dynamic_state : public d1::task {
 public:
@@ -86,7 +139,7 @@ public:
 
     task_dynamic_state* get_dynamic_state() {
         if (m_state == nullptr) {
-            m_state = m_allocator.new_object<task_dynamic_state>(m_allocator);
+            m_state = m_allocator.new_object<task_dynamic_state>(this, m_allocator);
         }
         return m_state;
     }
@@ -98,6 +151,44 @@ private:
     task_dynamic_state* m_state;
     d1::small_object_allocator& m_allocator;
 };
+
+class successor_vertex : public d1::reference_vertex {
+public:
+    successor_vertex(task_with_dynamic_state* task, d1::small_object_allocator& alloc)
+        // Reserving 1 for task_handle that owns the task for which the predecessors are added
+        // reference counter would be released when this task_handle would be submitted for execution
+        : d1::reference_vertex(nullptr, 1)
+        , m_task(task)
+        , m_next_successor(nullptr)
+        , m_allocator(alloc)
+    {}
+
+    task_with_dynamic_state* release_bypass(std::uint32_t delta = 1) {
+        task_with_dynamic_state* next_task = nullptr;
+
+        std::uint32_t ref = m_ref_count.fetch_sub(static_cast<std::uint64_t>(delta)) - static_cast<uint64_t>(delta);
+
+        if (ref == 0) {
+            // TODO: can we skip this step since the task would be spawned anyway
+            m_task->get_dynamic_state()->unset_dependency();
+            next_task = m_task;
+            m_allocator.delete_object(this);
+        }
+        return next_task;
+    }
+
+    void set_next_successor(successor_vertex* next) {
+        m_next_successor = next;
+    }
+
+    successor_vertex* get_next_successor() const {
+        return m_next_successor;
+    }
+private:
+    task_with_dynamic_state* m_task;
+    successor_vertex* m_next_successor;
+    d1::small_object_allocator m_allocator;
+}; // class successor_vertex
 #endif // __TBB_PREVIEW_TASK_GROUP_EXTENSIONS
 
 class task_handle_task
@@ -140,14 +231,13 @@ public:
     d1::task_group_context& ctx() const { return m_ctx; }
 };
 
-
 class task_handle {
     struct task_handle_task_finalizer_t{
         void operator()(task_handle_task* p){
-            p->finalize();
 #if __TBB_PREVIEW_TASK_GROUP_EXTENSIONS
-
+            p->get_dynamic_state()->release();
 #endif
+            p->finalize();
         }
     };
     using handle_impl_t = std::unique_ptr<task_handle_task, task_handle_task_finalizer_t>;
@@ -174,8 +264,7 @@ private:
 
     task_handle(task_handle_task* t) : m_handle {t} {
 #if __TBB_PREVIEW_TASK_GROUP_EXTENSIONS
-        task_dynamic_state* state = m_handle->get_dynamic_state();
-        state->reserve();
+        m_handle->get_dynamic_state()->reserve();
 #endif
     }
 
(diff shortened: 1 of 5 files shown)


include/oneapi/tbb/detail/_task_handle.h
```diff
@@ -191,6 +291,60 @@ inline bool operator!=(std::nullptr_t, task_handle const& th) noexcept {
...
+        successor_list_node* next_node = node->get_next_node();
+        task_handle_task* successor_task = node->get_continuation()->release_dependency();
+        node->finalize();
+        node = next_node;
+
+        if (successor_task) {
+            if (next_task == nullptr) {
+                next_task = successor_task;
+            } else {
+                d1::spawn(*successor_task, successor_task->ctx());
```
isaevil: It seems like it may lead to the situation, where task with a dependency was initially submitted into one task arena (using `enqueue` or just invoking task group interface inside the task arena), but was actually submitted into another task arena (the one where `release_successor_list` is invoked).

And this algorithm also breaks "mandatory concurrency" guarantee, if a task was initially submitted into arena using `enqueue`, because `spawn` will put task into regular task pool rather than FIFO task stream and `advertise_new_work` won't request extra "mandatory" worker thread from market.

I think we need some way of preserving the state of initial operation with a task (operation itself and a task arena where the task was submitted).

kboyarinov: We could save task arena pointer in `task_dynamic_state` to indicate that the task was submitted using `enqueue`, but the trickiest part is when the task was submitted using `this_task_arena::enqueue` and hence the task should be enqueued into the implicit arena. Since the thread that submits the task and the thread that releases the successor list can belong to different arena, we still need to save the pointer to the "correct" implicit arena. 
Current TBB ABI do not allow access to implicit arenas, so it may lead to introducing new entry points.

include/oneapi/tbb/detail/_task_handle.h
```diff
@@ -121,16 +197,32 @@ class task_handle_task : public d1::task {
         return current_state;
     }
 
-    void complete_task() {
+    task_handle_task* complete_task() {
```
isaevil: Since the class is already a "task", maybe rename this method to something like `complete_and_try_bypass` to better reflect what it does?

include/oneapi/tbb/detail/_task_handle.h
```diff
@@ -191,6 +291,60 @@ inline bool operator!=(std::nullptr_t, task_handle const& th) noexcept {
 }
 
 #if __TBB_PREVIEW_TASK_GROUP_EXTENSIONS
+inline task_handle_task* release_successor_list(successor_list_node* node) {
```
isaevil: Consider an alternative name `submit_successor_tasks`.

include/oneapi/tbb/detail/_task_handle.h
```diff
@@ -50,9 +87,48 @@ class task_dynamic_state {
...
+            ++m_num_dependencies;
+        }
     }
+
+    task_handle_task* release_dependency() {
+        task_handle_task* next_task = nullptr;
+        if (--m_num_dependencies == 0) {
+            next_task = m_task;
+        }
+        return next_task;
```
isaevil: It returns an actual task associated with dynamic state, not some kind of bypass task, right? I think `next_task` name is a little bit misleading.

include/oneapi/tbb/detail/_task_handle.h
```diff
@@ -121,16 +197,32 @@ class task_handle_task : public d1::task {
...
+    }
+
+    task_handle_task* release_dependency() {
+        task_dynamic_state* current_state = m_dynamic_state.load(std::memory_order_relaxed);
+        __TBB_ASSERT(current_state != nullptr && current_state->has_dependencies(),
+                     "release_dependency was called for task without dependencies");
+        task_handle_task* t = current_state->release_dependency();
+        __TBB_ASSERT(t == nullptr || t == this, nullptr);
+        return t;
+    }
```
akukanov: The API that returns either `this` or `nullptr` seems confusing to me. Why does it not return a bool then? Or perhaps the number of dependencies left?

include/oneapi/tbb/task_arena.h
```diff
@@ -109,7 +110,13 @@ inline void enqueue_impl(task_handle&& th, d1::task_arena_base* ta) {
...
 
     // Do not access th after release
-    r1::enqueue(*task_handle_accessor::release(th), ctx, ta);
+    task_handle_task* t = task_handle_accessor::release(th);
+#if __TBB_PREVIEW_TASK_GROUP_EXTENSIONS
+    if (!t->has_dependencies() || t->release_dependency() != nullptr)
+#endif
+    {
+        r1::enqueue(*t, ctx, ta);
+    }
```
akukanov:
```suggestion
#if __TBB_PREVIEW_TASK_GROUP_EXTENSIONS
    if (t->has_dependencies() && t->release_dependency() == nullptr) {
        return;
    }
#endif
    r1::enqueue(*t, ctx, ta);
```
Less negations, easier to understand.

include/oneapi/tbb/detail/_task_handle.h
```diff
@@ -35,10 +35,47 @@ class task_handle;
...
+public:
+    successor_list_node(task_dynamic_state* continuation, d1::small_object_allocator& alloc)
+        : m_next_node(nullptr)
+        , m_continuation(continuation)
+        , m_allocator(alloc)
+    {}
+
+    task_dynamic_state* get_continuation() const {
+        return m_continuation;
+    }
```
akukanov: Use of the term 'continuation' seemingly as a synonym to 'successor' is somewhat confusing. Why not calling it e.g. `m_successor_state` and `get_successor_state`?

include/oneapi/tbb/detail/_task_handle.h
kboyarinov: `submit_successor_tasks` sounds like the function always submits the tasks in the list. 
The idea of `release_successor_list` is that the function releases dependency in each successor in the list and submits only if the released dependency was the last. We can rename it to `release_successors` if it is more clear that releasing is made with nodes in the list, not with the list itself.

isaevil: `release_succesors` or something like `release_successors_dependencies` sound good to me.

include/oneapi/tbb/detail/_task_handle.h
```diff
@@ -50,9 +87,48 @@ class task_dynamic_state {
...
+
+    void add_successor(task_dynamic_state* successor);
+    void add_successor_node(successor_list_node* new_successor_node, successor_list_node* current_successor_list_head);
+
+    using successor_list_state_flag = std::uintptr_t;
+    static constexpr successor_list_state_flag COMPLETED_FLAG = ~std::uintptr_t(0);
+
+    static bool is_completed(successor_list_node* node) {
+        return node == reinterpret_cast<successor_list_node*>(COMPLETED_FLAG);
+    }
```
akukanov: I would change the name to better match the semantics. The function does not check if its argument is completed; it checks if the argument is a completion mark which indicates completion of the related task. But there is no task in expressions where `is_completed` is used.

As an option, I would suggest naming it `indicates_task_completion`, and likely there are other good choices as well.

include/oneapi/tbb/detail/_task_handle.h
kboyarinov: Renamed to `complete_and_try_bypass_successor`.

include/oneapi/tbb/detail/_task_handle.h
kboyarinov: Renamed to `release_successors`

include/oneapi/tbb/detail/_task_handle.h
kboyarinov: This part of code was rewritten.

include/oneapi/tbb/detail/_task_handle.h
kboyarinov: Changed this function to return bool, since having the exact number of dependencies outside of the function seems meaningless.

include/oneapi/tbb/task_arena.h
kboyarinov: Applied for `enqueue` and other submission functions

include/oneapi/tbb/detail/_task_handle.h

include/oneapi/tbb/detail/_task_handle.h
```diff
@@ -35,10 +35,47 @@ class task_handle;
...
+    void finalize() {
+        m_allocator.delete_object(this);
+    }
+private:
+    successor_list_node* m_next_node;
+    task_dynamic_state* m_successor_state;
+    d1::small_object_allocator m_allocator;
+};
+
+inline task_handle_task* release_successors(successor_list_node* list);
```
akukanov: Why is `release_successors` a free-standing function? I only see it used once in the implementation of `task_dynamic_state::complete_task`. If that is the only intended use, I would not even bother creating that function.

include/oneapi/tbb/detail/_task_handle.h
kboyarinov: Renamed to `represents_completed_task`

include/oneapi/tbb/detail/_task_handle.h
kboyarinov: Yes, you are right. The function was removed.

include/oneapi/tbb/detail/_task_handle.h
kboyarinov: The function was removed:)

include/oneapi/tbb/detail/_task_handle.h
kboyarinov: Opened a PR to add this issue to the RFC #1817

include/oneapi/tbb/detail/_task_handle.h
```diff
@@ -35,10 +35,45 @@ class task_handle;
 
 #if __TBB_PREVIEW_TASK_GROUP_EXTENSIONS
 
+class task_handle_task;
+class task_dynamic_state;
+
+class successor_list_node {
```
akukanov: `successor_list_node` is a simple internal class which API mostly consists of getters and setters. I wonder if encapsulation is really worth it in this case and whether we could just make it a struct with public fields instead.
```c++
struct successor_list_node {
    successor_list_node* next_node = nullptr;
    task_dynamic_state* successor_state = nullptr;
    d1::small_object_allocator allocator;

    successor_list_node(task_dynamic_state* state, d1::small_object_allocator& alloc)
        : successor_state(state), allocator(alloc)
    {}
    void destroy() {
        allocator.delete_object(this);
    }
};
```
No strong opinion on this, feel free to ignore.

include/oneapi/tbb/detail/_task_handle.h
```diff
@@ -50,9 +85,44 @@ class task_dynamic_state {
...
+    bool release_dependency() {
+        auto updated_dependency_counter = --m_num_dependencies;
+        return updated_dependency_counter == 0;
+    }
+
+    task_handle_task* complete_and_try_bypass_successor();
+
+    bool has_dependencies() const {
+        return m_num_dependencies.load(std::memory_order_acquire) != 0;
+    }
```
akukanov:
```suggestion
    bool has_dependencies() const {
        return m_num_dependencies.load(std::memory_order_acquire) != 0;
    }

    task_handle_task* complete_and_try_bypass_successor();
```
Swapping these two methods in the declaration seems to give a more logical grouping.

include/oneapi/tbb/task_group.h
```diff
@@ -494,7 +514,16 @@ class task_group_base : no_copy {
 
         bool cancellation_status = false;
         try_call([&] {
-            execute_and_wait(*acs::release(h), context(), m_wait_vertex.get_context(), context());
+            task_handle_task* t = acs::release(h);
```
akukanov: Please use consistent variable naming (either `task_ptr` or `t`) across various submission functions where you add it.

include/oneapi/tbb/detail/_task_handle.h
```diff
@@ -121,16 +191,30 @@ class task_handle_task : public d1::task {
...
         }
+        return next_task;
+    }
+
+    bool release_dependency() {
+        task_dynamic_state* current_state = m_dynamic_state.load(std::memory_order_relaxed);
+        __TBB_ASSERT(current_state != nullptr && current_state->has_dependencies(),
+                     "release_dependency was called for task without dependencies");
+        return current_state->release_dependency();
+    }
```
akukanov, line 196: Please add a comment describing the return values; ditto for `task_dynamic_state::release_dependency`.

include/oneapi/tbb/task_group.h
```diff
@@ -87,12 +100,13 @@ class function_task : public task_handle_task  {
 private:
     d1::task* execute(d1::execution_data& ed) override {
         __TBB_ASSERT(ed.context == &this->ctx(), "The task group context should be used for all tasks");
-        task* res = task_ptr_or_nullptr(m_func);
+        task* next_task = task_ptr_or_nullptr(m_func);
 #if __TBB_PREVIEW_TASK_GROUP_EXTENSIONS
-        this->complete_task();
+        task_handle_task* successor_task = this->complete_and_try_bypass_successor();
+        next_task = combine_tasks(next_task, successor_task);
```
akukanov: I think you do not really need `combine_tasks` as a function. I personally have found it more confusing than helpful.
The logic is not that complex, and comments might further simplify understanding.
```c++
if (next_task) {
    if (successor_task) d1::spawn(*successor_task, successor_task->ctx());
} else {
    next_task = successor_task;
}
```

include/oneapi/tbb/detail/_task_handle.h
```diff
@@ -121,16 +191,30 @@ class task_handle_task : public d1::task {
         return current_state;
     }
 
-    void complete_task() {
+    task_handle_task* complete_and_try_bypass_successor() {
```
akukanov: It's nitpicking, but I would rename it as `complete_and_try_get_successor` for both classes. Bypassing is is an action of returning a task directly to avoid excessive spawn & get for the task, it happens in `execute` and there is no "try" component there.

include/oneapi/tbb/detail/_task_handle.h
kboyarinov: Agree, let's simplify this. Done.

include/oneapi/tbb/task_group.h

include/oneapi/tbb/detail/_task_handle.h
kboyarinov: Done

include/oneapi/tbb/task_group.h
kboyarinov: Removed

include/oneapi/tbb/detail/_task_handle.h
kboyarinov: Renamed

PR #1694, opened 2025-04-06
Bazel support: Bump platform dependency

Vertexwahn:
This PR affects only the Bazel build support of oneTBB (CMake build stays untouched). In detail:

- Bump Bazel version from 7.2.1 to 8.1.1
- Improve visibility of targets
- Bump platforms to 0.0.11
- Make use of `rules_cc`
- Delete not needed `MODULE.bazel.lock` file
- Make use of `rules_license`


isaevil (approved):
LGTM

PR #1717, opened 2025-04-28
Propose NUMA-node allocator and memory resource

aleksei-fedotov:
_subj._


Alexandr-Konovalov:
Basic NUMA-aware data structure is an array with static interleaving between NUMA nodes (with granularity of page or user-defined). If we want to support such kind of structures, the memory provider should be extended for support of interleaving.

PR #1723, opened 2025-05-05
Reduce execution time of test_parallel_phase

isaevil:
### Description 

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/test/tbb/test_parallel_phase.cpp
+++ b/test/tbb/test_parallel_phase.cpp
@@ -45,6 +45,7 @@ struct dummy_func {
 template <typename F1 = dummy_func, typename F2 = dummy_func>
 std::size_t measure_median_start_time(tbb::task_arena* ta, const F1& start = F1{}, const F2& end = F2{}) {
     std::size_t num_threads = ta ? ta->max_concurrency() : tbb::this_task_arena::max_concurrency();
+    // TODO: propagate worker threads blocking time to the test
     std::size_t max_wait = 1000;
     std::size_t num_runs = 100;
     std::vector<std::size_t> longest_start_times;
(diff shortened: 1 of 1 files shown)


test/tbb/test_parallel_phase.cpp
```diff
@@ -45,7 +45,8 @@ struct dummy_func {
 template <typename F1 = dummy_func, typename F2 = dummy_func>
 std::size_t measure_median_start_time(tbb::task_arena* ta, const F1& start = F1{}, const F2& end = F2{}) {
     std::size_t num_threads = ta ? ta->max_concurrency() : tbb::this_task_arena::max_concurrency();
-    std::size_t num_runs = 500;
+    std::size_t max_wait = 1000;
+    std::size_t num_runs = 100;
```
aleksei-fedotov, line 49: Will something like the following not lose much on the testing quality and its stability while reducing the testing time even more?
```cpp
const std::size_t waitings = [0, 10, 50, 100, 500, 1000, 2000]`;
for (auto const& i : waitings) {
   if (ta) {
       ta->execute(work);
   } else {
       work();
   }
   active_wait_for(std::chrono::microseconds(i));
}
```

isaevil: 1. I'm leaning towards having more steps so test would output more accurate data on worker threads' new work start time.
2. 2000 waiting time won't make much sense here because it will exceed worker threads block time, meaning there will be no difference in start times between `fast_leave` and `automatic` policies.

aleksei-fedotov: 1. Okay.
2. Does it make sense to propagate the worker threads block time to the test as `max_wait` then? This will automatically cover the cases if the block time changes.

isaevil: Yes, it makes sense. Added the corresponding TODO.

aleksei-fedotov (approved):
Looks good to me!

PR #1729, opened 2025-05-14
Fix for overwrite_node to continue_node accounting

vossmjp:
### Description 
`overwrite_node` and `write_once_node` may send messages immediately to new successors before registering as a predecessor of the node. This can cause the predecessor count to be too low in a `continue_node` and therefore cause the node to trigger prematurely.

### Type of change

- [ X] bug fix - _change that fixes an issue_

### Tests

- [X ] added - _required for new features and some bug fixes_
- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [X ] No
- [ ] Unknown


test/tbb/test_overwrite_node.cpp
```diff
@@ -423,3 +423,39 @@ TEST_CASE("test overwrite_node try_put_and_wait") {
     test_overwrite_node_try_put_and_wait();
 }
 #endif
+
+//! Test for cancel register_predecessor_task
```
vossmjp, line 427:
```suggestion
//! Test for nested make_edge from predecessor of continue_node that adds edge to overwrite_node
```

test/tbb/test_overwrite_node.cpp
```diff
@@ -423,3 +423,39 @@ TEST_CASE("test overwrite_node try_put_and_wait") {
...
 #endif
+
+//! Test for nested make_edge from predecessor of continue_node that adds edge to overwrite_node
+//! \brief \ref error_guessing
+TEST_CASE("Nested make_edge to continue_node") {
+    tbb::flow::graph g;
+
+    using msg_t = tbb::flow::continue_msg;
+    using cnode_t = tbb::flow::continue_node<msg_t>;
+    using wnode_t = tbb::flow::write_once_node<msg_t>;
```
kboyarinov, line 434: Should we test both `overwrite` and `write_once_node` here?

include/oneapi/tbb/flow_graph.h
```diff
@@ -3318,6 +3318,11 @@ class overwrite_node : public graph_node, public receiver<T>, public sender<T> {
         spin_mutex::scoped_lock l( my_mutex );
         if (my_buffer_is_valid && is_graph_active( my_graph )) {
             // We have a valid value that must be forwarded immediately.
+            if (dynamic_cast<continue_receiver*>(&s)) {
```
kboyarinov, line 3321: I think we can use an existing method for this:

```suggestion
            if (s.is_continue_receiver()) {
```

include/oneapi/tbb/flow_graph.h
```diff
@@ -3318,6 +3318,11 @@ class overwrite_node : public graph_node, public receiver<T>, public sender<T> {
...
+                my_successors.register_successor( s );
+                s.try_put( my_buffer );
```
kboyarinov, line 3323: I think it makes sense to add a comment that we need to increase predecessors count in continue_node before putting the element. 
I also though about adding an assertion that `try_put` can't fail, but I am not sure it makes a lot of sense.

test/tbb/test_overwrite_node.cpp
vossmjp: Yes. If we agree the test is good then I can do the same for `overwrite_node`. `write_once_node` inherits from `overwrite_node` so its the same function, but even so, we should do both.

test/tbb/test_write_once_node.cpp
```diff
@@ -380,3 +380,11 @@ TEST_CASE("test write_once_node try_put_and_wait") {
...
 }
 #endif
+
+//! Test for nested make_edge from predecessor of continue_node that adds edge to write_once_node
+//! \brief \ref error_guessing
+TEST_CASE("Nested make_edge write_once_node to continue_node") {
+    using msg_t = tbb::flow::continue_msg;
+    using buffer_t = tbb::flow::write_once_node<msg_t>;
+    test_nested_make_edge_single_item_buffer_to_continue_receiver<buffer_t>();
+}
```
kboyarinov, line 390:
```suggestion
}

```

test/tbb/test_overwrite_node.cpp
```diff
@@ -423,3 +423,39 @@ TEST_CASE("test overwrite_node try_put_and_wait") {
...
+            tbb::flow::make_edge(w, execute_one_time);
+            return m;
+        }};
+
+    tbb::flow::make_edge(edge_adder, execute_one_time);
+    edge_adder.try_put(msg_t{});
+    g.wait_for_all();
+    execute_one_time.try_put(msg_t{}); // not enough with 2 predecessors
+    g.wait_for_all();
+    CHECK_MESSAGE ((count == 1), "node should only execute once");
```
kboyarinov, line 460: Should common test be used instead?

```suggestion
    using msg_t = tbb::flow::continue_msg;
    using buffer_t = tbb::flow::overwrite_node<msg_t>;
    test_nested_make_edge_single_item_buffer_to_continue_receiver<buffer_t>();
```

PR #1734, opened 2025-05-20
Flow Graph deduction guides simplification + support for newly added body types

kboyarinov:
### Description 
Simplification for deduction guides for Flow Graph functional nodes to avoid body  types detection twice.
Added support for bodies expressed as pointers to member functions and member objects (supported by std::invoke).
+ tests extensions

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown

### DIFF
--- a/include/oneapi/tbb/detail/_flow_graph_nodes_deduction.h
+++ b/include/oneapi/tbb/detail/_flow_graph_nodes_deduction.h
@@ -25,29 +25,29 @@ namespace d2 {
 
 template <typename Input>
 struct declare_input_type {
-    using input_type = Input;
+    using input_type = std::decay_t<Input>;
 };
 
 template <>
 struct declare_input_type<d1::flow_control> {};
 
 template <typename Input, typename Output>
-struct body_types : declare_input_type<std::decay_t<Input>> {
+struct body_types : declare_input_type<Input> {
     using output_type = std::decay_t<Output>;
 };
 
 template <typename P>
-struct extract_member_function_types;
+struct extract_callable_object_types;
 
 template <typename Body, typename Input, typename Output>
-struct extract_member_function_types<Output (Body::*)(Input)> : body_types<Input, Output> {};
+struct extract_callable_object_types<Output (Body::*)(Input)> : body_types<Input, Output> {};
 
 template <typename Body, typename Input, typename Output>
-struct extract_member_function_types<Output (Body::*)(Input) const> : body_types<Input, Output> {};
+struct extract_callable_object_types<Output (Body::*)(Input) const> : body_types<Input, Output> {};
 
 // Body is represented as a callable object - extract types from the pointer to operator()
 template <typename Body>
-struct extract_body_types : extract_member_function_types<decltype(&Body::operator())> {};
+struct extract_body_types : extract_callable_object_types<decltype(&Body::operator())> {};
 
 // Body is represented as a pointer to function
 template <typename Input, typename Output>
@@ -55,7 +55,7 @@ struct extract_body_types<Output (*)(Input)> : body_types<Input, Output> {};
 
 // Body is represented as a pointer to member function
 template <typename Input, typename Output>
-struct extract_body_types<Output (Input::*)()> : body_types<Input, Output> {};
+struct extract_body_types<Output (Input::*)() const> : body_types<Input, Output> {};
 
 // Body is represented as a pointer to member object
 template <typename Input, typename Output>
--- a/include/oneapi/tbb/detail/_flow_graph_nodes_deduction.h
+++ b/include/oneapi/tbb/detail/_flow_graph_nodes_deduction.h
@@ -23,83 +23,55 @@ namespace tbb {
 namespace detail {
 namespace d2 {
 
-template <typename Input, typename Output>
-struct declare_body_types {
+template <typename Input>
+struct declare_input_type {
     using input_type = Input;
-    using output_type = Output;
 };
 
-struct NoInputBody {};
+template <>
+struct declare_input_type<d1::flow_control> {};
 
-template <typename Output>
-struct declare_body_types<NoInputBody, Output> {
-    using output_type = Output;
+template <typename Input, typename Output>
+struct body_types : declare_input_type<std::decay_t<Input>> {
+    using output_type = std::decay_t<Output>;
 };
 
-template <typename T> struct body_types;
+template <typename P>
+struct extract_member_function_types;
 
-template <typename T, typename Input, typename Output>
-struct body_types<Output (T::*)(const Input&) const> : declare_body_types<Input, Output> {};
+template <typename Body, typename Input, typename Output>
+struct extract_member_function_types<Output (Body::*)(Input)> : body_types<Input, Output> {};
 
-template <typename T, typename Input, typename Output>
-struct body_types<Output (T::*)(const Input&)> : declare_body_types<Input, Output> {};
+template <typename Body, typename Input, typename Output>
+struct extract_member_function_types<Output (Body::*)(Input) const> : body_types<Input, Output> {};
 
-template <typename T, typename Input, typename Output>
-struct body_types<Output (T::*)(Input&) const> : declare_body_types<Input, Output> {};
+// Body is represented as a callable object - extract types from the pointer to operator()
+template <typename Body>
+struct extract_body_types : extract_member_function_types<decltype(&Body::operator())> {};
 
-template <typename T, typename Input, typename Output>
-struct body_types<Output (T::*)(Input&)> : declare_body_types<Input, Output> {};
-
-template <typename T, typename Output>
-struct body_types<Output (T::*)(d1::flow_control&) const> : declare_body_types<NoInputBody, Output> {};
-
-template <typename T, typename Output>
-struct body_types<Output (T::*)(d1::flow_control&)> : declare_body_types<NoInputBody, Output> {};
-
+// Body is represented as a pointer to function
 template <typename Input, typename Output>
-struct body_types<Output (*)(Input&)> : declare_body_types<Input, Output> {};
+struct extract_body_types<Output (*)(Input)> : body_types<Input, Output> {};
 
+// Body is represented as a pointer to member function
 template <typename Input, typename Output>
-struct body_types<Output (*)(const Input&)> : declare_body_types<Input, Output> {};
+struct extract_body_types<Output (Input::*)()> : body_types<Input, Output> {};
 
-template <typename Output>
-struct body_types<Output (*)(d1::flow_control&)> : declare_body_types<NoInputBody, Output> {};
+// Body is represented as a pointer to member object
+template <typename Input, typename Output>
+struct extract_body_types<Output Input::*> : body_types<Input, Output> {};
 
 template <typename Body>
-using input_t = typename body_types<Body>::input_type;
+using input_type = typename extract_body_types<Body>::input_type;
 
 template <typename Body>
-using output_t = typename body_types<Body>::output_type;
-
-template <typename T, typename Input, typename Output>
-auto decide_on_operator_overload(Output (T::*name)(const Input&) const)->decltype(name);
-
-template <typename T, typename Input, typename Output>
-auto decide_on_operator_overload(Output (T::*name)(const Input&))->decltype(name);
-
-template <typename T, typename Input, typename Output>
-auto decide_on_operator_overload(Output (T::*name)(Input&) const)->decltype(name);
-
-template <typename T, typename Input, typename Output>
-auto decide_on_operator_overload(Output (T::*name)(Input&))->decltype(name);
-
-template <typename Input, typename Output>
-auto decide_on_operator_overload(Output (*name)(const Input&))->decltype(name);
-
-template <typename Input, typename Output>
-auto decide_on_operator_overload(Output (*name)(Input&))->decltype(name);
-
-template <typename Body>
-decltype(decide_on_operator_overload(&Body::operator())) decide_on_callable_type(int);
-
-template <typename Body>
-decltype(decide_on_operator_overload(std::declval<Body>())) decide_on_callable_type(...);
+using output_type = typename extract_body_types<Body>::output_type;
 
 // Deduction guides for Flow Graph nodes
 
 template <typename GraphOrSet, typename Body>
 input_node(GraphOrSet&&, Body)
-->input_node<output_t<decltype(decide_on_callable_type<Body>(0))>>;
+->input_node<output_type<Body>>;
 
 #if __TBB_PREVIEW_FLOW_GRAPH_NODE_SET
 
@@ -134,7 +106,7 @@ queue_node(const NodeSet&)
 
 template <typename GraphOrProxy, typename Sequencer>
 sequencer_node(GraphOrProxy&&, Sequencer)
-->sequencer_node<input_t<decltype(decide_on_callable_type<Sequencer>(0))>>;
+->sequencer_node<input_type<Sequencer>>;
 
 #if __TBB_PREVIEW_FLOW_GRAPH_NODE_SET
 template <typename NodeSet, typename Compare>
@@ -181,17 +153,14 @@ join_node(const node_set<order::preceding, Successor, Successors...>)
 
 template <typename GraphOrProxy, typename Body, typename... Bodies>
 join_node(GraphOrProxy&&, Body, Bodies...)
-->join_node<std::tuple<input_t<decltype(decide_on_callable_type<Body>(0))>,
-                       input_t<decltype(decide_on_callable_type<Bodies>(0))>...>,
-            key_matching<join_key_t<output_t<decltype(decide_on_callable_type<Body>(0))>>>>;
+->join_node<std::tuple<input_type<Body>, input_type<Bodies>...>,
+            key_matching<join_key_t<output_type<Body>>>>;
 
 #if __TBB_PREVIEW_FLOW_GRAPH_NODE_SET
 template <typename... Predecessors>
 indexer_node(const node_set<order::following, Predecessors...>&)
 ->indexer_node<typename Predecessors::output_type...>;
-#endif
 
-#if __TBB_PREVIEW_FLOW_GRAPH_NODE_SET
 template <typename NodeSet>
 limiter_node(const NodeSet&, size_t)
 ->limiter_node<decide_on_set_t<NodeSet>>;
@@ -210,16 +179,12 @@ template <typename GraphOrSet, typename Body, typename Policy>
 function_node(GraphOrSet&&,
               size_t, Body,
               Policy, node_priority_t = no_priority)
-->function_node<input_t<decltype(decide_on_callable_type<Body>(0))>,
-                output_t<decltype(decide_on_callable_type<Body>(0))>,
-                Policy>;
+->function_node<input_type<Body>, output_type<Body>, Policy>;
 
 template <typename GraphOrSet, typename Body>
 function_node(GraphOrSet&&, size_t,
               Body, node_priority_t = no_priority)
-->function_node<input_t<decltype(decide_on_callable_type<Body>(0))>,
-                output_t<decltype(decide_on_callable_type<Body>(0))>,
-                queueing>;
+->function_node<input_type<Body>, output_type<Body>, queueing>;
 
 template <typename Output>
 struct continue_output {
--- a/test/conformance/conformance_flowgraph.h
+++ b/test/conformance/conformance_flowgraph.h
@@ -811,4 +811,65 @@ void test_with_reserving_join_node_class() {
         if at least one successor accepts the tuple must consume messages");
 }
 }
+
+#if __TBB_CPP17_DEDUCTION_GUIDES_PRESENT
+namespace deduction_guides_testing {
+
+template <typename Input, typename Output>
+struct callable_object_body_base {
+    using input_type = std::decay_t<Input>;
+    using output_type = std::decay_t<Output>;
+};
+
+template <typename Input, typename Output, bool IsConst>
+struct callable_object_body : callable_object_body_base<Input, Output> {
+    Output operator()(Input) { return Output{}; }
+};
+
+template <typename Input, typename Output>
+struct callable_object_body<Input, Output, /*IsConst = */true> : callable_object_body_base<Input, Output> {
+    Output operator()(Input) const { return Output{}; }
+};
+
+template <typename Input, typename Output>
+Output function_body(Input) { return Output{}; }
+
+template <typename Output>
+struct Input {
+    Output member_object_body;
+    Output member_function_body() const { return Output{}; }
+};
+
+template <typename Body>
+struct read_types {
+    using input_type = typename Body::input_type;
+    using output_type = typename Body::output_type;
+};
+
+template <typename Input, typename Output>
+struct read_types<Output (Input::*)> {
+    using input_type = std::decay_t<Input>;
+    using output_type = std::decay_t<Output>;
+};
+
+template <typename Input, typename Output>
+struct read_types<Output (Input::*)() const> {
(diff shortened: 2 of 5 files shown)


include/oneapi/tbb/detail/_flow_graph_nodes_deduction.h
```diff
@@ -1,5 +1,5 @@
 /*
-    Copyright (c) 2005-2024 Intel Corporation
+    Copyright (c) 2005-2025 Intel Corporation
```
vossmjp: You need to fix up the copyrights for UXL.

include/oneapi/tbb/detail/_flow_graph_nodes_deduction.h
```diff
@@ -23,83 +23,59 @@ namespace tbb {
...
+struct extract_body_types<Output (Input::*)() const> : body_types<Input, Output> {};
 
+// Body is represented as a pointer to member object
 template <typename Input, typename Output>
-auto decide_on_operator_overload(Output (*name)(Input&))->decltype(name);
+struct extract_body_types<Output Input::*> : body_types<Input, Output> {};
 
 template <typename Body>
-decltype(decide_on_operator_overload(&Body::operator())) decide_on_callable_type(int);
+using input_type = typename extract_body_types<Body>::input_type;
```
vossmjp: I wonder if `input_type` and `output_type` are too general as names here. I know these names are short and make things like line 78 look nice, `input_node<output_type<Body>>>`. It just strikes me as strange that the thing called input_type works specifically on bodies. Would something like `body_input_type` and `body_output_type` be too long, or misleading?

include/oneapi/tbb/detail/_flow_graph_nodes_deduction.h
kboyarinov: Fixed

include/oneapi/tbb/detail/_flow_graph_nodes_deduction.h
kboyarinov: What do you think of `input_type_of<Body>` and `output_type_of<Body>`?
I have applied it to the current PR.
If you think `body_input_type` and `body_output_type` is easier to understand, I don't have any objections for using this as well:)

include/oneapi/tbb/detail/_flow_graph_nodes_deduction.h
```diff
@@ -23,83 +24,100 @@ namespace tbb {
...
+template <typename Body>
+struct unary_operator_types_extractor {
+    // Overloads for const, noexcept and & qualified operator() with all possible combinations
+    template <typename B, typename Input, typename Output>
+    static auto check_args(Output (B::*name)(Input)) -> body_types<Input, Output>;
 
-template <typename Input, typename Output>
-struct body_types<Output (*)(Input&)> : declare_body_types<Input, Output> {};
+    template <typename B, typename Input, typename Output>
+    static auto check_args(Output (B::*name)(Input) noexcept) -> body_types<Input, Output>;
```
rarutyun:
```suggestion
    static auto check_args(Output (B::*)(Input) noexcept) -> body_types<Input, Output>;
```

`name` became redundant

include/oneapi/tbb/detail/_flow_graph_nodes_deduction.h
```diff
@@ -23,83 +24,100 @@ namespace tbb {
...
 
-template <typename T, typename Input, typename Output>
-struct body_types<Output (T::*)(const Input&)> : declare_body_types<Input, Output> {};
+// Body is a noexcept free function pointer
+template <typename Input, typename Output>
+struct body_traits<Output (*)(Input) noexcept> : body_types<Input, Output> {};
 
-template <typename T, typename Input, typename Output>
-struct body_types<Output (T::*)(Input&) const> : declare_body_types<Input, Output> {};
+// Body is a member object pointer
```
rarutyun:
```suggestion
// Body is a pointer to non-static data member
```

include/oneapi/tbb/detail/_flow_graph_nodes_deduction.h
```diff
@@ -23,83 +24,100 @@ namespace tbb {
...
+template <typename Input, typename Output>
+struct body_traits<Output Input::*> : body_types<Input, Output> {};
 
-template <typename T, typename Input, typename Output>
-struct body_types<Output (T::*)(Input&)> : declare_body_types<Input, Output> {};
+// Body is a member function pointer
```
rarutyun:
```suggestion
// Body is a pointer to non-static member function
```

include/oneapi/tbb/detail/_flow_graph_nodes_deduction.h
```diff
@@ -23,83 +24,100 @@ namespace tbb {
...
-    using input_type = Input;
-    using output_type = Output;
+template <typename Input>
+struct declare_input_type {
+    using input_type = std::decay_t<Input>;
 };
 
-struct NoInputBody {};
+template <>
+struct declare_input_type<d1::flow_control> {};
```
rarutyun: Please investigate, whether `invoke_result` is applicable here. It seems like we could make more mistakes with the current code

include/oneapi/tbb/detail/_flow_graph_nodes_deduction.h
```diff
@@ -23,83 +24,100 @@ namespace tbb {
...
 
-template <typename T, typename Input, typename Output>
-auto decide_on_operator_overload(Output (T::*name)(const Input&))->decltype(name);
+    template <typename B, typename Input, typename Output>
+    static auto check_args(Output (B::*name)(Input) const & noexcept) -> body_types<Input, Output>;
 
-template <typename T, typename Input, typename Output>
-auto decide_on_operator_overload(Output (T::*name)(Input&) const)->decltype(name);
+    // Fallback if no overload found
+    static void check_args(...);
```
rarutyun: This seems redundant.

```suggestion
```

include/oneapi/tbb/detail/_flow_graph_nodes_deduction.h
```diff
@@ -23,83 +24,100 @@ namespace tbb {
...
-struct declare_body_types<NoInputBody, Output> {
-    using output_type = Output;
+template <typename Input, typename Output>
+struct body_types : declare_input_type<Input> {
+    using output_type = std::decay_t<Output>;
 };
 
-template <typename T> struct body_types;
+template <typename Body, typename = void>
+struct body_traits;
```
rarutyun: Please check whether we can make a complete type and provide users with better diagnostics with `static_assert`

include/oneapi/tbb/detail/_flow_graph_nodes_deduction.h
```diff
@@ -23,83 +24,100 @@ namespace tbb {
...
 
-template <typename T, typename Input, typename Output>
-struct body_types<Output (T::*)(const Input&) const> : declare_body_types<Input, Output> {};
+// Body is a free function pointer
+template <typename Input, typename Output>
+struct body_traits<Output (*)(Input)> : body_types<Input, Output> {};
 
-template <typename T, typename Input, typename Output>
-struct body_types<Output (T::*)(const Input&)> : declare_body_types<Input, Output> {};
+// Body is a noexcept free function pointer
```
rarutyun:
```suggestion
// Body is a pointer to noexcept function 
```

include/oneapi/tbb/detail/_flow_graph_nodes_deduction.h
```diff
@@ -23,83 +24,100 @@ namespace tbb {
...
+    using output_type = std::decay_t<Output>;
 };
 
-template <typename T> struct body_types;
+template <typename Body, typename = void>
+struct body_traits;
 
-template <typename T, typename Input, typename Output>
-struct body_types<Output (T::*)(const Input&) const> : declare_body_types<Input, Output> {};
+// Body is a free function pointer
```
rarutyun:
```suggestion
// Body is a pointer to function
```

include/oneapi/tbb/detail/_flow_graph_nodes_deduction.h
```diff
@@ -23,83 +24,100 @@ namespace tbb {
...
-struct NoInputBody {};
+template <>
+struct declare_input_type<d1::flow_control> {};
 
-template <typename Output>
-struct declare_body_types<NoInputBody, Output> {
-    using output_type = Output;
+template <typename Input, typename Output>
+struct body_types : declare_input_type<Input> {
+    using output_type = std::decay_t<Output>;
```
rarutyun: I think it would be good to add static asserts (for both input and output) that check function signature making sure we provide good diagnostics if users pass something other than `T` or `const T&` and also we don't deduce types to anything unsupported

rarutyun (commented):
I haven't looked at tests yet.

I hope I didn't forget anything that we discussed

include/oneapi/tbb/detail/_flow_graph_nodes_deduction.h
kboyarinov: I have checked and unfortunately our previous assumption was incorrect.
Current `join_node` supports 3 different policies - `queueing`, `reserving` and `key_matching`. 
`queueing` and `reserving` have constructors with `node_set<Predecessors>` and `Policy` argument:

```cpp
template <typename... Predecessors>
join_node(const node_set<Predecessors...>& set, queueing);
```

and the corresponding CTAD is expected to work: `join_node j(flow::follows(pred1, pred2), queueing{});` should deduce `j` as a queueing node.

`key_matching` case accepts the set of bodies (1 for each input port) and the policy parameter is not needed:

```cpp
template <typename... Predecessors, typename... Bodies>
join_node(const node_set<Predecessors...>& set, Bodies... bodies);
```

and the CTAD is also expected to work: `join_node j(flow::follows(pred1, pred2), body1, body2);` should deduce `j` as a key_matching node.

We have the corresponding explicit deduction guides for each case:

```cpp
template <typename... Predecessors, typename Policy>
join_node(const node_set<Predecessors...>&, Policy)
-> join_node<tuple-of-predecessors-outputs, Policy>;

template <typename... Predecessors, typename Bodies>
join_node(const node_set<Predecessors...>&, Bodies...)
-> join_node<tuple-of-body-inputs, key_matching>; // tuple-of-body-inputs instantiates body_traits<Bodies...>
```

Even if we remove `Policy` parameter from the first guide and make it explicitly queueing, the deduction guide itself will remain template. Therefore, the second deduction guide will also be taken into account. And since instantiation of body_traits<queueing> results in hard-error, not a SFINAE, the compilation will break.

include/oneapi/tbb/detail/_flow_graph_nodes_deduction.h
kboyarinov: Added

include/oneapi/tbb/detail/_flow_graph_nodes_deduction.h
kboyarinov: Done

include/oneapi/tbb/detail/_flow_graph_nodes_deduction.h
kboyarinov: Fixed by adding `enable_if` into `join_node` deduction guide

include/oneapi/tbb/detail/_flow_graph_nodes_deduction.h
```diff
@@ -24,82 +25,103 @@ namespace detail {
...
 
-template <typename Input, typename Output>
-struct body_types<Output (*)(const Input&)> : declare_body_types<Input, Output> {};
+    template <typename B, typename Input, typename Output>
+    static auto check_args(Output (B::*)(Input) const) -> checked_body_types<Input, Output>;
 
-template <typename Output>
-struct body_types<Output (*)(d1::flow_control&)> : declare_body_types<NoInputBody, Output> {};
+    template <typename B, typename Input, typename Output>
+    static auto check_args(Output (B::*)(Input) &) -> checked_body_types<Input, Output>;
```
aleksei-fedotov: Can user specify &&-qualified member function? What can flow graph do about that? In what scenarios could that potentially provide advantage comparing to &-qualified ones?

aleksei-fedotov: I wonder what would happen if user specified &&-qualified non-static member function? Do flow graph nodes (should) support such a use case? Is there any benefit a user can leverage out of this? Seems not at the glance. Just wonder if we have thought about that in the past as I am unable to find anything w.r.t this.

include/oneapi/tbb/detail/_flow_graph_nodes_deduction.h
```diff
@@ -24,82 +25,103 @@ namespace detail {
...
-struct NoInputBody {};
-
-template <typename Output>
-struct declare_body_types<NoInputBody, Output> {
-    using output_type = Output;
+template <typename Input, typename Output>
+struct checked_body_types : body_types<Input, Output> {
+private:
+    using base_type = body_types<Input, Output>;
+    using const_lvalue_ref_input_type = std::add_lvalue_reference_t<std::add_const_t<typename base_type::input_type>>;
```
rarutyun: why not just `const typename base_type::input_type&`?

include/oneapi/tbb/detail/_flow_graph_nodes_deduction.h
rarutyun: The deduction guides will fail in this case, because we don't have a partial template specialization for && qualifier.

Whether FG nodes should support that or not, that's debatable. However, my understanding is that we store callables within FG nodes and thus, treat them as lvalues anyway. So, I guess it would not be that simple to store a particular value category. There are some technics that allow to do that in type-erased context. But I feel like we will overcomplicate the code without any particular reason.

I would suggest to wait for a specific request from customers to support `&&`

include/oneapi/tbb/detail/_flow_graph_nodes_deduction.h
```diff
@@ -146,6 +168,16 @@ priority_queue_node(const NodeSet&)
 ->priority_queue_node<decide_on_set_t<NodeSet>, std::less<decide_on_set_t<NodeSet>>>;
 #endif // __TBB_PREVIEW_FLOW_GRAPH_NODE_SET
 
+template <typename T>
+struct is_queueing_or_rejecting_tag
```
rarutyun: I wonder why do we need `struct`s here. Is not alias template sufficient? I just want to avoid extra template instantiation, because I don't see a reason for a class template and alias template is cheaper to compile.

include/oneapi/tbb/detail/_flow_graph_nodes_deduction.h
```diff
@@ -146,6 +168,16 @@ priority_queue_node(const NodeSet&)
...
+    : std::disjunction<std::is_same<T, queueing>, std::is_same<T, reserving>>
+{};
+
+template <typename... Args>
+struct is_queueing_or_rejecting_tags
+    : std::conjunction<is_queueing_or_rejecting_tag<Args>...>
+{};
```
rarutyun: do you really need an extra one? How about just having one with `template <typename T, typename... Args>` interface? I mean it could be just `Args` but I feel like it not the best idea because we want (I think) to guarantee that one argument is bare minimum.

include/oneapi/tbb/detail/_flow_graph_nodes_deduction.h
```diff
@@ -160,12 +192,14 @@ template <typename Key>
 using join_key_t = typename join_key<Key>::type;
 
 #if __TBB_PREVIEW_FLOW_GRAPH_NODE_SET
-template <typename Policy, typename... Predecessors>
+
+template <typename... Predecessors, typename Policy,
+          typename = std::enable_if_t<is_queueing_or_rejecting_tag<Policy>::value>>
```
rarutyun:
```suggestion
          std::enable_if_t<is_queueing_or_rejecting_tag<Policy>::value, int> = 0>
```

This is more robust `enable_if` assuming future extensions. If you agree, please apply for all three (new) `enable_if`s.

include/oneapi/tbb/detail/_flow_graph_nodes_deduction.h
rarutyun: I would group them slightly differently. `Body`, the `noexcept Body` or `const Body` (in whatever order), then `const noexcept Body`. And only then all the same with `&`, like `& Body`, then `const & Body`, etc. It's more logical to me because ref-qualifier sort of adds extra "dimension to all that".

It's a matter of taste, though. Please apply if you agree.

If you do apply that, perhaps it's worth reordering in the tests as well.

aleksei-fedotov: Thanks for confirming. Yeah, I have a similar opinion. I think it might be reasonable to have callable stored as an rvalue for a one-off nodes. However, the flow graph's very idea to build once and run multiple times just does not agree with that pattern.

kboyarinov: There are two possible places where && qualifiers could be applicable:
- If the body is a pointer to a non-static member function of `Input` qualified with `&&`. However, the named requirements specify that the input is passed to the body as `const Input&`. As a result, functions that are not const and/or qualified with && are implicitly disallowed. The current FG implementation also does not support this case without the use of deduction guides.
- If the body is a callable type whose unary operator() is qualified with `&&`. This case is also unsupported by the current FG implementation, because the body is always used as an lvalue.

Supporting `&&`-qualified function in either of these cases would require changes to how FG handles bodies and inputs in general, not merely updates to the deduction guides.

include/oneapi/tbb/detail/_flow_graph_nodes_deduction.h
kboyarinov: Changed

include/oneapi/tbb/detail/_flow_graph_nodes_deduction.h
kboyarinov: Fixed

include/oneapi/tbb/detail/_flow_graph_nodes_deduction.h
kboyarinov: I agree, we can use a single one with 1+ number of arguments, but I would prefer to have the second one as "impl". 
Otherwise, conjunction of disjunctions looks monstrous IMHO:

```
template <typename T, typename... Args>
using is_queueing_or_rejecting_tags =
    std::conjunction<std::disjunction<std::is_same<T, queueing>, std::is_same<T, rejecting>>,
                                std::disjunction<std::is_same<Args, queueing>, std::is_same<Args, rejecting>>...>;
```

In my vision, conjunction of impls looks better and helps avoid duplication in queueing/rejecting names:

```
template <typename T>
using is_queueing_or_rejecting_tag_impl = std::disjunction<std::is_same<T, queueing>, std::is_same<T, rejecting>>;

template <typename T, typename... Args>
using is_queueing_or_rejecting_tags = std::conjunction<is_queueing_or_rejecting_tag_impl<T>,
                                                                                          is_queueing_or_rejecting_tag_impl<Args>...>;
``` 

May be I would even preserve the original name `is_queueing_or_rejecting_tag` for the first one.

include/oneapi/tbb/detail/_flow_graph_nodes_deduction.h
kboyarinov: Reordered

include/oneapi/tbb/detail/_flow_graph_nodes_deduction.h
```diff
@@ -146,6 +167,13 @@ priority_queue_node(const NodeSet&)
 ->priority_queue_node<decide_on_set_t<NodeSet>, std::less<decide_on_set_t<NodeSet>>>;
 #endif // __TBB_PREVIEW_FLOW_GRAPH_NODE_SET
 
+template <typename T>
+using is_queueing_or_rejecting_tag_impl = std::disjunction<std::is_same<T, queueing>, std::is_same<T, reserving>>;
+
+template <typename T, typename... Args>
+using is_queueing_or_rejecting_tags = std::conjunction<is_queueing_or_rejecting_tag_impl<T>,
+                                                       is_queueing_or_rejecting_tag_impl<Args>...>;
```
kboyarinov:
```suggestion
using is_queueing_or_rejecting_tag = std::disjunction<std::is_same<T, queueing>, std::is_same<T, reserving>>;

template <typename T, typename... Args>
using are_queueing_or_rejecting_tags = std::conjunction<is_queueing_or_rejecting_tag<T>,
                                                       is_queueing_or_rejecting_tag<Args>...>;
```

include/oneapi/tbb/detail/_flow_graph_nodes_deduction.h
```diff
@@ -160,12 +188,14 @@ template <typename Key>
 using join_key_t = typename join_key<Key>::type;
 
 #if __TBB_PREVIEW_FLOW_GRAPH_NODE_SET
-template <typename Policy, typename... Predecessors>
+
+template <typename... Predecessors, typename Policy,
+          std::enable_if_t<is_queueing_or_rejecting_tags<Policy>::value, int> = 0>
```
kboyarinov:
```suggestion
          std::enable_if_t<are_queueing_or_rejecting_tags<Policy>::value, int> = 0>
```

include/oneapi/tbb/detail/_flow_graph_nodes_deduction.h
```diff
@@ -160,12 +188,14 @@ template <typename Key>
...
 join_node(const node_set<order::following, Predecessors...>&, Policy)
-->join_node<std::tuple<typename Predecessors::output_type...>,
-            Policy>;
+->join_node<std::tuple<typename Predecessors::output_type...>, Policy>;
 
-template <typename Policy, typename Successor, typename... Successors>
+template <typename Successor, typename... Successors, typename Policy,
+          std::enable_if_t<is_queueing_or_rejecting_tags<Policy>::value, int> = 0>
```
kboyarinov:
```suggestion
          std::enable_if_t<are_queueing_or_rejecting_tags<Policy>::value, int> = 0>
```

include/oneapi/tbb/detail/_flow_graph_nodes_deduction.h
```diff
@@ -179,19 +209,17 @@ join_node(const node_set<order::preceding, Successor, Successors...>)
 ->join_node<typename Successor::input_type, queueing>;
 #endif
 
-template <typename GraphOrProxy, typename Body, typename... Bodies>
+template <typename GraphOrProxy, typename Body, typename... Bodies,
+          std::enable_if_t<!is_queueing_or_rejecting_tags<Body, Bodies...>::value, int> = 0>
```
kboyarinov:
```suggestion
          std::enable_if_t<!are_queueing_or_rejecting_tags<Body, Bodies...>::value, int> = 0> 
```

include/oneapi/tbb/detail/_flow_graph_nodes_deduction.h
```diff
@@ -146,6 +167,13 @@ priority_queue_node(const NodeSet&)
 ->priority_queue_node<decide_on_set_t<NodeSet>, std::less<decide_on_set_t<NodeSet>>>;
 #endif // __TBB_PREVIEW_FLOW_GRAPH_NODE_SET
 
+template <typename T>
+using is_queueing_or_rejecting_tag = std::disjunction<std::is_same<T, queueing>, std::is_same<T, reserving>>;
+
```
akukanov: Note `rejecting` in the name but `reserving` in the implementation.

kboyarinov: Oops, too many tags:)
Fixed

PR #1631, opened 2025-02-10
Link tbbbind against static HWLOC instead of HWLOC 1.11

isaevil:
### Description 
Reserve "tbbbind" library name for the library, statically linked with HWLOC 2.x instead of HWLOC 1.11 as was proposed in https://github.com/uxlfoundation/oneTBB/blob/master/rfcs/proposed/numa_support/tbbbind-link-static-hwloc.org#proposal

Fixes # - _issue number(s) if exists_

### Type of change

_Choose one or multiple, leave empty if none of the other choices apply_

_Add a respective label(s) to PR if you have permissions_

- [ ] bug fix - _change that fixes an issue_
- [x] new feature - _change that adds functionality_
- [x] tests - _change in tests_
- [x] infrastructure - _change in infrastructure and CI_
- [ ] documentation - _documentation update_

### Tests

- [ ] added - _required for new features and some bug fixes_
- [x] not needed

### Documentation

- [ ] updated in # - _add PR number_
- [ ] needs to be updated
- [ ] not needed

### Breaks backward compatibility
- [ ] Yes
- [ ] No
- [x] Unknown

### Notify the following users
_List users with `@` to send notifications_

PR #1747, opened 2025-06-04
Docs: HWLOC linked with libtbbbind

aepanchi:
### Description 
Update GSG with information about static HWLOC linked with libtbbbind. 
Document changes from https://github.com/uxlfoundation/oneTBB/pull/1631

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/doc/GSG/next_steps.rst
+++ b/doc/GSG/next_steps.rst
@@ -129,31 +129,20 @@ Build and Run a Sample
 Enable Hybrid CPU and NUMA Support
 ***********************************
 
-If you need NUMA/Hybrid CPU support in oneTBB, you need to make sure that HWLOC* is installed on your system.
+To enable NUMA and hybrid CPU optimizations in oneTBB, complete the following steps:
+1. **Locate the ``tbbbind`` Library**
+   
+   Find the ``tbbbind`` library included in your oneTBB installation. Starting with oneTBB 2022.2, this library is statically linked with HWLOC* 2.x. You do not need to install HWLOC separately.
 
-HWLOC* (Hardware Locality) is a library that provides a portable abstraction of the hierarchical topology of modern architectures (NUMA, hybrid CPU systems, etc). oneTBB relies on HWLOC* to identify the underlying topology of the system to optimize thread scheduling and memory allocation.
+2. **Ensure ``tbbbind`` Is Accessible**
+   
+   Place the ``tbbbind`` library in a location where oneTBB can find it during execution. Use one of the following options:
+   
+   * Put it in the same directory as the core tbb library (``libtbb.so``, ``tbb.dll``, etc.)
+   * Add its location to the system’s library search path (``LD_LIBRARY_PATH`` on Linux* OS, ``%PATH%`` on Windows* OS)
 
-Without HWLOC*, oneTBB may not take advantage of NUMA/Hybrid CPU support. Therefore, it's important to make sure that HWLOC* is installed before using oneTBB on such systems.
+3. **Run Your Application**
+   
+   You do not need to link your application explicitly with ``tbbbind``. oneTBB automatically loads it at runtime when your application calls an API that requires NUMA or hybrid CPU support
 
-Check HWLOC* on the System
-^^^^^^^^^^^^^^^^^^^^^^^^^^^
-To check if HWLOC* is already installed on your system, run ``hwloc-ls``:
-
-* For Linux* OS, in the command line.
-* For Windows* OS, in the command prompt.
-
-If HWLOC* is installed, the command displays information about the hardware topology of your system. If it is not installed, you receive an error message saying that the command ``hwloc-ls`` could not be found.
-
-.. note:: For Hybrid CPU support, make sure that HWLOC* is version 2.5 or higher. For NUMA support, install HWLOC* version 1.11 or higher.
-
-Install HWLOC*
-^^^^^^^^^^^^^^
-
-To install HWLOC*, visit the official Portable Hardware Locality website (https://www-lb.open-mpi.org/projects/hwloc/).
-
-* For Windows* OS, binaries are available for download.
-* For Linux* OS, only the source code is provided and binaries should be built.
-
-On Linux* OS, HWLOC* can be also installed with package managers, such as APT*, YUM*, etc. To do so, run: sudo apt install hwloc.
-
-.. note:: For Hybrid CPU support, make sure that HWLOC* is version 2.5 or higher. For NUMA support, install HWLOC* version 1.11 or higher.
+.. tip:: To confirm that ``tbbbind`` is loaded successfully, set the environment variable ``TBB_VERSION=1`` before running your application.
--- a/doc/GSG/next_steps.rst
+++ b/doc/GSG/next_steps.rst
@@ -129,21 +129,17 @@ Build and Run a Sample
 Enable Hybrid CPU and NUMA Support
 ***********************************
 
-To enable NUMA and hybrid CPU optimizations in oneTBB, complete the following steps:
+To support Hybrid CPU and NUMA platforms, oneTBB relies on the HWLOC* library.
+ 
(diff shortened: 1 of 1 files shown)


doc/GSG/next_steps.rst
```diff
@@ -129,31 +129,21 @@ Build and Run a Sample
...
+   
+   Find the ``tbbbind`` library included in your oneTBB installation. Starting with oneTBB 2022.2, this library is statically linked with HWLOC* 2.x. You do not need to install HWLOC separately.
 
-* For Windows* OS, binaries are available for download.
-* For Linux* OS, only the source code is provided and binaries should be built.
+2. **Ensure ``tbbbind`` Is Accessible**
+   
+   Place the ``tbbbind`` library in a location where oneTBB can find it during execution. Use one of the following options:
+   
+   * Put it in the same directory as the core tbb library (``libtbb.so``, ``tbb.dll``, etc.)
```
aleksei-fedotov: Perhaps,
```suggestion
   * Put it in the same directory as the core oneTBB library (``libtbb.so``, ``tbb.dll``, etc.)
```
?

doc/GSG/next_steps.rst
```diff
@@ -129,31 +129,21 @@ Build and Run a Sample
...
+   * Add its location to the system’s library search path (``LD_LIBRARY_PATH`` on Linux* OS, ``%PATH%`` on Windows* OS)
```
aleksei-fedotov: `tbbbind` is still loaded by using its absolute path. It is the HWLOC library, which is dynamically linked to the other versions of `tbbbind` variants `tbbbind_2_5` and `tbbbind_2_0`, is searched using established search order, i.e. paths written in, depending on the platform, `LD_LIBRARY_PATH` or `PATH` environment variable.

doc/GSG/next_steps.rst
```diff
@@ -129,31 +129,21 @@ Build and Run a Sample
...
 
-On Linux* OS, HWLOC* can be also installed with package managers, such as APT*, YUM*, etc. To do so, run: sudo apt install hwloc.
+3. **Run Your Application**
+   
+   You do not need to link your application explicitly with ``tbbbind``. oneTBB automatically loads it at runtime when your application calls an API that requires NUMA or hybrid CPU support
```
aleksei-fedotov: I would suggest simplifying to avoid false interpretations:
```suggestion
   You do not need to link your application explicitly with ``tbbbind``. oneTBB automatically loads and uses it whenever necessary.
```

doc/GSG/next_steps.rst
```diff
@@ -129,31 +129,21 @@ Build and Run a Sample
...
-
-.. note:: For Hybrid CPU support, make sure that HWLOC* is version 2.5 or higher. For NUMA support, install HWLOC* version 1.11 or higher.
-
-Install HWLOC*
-^^^^^^^^^^^^^^
-
-To install HWLOC*, visit the official Portable Hardware Locality website (https://www-lb.open-mpi.org/projects/hwloc/).
+1. **Locate the ``tbbbind`` Library**
+   
+   Find the ``tbbbind`` library included in your oneTBB installation. Starting with oneTBB 2022.2, this library is statically linked with HWLOC* 2.x. You do not need to install HWLOC separately.
```
aleksei-fedotov: Do we want to mention still existing possibility to use user-specific version of HWLOC, which is currently done through the other variants of `tbbbind` library?
I think, the approach might be changed in the future (e.g. other `tbbbind_*` variants may be gone), but the possibility will stay in place. So, if we want to mention that, then it might make sense writing it in such a way that will not require us to update the description again if the approach changes.

doc/GSG/next_steps.rst
aepanchi: Thanks, what do you think about:
Place the ``tbbbind`` library in a location where oneTBB can find it during execution.
oneTBB loads the library using its absolute path. If a versioned variant of ``tbbbind`` is used (e.g., ``tbbbind_2_5`` or ``tbbbind_2_0``), and it dynamically links to HWLOC, the system resolves HWLOC using the standard dynamic library search order.
Use one of the following options to ensure proper loading:
* Put ``tbbbind`` in the same directory as the core oneTBB library (``libtbb.so``, ``tbb.dll``, etc.)
* Add its location to the system’s library search path (``LD_LIBRARY_PATH`` on Linux* OS, ``%PATH%`` on Windows* OS)

doc/GSG/next_steps.rst
aepanchi: @isaevil what do you think?

isaevil: I think it can be useful. Currently we have following tbbbind versions:
* tbbbind_2_5 - can utilize HWLOC version 2.5 and higher
* tbbbind_2_0 - can utilize HWLOC version from 2.1 to 2.4
* tbbbind - statically linked with HWLOC

We can write that tbbbind is able to use HWLOC provided by the user but with a notion that they may end up with feature set limited by that HWLOC version. For example, HWLOC versions prior to 2.5 don't support hybrid platforms.

aepanchi: Great, how about:
1. Locate the tbbbind library
Find the ``tbbbind`` library included in your oneTBB installation. Starting with oneTBB 2022.2, this library is statically linked with HWLOC* 2.x. You do not need to install HWLOC separately.

If you prefer to use a system-provided or custom-built version of HWLOC, oneTBB also provides versioned variants of the tbbbind library:
* tbbbind_2_5 — supports HWLOC version 2.5 and higher (required for hybrid CPU support)
* tbbbind_2_0 — supports HWLOC versions 2.1 to 2.4

Note: When using a versioned variant (tbbbind_2_5 or tbbbind_2_0), the available feature set may depend on the capabilities of the linked HWLOC version. For example, a hybrid CPU topology is only supported in HWLOC 2.5 and later.

isaevil: I think @aleksei-fedotov suggests using such wording that won't require updating if we change the approach. I am fine with suggested wording, but I wonder if @aleksei-fedotov have different in mind.

aleksei-fedotov: Here is my version =) I think it precisely describes the algorithm used by the library. And I tried to make it laconic.
I don't insist on it though, - you can use whichever you deem is best.

========

Enable Hybrid CPU and NUMA Support
***********************************

To support Hybrid CPU and NUMA platforms oneTBB relies on the HWLOC* library.

The HWLOC functionality is used through set of proxy libraries whose names start with `tbbbind` prefix. oneTBB automatically loads them at runtime when needed. To be found and loaded successfully, they need to be in the same directory as oneTBB library itself.

Starting with oneTBB 2022.2, `tbbbind` library is statically linked with HWLOC 2.x. This version is used if system-provided HWLOC could not be found.
If you prefer to use a system-provided or custom-built version of HWLOC, make sure it is available in search paths used by dynamic loader. Consult your dynamic loader documentation to find out what these paths are.
In addition, make sure one of the following `tbbbind` versions are placed in the directory with oneTBB library:

* `tbbbind_2_5` - depends on HWLOC version 2.5 or higher. Use it if support for hybrid CPU is required.
* `tbbbind_2_0` - depends on HWLOC versions 2.1 to 2.4

.. tip:: To confirm that ``tbbbind`` is loaded successfully, set the environment variable ``TBB_VERSION=1`` before running your application.

========

doc/GSG/next_steps.rst
aleksei-fedotov: Closing this comment since it somewhat duplicates the discussion in [another one](https://github.com/uxlfoundation/oneTBB/pull/1747/files#r2130377327)

aleksei-fedotov (approved):
Great! Thanks a lot! You are awesome! ❤️

PR #1768, opened 2025-07-01
Only enable fcf-protection on x86 based processors

lmarz:
### Description 
Fixes compilation errors on basically every arch other than amd64, x86 and aarch64.

The gcc man page in the section -fcf-protection states:
> Currently the x86 GNU/Linux target provides an implementation based on Intel Control-flow Enforcement Technology (CET) which works for i686 processor or newer.

and does not mention any other architecture.

Additionally, in the gcc testsuite, there is the file `gcc/testsuite/c-c++-common/fcf-protection-1.c`:
```
/* { dg-error "'-fcf-protection=full' is not supported for this target" "" { target { ! "i?86-*-* x86_64-*-*" } } 0 } */
```
### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

@omalyshe, since they introduced this flag in  #1680


isaevil (approved):
LGTM

Issue #1764, opened 2025-06-26
`concurrent_hash_map` Iterator Construction Allows Duplicate Keys


0xSqualo:
# Summary

Initialization of `concurrent_hash_map` allows duplicate key entries during iterator list construction.

# Version

2022.0.0-2

# Environment

* AMD Ryzen 9 7950X3D
* Ubuntu 25.04
* 2022.0.0-2

# Observed Behavior
Initializing the `concurrent_hash_map` from an iterator of a non-unique type container (e.g. `std::vector<T>`) initializes the map with duplicate keys. This does not happen through other initialization / constructor methods.

# Expected Behavior
That no duplicate key / value pairs are present or allowed in the `concurrent_hash_map`, irrespective of construction or post instantiation insertion (internally deduplicated based on insertion ordering as it would be if iterating post construction and inserting via an iterator loop).

# Steps To Reproduce

```c++
std::vector<std::pair<uint64_t, uint64_t>> x {
        {0, 0},
        {0, 1}, // Enter 0 as a key twice intentionally!
};
oneapi::tbb::concurrent_hash_map<uint64_t, uint64_t> y{ std::begin(x), std::end(x) };

// This prints two values e.g.
    // 0: 1 
    // 0: 0
for (auto const& [key, value] : y) {
    std::cout << key << ": " << value << std::endl;
}
    
// Check that we only have 1 item in the hashmap
if (y.size() != 1) {
    std::cout << "y.size() = " << y.size() << std::endl;
    abort(); // This triggers because the size is 2!
}
```

dnmokhov:
Thank you for reporting this, @0xSqualo!

It appears that the issue occurs with the iterator and initializer list constructors, as well as with the initializer list assignment operator.

kboyarinov:
Fixed by 9a771aafb61e3877e5a364322542f221b2f315cd.

PR #1771, opened 2025-07-04
Fix duplicating keys in concurrent_hash_map

kboyarinov:
### Description 
Fix a bug in concurrent_hash_map constructors accepting half-opened intervals [first, last) or initializer_list.

Fixes # - #1764 

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/include/oneapi/tbb/concurrent_hash_map.h
+++ b/include/oneapi/tbb/concurrent_hash_map.h
@@ -1,5 +1,5 @@
 /*
-    Copyright (c) 2005-2025 Intel Corporation
+    Copyright (c) 2025 UXL Foundation Contributors
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
--- a/include/oneapi/tbb/concurrent_hash_map.h
+++ b/include/oneapi/tbb/concurrent_hash_map.h
@@ -1520,9 +1520,12 @@ protected:
             hashcode_type h = my_hash_compare.hash( (*first).first );
             bucket *b = this->get_bucket( h & m );
             __TBB_ASSERT(!rehash_required(b->node_list.load(std::memory_order_relaxed)), "Invalid bucket in destination table");
-            node* node_ptr = create_node(base_type::get_allocator(), (*first).first, (*first).second);
-            this->add_to_bucket( b, node_ptr );
-            ++this->my_size; // TODO: replace by non-atomic op
+            const auto& key = (*first).first;
+            if (search_bucket(key, b) == nullptr) {
+                node* node_ptr = create_node(base_type::get_allocator(), *first);
+                this->add_to_bucket( b, node_ptr );
+                ++this->my_size; // TODO: replace by non-atomic op
+            }
         }
     }
 
--- a/include/oneapi/tbb/concurrent_hash_map.h
+++ b/include/oneapi/tbb/concurrent_hash_map.h
@@ -1,4 +1,5 @@
 /*
+    Copyright (c) 2005-2025 Intel Corporation
     Copyright (c) 2025 UXL Foundation Contributors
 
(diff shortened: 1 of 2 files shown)


isaevil:
Include <unordered_map> header to fix CI issues.

test/tbb/test_concurrent_hash_map.cpp
```diff
@@ -899,3 +899,52 @@ TEST_CASE("container_range concept for tbb::concurrent_hash_map ranges") {
 }
 
 #endif // __TBB_CPP20_CONCEPTS_PRESENT
+
+template <typename ChmapType, typename UnorderedMapType>
+void check_for_duplicated_keys(const ChmapType& chmap, const UnorderedMapType& unique_key_value_pairs) {
+    CHECK_MESSAGE(chmap.size() == unique_key_value_pairs.size(), "Incorrect number of keys in the hash map");
+    for (auto& pair : unique_key_value_pairs) {
+        CHECK_MESSAGE(chmap.count(pair.first) == 1, "Key from the unique set is not presented in the hash map");
+    }
```
aleksei-fedotov: Let's check not only for the keys but for associated values as well.

include/oneapi/tbb/concurrent_hash_map.h
```diff
@@ -1520,9 +1520,12 @@ class concurrent_hash_map
             hashcode_type h = my_hash_compare.hash( (*first).first );
             bucket *b = this->get_bucket( h & m );
             __TBB_ASSERT(!rehash_required(b->node_list.load(std::memory_order_relaxed)), "Invalid bucket in destination table");
-            node* node_ptr = create_node(base_type::get_allocator(), (*first).first, (*first).second);
-            this->add_to_bucket( b, node_ptr );
-            ++this->my_size; // TODO: replace by non-atomic op
+            const auto& key = (*first).first;
```
aleksei-fedotov: Move the `key` acquiring upper and also re-use in `my_hash_compare.hash( key )` above?

include/oneapi/tbb/concurrent_hash_map.h
```diff
@@ -1520,9 +1520,12 @@ class concurrent_hash_map
...
+            if (search_bucket(key, b) == nullptr) {
+                node* node_ptr = create_node(base_type::get_allocator(), *first);
+                this->add_to_bucket( b, node_ptr );
+                ++this->my_size; // TODO: replace by non-atomic op
```
aleksei-fedotov, line 1529: Seems like adding the associated value is missed.

isaevil: Do you mean that we don't pass a value when invoking `create_node`? If yes, then that's not the case here since `*first` is expected to be a pair at this point.

kboyarinov: Yes, @isaevil is right. At this point `*first` is expected to be something that can be used to construct a pair, i.e. a `std::pair` or pair-like object like `std::tuple`.

test/tbb/test_concurrent_hash_map.cpp
kboyarinov: Done

include/oneapi/tbb/concurrent_hash_map.h
kboyarinov: Applied

include/oneapi/tbb/concurrent_hash_map.h
aleksei-fedotov: Got it. Thanks! Although not as readable as the variant before the change. Especially, given such a diff :-)

PR #1777, opened 2025-07-10
Add cgroup num cpus support

aleksei-fedotov:
### Description 

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown

### DIFF
--- a/src/tbb/misc_ex.cpp
+++ b/src/tbb/misc_ex.cpp
@@ -1,5 +1,6 @@
 /*
-    Copyright (c) 2005-2024 Intel Corporation
+    Copyright (c) 2005-2025 Intel Corporation
+    Copyright (c) 2025 UXL Foundation Сontributors
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
--- a/src/tbb/misc_ex.cpp
+++ b/src/tbb/misc_ex.cpp
@@ -33,6 +33,10 @@
 #else
 #include <unistd.h>
 #if __unix__
+#include <thread>
+#include <fstream>
+#include <sstream>
+#include <cstring>
 #if __linux__
 #include <sys/sysinfo.h>
 #endif
@@ -130,6 +134,151 @@ static std::atomic<do_once_state> hardware_concurrency_info;
 
 static int theNumProcs;
 
+// Linux control groups support
+class cgroup_info {
+public:
+    cgroup_info()
+        : default_num_cpus(std::thread::hardware_concurrency()), // Fallback value
+          m_num_cpus(parse_num_cpus()) {}
+
+    unsigned num_cpus() const { return m_num_cpus; }
+
+private:
+    unsigned parse_num_cpus() {
+        const std::string cgroup_line = find_cgroup_mount(mounts_path);
+        if (cgroup_line.empty())
+            return default_num_cpus; // Return default if no cgroup mount found
+
+        unsigned version = 0;
+        std::string mount_point;
+        if (!parse_version_and_mount_point(cgroup_line.c_str(), mount_point, version))
+            return default_num_cpus; // Return default if parsing fails
+
+        if (1 == version)
+            return determine_num_cpus_for_cgroup_v1(mount_point);
+        else
+            return determine_num_cpus_for_cgroup_v2(mount_point);
+    }
+
+    std::string find_cgroup_mount(const char* file_path) {
+        std::ifstream f(file_path);
+        if (!f)
+            return std::string();
+
+        std::string line;
+        while (std::getline(f, line)) {
+            if (line.length() < 6)
+                continue; // Skip empty or too short lines
+
+            if (std::strncmp(line.data(), "cgroup", 6) != 0)
+                continue; // Skip non-cgroup lines
+
+            return line;
+        }
+
+        return std::string();
+    }
+
+    bool read_line_from_file(const std::string& file_path, std::string& line) {
+        std::ifstream f(file_path);
+        if (!f || !std::getline(f, line))
+            return false; // Failed to read line
+        return true;
+    }
+
+    template <typename T>
+    bool convert_token_from_istream(std::istringstream& istream, T& value) {
+        if (!(istream >> value))
+            return false; // Failed to read period
+        return true;
+    }
+
+    bool parse_version_and_mount_point(const char* input, std::string& mount_point,
+                                       unsigned& version)
+    {
+        std::istringstream iss(input);
+        std::string device, version_str;
+        std::string* strs[] = {&device, &mount_point, &version_str};
+        for (auto& str : strs) {
+            if (!convert_token_from_istream(iss, *str))
+                return false; // Failed to read expected fields
+        }
+
+        version = 2;            // default
+        if (version_str == "cgroup")
+            version = 1;
+
+        return true;
+    }
+
+    bool parse_cpu_quota(std::istringstream& istream, unsigned long long& cpu_quota) {
+        std::string quota_str;
+        if (!convert_token_from_istream(istream, quota_str))
+            return false; // Failed to read quota
+
+        if (quota_str == "max") // CPU quota may be set to "max", which means no limit
+            return false;
+
+        std::istringstream cpu_quota_stream(quota_str);
+        return convert_token_from_istream(cpu_quota_stream, cpu_quota);
+    }
+
+    bool parse_cpu_period(std::istringstream& istream, unsigned long long& cpu_period) {
+        return convert_token_from_istream(istream, cpu_period);
+    }
+
+    unsigned determine_num_cpus_for_cgroup_v1(const std::string& mount_point) {
+        std::string line;
+        if (!read_line_from_file(mount_point + "/cpu/cpu.cfs_quota_us", line))
+            return default_num_cpus;
+
+        std::istringstream iss(line);
+        unsigned long long cpu_quota = 0;
+        if (!parse_cpu_quota(iss, cpu_quota))
+            return default_num_cpus;
+
+        if (!read_line_from_file(mount_point + "/cpu/cpu.cfs_period_us", line))
+            return default_num_cpus;
+
+        iss = std::istringstream(line);
+        unsigned long long cpu_period = 0;
+        if (!parse_cpu_period(iss, cpu_period))
+            return default_num_cpus;
+
+        return determine_num_cpus(cpu_quota, cpu_period);
+    }
+
+    unsigned determine_num_cpus_for_cgroup_v2(const std::string& mount_point) {
+        std::string line;
+        if (!read_line_from_file(mount_point + "/cpu.max", line))
+            return default_num_cpus;
+
+        std::istringstream iss(line);
+        unsigned long long cpu_quota = 0;
+        if (!parse_cpu_quota(iss, cpu_quota))
+            return default_num_cpus;
+
+        unsigned long long cpu_period = 0;
+        if (!parse_cpu_period(iss, cpu_period))
+            return default_num_cpus;
+
+        return determine_num_cpus(cpu_quota, cpu_period);
+    }
+
+    unsigned determine_num_cpus(unsigned long long cpu_quota, unsigned long long cpu_period) {
+        if (0 == cpu_period)
+            return default_num_cpus; // Avoid division by zero, use the default number of CPUs
+
+        const unsigned num_cpus = static_cast<unsigned>((cpu_quota + cpu_period / 2) / cpu_period);
+        return num_cpus > 0 ? num_cpus : 1;
+    }
+
+    static constexpr const char* mounts_path = "/proc/self/mounts";
+    const unsigned default_num_cpus; // Fallback value if parsing fails
+
+    const unsigned m_num_cpus;
+};
+
 static void initialize_hardware_concurrency_info () {
     int err;
     int availableProcs = 0;
@@ -190,7 +339,9 @@ static void initialize_hardware_concurrency_info () {
         availableProcs = (maxProcs == INT_MAX) ? sysconf(_SC_NPROCESSORS_ONLN) : maxProcs;
         delete[] processMask;
     }
-    theNumProcs = availableProcs > 0 ? availableProcs : 1; // Fail safety strap
+    cgroup_info ci;
+    unsigned num_procs = availableProcs > 0 ? availableProcs : 1; // Fail safety strap
+    theNumProcs = min(ci.num_cpus(), num_procs);
     __TBB_ASSERT( theNumProcs <= sysconf(_SC_NPROCESSORS_ONLN), nullptr);
 }
 
--- a/test/common/utils_concurrency_limit.h
+++ b/test/common/utils_concurrency_limit.h
@@ -1,5 +1,6 @@
 /*
-    Copyright (c) 2020-2023 Intel Corporation
+    Copyright (c) 2005-2025 Intel Corporation
+    Copyright (c) 2025 UXL Foundation Сontributors
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
@@ -101,21 +102,20 @@ static int get_max_procs() {
     return maxProcs;
 }
 
-int get_start_affinity_process() {
+static std::vector<int> get_cpuset_indices() {
 #if __linux__
     cpu_set_t mask;
     sched_getaffinity(0, sizeof(cpu_set_t), &mask);
 
-    int result = -1;
+    std::vector<int> result;
 
     int nproc = sysconf(_SC_NPROCESSORS_ONLN);
     for (int i = 0; i < nproc; ++i) {
         if (CPU_ISSET(i, &mask)) {
-            result = i;
-            break;
+            result.push_back(i);
         }
     }
-    ASSERT(result != -1, nullptr);
+    ASSERT(!result.empty(), nullptr);
     return result;
 #else
     // TODO: add affinity support for Windows and FreeBSD
@@ -123,7 +123,7 @@ int get_start_affinity_process() {
 #endif
 }
 
-int limit_number_of_threads( int max_threads ) {
+static int limit_number_of_threads( int max_threads ) {
     ASSERT(max_threads >= 1,"The limited number of threads should be positive");
     maxProcs = get_max_procs();
     if (maxProcs < max_threads) {
@@ -155,9 +155,10 @@ int limit_number_of_threads( int max_threads ) {
     }
 
     ASSERT(max_threads <= int(sizeof(mask_t) * CHAR_BIT), "The mask size is not enough to set the requested number of threads.");
-    int st = get_start_affinity_process();
-    for (int i = st; i < st + max_threads; ++i) {
-        CPU_SET(i, &new_mask);
+    std::vector<int> cpuset_indices = get_cpuset_indices();
+
+    for (int i = 0; i < max_threads; ++i) {
+        CPU_SET(cpuset_indices[i], &new_mask);
     }
     int err = setaffinity(new_mask);
 #endif
--- a/test/common/utils_concurrency_limit.h
+++ b/test/common/utils_concurrency_limit.h
@@ -102,6 +102,159 @@ static int get_max_procs() {
     return maxProcs;
 }
 
+#if __TBB_USE_CGROUPS
+
+// Linux control groups support
+class cgroup_info {
+public:
+    cgroup_info()
+        : default_num_cpus(std::thread::hardware_concurrency()), // Fallback value
+          m_num_cpus(parse_num_cpus()) {}
+
+    unsigned num_cpus() const { return m_num_cpus; }
+
+private:
+    unsigned parse_num_cpus() {
+        const std::string cgroup_line = find_cgroup_mount(mounts_path);
+        if (cgroup_line.empty())
+            return default_num_cpus; // Return default if no cgroup mount found
+
+        unsigned version = 0;
+        std::string mount_point;
+        if (!parse_version_and_mount_point(cgroup_line.c_str(), mount_point, version))
+            return default_num_cpus; // Return default if parsing fails
+
+        if (1 == version)
+            return determine_num_cpus_for_cgroup_v1(mount_point);
+        else
+            return determine_num_cpus_for_cgroup_v2(mount_point);
+    }
+
+    std::string find_cgroup_mount(const char* file_path) {
+        std::ifstream f(file_path);
+        if (!f)
+            return std::string();
+
+        std::string line;
+        while (std::getline(f, line)) {
+            if (line.length() < 6)
+                continue; // Skip empty or too short lines
+
+            if (std::strncmp(line.data(), "cgroup", 6) != 0)
+                continue; // Skip non-cgroup lines
+
+            return line;
+        }
+
+        return std::string();
+    }
+
+    bool read_line_from_file(const std::string& file_path, std::string& line) {
+        std::ifstream f(file_path);
+        if (!f || !std::getline(f, line))
+            return false; // Failed to read line
+        return true;
+    }
+
+    template <typename T>
+    bool convert_token_from_istream(std::istringstream& istream, T& value) {
+        if (!(istream >> value))
+            return false; // Failed to read period
+        return true;
+    }
+
+    bool parse_version_and_mount_point(const char* input, std::string& mount_point,
+                                       unsigned& version)
+    {
+        std::istringstream iss(input);
+        std::string device, version_str;
+        std::string* strs[] = {&device, &mount_point, &version_str};
+        for (auto& str : strs) {
+            if (!convert_token_from_istream(iss, *str))
+                return false; // Failed to read expected fields
+        }
+
+        version = 2;            // default
+        if (version_str == "cgroup")
+            version = 1;
+
+        return true;
+    }
+
+    bool parse_cpu_quota(std::istringstream& istream, unsigned long long& cpu_quota) {
+        std::string quota_str;
+        if (!convert_token_from_istream(istream, quota_str))
+            return false; // Failed to read quota
+
+        if (quota_str == "max") // CPU quota may be set to "max", which means no limit
+            return false;
+
+        std::istringstream cpu_quota_stream(quota_str);
+        return convert_token_from_istream(cpu_quota_stream, cpu_quota);
+    }
+
+    bool parse_cpu_period(std::istringstream& istream, unsigned long long& cpu_period) {
+        return convert_token_from_istream(istream, cpu_period);
+    }
+
+    unsigned determine_num_cpus_for_cgroup_v1(const std::string& mount_point) {
+        std::string line;
+        if (!read_line_from_file(mount_point + "/cpu/cpu.cfs_quota_us", line))
+            return default_num_cpus;
+
+        std::istringstream iss(line);
+        unsigned long long cpu_quota = 0;
+        if (!parse_cpu_quota(iss, cpu_quota))
+            return default_num_cpus;
+
+        if (!read_line_from_file(mount_point + "/cpu/cpu.cfs_period_us", line))
+            return default_num_cpus;
+
+        iss = std::istringstream(line);
+        unsigned long long cpu_period = 0;
+        if (!parse_cpu_period(iss, cpu_period))
+            return default_num_cpus;
+
+        return determine_num_cpus(cpu_quota, cpu_period);
+    }
+
+    unsigned determine_num_cpus_for_cgroup_v2(const std::string& mount_point) {
+        std::string line;
+        if (!read_line_from_file(mount_point + "/cpu.max", line))
+            return default_num_cpus;
+
+        std::istringstream iss(line);
+        unsigned long long cpu_quota = 0;
+        if (!parse_cpu_quota(iss, cpu_quota))
+            return default_num_cpus;
+
+        unsigned long long cpu_period = 0;
+        if (!parse_cpu_period(iss, cpu_period))
+            return default_num_cpus;
+
+        return determine_num_cpus(cpu_quota, cpu_period);
+    }
+
+    unsigned determine_num_cpus(unsigned long long cpu_quota, unsigned long long cpu_period) {
+        if (0 == cpu_period)
+            return default_num_cpus; // Avoid division by zero, use the default number of CPUs
+
+        const unsigned num_cpus = static_cast<unsigned>((cpu_quota + cpu_period / 2) / cpu_period);
+        return num_cpus > 0 ? num_cpus : 1;
+    }
+
+    static constexpr const char* mounts_path = "/proc/self/mounts";
+    const unsigned default_num_cpus; // Fallback value if parsing fails
+
+    const unsigned m_num_cpus;
+};
+
+static unsigned get_cgroups_max_concurrency() {
+    static cgroup_info cgroup_info_instance;
+    return cgroup_info_instance.num_cpus();
+}
+#endif
+
 static std::vector<int> get_cpuset_indices() {
 #if __linux__
     cpu_set_t mask;
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,5 @@
 # Copyright (c) 2021-2025 Intel Corporation
+# Copyright (c) 2025 UXL Foundation Сontributors
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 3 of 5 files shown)


src/tbb/misc_ex.cpp
```diff
@@ -130,6 +133,222 @@ static std::atomic<do_once_state> hardware_concurrency_info;
...
+        if (!*paths_cache.v1_relative_path)
+            cache_relative_path_for(cgroup_fd, paths_cache);
+
+        if (snprintf(dir, PATH_MAX, "%s/%s", mnt_dir, paths_cache.v1_relative_path) >= 0)
+            try_read_cgroup_v1_num_cpus_from(dir, num_cpus);
+        return num_cpus;
+    }
+
+    static int parse_cpu_constraints() {
+        using unique_mounts_file_t = std::unique_ptr<FILE, decltype(&close_mounts_file)>;
```
isaevil: Consider declaring `close_mounts_file` as lambda inside this function since it is not used anywhere else.

src/tbb/misc_ex.cpp
```diff
@@ -130,6 +133,222 @@ static std::atomic<do_once_state> hardware_concurrency_info;
...
+        if (!cgroup_file_ptr)
+            return error_value; // Failed to open cgroup file
+
+        unique_mounts_file_t mounts_file_ptr(setmntent("/proc/self/mounts", "r"), &close_mounts_file);
+        if (!mounts_file_ptr)
+            return error_value;
+
+        cgroup_paths relative_paths_cache;
+        struct mntent mntent;
+        const std::size_t buffer_size = 4096; // Allocate a buffer for reading mount entries
```
isaevil:
```suggestion
        constexpr std::size_t buffer_size = 4096; // Allocate a buffer for reading mount entries
```

src/tbb/misc_ex.cpp
```diff
@@ -130,6 +133,223 @@ static std::atomic<do_once_state> hardware_concurrency_info;
 
 static int theNumProcs;
 
+
+// Linux control groups support
+class cgroup_info {
```
akukanov: I'd recommend to move this class into a separate source file.

src/tbb/cgroup_info.h
```diff
@@ -0,0 +1,254 @@
...
+    limitations under the License.
+*/
+
+#ifndef _TBB_cgroup_info_H
+#define _TBB_cgroup_info_H
+
+#include <stdio.h>
+#include <limits.h>
+
+#include <cstring>
```
akukanov: Why are the C header files included in different ways: `stdio.h` and `limits.h` vs. `cstring`?

src/tbb/misc_ex.cpp
```diff
@@ -33,6 +34,7 @@
 #else
 #include <unistd.h>
 #if __unix__
+#include "cgroup_info.h"
```
akukanov: Is it really for `__unix__`, and not just `__linux__`?

isaevil: Moved under `__linux__`

src/tbb/cgroup_info.h
isaevil: Made it consistent

rfcs/supported/control_groups/support_for_cpu_resource_constraints.org
```diff
@@ -0,0 +1,153 @@
...
+      | grep "period\|quota"
+  cpu.cfs_period_us
+  cpu.cfs_quota_us
+#+end_src
+
+*** Minimizing File Reads
+On many modern systems, controllers are automatically mounted under =/sys/fs/cgroup= (v1
+controllers) and =/sys/fs/cgroup/unified= (v2 controllers) [4]. In practice, v2 controllers usually
+automatically mounted under v1 controllers path, that is =/sys/fs/cgroup=. So, to avoid parsing
+=/proc/self/mounts=, these paths can be tried first.
```
akukanov, line 150: I am still not sure that this is the right approach to optimize the procedure.
The expected most common use case, and so the one we should be optimizing for, is "there are no cgroup-based CPU restrictions". In other words, the sooner we can reliably detect that cgroups do not apply to the process, the better.

If parsing `/proc/self/mounts` is required to prove absence of cgroup restrictions, exploring `/sys/fs/cgroup` before will make the common path longer, not shorter.

rfcs/supported/control_groups/support_for_cpu_resource_constraints.org
```diff
@@ -0,0 +1,153 @@
...
+
+Since each of the v1 controllers has an associated configuration option that must be set in order to
+employ that controller [4], it makes sense to filter only those mount points that contain =cpu=
+controller in the mount options.
+
+*** The Algorithm Outline
+Based on the described considerations, the algorithm outline should be the following:
+
+1. Find relative path in =/proc/self/cgroup=
+2. Apply it to the most probable paths and look for the information there:
```
akukanov: Maybe parыing `/proc/self/cgroup` in the step (1) can reliably answer if cgroup CPU restrictions apply or not.

The format of this file is described in https://man7.org/linux/man-pages/man7/cgroups.7.html:

> For each cgroup hierarchy of which the process is a member, there is one entry containing three colon-separated fields: `hierarchy-ID:controller-list:cgroup-path`
>  For example: `5:cpuacct,cpu,cpuset:/daemons`

In the case of cgroups-v1, according to my understanding the CPU controller is used if and only if there is a record in this file containing `cpu` in the second field.

In the case of cgroups-v2, the first and the second fields are always 0 and empty, respectively: `0::/cgroup-path`. However, I believe the path referring to the root of the cgroups hierarchy: `0::/` also indicates that the CPU controller is not active for the current process, based on the information in https://www.kernel.org/doc/html/v6.15/admin-guide/cgroup-v2.html#cgroup-v2-cpu:

> cpu.max
A read-write two value file which exists on non-root cgroups. The default is “max 100000”.

`cpu.max` is the file describing the quota (comes first) and the period (comes second) values for the CPU controller. If this file only exists on non-root cgroups, then in the root cgroup directory we will not find the information we need. Therefore, `0::/` in `/proc/self/cgroup` is, I believe, an indication that there is no restrictions set and it makes no sense to search for the cgroup mount points etc.

isaevil: Right, `cpu.max` can be set only in non-root cgroups. But my understanding is that `0::/` in `/proc/self/cgroup` doesn't always mean it is an actual root cgroup. 
There is such thing as [cgroup namespaces](https://www.man7.org/linux/man-pages/man7/cgroup_namespaces.7.html) which allow to virtualize the view of a cgroups, meaning that nested cgroups can appear as `0::/`. Essentially, that's what docker does when you start a docker container. This is the example of docker container started with `--cpus 24` flag on HCC system on Ubuntu 25.04:
```bash
## Inside the container
# cat /proc/self/cgroup
0::/

# cat /sys/fs/cgroup/cpu.max # file exists
2400000 100000

## Outside the container
$ cat /proc/1208233/cgroup # PID of docker container
0::/system.slice/docker-38c1b7aefa88e31ba9762658a6fca86db854cd583d2640675dfb514b6a479828.scope

$ cat /sys/fs/cgroup/system.slice/docker-38c1b7aefa88e31ba9762658a6fca86db854cd583d2640675dfb514b6a479828.scope/cpu.max
2400000 100000
```
So it seems like we can't safely assume that `0::/` means no scheduling restrictions.

akukanov: I see. Thanks for exploring the question.

rfcs/supported/control_groups/support_for_cpu_resource_constraints.org
isaevil: It seems like we can try to optimize it by detecting whether the current process is running as part of container or inside namespace (see https://www.man7.org/linux/man-pages/man1/unshare.1.html). But it seems like the detection algorithm may take up to 3 VFS reads (`/proc/1/comm`, `/proc/1/cgroup`, `/proc/self/cgroup`), which perhaps should be faster since `/proc/self/mounts` may contain a lot of entries. Consider the following algorithm:
1. Read and inspect `/proc/self/cgroup`.
2. If `/proc/self/cgroup` has **NO** `cpu` controller entry (cgroup v1) and **NO** cgroup v2 entries `0::cgroup-path`, then we can safely say that there are no CFS constraints set.
3. If there **IS**  `cpu` controller entry (cgroup v1) or cgroup v2 entry (`0::some-relative-cgroup-path`, not `0::/`), then we should proceed and inspect `/proc/self/mounts`.
4. If there is `0::/` entry, it may mean that the process uses root cgroup and there can't be any `cpu.max` file if it is actual host root cgroup (see discussion https://github.com/uxlfoundation/oneTBB/pull/1777#discussion_r2273309051). So we need to determine if it is an actual host and not cgroup namespace. We can do that by inspecting files in `/proc/1` which usually refer to init system (like `systemd`). For example, when running inside Docker container `/proc/1/` will refer to either shell or program that was run under container. So the algorithm can be as follows.
   1. Inspect `/proc/1/comm`. If it is either `systemd` or `init`, then we definitely are not inside Docker at least. However, it doesn't guarantee that the process wasn't run under namespace (because `unshare` can be run without PID namespace), that is why step 2 is required
   2. Inspect `/proc/1/cgroup`. If process is running under namespace, it should show some variation of path which is relative to "the new virtual root". We could check for `0::/init.scope` (should be default when using `systemd`) or `0::/` or simply for `../` occurrences.

isaevil:
@kboyarinov @vossmjp please take a look.

rfcs/supported/control_groups/support_for_cpu_resource_constraints.org
akukanov: Thanks Ilya.
Seems the step 4ii gives more useful information than 4i, so maybe if those are swapped, we can sometimes have just two file reads instead of 3?

isaevil: > Seems the step 4ii gives more useful information than 4i, so maybe if those are swapped, we can sometimes have just two file reads instead of 3?

I am not sure if `/proc/1/cgroup` alone can answer that without checking `/proc/1/comm`. I think we still need to have both `systemd` or `init` in and `0::/init.scope` or `0::/` in `/proc/1/cgroup`.

By the way, I've run some measurements of latency of `initialize_hardware_concurrency_info` function in misc_ex.cpp. And I have following results for HOST case (median time after 100 runs):
| With algorithm mentioned in discussion (msec) | Without the algorithm (msec) | Without cgroup patch - master branch (msec) |
|--------|--------|--------|
| 0.0489575 | 0.086152 | 0.030635 |

rfcs/supported/control_groups/support_for_cpu_resource_constraints.org
```diff
@@ -0,0 +1,153 @@
...
+   - =/sys/fs/cgroup= for cgroup v1 and v2
+   - =/sys/fs/cgroup/unified= for cgroup v2
+3. If not found, loop over =/proc/self/mounts= and find a mount point for cgroupfs
+   - For cgroup v1, make sure the mount options contain =cpu= controller
+4. Concatenate the found mount path with the path from (1) and look for the information there
+5. Repeat from (3) until the info is found or the mount points are exhausted.
```
isaevil: Apparently, that is not what the current implementation does. The actual algorithm implemented is as follows:
1. Start looping over `/proc/self/mounts` straight away.
2. If found cgroup mount (either version of cgroup), try to parse this entry
    1. Parse `/proc/self/cgroup` to fill relatives path cache (interestingly, the algorithm will stop if it found either cgroup v1 cpu controller or cgroup v2)
    2. Try read from cgroup mount dir (I think that is incorrect and you should always try to read from relative path first)
    3. Try read from cgroup mount dir concatenated with relative path in `/proc/self/cgroup`

akukanov: I think @aleksei-fedotov has changed the RFC in response to the feedback, but not yet the implementation.
Anyway, as we discussed above, we probably want to change both, for faster exit path.

rfcs/supported/control_groups/support_for_cpu_resource_constraints.org
isaevil: I've implemented the algorithm (but only left 4ii part from step 4 to detect systemd init process). I also added the attempt to prematurely read from /sys/fs/cgroup + cgroup relative path. See the commit [46a52a2](https://github.com/uxlfoundation/oneTBB/pull/1777/commits/46a52a21d2a19feca360e4f8d7060572a82f8145)

src/tbb/cgroup_info.h
```diff
@@ -0,0 +1,308 @@
...
+            }
+
+            if (path_start) {
+                std::strncpy(relative_path, path_start, rel_path_size);
+                relative_path[rel_path_size - 1] = '\0'; // Ensure null-termination after copy
+
+                char* new_line = std::strrchr(relative_path, '\n');
+                if (new_line)
+                    *new_line = '\0'; // Ensure no new line at the end of the path is copied
+            }
```
kboyarinov: Can we swap finding the "\n" and copying to copy only the real number of bytes?

```suggestion
                std::size_t real_rel_path_size = std::strcspn(path_start, '\n');
                std::strncpy(relative_path, path_start, real_rel_path_size);
                relative_path[real_rel_path_size] = '\0';
```

src/tbb/cgroup_info.h
```diff
@@ -0,0 +1,308 @@
...
+    static void cache_relative_path_for(FILE* cgroup_fd, cgroup_paths& paths_cache) {
+        char* relative_path = nullptr;
+        char line[rel_path_size] = {0};
+        const char* last_char = line + rel_path_size - 1;
+
+        const char* path_start = nullptr;
+        while (std::fgets(line, rel_path_size, cgroup_fd)) {
+            path_start = nullptr;
+
+            if (std::strncmp(line, "0::", 3) == 0) {
```
kboyarinov: Am I correct that there cannot be two "v1" or "v2" entries in the file?
If so, I would add an assertion that the second found entry is invalid.

src/tbb/cgroup_info.h
```diff
@@ -0,0 +1,308 @@
...
+                path_start = line + 3; // cgroup v2 unified path
```
kboyarinov: Can we avoid using "magic" constants and create variables with meaningful name for such cases?
Same applies to `look_for_cpu_controller_path`.

src/tbb/cgroup_info.h
isaevil: Do you mean having both v1 and v2 cgroup at /proc/self/cgroup? My understanding is that there can be. See https://www.man7.org/linux/man-pages/man7/cgroups.7.html.
> The two systems are implemented so that both v1 controllers and v2
       controllers can be mounted on the same system.  Thus, for example,
       it is possible to use those controllers that are supported under
       version 2, while also using version 1 controllers where version 2
       does not yet support those controllers.  The only restriction here
       is that a controller can't be simultaneously employed in both a
       cgroups v1 hierarchy and in the cgroups v2 hierarchy.

So you can have both v1 and v2 cgroup in /proc/self/cgroup, you can't only have the same controller in both versions simultaneously.

kboyarinov: No, I mean if some entry was found (v1 or v2), there cannot be more entries of the same type. It seems like the algorithm relies on that, otherwise you overwrite the relative path in the cache.

isaevil: Yes, cgroup v1 CPU controller should be present only once and there should be only one entry of cgroup v2.

src/tbb/cgroup_info.h
```diff
@@ -0,0 +1,308 @@
...
+        if (std::strncmp(cpu_quota_str, "max", 3) == 0) {
+            num_cpus = unlimited_num_cpus;  // "max" means no CPU constraint
+            return true;
+        }
+
+        errno = 0; // Reset errno before strtoll
+        char* str_end = nullptr;
+        long long cpu_quota = std::strtoll(cpu_quota_str, &str_end, /*base*/ 10);
+        if (errno == ERANGE || str_end == cpu_quota_str)
+            return false;
```
kboyarinov, line 197: Should not we reset `errno` before exiting?

isaevil: I don't think that it is strictly necessary, because usage pattern of `errno` implies that before any call that sets `errno`, it is better to reset `errno` to 0.

src/tbb/cgroup_info.h
isaevil: Done.

src/tbb/cgroup_info.h
isaevil: Added assertions.

src/tbb/cgroup_info.h
isaevil: Done.

kboyarinov (approved):
LGTM

src/tbb/misc_ex.cpp
aleksei-fedotov: Actually, it should be for `__unix__`.

rfcs/supported/control_groups/support_for_cpu_resource_constraints.org
```diff
@@ -119,6 +119,30 @@ all of the possible options and stop once the data is found and successfully rea
   cpu.cfs_quota_us
 #+end_src
 
+*** Determining If Scheduling Constraints Are Set
+As an optimization, it makes sense to first check whether any CPU scheduling constraints can be set.
```
aleksei-fedotov: Probably, telling about state of things rather their possibility is better here.
```suggestion
As an optimization, it makes sense to first check whether any CPU scheduling constraints are set.
```

rfcs/supported/control_groups/support_for_cpu_resource_constraints.org
```diff
@@ -132,8 +156,12 @@ controller in the mount options.
 *** The Algorithm Outline
 Based on the described considerations, the algorithm outline should be the following:
 
-1. Find relative path in =/proc/self/cgroup=
-2. Apply it to the most probable paths and look for the information there:
+1. Parse =/proc/self/cgroup= to:
+   - Find relavive paths to cgroup controllers
```
aleksei-fedotov: A typo:
```suggestion
   - Find relative paths to cgroup controllers
```

src/tbb/cgroup_info.h
```diff
@@ -0,0 +1,314 @@
...
+        }
+
+        auto close_mounts_file = [](std::FILE* file) { endmntent(file); };
+        using unique_mounts_file_t = std::unique_ptr<std::FILE, decltype(close_mounts_file)>;
+        unique_mounts_file_t mounts_file_ptr(setmntent("/proc/self/mounts", "r"), close_mounts_file);
+        if (!mounts_file_ptr)
+            return error_value;
+
+        // TODO: To avoid parsing /proc/self/mounts, read relative paths first and try opening
+        //       necessary files in the most probable mount points (see RFC for details)
```
akukanov: Is not that TODO already done in `try_common_cgroup_mount_path`?

src/tbb/cgroup_info.h
```diff
@@ -0,0 +1,314 @@
...
+
+        const long long num_cpus = (cpu_quota + cpu_period - 1) / cpu_period;
+        return num_cpus > 0 ? int(num_cpus) : 1; // Ensure at least one CPU is returned
+    }
+
+    static constexpr std::size_t rel_path_size = 256; // Size of the relative path buffer
+    struct cgroup_paths {
+        char v1_relative_path[rel_path_size] = {0};
+        char v2_relative_path[rel_path_size] = {0};
+    };
```
akukanov: Why both v1 and v2 paths are needed? I thought that parsing `/proc/self/cgroup` also gives the answer whether it is v1 or v2 controller, and it can't be both. A single char-size enum with the values of 0, 1, and 2 would be sufficient to distinguish no constraints vs. v1 vs, v2 for subsequent procedures.

src/tbb/cgroup_info.h
```diff
@@ -0,0 +1,314 @@
...
+        bool is_cgroup_v2_with_root_path =
+            relative_paths_cache.v2_relative_path[0] == '/' && !relative_paths_cache.v2_relative_path[1];
+        if (!has_cgroup_v2 && !has_cgroup_v1_cpu_controller) {
+            return false; // No CPU constraints found
+        } else if (has_cgroup_v1_cpu_controller || !is_cgroup_v2_with_root_path) {
+            return true;
+        }
+        __TBB_ASSERT(!has_cgroup_v1_cpu_controller && has_cgroup_v2 && is_cgroup_v2_with_root_path, nullptr);
+
+        // Usually PID 1 is the init process
```
akukanov: Are there exceptions? :)

isaevil: I think I forgot to elaborate this comment, but yes. Exception is when [pid namespaces](https://www.man7.org/linux/man-pages/man7/pid_namespaces.7.html) are used. They are used by default in containerized environment and when are used, PID 1 becomes shell if the container is run in interactive mode or  application that was run under container.

src/tbb/cgroup_info.h
isaevil: My understanding is that process can be mount to both cgroup v1 and v2 (so it is possible to find entry of cgroup v2 and entries of cgroup v1 controllers), but it actually can't have the same controller (in our case, CPU controller) in both places, it has to be in the only one of them. That makes me think that we can actually have only one buffer for relative path as you suggest with version indication.

The current implementation uses the parsed paths as an version indication, what cgroup version is mounted to the process. For example, if `v2_relative_path` is "empty", check `v1_relative_path`.

Perhaps @aleksei-fedotov can describe better the idea.

aleksei-fedotov: As far as I remember, the idea was to make the algorithm robust and yet read as little as possible from the files, taking into account that more then one line about CPU controller for certain version of cgroup cannot be present, but multiple mount points in `/proc/self/mounts` can. The idea was that once relative path of whatever cgroup version is found in `/proc/self/cgroup` file, it gets concatenated with the path from `/proc/self/mounts` file and corresponding CPU controller file(s), and the resulting file path is checked. If CPU controller file does not exist, the algorithm continues to read `/proc/self/cgroup` and `/proc/self/mounts` until the resulting file is found.

src/tbb/cgroup_info.h
```diff
@@ -0,0 +1,314 @@
...
+                             "Found second entry of cgroup v1 cpu controller");
+
+                relative_path = paths_cache.v1_relative_path;
+            }
+
+            if (path_start) {
+                // Ensure no new line at the end of the path is copied
+                std::size_t real_rel_path_size = std::strcspn(path_start, "\n");
+                std::strncpy(relative_path, path_start, real_rel_path_size);
+                relative_path[real_rel_path_size] = '\0';
```
aleksei-fedotov: How much of initialization time can be saved if the memory is passed only once here?

src/tbb/cgroup_info.h
akukanov: Ah, you meant the hosting OS init process. Well, then perhaps "usually" is not the best word since containers are also usual nowadays. Anyway, expanding the comment a bit might be useful.

src/tbb/cgroup_info.h
akukanov: I think having just one relative path string does not contradict to the overall design Aleksei has described, and a dedicated "cgroup version" field could simplify some processing logic. 

For the parsing of `/proc/self/cgroup`, a v1 CPU controller path takes precedence over a v2 unified controller path (i.e., overrides it), but not vice versa. On the other hand, for the parsing of `/proc/1/cgroup` we are not interested in the v1 controller at all, we want to get the v2 path. That's additional complication - though of course it's doable, e.g. with an extra parameter indicating that v1 records should be ignored.

src/tbb/cgroup_info.h
```diff
@@ -96,56 +100,58 @@ class cgroup_info {
         const char* last_char = line + rel_path_size - 1;
 
         const char* path_start = nullptr;
-        while (fgets(line, rel_path_size, cgroup_fd)) {
+        constexpr std::size_t cgroup_v2_prefix_size = 3;
+        while (std::fgets(line, rel_path_size, cgroup_fd)) {
```
aleksei-fedotov, line 112: Shall we stop reading if both paths have been found?

src/tbb/misc_ex.cpp
isaevil: Well, while the implementation is designed in a UNIX-compatible way, it doesn't make sense to me to invoke cgroup parsing somewhere besides Linux. CFS is the process scheduler of the Linux kernel scheduler and we are interested in only scheduling constraints of CFS.

src/tbb/cgroup_info.h
```diff
@@ -209,31 +215,85 @@ class cgroup_info {
...
+        bool has_cgroup_v1_cpu_controller = *relative_paths_cache.v1_relative_path;
+        bool has_cgroup_v2 = *relative_paths_cache.v2_relative_path;
+        bool is_cgroup_v2_with_root_path =
+            relative_paths_cache.v2_relative_path[0] == '/' && !relative_paths_cache.v2_relative_path[1];
+        if (!has_cgroup_v2 && !has_cgroup_v1_cpu_controller) {
+            return false; // No CPU constraints found
+        } else if (has_cgroup_v1_cpu_controller || !is_cgroup_v2_with_root_path) {
+            return true;
+        }
+        __TBB_ASSERT(!has_cgroup_v1_cpu_controller && has_cgroup_v2 && is_cgroup_v2_with_root_path, nullptr);
```
aleksei-fedotov: The number of checks can be reduced:
```suggestion
        if (/*has_cgroup_v1=*/ *relative_paths_cache.v1_relative_path)
            return true;
        else if (/*no_cgroup_v2=*/ !*relative_paths_cache.v2_relative_path)
            return false;
        else if (/*no_root_path_for_cgroup_v2=*/ relative_paths_cache.v2_relative_path[1])
            return true;
        
        __TBB_ASSERT('/' == relative_paths_cache.v2_relative_path[0], "Incorrect cgroup file?");
```

src/tbb/cgroup_info.h
```diff
@@ -209,31 +215,85 @@ class cgroup_info {
...
+
+        // Usually PID 1 is the init process
+        unique_file_t init_process_cgroup_file(std::fopen("/proc/1/cgroup", "r"), &close_file);
```
aleksei-fedotov, line 244: At first, check the file is opened w/o error?

src/tbb/cgroup_info.h
```diff
@@ -209,31 +215,85 @@ class cgroup_info {
...
+        cgroup_paths init_process_cgroup_cache{};
+        cache_relative_path_for(init_process_cgroup_file.get(), init_process_cgroup_cache);
```
aleksei-fedotov: Read only the v2 path, since it's what needed below?

src/tbb/cgroup_info.h
```diff
@@ -209,31 +215,85 @@ class cgroup_info {
...
+        if (*init_process_cgroup_cache.v2_relative_path) {
+            // If the init process cgroup path is "/init.scope", it means systemd is used
+            if (!std::strncmp(init_process_cgroup_cache.v2_relative_path, "/init.scope", 11)) {
+                return false;
+            }
+        }
```
aleksei-fedotov: Why double check here? Does it give any speedup comparing to invocation of `std::strncmp` as in:
```suggestion
        // If the init process cgroup path is "/init.scope", it means systemd is used
        if (!std::strncmp(init_process_cgroup_cache.v2_relative_path, "/init.scope", 11)) {
            return false;
        }
```

src/tbb/cgroup_info.h
aleksei-fedotov: If I understand your idea correctly, you are proposing to look only for v1 CPU limitations once v1 CPU controller is found in `/proc/self/cgroup` file, and ignore and do not even try to find v2 CPU limitations in this case.
However, what if v2 line is first found in `/proc/self/cgroup` file? I don't remember documentation saying that it is impossible. At this point, we are still not sure whether CPU limitation from either cgroup version exists. So, we still need to keep it somewhere, and yet anticipate that v1 line can appear later.

akukanov: Exactly, this is what I meant by saying that a v1 path takes precedence over a v2 path. If the latter is found, we should keep looking - unless the procedure is invoked for `/proc/1/cgroup` and v1 paths are of no interest. And if we do find a v1 path later on, it overwrites the v2 path found earlier.

src/tbb/cgroup_info.h
akukanov: "The behavior [of `strncmp`] is undefined when either `lhs` or `rhs` is the null pointer."
https://en.cppreference.com/w/cpp/string/byte/strncmp.html

src/tbb/cgroup_info.h
isaevil: I tried to find some information on order of entries at `/proc/pid/cgroup`, but seems like it is not strictly defined, though I've seen that on systems with both cgroup v1 and cgroup v2 mounted, entries are in descending order based on hierarchy ID. But I don't think we can rely on that.

So, as I understand, the idea is:
1. Parse `/proc/self/cgroup` kind of "asymmetrically"
    1. If CPU controller of cgroup v1 is found -> stop parsing
    2. If cgroup v2 controller is found -> proceed with parsing to make sure no cgroup v1 CPU controller is present (or proceed only if it is presumably root cgroup `0::/`?).
    3. Since parsing `/proc/1/cgroup` is only needed for cgroup v2 we can parameterize parsing function to skip follow up from 1ii bullet and look only for cgroup v2 entry.
 2. At this point we only have one relative path and the cgroup version used, thus try to read quota/period using this path
     1. If failed we can proceed parsing `/proc/self/cgroup` further (for example, we found cgroup v1 first and try to find cgroup v2) to find another entry to improve robustness

akukanov: My current understanding is that parsing `/proc/self/cgroup` is done once and gives a definite answer where to look for CPU controller data and which version of cgroups is used for that. If later there is nothing found in the path, then the controller is not activated.

Or am I wrong to think that if a v1 CPU controller is present in `/proc/self/cgroup` then a v2 CPU controller is effectively disabled? Do you have to check both paths to find out which controller is active?

isaevil: > Or am I wrong to think that if a v1 CPU controller is present in /proc/self/cgroup then a v2 CPU controller is effectively disabled? Do you have to check both paths to find out which controller is active?

I think you are correct. [cgroup manual](https://www.man7.org/linux/man-pages/man7/cgroups.7.html) states:

> The two systems are implemented so that both v1 controllers and v2 controllers can be mounted on the same system.  Thus, for example, it is possible to use those controllers that are supported under version 2, while also using version 1 controllers where version 2 does not yet support those controllers.  The only restriction here is that a controller can't be simultaneously employed in both a cgroups v1 hierarchy and in the cgroups v2 hierarchy.

So I think we can safely assume that if cgroup v1 CPU controller is present, we won't find the same controller in cgroup v2 mount.

src/tbb/cgroup_info.h
aleksei-fedotov: `v2_relative_path` is an array of chars, so it is always non-null here.

src/tbb/cgroup_info.h
isaevil: Removed.

src/tbb/cgroup_info.h
isaevil: Updated implementation in [7c3feb3](https://github.com/uxlfoundation/oneTBB/pull/1777/commits/7c3feb3f69c44ac866594273051c41967eac0368)

src/tbb/cgroup_info.h
isaevil: Updated the comment

src/tbb/cgroup_info.h
isaevil: Done.

src/tbb/cgroup_info.h
isaevil: Fixed

src/tbb/cgroup_info.h
```diff
@@ -0,0 +1,331 @@
...
+                    path_start = v1_path_start;
+                    break;
+                }
+            }
+        }
+        if (path_start) {
+            // Ensure no new line at the end of the path is copied
+            std::size_t real_rel_path_size = std::strcspn(path_start, "\n");
+            std::strncpy(pcd.relative_path, path_start, real_rel_path_size);
+            pcd.relative_path[real_rel_path_size] = '\0';
```
aleksei-fedotov: Technically, the `real_rel_path_size` can be longer than the `rel_path_size`.
Shall we at least add an assert for this? We also can extend the `relative_path` array to hold up to PATH_MAX characters?

aleksei-fedotov: Sorry, I was wrong. It cannot be longer because `fgets` reads at most `rel_path_size - 1` and then all the work happens within this limit. I guess, adding an assertion `__TBB_ASSERT(real_rel_path_size < rel_path_size, nullptr)` may be enough or just leave a comment?

test/common/utils_concurrency_limit.h
```diff
@@ -101,26 +107,242 @@ static int get_max_procs() {
     return maxProcs;
 }
 
-int get_start_affinity_process() {
+#if __TBB_USE_CGROUPS
+
+// Linux control groups support
+class cgroup_info {
```
aleksei-fedotov, line 113: I think the implementation is good. Shall we copy it to the test for reference before merging the patch or maybe use some dumb version in tests that goes over the official paths without tricks to avoid parsing `/proc/self/mounts`?

isaevil: I think leaving more brute-force implementation in tests is good, because it will allow us to better verify whether our current more optimized implementation works correctly or not.

aleksei-fedotov: Yeah. I have the same opinion.

test/common/utils_concurrency_limit.h
```diff
@@ -101,26 +107,242 @@ static int get_max_procs() {
...
+            return error_value; // Failed to open cgroup file
+
+        auto close_mounts_file = [](std::FILE* file) { endmntent(file); };
+        using unique_mounts_file_t = std::unique_ptr<std::FILE, decltype(close_mounts_file)>;
+        unique_mounts_file_t mounts_file_ptr(setmntent("/proc/self/mounts", "r"), close_mounts_file);
+        if (!mounts_file_ptr)
+            return error_value;
+
+        // TODO: To avoid parsing /proc/self/mounts, read relative paths first and try opening
+        //       "/sys/fs/cgroup/<relative_path>/cpu.max" for cgroup v2.
```
aleksei-fedotov: Remove this unneeded comment?
```suggestion
```

test/common/utils_concurrency_limit.h
```diff
@@ -101,26 +107,242 @@ static int get_max_procs() {
...
+
+            // Now try reading including relative path
+            if (std::snprintf(dir, PATH_MAX, "%s/%s", mnt_dir, paths_cache.v2_relative_path) >= 0)
+                try_read_cgroup_v2_num_cpus_from(dir, num_cpus);
+            return num_cpus;
+        }
+
+        __TBB_ASSERT(std::strncmp(mnt_type, "cgroup", 6) == 0, "Unexpected cgroup type");
+
+        // TODO: Before opening the path, make sure the mnt_opts entry contains "cpu" controller
```
aleksei-fedotov: This one is also unneeded.
```suggestion
```

src/tbb/cgroup_info.h
isaevil: Added assert.

test/common/utils_concurrency_limit.h
```diff
@@ -101,26 +107,242 @@ static int get_max_procs() {
     return maxProcs;
 }
 
-int get_start_affinity_process() {
+#if __TBB_USE_CGROUPS
+
+// Linux control groups support
+class cgroup_info {
+public:
+    static bool is_cpu_constrained(int& constrained_num_cpus) {
```
aleksei-fedotov, line 117:
```suggestion
    // The algorithm deliberately goes over slow but reliable paths to determine possible CPU
    // constraints. This helps to make sure the optimizations in the code are correct.
    static bool is_cpu_constrained(int& constrained_num_cpus) {
```

kboyarinov (approved):
LGTM

PR #1778, opened 2025-07-11
Add the document describing isolated_task_group

akukanov:
This PR adds the "RFC"/design document for `isolated_task_group` as an experimental feature of the library.

### DIFF
--- a/rfcs/experimental/isolated_task_group/readme.md
+++ b/rfcs/experimental/isolated_task_group/readme.md
@@ -87,6 +87,10 @@ This experimental feature was added in [TBB 2019 Update 9](https://github.com/ux
 For unknown reasons (most likely just by oversight) the documentation for the class is missed
 for oneTBB 2021 and beyond.
 
+`isolated_task_group` uses the ABI function `isolate_within_arena`, passing `this` (the task group
+ object address) as the isolation tag.  For comparison, `this_task_arena::isolate` uses the same
+ ABI function, but the isolation tag is the address of `isolate`'s function object argument.
+
 ## Does oneTBB still need `isolated_task_group`?
 
 ### Use cases


rfcs/experimental/isolated_task_group/readme.md
```diff
@@ -0,0 +1,201 @@
+# Task group as an isolation scope
+
+## Introduction
+
+In-arena task isolation scopes were introduced into TBB to restrict task stealing within arena
+in scenarios where stealing arbitrary tasks lead to correctness or performance issues.
```
vossmjp:
```suggestion
in scenarios where stealing arbitrary tasks leads to correctness or performance issues.
```

rfcs/experimental/isolated_task_group/readme.md
```diff
@@ -0,0 +1,201 @@
+# Task group as an isolation scope
+
+## Introduction
+
+In-arena task isolation scopes were introduced into TBB to restrict task stealing within arena
```
vossmjp:
```suggestion
[In-arena task isolation scopes](https://oneapi-spec.uxlfoundation.org/specifications/oneapi/latest/elements/onetbb/source/task_scheduler/task_arena/this_task_arena_ns) were introduced into TBB to restrict task stealing within an arena
```

rfcs/experimental/isolated_task_group/readme.md
```diff
@@ -0,0 +1,201 @@
...
+in scenarios where stealing arbitrary tasks lead to correctness or performance issues.
+
+On the implementation level, an isolation scope is represented by a unique tag (usually an address)
+that is assigned to any task created within the dynamic extent of the scope and used by
+task dispatchers called within that extent. Other task dispatchers, and specifically those of
+oneTBB worker threads, are not restricted by isolation and can take any task, including those
+with an isolation tag.
+
+On the API level, the main way to create an isolation scope is to call `task_arena::isolate`.
```
vossmjp:
```suggestion
On the API level, the main way to create an isolation scope is to call `this_task_arena::isolate`.
```

rfcs/experimental/isolated_task_group/readme.md
```diff
@@ -0,0 +1,201 @@
...
+Typically, it wraps one or more parallel algorithm invocations, and the calling thread
```
vossmjp:
```suggestion
Typically, it is used to wrap one or more parallel algorithm invocations, and the calling thread
```

rfcs/experimental/isolated_task_group/readme.md
```diff
@@ -0,0 +1,201 @@
...
+is prohibited to take any task created outside of those algorithms. A typical usage pattern
+for this is to wrap nested parallel calls within an outer-level parallel algorithm which tasks
+depend on a thread-specific state (e.g., use TLS or acquire a lock).
```
vossmjp:
```suggestion
for this is to wrap nested parallel calls within an outer-level parallel algorithm in which tasks
depend on a thread-specific state (e.g., use TLS or acquire a lock).
```

rfcs/experimental/isolated_task_group/readme.md
```diff
@@ -0,0 +1,201 @@
...
+same task group.
+
+In this example the use of `try_lock` eliminates the risk of deadlock via unrestricted stealing.
+Still, isolation can be helpful for both the thread doing the initialization and those helping.
+The main advantage is to prevent tasks not related to the initialization of `some_obj` to be taken
+by threads that wait for that initialization to complete - so that it can be complete sooner
+and the threads can make progress on their outer level tasks.
+
+However, for that to happen each participating thread should use the same isolation tag, which is
+not possible with `task_arena::isolate`. And since a task group is used for collaborative waiting
```
vossmjp:
```suggestion
not possible with `this_task_arena::isolate`. And since a task group is used for collaborative waiting
```

rfcs/experimental/isolated_task_group/readme.md
```diff
@@ -0,0 +1,201 @@
...
+### Alternatives
+
+Over time, alternative ways to provide "shareable" in-arena isolation scopes were also considered.
+
+#### A `task_group_context` trait
+
+One of early ideas for in-arena isolation was to add a special trait to the `task_group_context`
+class that would communicate to the task scheduler that arbitrary stealing is not allowed
+for a certain parallel construct. The benefits of this approach are: a) the possibility to use
+with all TBB parallel constructs (algorithms, task groups, flow graphs), and b) the dynamic tree
```
vossmjp:
```suggestion
it with all TBB parallel constructs (algorithms, task groups, flow graphs), and b) the dynamic tree
```

rfcs/experimental/isolated_task_group/readme.md
```diff
@@ -0,0 +1,201 @@
...
+  therefore have minimal overheads.
+- At the task scheduler level, the high-level semantic information is mostly lost, and
+  significantly different usage patterns are often indistinguishable for the scheduler.
+  A task with an isolation property might spawn tasks which belong to the same parallel construct,
+  to the outer level one, or to a new nested one. It essentially means that, in order to support
+  isolation specified in `task_group_context` (and not by dedicated APIs), each task spawn should
+  check/obtain an isolation tag, causing extra overhead and breaking the pay-as-you-go principle.
+
+Seemingly, making task isolation a property of `task_group_context` (or something similar) would
+require a significant rehaul of both the scheduler internals and likely the public oneTBB APIs.
```
vossmjp:
```suggestion
require a significant overhaul of both the scheduler internals and likely the public oneTBB APIs.
```

rfcs/experimental/isolated_task_group/readme.md
```diff
@@ -0,0 +1,201 @@
...
+dynamic extent, i.e., the nested parallelism within) from any other work in that arena.
+
+Compared to the user-specified isolation tags, this approach would not have most of the downsides
+but also would lose the flexibility and broader usage potential. One issue would however remain
+the same - the regular methods of `task_group` can be used to create tasks that do not carry
+any isolation tag; and so user errors are still possible.
+
+## Exit criteria
+
+To decide on making `isolated_task_arena` a fully supported feature of the library, we need to
```
vossmjp:
```suggestion
To decide on making `isolated_task_group` a fully supported feature of the library, we need to
```

rfcs/experimental/isolated_task_group/readme.md
```diff
@@ -0,0 +1,201 @@
...
+  while (!initialized(some_obj)) {
+    if (try_lock(init_mtx)) {
+      tg.run_and_wait {
+        initialize(some_obj); // uses parallel_for, etc.
+      };
+      unlock(init_mtx);
+    } else { tg.wait(); }
+  }
+  ... // do the work that uses some_obj
+}
```
kboyarinov, line 38: I have a general question regarding the isolation. 
Will the thread that execute the isolated region and enters another isolated region temporary change the isolation?

I have imagined the example, when we have two lazily initialized objects `A` and `B` and initialization for both of them is done shown in this code snippet, but using the `isolated_task_group`. But the object `B` needs `A` to be initialized before running the parallel section of the initialization:

```cpp
A::initialize() { parallel_for(...) }

B::initialize() {
    A& a = getA();
    parallel_for(...);
}

A& getA() { /* done as in the code snippet from the RFC, but with isolated group */ }
B& getB() { ... }

int main() {
    parallel_for( {
        B& b = getB();
    })
}
```

As far as I understand, workers performing tasks of the `parallel_for` in `main` will be able to help initializing `B`. 
But will they be able to help initializing `A`?

kboyarinov, line 39: Can it make sense for two `isolated_task_group`s to share one isolation tag?
I thought of the case of having two objects `A` and `B`, both are lazy initialized using the pattern described, but using the `isolated_task_group`. But an object `B` needs `A` to be initialized before running the parallel phase of the initialization:

```cpp
A::initialize() { parallel_for(...) }

B::initialize() {
    A& a = getA();
    parallel_for(...);
}

A& getA() {
    while (!initialized(a_obj)) {
        if (try_lock(a_mutex)) {
           a_isolated_tg.run_and_wait( { initialize(a_obj); });
           unlock(a_mutex);
        } else { a_isolated_tg.wait(); }
    }
    return a_obj;
}
B& getB() { /*same as getA, but using b_mutex, b_isolated_tg and b_obj*/ }

int main() {
    parallel_for( {
        B& b = getB();
    })
}
```

As far as I understand, workers performing tasks of the `parallel_for` in `main` will be able to help initializing `B`.
But would they be able to help initializing `A`, as it is required to initialize `B`?

rfcs/experimental/isolated_task_group/readme.md
akukanov: Added a link to https://uxlfoundation.github.io/oneTBB/main/tbb_userguide/work_isolation.html.

rfcs/experimental/isolated_task_group/readme.md
akukanov: If the initialization of A is in its own isolation scope *and* is called only once in `B::initialize` - then no, threads that call `getB` will not be able to help with initializing A. It is prevented by the implementation mechanics: a task dispatcher that runs in an isolation scope can only take tasks with that scope's tag, not any other tasks.

If however `getA` would be called in the parallel loop of `B::initialize`, then threads working on that loop could possibly enter the isolation scope in `getA` and help.

kboyarinov: I just thought if it makes sense to have API in the future that allow sharing the same isolation for several `task_group`s?
In my understanding, in this example having `a_isolated_tg` and `b_isolated_tg` to share the same isolation tag will allow  workers in `main`'s `parallel_for` to participate in `A` initialization as well.

akukanov: Sharing the same tag means having no isolation between layers, in this case between `A::initialize` and `B::initialize`. If that's the goal, the simplest thing is to use a regular task group for A.

I guess you might want to allow `B::initialize` to take tasks from `A::initialize`, but not vice versa - i.e., to have one-way isolation. The current implementation of isolated tasks simply does not allow to do that. For that to be possible, we would need something similar to the binding of task_group_contexts.

Updated: perhaps even something more complicated, because task_group_contexts form a tree, but with task groups  it is not hard to build "isolation DAGs". Imagine in the example above there is an object C which lazy initialization also needs A. That means, `A::initialize` can be "nested" into both `B::initialize` and `C::initialize` at the same time.

PR #1779, opened 2025-07-11
Change get_task to not reset the task pool if some tasks were skipped

akukanov:
### Description 

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown


src/tbb/arena_slot.cpp
```diff
@@ -96,52 +102,40 @@ d1::task* arena_slot::get_task(execution_data_ext& ed, isolation_type isolation)
...
-            // We just restore the bounds for the available tasks.
-            // TODO: Does it have sense to move them to the beginning of the task pool?
-            __TBB_ASSERT( is_quiescent_local_task_pool_reset(), nullptr );
+            // All tasks have been checked.
             if ( result ) {
                 // If we have a task, it should be at H0 position.
                 __TBB_ASSERT( H0 == T, nullptr );
                 ++H0;
             }
             __TBB_ASSERT( H0 <= T0, nullptr );
```
akukanov: Seemingly, there is nothing useful done in the `if (task_pool_empty)` branch anymore, as it only has assertions and invariant preparations for these. Perhaps It can be removed completely.

Updated: done.

akukanov:
@kboyarinov @isaevil @dnmokhov Please take a look.

src/tbb/arena_slot.cpp
```diff
@@ -81,67 +83,54 @@ d1::task* arena_slot::get_task(execution_data_ext& ed, isolation_type isolation)
...
-            }
-            __TBB_ASSERT( H0 <= T0, nullptr );
-            if ( H0 < T0 ) {
-                // Restore the task pool if there are some tasks.
-                head.store(H0, std::memory_order_relaxed);
-                tail.store(T0, std::memory_order_relaxed);
-                // The release fence is used in publish_task_pool.
-                publish_task_pool();
-                // Synchronize with snapshot as we published some tasks.
-                ed.task_disp->m_thread_data->my_arena->advertise_new_work<arena::wakeup>();
```
kboyarinov, line 129: Just to clarify my understanding of the problem statement and the proposed fix (may duplicate the description in a bit:) ).

Let's assume that the current slot contain a single task, isolation of which prevents the owning thread to execute it. 
The owning thread skips this task while looking for a work to do and temporarily decrements the tail (head equal to tail in our case). 
Between the decrement and the restore of the tail, the stealing thread can inspect the current slot in `arena::out_of_work` and see head equal to tail meaning the current slot is out of tasks.

If all other slots in arena are out of tasks, the stealing thread signals to workers to leave the arena. 

When the restoration of head and tail is done by the owning thread, the `advertise_new_work` will request the workers to re-join the arena.

Am I correct that the idea of the patch is to prevent workers from leaving the arena in this case and double-check by reading the flag `has_skipped_tasks` that should be set by the owning thread while the tasks are skipped?

akukanov: That's correct overall, with a couple of comments to be made.

> Between the decrement and the restore of the tail, the **stealing** thread can inspect the current slot in arena::out_of_work...

Technically, it's not a stealing thread, as it does not at that point attempt to steal anything. I would call it an inspecting thread, to disambiguate with other threads that can really try stealing at the same time.

The second comment is that stealing threads might also skip tasks and temporarily make the slot appear empty. In the current implementation, they also call `advertise_new_work` afterwards. In the patch, the inspecting thread has to lock the pool for final inspection, therefore no stealing can happen at the same time.

src/tbb/arena_slot.cpp
```diff
@@ -81,67 +83,54 @@ d1::task* arena_slot::get_task(execution_data_ext& ed, isolation_type isolation)
                 __TBB_ASSERT( H0 == head.load(std::memory_order_relaxed)
                     && T == tail.load(std::memory_order_relaxed)
                     && H0 == T + 1, "victim/thief arbitration algorithm failure" );
-                reset_task_pool_and_leave();
+                (tasks_skipped) ? release_task_pool() : reset_task_pool_and_leave();
                 // No tasks in the task pool.
```
kboyarinov: It seems like the comment do not reflect the reality

```suggestion
```

src/tbb/arena_slot.cpp
```diff
@@ -81,67 +83,54 @@ d1::task* arena_slot::get_task(execution_data_ext& ed, isolation_type isolation)
...
+                all_tasks_checked = true;
             } else {
                 // Release task pool if there are still some tasks.
                 // After the release, the tail will be less than T, thus a thief
                 // will not attempt to get a task at position T.
                 release_task_pool();
             }
         }
-        result = get_task_impl( T, ed, tasks_omitted, isolation );
+        result = get_task_impl( T, ed, tasks_skipped, isolation );
```
kboyarinov: Does it make sense to run `get_task_impl` and re-check the isolation if there is only one task in the pool? Seems like we can take `task_pool_ptr[T]` directly here in this case.

src/tbb/arena_slot.cpp
```diff
@@ -81,67 +83,54 @@ d1::task* arena_slot::get_task(execution_data_ext& ed, isolation_type isolation)
...
-            tail.store(T0, std::memory_order_release);
-            // Synchronize with snapshot as we published some tasks.
-            // TODO: consider some approach not to call wakeup for each time. E.g. check if the tail reached the head.
-            ed.task_disp->m_thread_data->my_arena->advertise_new_work<arena::wakeup>();
-        }
+    } while ( !result && !all_tasks_checked );
+
+    if ( tasks_skipped ) {
+        __TBB_ASSERT( is_task_pool_published(), nullptr ); // the pool was not reset
+        tail.store(T0, std::memory_order_release);
```
kboyarinov, line 132: Do I understand correctly that we do not need to restore head here since it is a stealing thread responsibility and is done in the `steal_task`?

src/tbb/arena_slot.cpp
akukanov: We still need to do all the proxy related stuff, and duplicating it makes no sense. And I think calling `get_task_impl` for the last task while holding the pool lock is far from ideal either.

src/tbb/arena_slot.cpp
akukanov: That is correct.

Generally, H0 represents the state of the pool head as it was seen by the owner; it might get outdated at any time. The core principle therefore is that the owner only works with the tail and does not change the head.

Indeed, if there was no conflict for the last task, the owner has no idea what the proper value for the head should be. And in case of a conflict the pool lock is taken and the head is re-read, and we can be sure that there is no skipped tasks beyond the head, so there is no need to change anything.

Prior to the patch, there is a store of H0 to the head - but it is done at the point where the pool is temporarily quiescent, and therefore it is safe. It "optimizes" the case when the task at the head is taken while others were skipped. In the patch, the pool is not reset if tasks were skipped, as that would also mislead observers. So this optimization cannot be safely performed anymore.

src/tbb/arena_slot.cpp
akukanov: I refactored the code in 796db43. Check if you like it more the new way :)

akukanov:
The commit 796db43 is for code refactoring and is not strictly necessary. It significantly changes `get_task`, and though the core logic there remains the same, some code blocks and checks had to be reordered. It's likely better to review it separately, on top of the previous commits. I can as well revert it in case you prefer to keep refactoring separate from the substantial changes.

src/tbb/arena_slot.h
```diff
@@ -169,9 +174,28 @@ class arena_slot : private arena_slot_shared_state, private arena_slot_private_s
...
+            if ( the_task_pool == EmptyTaskPool ) {
+                return false;
+            }
+            tail_unstable = accessed_by_owner.load(std::memory_order_acquire);
+            hd = head.load(std::memory_order_relaxed);
+            tl = tail.load(std::memory_order_relaxed);
+            unlock_task_pool(the_task_pool);
+            if ( (std::intptr_t)hd > (std::intptr_t)tl || ( hd == tl && !tail_unstable ) ) {
+                return false;
+            }
```
isaevil: I am trying to understand why is it assumed that if `head` > `tail`, then the task pool is empty? As I understand, it is still possible that the owner is still yet to restore tail if tasks were skipped, meaning that `accessed_by_owner` can still be true?

akukanov: Actually, you are correct. Thanks for catching this!

isaevil (approved):
The PR looks good to me, though I think we probably should do some performance evaluation prior merging. Do we need to review #1805 first or can we merge this one and rebase #1805 on latest master?
I also have one more observation. So the inspecting thread checks whether task pools are empty during `out_of_work` call. And if during thorough inspection in case of skipped tasks it determines that the arena, in fact, still has some tasks (they will become available once owning thread restores the tail), the thread won't leave the arena since `waiter.continue_execution(slot, t)` will return `true` and the thread will be in the stealing loop again. But as I understand, tail might still not be restored, meaning that `waiter.pause(slot)` will again be invoked. But the backoff of `waiter` object is already in its reached threshold, meaning `out_of_work` will be instantly called. Shouldn't the backoff be reset once it returns back or that is not a big deal?

akukanov:
> ... tail might still not be restored, meaning that `waiter.pause(slot)` will again be invoked. But the backoff of `waiter` object is already in its reached threshold, meaning `out_of_work` will be instantly called. Shouldn't the backoff be reset once it returns back or that is not a big deal?

I tend to think that it's not a big deal, and will in practice happen rather rarely - as there is seemingly much more for that inspecting thread to do in `receive_or_steal_task` than for the owner thread that at most has to check the last task in its pool. Also, since the inspection gives no clue where specifically a task was found, the thread after inspection will still do random victim selection, likely missing the slot where the tasks are - so the problem you described is already there it seems.

> Do we need to review #1805 first or can we merge this one and rebase #1805 on latest master?

Unless this patch adds much overhead - which I do not think it does - merging it first seems better to me.

> ... I think we probably should do some performance evaluation prior merging.

Agreed.

src/tbb/arena_slot.h
```diff
@@ -169,9 +174,28 @@ class arena_slot : private arena_slot_shared_state, private arena_slot_private_s
...
-               head.load(std::memory_order_relaxed) >= tail.load(std::memory_order_relaxed);
+    bool has_tasks() {
+        d1::task** the_task_pool = task_pool.load(std::memory_order_relaxed);
+        if ( the_task_pool == EmptyTaskPool ) {
+            return false;
+        }
+        std::size_t hd = head.load(std::memory_order_relaxed), tl = tail.load(std::memory_order_relaxed); 
+        if ( (std::intptr_t)hd >= (std::intptr_t)tl ) {
+            // Since some tasks might be temporary out of the visible pool bounds, lock the pool to examine closely
+            bool tail_unstable = false;
```
aleksei-fedotov: Consider renaming to exclude the negation. This will improve readability of the constructions that negate the negations, one of which is used below.
```suggestion
            bool has_stable_tail = false;
```

src/tbb/arena_slot.h
```diff
@@ -169,9 +174,28 @@ class arena_slot : private arena_slot_shared_state, private arena_slot_private_s
...
+            the_task_pool = lock_task_pool();
+            if ( the_task_pool == EmptyTaskPool ) {
+                return false;
+            }
+            tail_unstable = accessed_by_owner.load(std::memory_order_acquire);
```
aleksei-fedotov: The `lock_task_pool()` uses sequential consistency memory semantics. Why is `memory_order_relaxed` is not enough here? Also, consider leaving a comment in the code about that.

aleksei-fedotov (commented):
Not a full review.

src/tbb/arena_slot.h
akukanov: It is the "receiver" part of the classic release-acquire synchronization pattern with the pool owner thread which does not usually lock the pool.

src/tbb/arena_slot.cpp
```diff
@@ -81,67 +81,61 @@ d1::task* arena_slot::get_task(execution_data_ext& ed, isolation_type isolation)
...
-            // A task has been obtained. We need to make a hole in position T.
-            __TBB_ASSERT( is_task_pool_published(), nullptr );
-            __TBB_ASSERT( result, nullptr );
-            task_pool_ptr[T] = nullptr;
-            tail.store(T0, std::memory_order_release);
-            // Synchronize with snapshot as we published some tasks.
-            // TODO: consider some approach not to call wakeup for each time. E.g. check if the tail reached the head.
-            ed.task_disp->m_thread_data->my_arena->advertise_new_work<arena::wakeup>();
-        }
+    } while ( /*!result &&*/ !all_tasks_checked );
```
aleksei-fedotov, line 128: Is this intentional or you just forgot to remove unnecessary code?
```suggestion
    } while ( !all_tasks_checked );
```
At a glance it looks a bit suspicious, especially because there is a similar assert at line 122. I would remove this commented out code to avoid confusion, but move the assert just before the loop's `while` statement to clearly tell about the expectations in the code.

akukanov: Asserts are better put at the beginning of a block for which the asserted invariant holds, so I think moving it would not be right.
I wanted the comment to sort of inform that this is still the loop exit condition, even though not directly checked. I agree that it might look confusing as well; but let's address that later.

PR #1780, opened 2025-07-15
Make task_group::wait more thread safe

vossmjp:
### Description 

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown

### DIFF
--- a/test/tbb/test_task_group.cpp
+++ b/test/tbb/test_task_group.cpp
@@ -1376,7 +1376,9 @@ TEST_CASE("Arena exception handling race condition test") {
 }
 
 //! \brief \ref error_guessing
-TEST_CASE("Task dispatcher exception handling race condition test") {   
+TEST_CASE("Task dispatcher exception handling race condition test") {
+    if (tbb::this_task_arena::max_concurrency() < 2) return;
+    
     const int num_iterations = 50;
     const int tasks_to_throw = 20;
     const int num_concurrent_waits = 8;


test/tbb/test_task_group.cpp
```diff
@@ -1221,4 +1222,78 @@ TEST_CASE("task_handle cannot be scheduled into other task_group of the same con
...
+                    tbb::task_group_status status = tg.wait();
+                    if (status != tbb::not_complete)
+                        ++normal_completions;
+                } catch (const test_exception& e) {
+                    CHECK_MESSAGE(std::string(e.what()) == "Concurrent wait exception test", 
+                                "Unexpected exception message");
+                    ++exceptions_caught;
+                } catch (...) {
+                    CHECK_MESSAGE(false, "Unexpected exception type caught");
+                }
```
aleksei-fedotov, line 1276: Make up an additional lambda of a duplicated part?

test/tbb/test_task_group.cpp
```diff
@@ -1221,4 +1222,78 @@ TEST_CASE("task_handle cannot be scheduled into other task_group of the same con
...
+        } catch (...) {
+            CHECK_MESSAGE(false, "Unexpected exception type caught");
+        }
+        
+        // Wait for all threads to complete
+        for (auto& waiter : waiters) {
+            waiter.join();
+        }
+        
+        MESSAGE("thrown == ", task_thrown.load(), " caught == ", exceptions_caught.load(), " completed == ", normal_completions.load() );
```
aleksei-fedotov: Can we split to avoid too wide lines?
```suggestion
        MESSAGE("thrown == ", task_thrown.load(), " caught == ", exceptions_caught.load(),
                " completed == ", normal_completions.load() );
```

aleksei-fedotov (commented):
I found the following lines that seem requiring some updates as well:
* [task_dispatcher.cpp:176-180](https://github.com/uxlfoundation/oneTBB/blob/792d31223300bc849b20722aa1d79d421b6631b8/src/tbb/task_dispatcher.cpp#L176-L180)
* [arena.cpp:847-851](https://github.com/uxlfoundation/oneTBB/blob/792d31223300bc849b20722aa1d79d421b6631b8/src/tbb/arena.cpp#L847-L851)

src/tbb/task_group_context.cpp
```diff
@@ -240,19 +243,15 @@ bool task_group_context_impl::is_group_execution_cancelled(const d1::task_group_
     return ctx.my_cancellation_requested.load(std::memory_order_relaxed) != 0;
 }
 
-// IMPORTANT: It is assumed that this method is not used concurrently!
+// Thread-safe reset implementation using compare_exchange
```
akukanov: The new comment is misleading, as there is no CAS on this path anymore.

And I think the old comment has some value. Even if the method will allow concurrent invocations, it still should not be used while there are tasks in the context, as it can erase the cancellation signal.

src/tbb/task_group_context.cpp
```diff
@@ -65,11 +72,7 @@ void task_group_context_impl::destroy(d1::task_group_context& ctx) {
 #endif
     ctl->~cpu_ctl_env();
 
-    auto exception = ctx.my_exception.load(std::memory_order_relaxed);
-    if (exception) {
-        exception->destroy();
-    }
-    ITT_STACK_DESTROY(ctx.my_itt_caller);
```
akukanov, line 72: The ITT notification should be preserved.

src/tbb/task_group_context.cpp
```diff
@@ -47,6 +48,12 @@ void tbb_exception_ptr::throw_self() {
     std::rethrow_exception(my_ptr);
 }
 
+void tbb_exception_ptr::throw_self_and_destroy() {
+    auto temp_ptr = my_ptr;
+    destroy();
+    std::rethrow_exception(temp_ptr);
+}
```
akukanov: I would rename the method to `rethrow_and_destroy`.

Also take a look at `throw_self()` right above; there is a `fix_broken_rethrow` workaround which I believe should be used in the new method as well.

Finally, if `throw_self` is not used anymore, let's remove it.

src/tbb/scheduler_common.h
```diff
@@ -339,10 +339,30 @@ class tbb_exception_ptr {
...
+    tbb_exception_ptr* exception = ctx.my_exception.load(std::memory_order_acquire);
+    while (exception != nullptr) {
+        // Atomically clear the exception if it's still the same one
+        if (ctx.my_exception.compare_exchange_weak(exception, nullptr, 
+                                                   std::memory_order_acq_rel,
+                                                   std::memory_order_acquire)) {
+            exception->throw_self_and_destroy();
+            return true;
+        }
+    }
```
akukanov: The code contradicts the comment, Since `exception` is updated by `compare_exchange_weak` if CAS fails, the loop actually continues until CAS succeeds - even if the exception is not "the same one" that was there initially.

Is it possible for a new exception to be recorded to the context while this function is in progress? If not, and so the function can only be concurrent to itself, then the loop is unnecessary as `my_exception` can only become `nullptr`. If it is possible, and a new exception has been added concurrently, should it be "handled" by the call?

src/tbb/scheduler_common.h
```diff
@@ -336,13 +336,27 @@ class tbb_exception_ptr {
...
 
+//! Utility function to atomically clear and handle exceptions from task_group_context
+/** This function implements thread-safe pattern for clearing exceptions
+    from a task_group_context and either destroying or throwing them. **/
+inline void handle_context_exception(d1::task_group_context& ctx) {
+    tbb_exception_ptr* exception = ctx.my_exception.load(std::memory_order_acquire);
+    if (exception) {
+        if (ctx.my_exception.compare_exchange_strong(exception, nullptr, 
+                                                     std::memory_order_acq_rel,
+                                                     std::memory_order_acquire)) {
```
kboyarinov: As far as I understand, the second acquire fence is redundant. See [cppreference](https://en.cppreference.com/w/cpp/atomic/atomic/compare_exchange.html#:~:text=order-,std%3A%3Amemory_order_acquire,order,-If).
```suggestion
        if (ctx.my_exception.compare_exchange_strong(exception, nullptr, 
                                                     std::memory_order_acq_rel)) {
```

src/tbb/scheduler_common.h
```diff
@@ -336,13 +336,27 @@ class tbb_exception_ptr {
...
 private:
     tbb_exception_ptr(const std::exception_ptr& src) : my_ptr(src) {}
 }; // class tbb_exception_ptr
 
+//! Utility function to atomically clear and handle exceptions from task_group_context
+/** This function implements thread-safe pattern for clearing exceptions
+    from a task_group_context and either destroying or throwing them. **/
+inline void handle_context_exception(d1::task_group_context& ctx) {
+    tbb_exception_ptr* exception = ctx.my_exception.load(std::memory_order_acquire);
+    if (exception) {
```
kboyarinov, line 352: Should we keep the assert that `ctx` is cancelled? It is present in arena.cpp and task_dispatcher.cpp

src/tbb/scheduler_common.h
```diff
@@ -336,13 +336,27 @@ class tbb_exception_ptr {
...
+    //! Throws the contained exception and then destroys this object.
+    void rethrow_and_destroy();
 
 private:
     tbb_exception_ptr(const std::exception_ptr& src) : my_ptr(src) {}
 }; // class tbb_exception_ptr
 
+//! Utility function to atomically clear and handle exceptions from task_group_context
+/** This function implements thread-safe pattern for clearing exceptions
+    from a task_group_context and either destroying or throwing them. **/
```
akukanov, line 348:
```suggestion
/** This function implements thread-safe pattern for clearing exceptions
    from a task_group_context and rethrowing them. **/
```

src/tbb/task_group_context.cpp
```diff
@@ -1,5 +1,6 @@
 /*
     Copyright (c) 2005-2023 Intel Corporation
```
kboyarinov:
```suggestion
    Copyright (c) 2005-2025 Intel Corporation
```

test/tbb/test_task_group.cpp
```diff
@@ -1228,6 +1228,322 @@ TEST_CASE("task_handle cannot be scheduled into other task_group of the same con
...
+        // At least one thread should have caught the exception
+        CHECK_MESSAGE(exceptions_caught.load() > 0, 
+                     "At least one waiter should have caught the exception");
+        CHECK_MESSAGE(normal_completions.load() + exceptions_caught.load() == num_waiters + 1,
+                     "The total caught or completed is num_waiters + 1");
+    }
+}
+
+
+TEST_CASE("Arena exception handling race condition test") {
```
kboyarinov:
```suggestion
//! \brief \ref error_guessing
TEST_CASE("Arena exception handling race condition test") { 
```

test/tbb/test_task_group.cpp
```diff
@@ -1228,6 +1228,322 @@ TEST_CASE("task_handle cannot be scheduled into other task_group of the same con
...
+		if (throw_to_arena) {
+			CHECK_MESSAGE(expected_arena_caught.load(), "Should have caught exceptions in arena");
+			CHECK_MESSAGE(expected_arena_caught.load() <= expected_task_group_caught.load(), "Should not catch more exceptions than are thrown");
+		} else {
+			CHECK_MESSAGE(expected_arena_caught.load() == 0, "Should not catch exceptions in arena when not throwing to arena");
+		}
+    }
+}
+
+TEST_CASE("Task dispatcher exception handling race condition test") {
```
kboyarinov:
```suggestion

//! \brief \ref error_guessing
TEST_CASE("Task dispatcher exception handling race condition test") {   
```

test/tbb/test_task_group.cpp
```diff
@@ -1228,6 +1228,322 @@ TEST_CASE("task_handle cannot be scheduled into other task_group of the same con
...
+        
+        // Verify all waits completed successfully without crashes
+        CHECK_MESSAGE(exceptions_thrown.load(), "Should have thrown at least one exception");
+        CHECK_MESSAGE(expected_exceptions_caught.load(), "Should have caught at least one exception");
+        CHECK_MESSAGE(expected_exceptions_caught.load() <= exceptions_thrown.load(), "Should not catch more exceptions than are thrown");
+        CHECK_MESSAGE(unexpected_exceptions_caught.load() == 0, "Should not catch unknown exceptiom types");
+    }
+}
+
+TEST_CASE("Mixed arena and task_group race test") {
```
kboyarinov, line 1458:
```suggestion
//! \brief \ref error_guessing
TEST_CASE("Mixed arena and task_group race test") {
```

src/tbb/task_group_context.cpp
```diff
@@ -240,19 +239,15 @@ bool task_group_context_impl::is_group_execution_cancelled(const d1::task_group_
...
+    __TBB_ASSERT(!ctx.my_exception.load(std::memory_order_relaxed), "reset reached with non-null exception");
 
-    auto exception = ctx.my_exception.load(std::memory_order_relaxed);
-    if (exception) {
-        exception->destroy();
-        ctx.my_exception.store(nullptr, std::memory_order_relaxed);
-    }
-    ctx.my_cancellation_requested = 0;
+    // Thread-safe cancellation reset using atomic store
+    ctx.my_cancellation_requested.store(0, std::memory_order_relaxed);
```
akukanov: Do we have any formal definition of "thread-safe"? In the C++ standard, the term is occasionally used but not really defined. Is it the same as "formal absence of data races"?

src/tbb/task_dispatcher.h
```diff
@@ -380,10 +380,8 @@ d1::task* task_dispatcher::local_wait_for_all(d1::task* t, Waiter& waiter ) {
             if (global_control::active_value(global_control::terminate_on_exception) == 1) {
                 do_throw_noexcept([] { throw; });
             }
-            if (ed.context->cancel_group_execution()) {
-                /* We are the first to signal cancellation, so store the exception that caused it. */
-                ed.context->my_exception.store(tbb_exception_ptr::allocate(), std::memory_order_release);
-            }
+            ed.context->cancel_group_execution();
+            ed.context->my_exception.store(tbb_exception_ptr::allocate(), std::memory_order_release);
```
aleksei-fedotov: Cancellation may be called multiple times here, as far as I understand. Why don't we check  the returning value of `ed.context->cancel_group_execution()` anymore and store the captured exception in `my_exception` only once?

src/tbb/scheduler_common.h
```diff
@@ -336,13 +336,26 @@ class tbb_exception_ptr {
     /** Note that objects of this type can be created only by the allocate() method. **/
     void destroy() noexcept;
 
-    //! Throws the contained exception .
-    void throw_self();
+    //! Throws the contained exception and then destroys this object.
+    void rethrow_and_destroy();
```
aleksei-fedotov, line 340: To match the implementation:
```suggestion
    //! Destroys this object and re-throws the exception it had been containing
    void destroy_and_rethrow();
```

src/tbb/task_dispatcher.h
akukanov: It's an interesting question if the check should or should not "guard" storing the caught exception. The risk is that the cancellation flag might have been set without any exception captured, So the check, if any, should probably be for `my_exception` being non-null. But also, since we store only one exception (of all caught in the given group of tasks) anyway and cannot know which exactly one, does the check matter much?

kboyarinov: As far as I understand, if we have several threads in `local_wait_for_all` within the same `task_group_context` and all of them catch different exceptions and then the line 384 happens in the catch block, both of them allocate `tbb_exception_ptr`, but only one last stored instance will be deallocated by `handle_context_exception`, which results in a memory leak.

src/tbb/scheduler_common.h
```diff
@@ -336,13 +336,26 @@ class tbb_exception_ptr {
...
     tbb_exception_ptr(const std::exception_ptr& src) : my_ptr(src) {}
 }; // class tbb_exception_ptr
 
+//! Utility function to atomically clear and handle exceptions from task_group_context
+/** This function implements thread-safe pattern for clearing exceptions
+    from a task_group_context and rethrowing them. **/
+inline void handle_context_exception(d1::task_group_context& ctx) {
+    tbb_exception_ptr* exception = ctx.my_exception.load(std::memory_order_acquire);
+    if (exception) {
+        if (ctx.my_exception.compare_exchange_strong(exception, nullptr,
```
kboyarinov, line 353: As far as I understand, reset the exception to `nullptr` caused issues in `test_eh_algorithms` for `parallel_for_each` when storing the exception in `local_wait_for_all` was made conditionally only if the thread successfully requested the cancellation:

```cpp
// From task_dispatcher.h, catch block in local_wait_for_all
if (ed.context->cancel_group_execution()) {
    /* We are the first to signal cancellation, so store the exception that caused it. */
    ed.context->my_exception.store(tbb_exception_ptr::allocate(), std::memory_order_release);
}
```

Consider a `parallel_for_each` algorithm executed with a pair of forward iterators. In this case, a set of `forward_block_handling_task`s is spawned using the context of `parallel_for_each`.
Consider a single block for simplicity:

```cpp
void parallel_for_each(forward_it, forward_it) {
    task_group_context ctx;
    block_handling_task task(ctx);
    execute_and_wait(task, ctx); // outer execute_and_wait
}
```
Block handling task creates a `for_each_iteration_task` for each single element of the block and waits for block completion:

```cpp
task* block_handling_task::execute() {
    for (each task in block except one) {
       spawn(task, ctx); // ctx is a parallel_for_each task_group_context
    }
    execute_and_wait(last_task, ctx); // inner execute_and_wait
}
```

So, `parallel_for_each` models the nested `execute_and_wait` calls.
Imagine one of the tasks in the block throws an exception. 
The exception is caught in `local_wait_for_all` (inner `execute_and_wait`). Since the thread is the first to request the cancellation on `ctx`, the `tbb_exception_ptr` is created and stored into `ctx`. 

While exiting the inner `execute_and_wait`, `handle_context_exception` is called. The `ctx.my_exception` is stored to `nullptr` and the exception is rethrown.

Since we are simultaneously in another `local_wait_for_all` in the outer `execute_and_wait`, the same catch block above will be called. Since the cancellation was already requested on the inner level, the exception is not stored. Hence while exiting the outer `execute_and_wait`, `ctx.my_exception` is `nullptr` and hence no exception is thrown to the user level. 

In the current master, it works because we never reset `ctx.my_exception` to `nullptr`, and any exit from `execute_and_wait` will throw the exception to the next level.

Removing the condition in `local_wait_for_all` catch block fixes this issue, but introduces a memory leak I mentioned earlier in another discussion thread.

src/tbb/task_dispatcher.h
vossmjp: Updated to make check around null my_exception and only if that is successfully updated then cancel is attempted. If it fails, the new exception is destroyed. Both test_task_group concurrent wait tests and test_eh_algorithms now pass.

src/tbb/scheduler_common.h
vossmjp: I removed asserts and now destroy if not null. I believe its possible that a new exception appears in racy situations such as concurrent reset().

src/tbb/task_group_context.cpp
vossmjp: Done

src/tbb/scheduler_common.h
```diff
@@ -336,13 +336,30 @@ class tbb_exception_ptr {
...
+inline void handle_context_exception(d1::task_group_context& ctx, bool rethrow = true) {
+
+    tbb_exception_ptr* exception = ctx.my_exception.load(std::memory_order_acquire);
+    if (exception) {
+        if (ctx.my_exception.compare_exchange_strong(exception, nullptr, 
+                                                     std::memory_order_acq_rel)) {
+            if (rethrow)
+                exception->rethrow_and_destroy();
+            else
+                exception->destroy();
```
akukanov: I do not like an exception being captured and not later rethrown, and I think it should be fixed at some point. Add a TODO?

src/tbb/scheduler_common.h
akukanov: Here Konstantin asks about a different assertion - a one that the context is surely in the cancelled state when we are going to rethrow an exception.

src/tbb/task_group_context.cpp
akukanov: Perhaps I would simply remove the comment to avoid misinterpretations of "thread-safe"

src/tbb/scheduler_common.h
vossmjp: Done

test/tbb/test_task_group.cpp
```diff
@@ -1796,3 +2123,7 @@ TEST_CASE("Test stateful task body") {
     }
     CHECK_MESSAGE(placeholder == 1, "Not submitted task was executed");
 }
+
+#if _MSC_VER
+#pragma warning (pop)
+#endif
```
kboyarinov:
```suggestion
#endif

```

src/tbb/task_group_context.cpp
```diff
@@ -65,10 +68,7 @@ void task_group_context_impl::destroy(d1::task_group_context& ctx) {
 #endif
     ctl->~cpu_ctl_env();
 
-    auto exception = ctx.my_exception.load(std::memory_order_relaxed);
-    if (exception) {
-        exception->destroy();
-    }
+    handle_context_exception(ctx, /* do not rethrow */ false);
```
akukanov:
```suggestion
    handle_context_exception(ctx, /* rethrow = */ false);
```

src/tbb/task_group_context.cpp
```diff
@@ -240,19 +240,14 @@ bool task_group_context_impl::is_group_execution_cancelled(const d1::task_group_
...
 
-// IMPORTANT: It is assumed that this method is not used concurrently!
+// IMPORTANT: If used while tasks are in the context, the cancellation signal can be lost
 void task_group_context_impl::reset(d1::task_group_context& ctx) {
     __TBB_ASSERT(!is_poisoned(ctx.my_context_list), nullptr);
     //! TODO: Add assertion that this context does not have children
-    // No fences are necessary since this context can be accessed from another thread
-    // only after stealing happened (which means necessary fences were used).
+    
+    handle_context_exception(ctx, /* do not rethrow */ false);
```
akukanov:
```suggestion
    handle_context_exception(ctx, /* rethrow = */ false);
```

test/tbb/test_task_group.cpp
```diff
@@ -1228,6 +1237,324 @@ TEST_CASE("task_handle cannot be scheduled into other task_group of the same con
     CHECK_THROWS_WITH_AS(tg1.run(tg.defer([]{})), "Attempt to schedule task_handle into different task_group", std::runtime_error);
 }
 
+//! Test for thread-safe exception handling in concurrent wait scenarios
+//! \brief \ref error_guessing \ref interface
+TEST_CASE("Concurrent exception handling in task_group wait") {
+    if (tbb::this_task_arena::max_concurrency() < 3) {
+        // Test requires at least 3 threads: main + 2 waiters
```
akukanov:
```suggestion
        // Test requires at least 3 threads: main + 2 workers
```
I assume that was the intended comment, because waiters in this test are external threads calling `wait()`.

Also it is not clear why this test needs two workers.

test/tbb/test_task_group.cpp
```diff
@@ -1228,6 +1237,319 @@ TEST_CASE("task_handle cannot be scheduled into other task_group of the same con
...
+        for (auto& waiter : waiters) {
+            waiter.join();
+        }
+        
+        // Verify that the exception was properly handled
+        CHECK_MESSAGE(task_thrown.load(), "Task should have thrown an exception");
+        
+        // At least one thread should have caught the exception
+        CHECK_MESSAGE(exceptions_caught.load() > 0, 
+                     "At least one waiter should have caught the exception");
```
aleksei-fedotov, line 1302: The implementation of the library (at least in the patch) and test suggests that the exception will be thrown and caught only once, does not it? Can we expect `exceptions_caught == 1` then? And also expect the invariant `normal_completions == num_waiters`?

src/tbb/scheduler_common.h
```diff
@@ -336,13 +336,32 @@ class tbb_exception_ptr {
     /** Note that objects of this type can be created only by the allocate() method. **/
     void destroy() noexcept;
 
-    //! Throws the contained exception .
-    void throw_self();
+    //! Throws the contained exception and then destroys this object.
+    void rethrow_and_destroy();
```
kboyarinov, line 340: It seems like the old comment about the function renaming was not applied.

PR #1784, opened 2025-07-17
Improve task_arena interoperability with task_groups, fix lifetime issue in reference_vertex

isaevil:
### Description 

### Type of change

### Tests

- [ ] not needed

### Documentation

- [x] [Spec PR](https://github.com/uxlfoundation/oneAPI-spec/pull/624)
- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown


akukanov:
It would be better not to mix the implementation of a new feature and a bug fix in the same PR.
And for the bug fix, it would be good to provide some description/analysis of the problem.

isaevil:
> It would be better not to mix the implementation of a new feature and a bug fix in the same PR. And for the bug fix, it would be good to provide some description/analysis of the problem.

Yeah, I have some doubts about my decision to do it in a single branch. On the one hand, this fix is kind of a prerequisite for this feature to work properly and guarantee correct `enqueue` with `task_group` into arena from external thread and the fix is small (at least for now), but on the other hand it is pretty much independent. I think I will move this fix into a separate PR and highlight that it must be merged before the feature implementation.

include/oneapi/tbb/task_arena.h
```diff
@@ -295,6 +298,13 @@ class task_arena : public task_arena_base {
         r1::execute(*this, func);
         return func.consume_result();
     }
+
+    d2::task_group_status wait_for_impl(d2::task_group& tg) {
+        d2::task_group_status status;
```
akukanov:
```suggestion
        d2::task_group_status status = d2::task_group_status::not_complete;
```
In the implementation of `isolated_task_group` there is also an assertion that the resulting status is not `not_complete`. Not sure if it's useful, given the implementation of `task_group::wait`; on the other hand, it would certainly describe the expected invariant.

akukanov (commented):
The implementation looks good to me.

include/oneapi/tbb/task_arena.h
isaevil: It'll definitely won't hurt :)
Added the assertion.

kboyarinov (approved):
LGTM

PR #1785, opened 2025-07-18
Fix lifetime issue with reference_vertex

isaevil:
### Description 
To improve the scalability of `wait_context` usage across several oneTBB parallel algorithms and structures, the `wait_context_vertex` class was introduced. In particular, a special `reference_vertex` class was added to track the number of references to a `wait_context` from a specific thread.

For example, `task_group` has a `wait_context_vertex` that is bound to a single `wait_context`. When a task is submitted to the task_group, `get_thread_reference_vertex` is invoked to retrieve a `reference_vertex` bound to the calling thread and associated with the parent `wait_context_vertex`. For each `reference_vertex`, the first call to `reserve` acquires the actual wait_context, and the last call to `release` releases it accordingly. When the corresponding arena associated with the calling thread is destroyed, the `reference_vertex` objects are also deallocated.

While this approach works well for blocking APIs like parallel_for_each, it poses issues for asynchronous APIs like task_group. A task may be deferred on one thread but submitted for execution later in a different arena. For example:
```cpp
    tbb::task_group tg;
    tbb::task_handle task;
    std::thread([&] { task = tg.defer(body); }).join();
    tg.run(std::move(task));
    tg.wait();
```
In this scenario, the lifetime of the task's `reference_vertex` is tied to the thread that called `task_group::defer`. Consequently, when that thread terminates, the vertex may be deallocated before the task is finalized. During finalization, the destructor of `task_handle_task` attempts to access the now-invalid vertex to release its reference, resulting in a failure.

This patch addresses the issue by introducing an additional level of reference counting specifically for the lifetime management of the vertex. By default, a reference is held by the thread that allocates the vertex, and a second reference is added upon the first call to `reserve()`. During the final `release()` call, the implementation checks whether the owning thread has already abandoned the vertex. If so, the responsibility of deallocating the vertex falls to the current thread.

Is a prerequisite for #1784

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No - `reference_vertex` class was created only in library and the entry point returned a pointer to the interface class `wait_vertex_interface`, and the code in headers interact with it via virtual methods. So, it should be safe to return a pointer to new `thread_reference_vertex` instead because the correct methods will be resolved through polymorphism.
- [ ] Unknown

### DIFF
--- a/include/oneapi/tbb/detail/_flow_graph_impl.h
+++ b/include/oneapi/tbb/detail/_flow_graph_impl.h
@@ -159,9 +159,6 @@ public:
         auto last_iterator = my_msg_reference_vertices.cbefore_begin();
 
         for (auto& msg_waiter : my_msg_wait_context_vertices) {
-            // If the task is created by the thread outside the graph arena, the lifetime of the thread reference vertex
-            // may be shorter that the lifetime of the task, so thread reference vertex approach cannot be used
-            // and the task should be associated with the msg wait context itself
             d1::wait_tree_vertex_interface* ref_vertex = r1::get_thread_reference_vertex(msg_waiter);
             last_iterator = my_msg_reference_vertices.emplace_after(last_iterator,
                                                                     ref_vertex);
@@ -449,10 +446,6 @@ inline graph_task::graph_task(graph& g, d1::small_object_allocator& allocator,
     , priority(node_priority)
     , my_allocator(allocator)
 {
-    // If the task is created by the thread outside the graph arena, the lifetime of the thread reference vertex
-    // may be shorter that the lifetime of the task, so thread reference vertex approach cannot be used
-    // and the task should be associated with the graph wait context itself
-    // TODO: consider how reference counting can be improved for such a use case. Most common example is the async_node
     d1::wait_context_vertex* graph_wait_context_vertex = &my_graph.get_wait_context_vertex();
     my_reference_vertex = r1::get_thread_reference_vertex(graph_wait_context_vertex);
     __TBB_ASSERT(my_reference_vertex, nullptr);
--- a/include/oneapi/tbb/detail/_flow_graph_impl.h
+++ b/include/oneapi/tbb/detail/_flow_graph_impl.h
@@ -1,5 +1,6 @@
 /*
     Copyright (c) 2005-2025 Intel Corporation
+    Copyright (c) 2025 UXL Foundation Contributors
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
@@ -146,8 +147,6 @@ private:
     friend graph_task* prioritize_task(graph& g, graph_task& gt);
 };
 
-inline bool is_this_thread_in_graph_arena(graph& g);
-
 #if __TBB_PREVIEW_FLOW_GRAPH_TRY_PUT_AND_WAIT
 class trackable_messages_graph_task : public graph_task {
 public:
@@ -163,9 +162,7 @@ public:
             // If the task is created by the thread outside the graph arena, the lifetime of the thread reference vertex
             // may be shorter that the lifetime of the task, so thread reference vertex approach cannot be used
             // and the task should be associated with the msg wait context itself
-            d1::wait_tree_vertex_interface* ref_vertex = is_this_thread_in_graph_arena(g) ?
-                                                         r1::get_thread_reference_vertex(msg_waiter) :
-                                                         msg_waiter;
+            d1::wait_tree_vertex_interface* ref_vertex = r1::get_thread_reference_vertex(msg_waiter);
             last_iterator = my_msg_reference_vertices.emplace_after(last_iterator,
                                                                     ref_vertex);
             ref_vertex->reserve(1);
@@ -420,7 +417,6 @@ private:
     friend void activate_graph(graph& g);
     friend void deactivate_graph(graph& g);
     friend bool is_graph_active(graph& g);
-    friend bool is_this_thread_in_graph_arena(graph& g);
     friend graph_task* prioritize_task(graph& g, graph_task& arena_task);
     friend void spawn_in_graph_arena(graph& g, graph_task& arena_task);
     friend void enqueue_in_graph_arena(graph &g, graph_task& arena_task);
@@ -458,8 +454,7 @@ inline graph_task::graph_task(graph& g, d1::small_object_allocator& allocator,
     // and the task should be associated with the graph wait context itself
     // TODO: consider how reference counting can be improved for such a use case. Most common example is the async_node
     d1::wait_context_vertex* graph_wait_context_vertex = &my_graph.get_wait_context_vertex();
-    my_reference_vertex = is_this_thread_in_graph_arena(g) ? r1::get_thread_reference_vertex(graph_wait_context_vertex)
-                                                           : graph_wait_context_vertex;
+    my_reference_vertex = r1::get_thread_reference_vertex(graph_wait_context_vertex);
     __TBB_ASSERT(my_reference_vertex, nullptr);
     my_reference_vertex->reserve();
 }
@@ -511,11 +506,6 @@ inline void deactivate_graph(graph& g) {
 
 inline bool is_graph_active(graph& g) {
     return g.my_is_active;
-}
-
-inline bool is_this_thread_in_graph_arena(graph& g) {
-    __TBB_ASSERT(g.my_task_arena && g.my_task_arena->is_active(), nullptr);
-    return r1::execution_slot(*g.my_task_arena) != d1::slot_id(-1);
 }
 
 inline graph_task* prioritize_task(graph& g, graph_task& gt) {
--- a/include/oneapi/tbb/detail/_task.h
+++ b/include/oneapi/tbb/detail/_task.h
@@ -197,18 +197,18 @@ public:
     {}
 
     void reserve(std::uint32_t delta = 1) override {
-        reserve_impl(delta);
+        reserve_post_increment(delta);
     }
 
     void release(std::uint32_t delta = 1) override {
-        release_impl(delta);
+        release_pre_decrement(delta);
     }
 
-    std::uint32_t get_num_child() {
+    std::uint32_t get_num_children() {
         return static_cast<std::uint32_t>(m_ref_count.load(std::memory_order_acquire));
     }
 protected:
-    std::uint64_t reserve_impl(std::uint32_t delta) {
+    std::uint64_t reserve_post_increment(std::uint32_t delta) {
         auto ref = m_ref_count.fetch_add(static_cast<std::uint64_t>(delta));
         if (ref == 0) {
             my_parent->reserve();
@@ -216,7 +216,7 @@ protected:
         return ref;
     }
 
-    std::uint64_t release_impl(std::uint32_t delta) {
+    std::uint64_t release_pre_decrement(std::uint32_t delta) {
         auto parent = my_parent;
         std::uint64_t ref = m_ref_count.fetch_sub(static_cast<std::uint64_t>(delta)) - static_cast<std::uint64_t>(delta);
         if (ref == 0) {
--- a/include/oneapi/tbb/detail/_task.h
+++ b/include/oneapi/tbb/detail/_task.h
@@ -1,5 +1,5 @@
 /*
-    Copyright (c) 2020-2024 Intel Corporation
+    Copyright (c) 2020-2025 Intel Corporation
     Copyright (c) 2025 UXL Foundation Contributors
 
     Licensed under the Apache License, Version 2.0 (the "License");
@@ -198,38 +198,36 @@ public:
     {}
 
     void reserve(std::uint32_t delta = 1) override {
-        if (m_ref_count.fetch_add(static_cast<std::uint64_t>(delta)) == 0) {
-            my_parent->reserve();
-            m_lifetime_ref_count.fetch_add(1);
-        }
+        reserve_impl(delta);
     }
 
     void release(std::uint32_t delta = 1) override {
-        auto parent = my_parent;
-        std::uint64_t ref = m_ref_count.fetch_sub(static_cast<std::uint64_t>(delta)) - static_cast<std::uint64_t>(delta);
-        if (ref == 0) {
-            parent->release();
-            destroy();
-        }
+        release_impl(delta);
     }
 
     std::uint32_t get_num_child() {
         return static_cast<std::uint32_t>(m_ref_count.load(std::memory_order_acquire));
     }
-
-    void destroy() {
-        auto ref = m_lifetime_ref_count.fetch_sub(1);
-        if (ref == 1) {
-            this->~reference_vertex();
-            r1::cache_aligned_deallocate(this);
+protected:
+    std::uint64_t reserve_impl(std::uint32_t delta) {
+        auto ref = m_ref_count.fetch_add(static_cast<std::uint64_t>(delta));
+        if (ref == 0) {
+            my_parent->reserve();
         }
+        return ref;
     }
 
+    std::uint64_t release_impl(std::uint32_t delta) {
+        auto parent = my_parent;
+        std::uint64_t ref = m_ref_count.fetch_sub(static_cast<std::uint64_t>(delta)) - static_cast<std::uint64_t>(delta);
+        if (ref == 0) {
+            parent->release();
+        }
+        return ref;
+    }
 private:
     wait_tree_vertex_interface* my_parent;
     std::atomic<std::uint64_t> m_ref_count;
-    // Reserve for TLS in advance
-    std::atomic<std::uint64_t> m_lifetime_ref_count{1};
 };
 
 struct execution_data {
--- a/include/oneapi/tbb/detail/_task.h
+++ b/include/oneapi/tbb/detail/_task.h
@@ -197,34 +197,24 @@ public:
     {}
 
     void reserve(std::uint32_t delta = 1) override {
-        reserve_post_increment(delta);
+        auto ref = m_ref_count.fetch_add(static_cast<std::uint64_t>(delta));
+        if (ref == 0) {
+            my_parent->reserve();
+        }
     }
 
     void release(std::uint32_t delta = 1) override {
-        release_pre_decrement(delta);
+        auto parent = my_parent;
+        std::uint64_t ref = m_ref_count.fetch_sub(static_cast<std::uint64_t>(delta)) - static_cast<std::uint64_t>(delta);
+        if (ref == 0) {
+            parent->release();
+        }
     }
 
     std::uint32_t get_num_children() {
         return static_cast<std::uint32_t>(m_ref_count.load(std::memory_order_acquire));
     }
 protected:
-    std::uint64_t reserve_post_increment(std::uint32_t delta) {
-        auto ref = m_ref_count.fetch_add(static_cast<std::uint64_t>(delta));
-        if (ref == 0) {
-            my_parent->reserve();
-        }
-        return ref;
-    }
-
-    std::uint64_t release_pre_decrement(std::uint32_t delta) {
-        auto parent = my_parent;
-        std::uint64_t ref = m_ref_count.fetch_sub(static_cast<std::uint64_t>(delta)) - static_cast<std::uint64_t>(delta);
-        if (ref == 0) {
-            parent->release();
-        }
-        return ref;
-    }
-private:
     wait_tree_vertex_interface* my_parent;
     std::atomic<std::uint64_t> m_ref_count;
 };
--- a/include/oneapi/tbb/detail/_task.h
+++ b/include/oneapi/tbb/detail/_task.h
@@ -1,5 +1,6 @@
 /*
     Copyright (c) 2020-2024 Intel Corporation
+    Copyright (c) 2025 UXL Foundation Contributors
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
@@ -21,6 +22,7 @@
 #include "_assert.h"
 #include "_template_helpers.h"
 #include "_small_object_pool.h"
+#include "../cache_aligned_allocator.h"
 
 #include "../profiling.h"
 
@@ -198,6 +200,7 @@ public:
     void reserve(std::uint32_t delta = 1) override {
         if (m_ref_count.fetch_add(static_cast<std::uint64_t>(delta)) == 0) {
             my_parent->reserve();
+            m_lifetime_ref_count.fetch_add(1);
         }
     }
 
@@ -206,15 +209,27 @@ public:
         std::uint64_t ref = m_ref_count.fetch_sub(static_cast<std::uint64_t>(delta)) - static_cast<std::uint64_t>(delta);
         if (ref == 0) {
             parent->release();
+            destroy();
         }
     }
 
     std::uint32_t get_num_child() {
         return static_cast<std::uint32_t>(m_ref_count.load(std::memory_order_acquire));
     }
+
+    void destroy() {
+        auto ref = m_lifetime_ref_count.fetch_sub(1);
+        if (ref == 1) {
+            this->~reference_vertex();
+            r1::cache_aligned_deallocate(this);
+        }
+    }
+
 private:
     wait_tree_vertex_interface* my_parent;
     std::atomic<std::uint64_t> m_ref_count;
+    // Reserve for TLS in advance
+    std::atomic<std::uint64_t> m_lifetime_ref_count{1};
 };
 
 struct execution_data {
--- a/include/oneapi/tbb/detail/_task.h
+++ b/include/oneapi/tbb/detail/_task.h
@@ -191,34 +191,6 @@ private:
     wait_context m_wait;
 };
 
-class reference_vertex : public wait_tree_vertex_interface {
-public:
-    reference_vertex(wait_tree_vertex_interface* parent, std::uint32_t ref_count) : my_parent{parent}, m_ref_count{ref_count}
-    {}
-
-    void reserve(std::uint32_t delta = 1) override {
-        auto ref = m_ref_count.fetch_add(static_cast<std::uint64_t>(delta));
-        if (ref == 0) {
-            my_parent->reserve();
-        }
-    }
-
-    void release(std::uint32_t delta = 1) override {
-        auto parent = my_parent;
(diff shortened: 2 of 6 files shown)


akukanov:
> Breaks backward compatibility?
 No - Although new member has been added to the reference_vertex class, it shouldn't be an issue for backward compatibility, since it is the last in order so older oneTBB versions shouldn't access it.

Generally, it is not true. Though maybe it's true in this case, as described at the end of this comment.

Imagine that the code built with old headers (and the old definition of `reference_vertex`) is run with the new TBB binary that uses the new definition. If an instance created in the headers is then passed "by reference" and modified in the binary, and the last member is changed, it will be an "out-of-bound" modification that corrupts some possibly valid value.

In that case, to add a new data member "safely" you'd need to either make sure that the size of the class does not increase (i.e., the space for that member was in the class, just not used), or that there is a "version" of the class that can be checked within the binary to prevent access to non-existent data members, or better both. That's not even accounting for potential impact on layouts of derived classes or classes containing a member of that type.

If `reference_vertex` objects are only created within the library and never in the headers, then it is safer to add new data members at the "bottom" of the class, as we can be sure that all instances have the same (new) layout. If that is the case, then the question is why the class is defined in public headers (given that its methods are virtual).

akukanov:
Seems it should be possible, if needed for backward compatibility, to avoid adding another counter by using instead the topmost bit(s) of the existing counter as an indication that the object is kept in some internal storage(s). These bits would need to be masked off for task reference counting, but checked when deciding if the object should be destroyed.

isaevil:
> > Breaks backward compatibility?
> > No - Although new member has been added to the reference_vertex class, it shouldn't be an issue for backward compatibility, since it is the last in order so older oneTBB versions shouldn't access it.
> 
> Generally, it is not true. Though maybe it's true in this case, as described at the end of this comment.
> 
> Imagine that the code built with old headers (and the old definition of `reference_vertex`) is run with the new TBB binary that uses the new definition. If an instance created in the headers is then passed "by reference" and modified in the binary, and the last member is changed, it will be an "out-of-bound" modification that corrupts some possibly valid value.
> 
> In that case, to add a new data member "safely" you'd need to either make sure that the size of the class does not increase (i.e., the space for that member was in the class, just not used), or that there is a "version" of the class that can be checked within the binary to prevent access to non-existent data members, or better both. That's not even accounting for potential impact on layouts of derived classes or classes containing a member of that type.
> 
> If `reference_vertex` objects are only created within the library and never in the headers, then it is safer to add new data members at the "bottom" of the class, as we can be sure that all instances have the same (new) layout. If that is the case, then the question is why the class is defined in public headers (given that its methods are virtual).

Yes, `reference_vertex` object is created only inside the library and can be retrieved using `get_thread_reference_vertex` entry point which returns the pointer to `wait_tree_vertex_interface`. It seems to me that `task.cpp` is more appropriate place to have `reference_vertex` defined unless we want to use `reference_vertex` in headers as well in the future.

kboyarinov:
> > > Breaks backward compatibility?
> > > No - Although new member has been added to the reference_vertex class, it shouldn't be an issue for backward compatibility, since it is the last in order so older oneTBB versions shouldn't access it.
> > 
> > 
> > Generally, it is not true. Though maybe it's true in this case, as described at the end of this comment.
> > Imagine that the code built with old headers (and the old definition of `reference_vertex`) is run with the new TBB binary that uses the new definition. If an instance created in the headers is then passed "by reference" and modified in the binary, and the last member is changed, it will be an "out-of-bound" modification that corrupts some possibly valid value.
> > In that case, to add a new data member "safely" you'd need to either make sure that the size of the class does not increase (i.e., the space for that member was in the class, just not used), or that there is a "version" of the class that can be checked within the binary to prevent access to non-existent data members, or better both. That's not even accounting for potential impact on layouts of derived classes or classes containing a member of that type.
> > If `reference_vertex` objects are only created within the library and never in the headers, then it is safer to add new data members at the "bottom" of the class, as we can be sure that all instances have the same (new) layout. If that is the case, then the question is why the class is defined in public headers (given that its methods are virtual).
> 
> Yes, `reference_vertex` object is created only inside the library and can be retrieved using `get_thread_reference_vertex` entry point which returns the pointer to `wait_tree_vertex_interface`. It seems to me that `task.cpp` is more appropriate place to have `reference_vertex` defined unless we want to use `reference_vertex` in headers as well in the future.

I have realized that `reference_vertex` is used in my task group dependencies implementation (for `continuation_vertex`) and this vertex is not created on the binary side and also the extra lifetime management is not needed for this object.

I propose to move the fix to the separate class inherited from `reference_vertex`, located in the `src` and used only in the map, located in the `task_dispatcher`. 

In this case, reusage of the topmost bits in the existing reference counter is not needed, since the object is defined and used only in the library.

include/oneapi/tbb/detail/_flow_graph_impl.h
```diff
@@ -163,9 +162,7 @@ class trackable_messages_graph_task : public graph_task {
             // If the task is created by the thread outside the graph arena, the lifetime of the thread reference vertex
             // may be shorter that the lifetime of the task, so thread reference vertex approach cannot be used
             // and the task should be associated with the msg wait context itself
-            d1::wait_tree_vertex_interface* ref_vertex = is_this_thread_in_graph_arena(g) ?
-                                                         r1::get_thread_reference_vertex(msg_waiter) :
-                                                         msg_waiter;
+            d1::wait_tree_vertex_interface* ref_vertex = r1::get_thread_reference_vertex(msg_waiter);
```
akukanov: Seemingly the change might make the comment above outdated; please check.

include/oneapi/tbb/detail/_flow_graph_impl.h
```diff
@@ -458,8 +454,7 @@ inline graph_task::graph_task(graph& g, d1::small_object_allocator& allocator,
     // and the task should be associated with the graph wait context itself
     // TODO: consider how reference counting can be improved for such a use case. Most common example is the async_node
     d1::wait_context_vertex* graph_wait_context_vertex = &my_graph.get_wait_context_vertex();
-    my_reference_vertex = is_this_thread_in_graph_arena(g) ? r1::get_thread_reference_vertex(graph_wait_context_vertex)
-                                                           : graph_wait_context_vertex;
+    my_reference_vertex = r1::get_thread_reference_vertex(graph_wait_context_vertex);
```
akukanov, line 450: Same here.

include/oneapi/tbb/detail/_task.h
```diff
@@ -196,21 +197,32 @@ class reference_vertex : public wait_tree_vertex_interface {
...
+        reserve_impl(delta);
+    }
+
+    void release(std::uint32_t delta = 1) override {
+        release_impl(delta);
+    }
+
+    std::uint32_t get_num_child() {
+        return static_cast<std::uint32_t>(m_ref_count.load(std::memory_order_acquire));
+    }
```
akukanov: Could you please rename it to `get_num_children`?

include/oneapi/tbb/detail/_task.h
```diff
@@ -196,21 +197,32 @@ class reference_vertex : public wait_tree_vertex_interface {
...
         auto parent = my_parent;
         std::uint64_t ref = m_ref_count.fetch_sub(static_cast<std::uint64_t>(delta)) - static_cast<std::uint64_t>(delta);
         if (ref == 0) {
             parent->release();
         }
-    }
-
-    std::uint32_t get_num_child() {
-        return static_cast<std::uint32_t>(m_ref_count.load(std::memory_order_acquire));
+        return ref;
```
akukanov: What would be more descriptive names that explain the semantics better - specifically, that indicate which values are returned? Maybe `fetch_and_reserve` and `release_and_fetch`?

isaevil: What about something similar to `fetch_add` and `fetch_sub` wording-wise? For example, `fetch_reserve` and `fetch_release`?

akukanov: Ordering is important. In one case, you fetch, do the operation, then return what you have fetched. In the other case though, you return an adjusted value as if the fetch was done after the operation.

Maybe some other names can better reflect these nuances.

test/tbb/test_task_group.cpp
```diff
@@ -1221,4 +1222,29 @@ TEST_CASE("task_handle cannot be scheduled into other task_group of the same con
     CHECK_THROWS_WITH_AS(tg1.run(tg.defer([]{})), "Attempt to schedule task_handle into different task_group", std::runtime_error);
 }
 
+//! \brief \ref requirement
+TEST_CASE("Test safe task submit from external thread") {
```
aleksei-fedotov, line 1226: Shall we vary the test a bit more? How about permuting deferring in one thread, submitting in another, and waiting in the third?

Something like: 
```cpp
for (auto i = 0; i < 3; ++i) {
    for (auto j = 0; j < 3; ++j) {
       for (auto k = 0; k < 3; ++k) {
           thread[i]( { ta.execute({ t = tg.defer();  }) );
           thread[j]( { ta.execute({ tg.run(move(t)); }) );
           thread[k]( { ta.execute({ tg.wait();       }) );
       }
    }
}
```
Need to think about thread synchronization though.

aleksei-fedotov, line 1226: How about renaming to?
```suggestion
TEST_CASE("Work with a task from different threads") {
```

src/tbb/scheduler_common.h
```diff
@@ -439,6 +440,33 @@ struct suspend_point_type {
...
+
+    void destroy() {
+        auto ref = m_lifetime_ref_counter.fetch_sub(1);
+        if (ref == 1) {
+            this->~thread_reference_vertex();
+            cache_aligned_deallocate(this);
+        }
+    }
+private:
+    std::atomic<uint32_t> m_lifetime_ref_counter{1};
```
kboyarinov: Could you please add a comment that we reserve 1 in this reference counter in the constructor for the task_dispatcher that own the vertex map.

src/tbb/scheduler_common.h
```diff
@@ -439,6 +440,33 @@ struct suspend_point_type {
...
+        }
+    }
+
+    void release(std::uint32_t delta = 1) override {
+        if (release_impl(delta) == 0) {
+            destroy();
+        }
+    }
+
+    void destroy() {
```
kboyarinov, line 485: I would suggest alternative naming for this method, since it does not actually destroy the vertex any time you call it. 
May be something like `release_ownership`?

include/oneapi/tbb/detail/_task.h
isaevil: It kind of resembles pre/post increment and decrement. Here we're doing the similar thing but combined with corresponding operations: we post-increment with reserve operation and pre-decrement with the release operation. So, while `fetch_and_reserve` and `release_and_fetch` are good enough as for me, as an alternative I can suggest `post_increment_and_reserve` and `pre_decrement_and_release`  but they are more verbose.

test/tbb/test_task_group.cpp
isaevil: My idea for this test is to have a sort of a minimal reproducer that can expose the original issue. You can take a look at the new test "Stress test enqueue with task_group from multiple threads" that actually does defer and submit in one group of threads, but waits from another group of threads.

isaevil: To be fair, I prefer the original name more :)

include/oneapi/tbb/detail/_task.h
akukanov: These names as well as variations like `reserve_post_increment`/`release_pre_increment` also sound good to me.

include/oneapi/tbb/detail/_flow_graph_impl.h
isaevil: Removed, thank you

include/oneapi/tbb/detail/_task.h
isaevil: Done

include/oneapi/tbb/detail/_task.h
isaevil: Renamed

src/tbb/scheduler_common.h
isaevil: Done

src/tbb/scheduler_common.h
isaevil: Done

akukanov:
> Seems it should be possible, if needed for backward compatibility, to avoid adding another counter by using instead the topmost bit(s) of the existing counter as an indication that the object is kept in some internal storage(s). These bits would need to be masked off for task reference counting, but checked when deciding if the object should be destroyed.

I now think that the approach with "reserving" the topmost bit of the reference counter is better, and not because of backward compatibility but because it avoids an extra atomic operation and maybe does not require a new class. The bit is not for a "lifetime counter" though, but for a flag indicating that the vertex has been "orphaned" by its owning thread. That flag would be set at `task_dispatcher` destruction, and read by the thread that decrements the reference count to 0 (except for that topmost bit). Whichever of the two gets `100...0` as the result, should destroy the vertex.

isaevil:
> > Seems it should be possible, if needed for backward compatibility, to avoid adding another counter by using instead the topmost bit(s) of the existing counter as an indication that the object is kept in some internal storage(s). These bits would need to be masked off for task reference counting, but checked when deciding if the object should be destroyed.
> 
> I now think that the approach with "reserving" the topmost bit of the reference counter is better, and not because of backward compatibility but because it avoids an extra atomic operation and maybe does not require a new class. The bit is not for a "lifetime counter" though, but for a flag indicating that the vertex has been "orphaned" by its owning thread. That flag would be set at `task_dispatcher` destruction, and read by the thread that decrements the reference count to 0 (except for that topmost bit). Whichever of the two gets `100...0` as the result, should destroy the vertex.

I think that it is still better to use a separate class inside the library (even if some slight code duplication is required for that) because lifetime information and logic is kind of unrelated to the vertex class that is/can be used in public headers (for example, @kboyarinov implementation of task group dependencies). What do you think?

src/tbb/scheduler_common.h
```diff
@@ -439,6 +440,34 @@ struct suspend_point_type {
...
+class thread_reference_vertex: public d1::reference_vertex {
+public:
+    using d1::reference_vertex::reference_vertex;
+
+    void release(std::uint32_t delta = 1) override {
+        auto parent = my_parent;
+        std::uint64_t ref = m_ref_count.fetch_sub(static_cast<std::uint64_t>(delta)) - static_cast<std::uint64_t>(delta);
+        if ((ref & ~m_orphaned_bit) == 0) {
+            parent->release();
+            if (ref & m_orphaned_bit) {
```
kboyarinov: Please add comments on lines 448, 450 and 452

src/tbb/scheduler_common.h
```diff
@@ -439,6 +440,34 @@ struct suspend_point_type {
...
+                cache_aligned_deallocate(this);
+            }
+        }
+    }
+
+    void release_ownership() {
+        auto ref = m_ref_count.fetch_or(m_orphaned_bit);
+        if (ref == 0) {
+            this->~thread_reference_vertex();
+            cache_aligned_deallocate(this);
```
kboyarinov: Lets move these two lines into separate function and reuse here and in `release`

src/tbb/task.cpp
```diff
@@ -236,9 +237,10 @@ d1::wait_tree_vertex_interface* get_thread_reference_vertex(d1::wait_tree_vertex
         if (reference_map.size() > max_reference_vertex_map_size) {
             // TODO: Research the possibility of using better approach for a clean-up
             for (auto it = reference_map.begin(); it != reference_map.end();) {
-                if (it->second->get_num_child() == 0) {
-                    it->second->~reference_vertex();
-                    cache_aligned_deallocate(it->second);
+                if (it->second->get_num_children() == 0) {
```
kboyarinov: Please add a comment that `get_num_children` can be used here since this function is executed by the owner thread and orphaned bit cannot be set.

akukanov: There is no assertion for "this function is executed by the owner thread", is there? Can it be added? Or at least that "the orphaned bit cannot be set".

include/oneapi/tbb/detail/_task.h
```diff
@@ -209,10 +211,10 @@ class reference_vertex : public wait_tree_vertex_interface {
         }
     }
 
-    std::uint32_t get_num_child() {
+    std::uint32_t get_num_children() {
         return static_cast<std::uint32_t>(m_ref_count.load(std::memory_order_acquire));
     }
-private:
+protected:
```
kboyarinov: Since `reference_vertex` is only used in the implementation of `thread_reference_vertex` which overrides the `release` function, I think it makes sense to completely remove `reference_vertex` and merge everything into `thread_reference_vertex` implementation in the library. We will as well avoid the uncovered `release` in code coverage by doing this.

The latest version of task_group dynamic dependencies do not depend on `reference_vertex`.

src/tbb/task.cpp
akukanov: And of course we can as well fix `get_num_children` to mask the orphaned bit out.

src/tbb/scheduler_common.h
isaevil: Done.

src/tbb/scheduler_common.h
isaevil: Done.

src/tbb/task.cpp
isaevil: Fixed get_num_children and added assertion.

include/oneapi/tbb/detail/_task.h
isaevil: Removed `reference_vertex`.

src/tbb/task.cpp
```diff
@@ -236,9 +237,11 @@ d1::wait_tree_vertex_interface* get_thread_reference_vertex(d1::wait_tree_vertex
...
+                    auto& node = it->second;
+                    __TBB_ASSERT(!(node->m_ref_count.load(std::memory_order_relaxed) & node->m_orphaned_bit),
+                        "the orphaned bit cannot be set");
```
akukanov:
```suggestion
                        "the orphaned bit should not yet be set");
```

src/tbb/scheduler_common.h
```diff
@@ -439,6 +440,58 @@ struct suspend_point_type {
...
+
+    void release_ownership() {
+        auto ref = m_ref_count.fetch_or(m_orphaned_bit);
+        __TBB_ASSERT(!(ref & m_orphaned_bit), "cannot release ownership twice");
+        if (ref == 0) {
+            finalize();
+        }
+    }
+
+    void finalize() {
```
akukanov: I would call this method `destroy`, to make it clear that the object is no more valid afterwards.

src/tbb/scheduler_common.h
```diff
@@ -439,6 +440,58 @@ struct suspend_point_type {
...
+        if (ref == 0) {
+            m_parent->reserve();
+        }
+    }
+
+    void release(std::uint32_t delta = 1) override {
+        // Saving the pointer to the parent before decrementing the reference count
+        // since it won't be safe access any members after that
+        auto parent = m_parent;
+        std::uint64_t ref = m_ref_count.fetch_sub(static_cast<std::uint64_t>(delta)) - static_cast<std::uint64_t>(delta);
```
akukanov: Are the static casts necessary here, especially the second one? I mean, would not the 32-bit value be automatically promoted to 64 bit?

src/tbb/scheduler_common.h
```diff
@@ -439,6 +440,58 @@ struct suspend_point_type {
...
+        // Masking out the orphaned bit to check actual number of references
+        if ((ref & ~m_orphaned_bit) == 0) {
+            parent->release();
+            // The owning thread has abandoned this vertex so it is our responsibility to finalize it
+            if (ref & m_orphaned_bit) {
```
akukanov:
```suggestion
            // If the owning thread has abandoned this vertex, it is our responsibility to destroy it
            if (ref & m_orphaned_bit) {
```

src/tbb/scheduler_common.h
```diff
@@ -439,6 +440,58 @@ struct suspend_point_type {
 #pragma warning( disable: 4324 )
 #endif
 
+class thread_reference_vertex : public d1::wait_tree_vertex_interface {
+public:
+    thread_reference_vertex(wait_tree_vertex_interface *parent, std::uint32_t ref_count)
+        : m_parent{parent}, m_ref_count{ref_count} {}
```
akukanov: What prevents `m_parent` from becoming `nullptr` here? It is never checked for validity afterwards. Should we change the ctor parameter to be a reference instead of a pointer? Should we also make `m_parent` a reference?

isaevil: I think it is OK to make `m_parent` a reference, since we set it only during construction of `thread_reference_vertex`.

src/tbb/scheduler_common.h
```diff
@@ -439,6 +440,58 @@ struct suspend_point_type {
...
+    void reserve(std::uint32_t delta = 1) override {
+        auto ref = m_ref_count.fetch_add(static_cast<std::uint64_t>(delta));
+        if (ref == 0) {
+            m_parent->reserve();
+        }
+    }
+
+    void release(std::uint32_t delta = 1) override {
+        // Saving the pointer to the parent before decrementing the reference count
+        // since it won't be safe access any members after that
```
kboyarinov: I would add more details about the nature of unsafe access:

```suggestion
        // Saving the pointer to the parent before decrementing the reference count
        // since other thread can destroy the vertex after the decrement
```

Feel free to change the wording:)

src/tbb/scheduler_common.h
isaevil: Done.

src/tbb/scheduler_common.h
```diff
@@ -513,9 +566,8 @@ class alignas (max_nfs_size) task_dispatcher {
         }
 
         for (auto& elem : m_reference_vertex_map) {
-            d1::reference_vertex*& node = elem.second;
-            node->~reference_vertex();
-            cache_aligned_deallocate(node);
+            thread_reference_vertex*& node = elem.second;
+            node->release_ownership();
```
kboyarinov, line 578: It seems like we don't need a `node` variable any more. 
```suggestion
            elem.second->release_ownership();
```
In contrast, having a type here may add useful information, so feel free to keep as-is.

src/tbb/task.cpp
kboyarinov: Since `get_num_children` was fixed to mask out the orphaned bit, I think we can remove the assertion here. Now it seems like we are accessing private field (with a friend) just to do an assertion.

src/tbb/scheduler_common.h
isaevil: Yeah, you are right. The cast was needed in `wait_context::add_reference`, because delta was int64 and it seems like this code was a copy-paste code from there. I also added missing overflow checks because as I understand, the initial idea was to have 32 bit delta, so the most significant 32 bits remain reserved (just like in `wait_context`)

src/tbb/task.cpp
isaevil: I think even a little bit silly assert is still good to have in case if we somehow mess up with `get_num_children` in future.

src/tbb/scheduler_common.h
isaevil: Done.

src/tbb/task.cpp
akukanov: But we can move the assertion out of `if`, can't we? In which case it would make more sense.

isaevil: Moved.

src/tbb/task.cpp
```diff
@@ -236,9 +237,11 @@ d1::wait_tree_vertex_interface* get_thread_reference_vertex(d1::wait_tree_vertex
...
             // TODO: Research the possibility of using better approach for a clean-up
             for (auto it = reference_map.begin(); it != reference_map.end();) {
-                if (it->second->get_num_child() == 0) {
-                    it->second->~reference_vertex();
-                    cache_aligned_deallocate(it->second);
+                __TBB_ASSERT(!(it->second->m_ref_count.load(std::memory_order_relaxed) & it->second->m_orphaned_bit),
+                    "the orphaned bit should not yet be set");
+                if (it->second->get_num_children() == 0) {
+                    auto& node = it->second;
+                    node->destroy();
```
akukanov:
```suggestion
                __TBB_ASSERT(it->second, nullptr);
                thread_reference_vertex*& node = it->second;
                __TBB_ASSERT(!(node->m_ref_count.load(std::memory_order_relaxed) & node->m_orphaned_bit),
                    "the orphaned bit should not yet be set");
                if (node->get_num_children() == 0) {
                    node->destroy();
```

src/tbb/task.cpp
```diff
@@ -236,9 +237,11 @@ d1::wait_tree_vertex_interface* get_thread_reference_vertex(d1::wait_tree_vertex
         if (reference_map.size() > max_reference_vertex_map_size) {
             // TODO: Research the possibility of using better approach for a clean-up
             for (auto it = reference_map.begin(); it != reference_map.end();) {
-                if (it->second->get_num_child() == 0) {
-                    it->second->~reference_vertex();
-                    cache_aligned_deallocate(it->second);
+                __TBB_ASSERT(!(it->second->m_ref_count.load(std::memory_order_relaxed) & it->second->m_orphaned_bit),
+                    "the orphaned bit should not yet be set");
```
kboyarinov: May be it makes sense to introduce a `is_orphaned` method in `thread_reference_vertex` and modify the assert?
```suggestion
                __TBB_ASSERT(!it->second->is_orphaned()),
                    "the orphaned bit should not yet be set");
```

akukanov (commented):
The implementation looks good to me.
I have not reviewed the tests.

src/tbb/task.cpp
isaevil: Combined with @kboyarinov suggestions.

src/tbb/task.cpp
isaevil: Combined with https://github.com/uxlfoundation/oneTBB/pull/1785#discussion_r2266243934

src/tbb/scheduler_common.h
isaevil: Decided to keep as-is to be more consistent with the cleanup in `get_thread_reference_vertex`.

kboyarinov (approved):
LGTM

PR #1788, opened 2025-07-21
Implementation of task_group dynamic dependencies - part 4 - support for task cancellation

kboyarinov:
### Description 
Improve `function_task::cancel` to propagate the cancellation signal

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown


include/oneapi/tbb/task_group.h
```diff
@@ -109,8 +109,13 @@ class function_task : public task_handle_task  {
         return next_task;
     }
     d1::task* cancel(d1::execution_data& ed) override {
+        task* t = nullptr;
+#if __TBB_PREVIEW_TASK_GROUP_EXTENSIONS
+        // TODO: complete_task returns one ready successor task, others are spawned. Should cancel() be called instead?
+        t = this->complete_task();
```
isaevil:
```suggestion
        t = this->complete_and_try_bypass_successor();
```

test/tbb/test_task_group.cpp
```diff
@@ -1693,4 +1694,42 @@ TEST_CASE("test task_completion_handle in concurrent environment") {
     CHECK_MESSAGE(task_placeholder == 1, "task body was not executed");
     CHECK_MESSAGE(succ_counter == n, "Not all of the successors were executed");
 }
+
+TEST_CASE("test dependencies and cancellation") {
+    tbb::task_group tg;
+
+    std::size_t execution_placeholder = 0;
+    auto body = [&] { ++execution_placeholder; };
```
isaevil: Instead of counting you can just assert inside of this body because it is not supposed to be executed.
```suggestion
    auto body = [&] { CHECK_MESSAGE(false, "Some tasks were executed in the cancelled task_group"); };
```

include/oneapi/tbb/task_group.h

test/tbb/test_task_group.cpp
kboyarinov: Done

include/oneapi/tbb/task_group.h
```diff
@@ -103,8 +103,14 @@ class function_task : public task_handle_task  {
         return next_task;
     }
     d1::task* cancel(d1::execution_data& ed) override {
+        task* task_ptr = nullptr;
+#if __TBB_PREVIEW_TASK_GROUP_EXTENSIONS
+        // TODO: complete_and_try_get_successor returns one ready successor task, others are spawned and cancelled by the scheduler
+        // Should cancel() be called directly instead?
```
akukanov, line 109: The tradeoff is not obvious. Spawning carries some overhead but allows finalization to be done by multiple threads. And we cannot predict how many tasks are there for finalization. Perhaps some experimentation is needed to see how many (in the order of magnitude) tasks are needed for the break-even point between the two approaches.

include/oneapi/tbb/task_group.h
```diff
@@ -103,8 +103,14 @@ class function_task : public task_handle_task  {
...
+        task_ptr = this->complete_and_try_get_successor();
```
akukanov, line 110: In the case of cancellation, do we even need to preserve the dependencies? Can the dependency counting to be fully skipped, and instead some marker in `task_dynamic_state` put in place of the task pointer to indicate that the task has been taken for cancellation?

kboyarinov: It seems like this is not that straightforward to implement. Consider submitting the predecessor task_handle and not submitting the successor handle with the cancelled execution. 
Since there are not dependencies for the predecessor task, it would be cancelled by the scheduler. Yes, we can mark the successor list of the predecessor as completed (or using the special flag), but what should we do with the successor list. 
I don't think that cancelling all of the tasks without dependency counting is a good idea, since the successor would be cancelled before being submitted for execution in this case.

kboyarinov: By the way, the "correct" support for cancellation needs to be reconsidered in a similar way with the partial graph destruction. Imagine cancelling some of the tasks and then resetting the `task_group_context`. 
The idea of this patch is to avoid hangs on waiting if the execution is cancelled as a "preliminary" support for cancellation.

akukanov: So, we have more open questions to add to the design RFC it seems.

src/tbb/task_dispatcher.h
```diff
@@ -338,6 +338,7 @@ d1::task* task_dispatcher::local_wait_for_all(d1::task* t, Waiter& waiter ) {
 
                     ITT_CALLEE_ENTER(ITTPossible, t, itt_caller);
 
+                    m_innermost_running_task = t;
```
vossmjp: Same double setting of m_innermost_running_task issue.

kboyarinov: Rebased to part1 after merging the transferring support. Resolved.

vossmjp (approved):
I would suggest (at some point) adding a test that runs at least some part of the task graph before the task_group is canceled. So probably, cancel from within a task at an intermediate level.

PR #1791, opened 2025-07-25
WIP: Add tasking documentation to user guide

vossmjp:
### Description 

### Type of change

- [X ] documentation - _documentation update_

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown


doc/main/tbb_userguide/Parallelizing_with_Tasks.rst
```diff
@@ -0,0 +1,18 @@
+.. _Parallelizing_with_Tasks:
+
+Parallelizing with Tasks
+========================
+
+When parallel loops or the flow graph are not sufficient, the |full_name|
+library supports parallelization directly with tasks. Independent tasks
```
aleksei-fedotov: In light of the introduction of dependencies between the tasks of the task group, shall we consider omitting the word "independent"?
```suggestion
library supports parallelization directly with tasks. Tasks
```

doc/main/tbb_userguide/examples/task_examples.cpp
```diff
@@ -0,0 +1,318 @@
+/*
+    Copyright (c) 2025 Intel Corporation
```
aleksei-fedotov: As far as I know, it should be "Copyright (c) 2025 UXL Foundation Contributors" nowadays.

doc/main/tbb_userguide/examples/task_examples.cpp
```diff
@@ -0,0 +1,318 @@
...
+            sequential_tree_search(node->right, target, result);
+        }
+    }
+}
+
+void parallel_invoke_search(TreeNode* node, int target, std::atomic<TreeNode*>& result,
+                            size_t depth_threshold = initial_depth_threshold) {
+    if (node && !result.load()) { 
+        if (node->value == target) {
+            result.store(node);
```
aleksei-fedotov: I suppose for this example it is okay to have logical races here in case of two nodes satisfying the search criteria have been found simultaneously? Though, I would suggest describing such behavior at least so that readers know it is expected here. Or, probably better, just avoid it using CAS operation?

doc/main/tbb_userguide/examples/task_examples.cpp
```diff
@@ -0,0 +1,318 @@
...
+/*begin_parallel_search*/
+void parallel_tree_search_impl(tbb::task_group& tg, TreeNode* node, int target, 
+                               std::atomic<TreeNode*>& result,
+                               size_t depth_threshold = initial_depth_threshold) {
+    if (node && !result.load()) { 
+        if (node->value == target) {
+            result.store(node);
+        } else if (depth_threshold == 0) {
+            // at parallel cutoff
+            sequential_tree_search(node, target, result);
```
aleksei-fedotov, line 97: The code for `sequential_tree_search` is shown in different .rst file than from which it is referenced. Unless it is supposed and clear for readers that they need to read these documentation pages continuously, I suggest including reference to the implementation of this function on all of the pages in which it is mentioned. This also applies to the `TreeNode` structure. It may be not clear to readers what these things are.

doc/main/tbb_userguide/examples/task_examples.cpp
```diff
@@ -0,0 +1,318 @@
...
+    std::cout << "Serial search:                   " << serial_duration.count() << " us\n";
+    std::cout << "Parallel search (invoke):       " << parallel_invoke_duration.count() << " us\n";
+    std::cout << "Parallel search (basic):         " << parallel_duration.count() << " us\n";
+    std::cout << "Parallel search (cancellation):  " << cancellation_duration.count() << " us\n";
+    
+    if (serial_duration.count() > 0) {
+        double speedup_invoke = static_cast<double>(serial_duration.count()) / parallel_invoke_duration.count();
+        double speedup_basic = static_cast<double>(serial_duration.count()) / parallel_duration.count();
+        double speedup_cancel = static_cast<double>(serial_duration.count()) / cancellation_duration.count();
+         std::cout << "Speedup (invoke):                " << speedup_invoke << "x\n";
```
aleksei-fedotov:
```suggestion
        std::cout << "Speedup (invoke):                " << speedup_invoke << "x\n";
```

doc/main/tbb_userguide/examples/task_examples.cpp
```diff
@@ -0,0 +1,318 @@
...
+    unique_values.reserve(num_nodes);
+    
+    while (unique_values.size() < num_nodes) {
+        int value = dist(gen);
+        // Check if value is already used (O(1) average time complexity)
+        if (used_values.find(value) == used_values.end()) {
+            unique_values.push_back(value);
+            used_values.insert(value);
+        }
+    }
```
aleksei-fedotov: Optimize at the expense of memory footprint?
```suggestion
    // Generate unique values for all nodes
    std::size_t universe_size = dist.b();
    std::vector<int> universe(universe_size);
    std::iota(universe.begin(), universe.end(), 1);
    std::vector<int> unique_values(num_nodes);
    for (size_t i = 0; i < num_nodes; ++i, --universe_size) {
        const std::size_t index = (dist(gen) - 1) % universe_size;
        unique_values[i] = universe[index];
        std::swap(universe[index], universe[universe_size - 1]);
    }

```

test/tbb/test_task_group.cpp
```diff
@@ -1221,4 +1222,78 @@ TEST_CASE("task_handle cannot be scheduled into other task_group of the same con
     CHECK_THROWS_WITH_AS(tg1.run(tg.defer([]{})), "Attempt to schedule task_handle into different task_group", std::runtime_error);
 }
 
+//! Test for thread-safe exception handling in concurrent wait scenarios
+//! \brief \ref error_guessing \ref interface
+TEST_CASE("Concurrent exception handling in task_group wait") {
```
aleksei-fedotov, line 1242: I saw similar change in another PR.

doc/main/tbb_userguide/examples/task_examples.cpp
vossmjp: Done

doc/main/tbb_userguide/examples/task_examples.cpp
vossmjp: Ok, will add.

doc/main/tbb_userguide/examples/task_examples.cpp
vossmjp: Ok, I will add CAS

doc/main/tbb_userguide/task_group_cancellation.rst
```diff
@@ -0,0 +1,43 @@
+.. _task_group_cancellation:
+
+task_group cancellation
```
akukanov: Though I can't stand the capitalization in headings :), especially when they also have API names, let's be consistent:
```suggestion
task_group Cancellation
```
But also, we have a special chapter for cancellation - so I would move the `task_group` example there as well, maybe merging it with `Cancellation_Without_An_Exception.rst`

doc/main/tbb_userguide/creating_tasks_with_parallel_invoke.rst
```diff
@@ -0,0 +1,55 @@
+.. _creating_tasks_with_parallel_invoke:
+
+Creating Tasks with parallel_invoke
+===================================
+
+Suppose you want to search a binary tree for the node that contains a specific value.
+Here is sequential code to do this:
```
akukanov, line 7:
```suggestion
Here is sequential code to do this.
```

doc/main/tbb_userguide/creating_tasks_with_parallel_invoke.rst
```diff
@@ -0,0 +1,55 @@
...
+and, if the value is not found, calls itself on the left and right subtree. 
+
+.. literalinclude:: ./examples/task_examples.cpp
+    :language: c++
+    :start-after: /*begin_search_serial*/
+    :end-before: /*end_search_serial*/
+
+
+To improve performance, you can use ``oneapi::tbb::parallel_invoke`` to search the tree 
+in parallel:
```
akukanov, line 26:
```suggestion
To improve performance, you can use ``oneapi::tbb::parallel_invoke`` to search the tree in parallel.
```

doc/main/tbb_userguide/examples/task_examples.cpp
```diff
@@ -0,0 +1,319 @@
...
+}
+/*end_search_serial*/
+
+/*begin_sequential_tree_search*/
+void sequential_tree_search(TreeNode* node, int target, std::atomic<TreeNode*>& result) {
+    if (node && !result.load()) { 
+        if (node->value == target) {
+            TreeNode* expected = nullptr;
+            result.compare_exchange_strong(expected, node);
```
akukanov: If the result of CAS is not checked, do we even need CAS here?
The target value is the same; if there are multiple instances of it in the tree, the result is non-deterministic anyway, and if there is only one instance, then only one thread writes the result.

doc/main/tbb_userguide/creating_tasks_with_task_group.rst
```diff
@@ -0,0 +1,51 @@
+.. _creating_tasks_with_task_group:
+
+Creating Tasks with task_group
+==============================
+
+The |full_name| library supports parallelization directly with tasks. The class 
+``oneapi::tbb::task_group`` is used to run and wait for independent tasks in a less
```
akukanov:
```suggestion
``oneapi::tbb::task_group`` is used to run and wait for tasks in a less
```

doc/main/tbb_userguide/creating_tasks_with_task_group.rst
```diff
@@ -0,0 +1,51 @@
...
+structured way than ``oneapi::tbb::parallel_invoke``. It is useful when you want to
+create a set of tasks that can be run in parallel, but you do not know how many there
+will be in advance.
+
+Here is code that uses ``oneapi::tbb::task_group`` to implement a parallel search in
+a binary tree:
```
akukanov:
```suggestion
Here is code that uses ``oneapi::tbb::task_group`` to implement a parallel search in
a binary tree.
```

doc/main/tbb_userguide/task_group_cancellation.rst
```diff
@@ -0,0 +1,43 @@
...
+=======================
+
+The |full_name| library provides a way to cancel tasks using
+``oneapi::tbb::task_group_context``. This is useful when you want
+to stop the execution of tasks that are no longer needed or when a
+condition is met that makes further execution unnecessary. 
+
+Here is an example of how to use task cancellation with ``oneapi::tbb::task_group``
+using an explicit ``oneapi::tbb::task_group_context``:
```
akukanov:
```suggestion
Here is an example of how to use task cancellation with ``oneapi::tbb::task_group``
using an explicit ``oneapi::tbb::task_group_context``.
```

doc/main/tbb_userguide/task_group_cancellation.rst
```diff
@@ -0,0 +1,43 @@
...
+
+
+Nodes are represented by ``struct TreeNode``.
+
+.. literalinclude:: ./examples/task_examples.cpp
+    :language: c++
+    :start-after: /*begin_treenode*/
+    :end-before: /*end_treenode*/
```
akukanov: I saw your conversation with @aleksei-fedotov about referring to the basic code on all the pages - but instead of duplication, why not add a link to `creating_tasks_with_task_group.rst`?

doc/main/tbb_userguide/examples/task_examples.cpp
```diff
@@ -0,0 +1,319 @@
...
+       if (node->value == target) {
+            TreeNode* expected = nullptr;
+            if (result.compare_exchange_strong(expected, node)) {
+                ctx.cancel_group_execution();
+            }
+       } else if (depth_threshold == 0) {
+            sequential_tree_search(node, target, result);
+            if (result.load() != nullptr) {
+                ctx.cancel_group_execution();
+            }
```
akukanov: The first branch seemingly tries to avoid multiple cancellations by using CAS. Yet in the second branch it is not known if `result` was set in the `sequential_tree_search` call or by another thread, and so multiple cancellations are possible. Since `ctx.cancel_group_execution` is thread-safe if I am not mistaken, maybe CAS is unnecessary in this code variation as well?

doc/main/tbb_userguide/task_group_thread_safety.rst
```diff
@@ -0,0 +1,88 @@
+.. _task_group_thread_safety:
+
+task_group Thread Safety
+========================
+
+The use of a shared ``task_group`` object across different threads is safe and easy to reason about
+in many common cases, such as recursive algorithms. But sometimes it is difficult
+to reason about the concurrent use of a single shared ``task_group`` across threads.
+
+In :ref:`creating_tasks_with_task_group`, tasks that are executing within a ``task_group`` add additional tasks
```
akukanov:
```suggestion
In :ref:`creating_tasks_with_task_group`, tasks that are executing within a ``task_group`` add more tasks
```

doc/main/tbb_userguide/task_group_thread_safety.rst
```diff
@@ -0,0 +1,88 @@
...
+by calling ``run`` on the same ``task_group`` object. The call to ``task_group::wait`` is then made
+from the single thread that started the recursive parallel algorithm. The calls to ``run`` are logically nested
+within a recursive algorithm and that single call to ``wait`` is guaranteed to wait
+for all of the children tasks, even those added from other worker threads.
+
+In less structured cases, calls to ``task_group::run`` and ``task_group::wait`` on the same object
+from different threads do not offer such simple guarantees. Here is a diagram that shows a single ``task_group``
+object that is accessed in parallel by three different threads. Each thread runs some tasks and then calls wait
+on the shared ``task_group``:
```
akukanov, line 19:
```suggestion
on the shared ``task_group``.
```

doc/main/tbb_userguide/task_group_thread_safety.rst
```diff
@@ -0,0 +1,88 @@
...
+are guaranteed to be complete when the call to wait returns. So for example, the thread that runs the
+`A` tasks is guaranteed to wait for all the `A` tasks in its call to ``wait``. 
+
+Second, any ``run`` that *inter-thread happens before* a call to ``wait`` on another thread will be complete
+when that call to ``wait`` returns.
+
+Both of these guarantees mean that if you use C++ mechanisms to order the calls to ``run``
+and ``wait`` on the same ``task_group``, this ordering will be respected. But if you do not
+enforce an ordering, then there can be races between when a task is ``run`` in a ``task_group`` and
+the ``wait`` on that ``task_group`` in another thread.
```
akukanov:
```suggestion
and ``wait`` on the same ``task_group``, this ordering will be respected. But if you do not
enforce an ordering, task submissions and waits on different threads are not synchronized.
That is, it is unknown if a task in the ``task_group`` will be complete when
a ``wait`` on that ``task_group`` by another thread returns.
```

doc/main/tbb_userguide/creating_tasks_with_parallel_invoke.rst
vossmjp: The use of ":" is consistent with other pages, such as [parallel_for](https://github.com/uxlfoundation/oneTBB/blob/master/doc/main/tbb_userguide/parallel_for_os.rst). I don't really have a personal preference, but we should probably be consistent.

doc/main/tbb_userguide/creating_tasks_with_parallel_invoke.rst
vossmjp: Same as above.

doc/main/tbb_userguide/examples/task_examples.cpp
vossmjp: The CAS is not strictly necessary anywhere since any result is as good as any other. I added CAS in response to a previous suggestion, but I agree it perhaps just complicates the example. I think I should remove and instead add a comment that notes that any race is benign.

doc/main/tbb_userguide/task_group_cancellation.rst
vossmjp: Do you think the page is too long with these added? There's some advantage to having all of the relevant code on a single page, as long as it doesn't become too long. I initially omitted the code because I wanted to keep it short, but after putting it there, it seems ok to me.

doc/main/tbb_userguide/examples/task_examples.cpp
vossmjp: As mentioned above, I'll probably just remove all the CASes and add comments.

doc/main/tbb_userguide/examples/task_examples.cpp
akukanov: Sounds good to me.

doc/main/tbb_userguide/examples/task_examples.cpp
akukanov: See the comments https://github.com/uxlfoundation/oneTBB/pull/1791#discussion_r2351981425 and https://github.com/uxlfoundation/oneTBB/pull/1791#discussion_r2352196632. First, CAS does not remove the "logical race", it only changes the "winner" of that race to be the first thread and not the last. And second, in the code with cancellation even that is not achieved.

doc/main/tbb_userguide/task_group_cancellation.rst
akukanov: I think it's more of a "philosophical" question of considering the doc as a set of independent pages or a single "book". In the first case, each page is as self-sufficient as it can be, and may duplicate information. In the second case, it gives enough context but refers to another part of the book for details. I have a preference towards that second case and find repeating the same words 3 or 4 times a bit awkward. But I can live with that :)

doc/main/tbb_userguide/examples/task_examples.cpp
```diff
@@ -134,10 +131,8 @@ void parallel_tree_search_cancellable_impl(tbb::task_group& tg, TreeNode* node,
                                            size_t depth_threshold = initial_depth_threshold) {
     if (node && !ctx.is_group_execution_cancelled()) {
        if (node->value == target) {
-            TreeNode* expected = nullptr;
-            if (result.compare_exchange_strong(expected, node)) {
-                ctx.cancel_group_execution();
-            }
+            result.store(node); // overwrite is ok since any result is valid
+            ctx.cancel_group_execution();
```
akukanov: Also add a comment (or maybe a text paragraph) that multiple cancellations are OK if there is only a single wait and no resets?

doc/main/tbb_userguide/examples/task_examples.cpp
```diff
@@ -0,0 +1,314 @@
...
+        } else if (node->left) {
+            if (node->right) {
+                tg.run([node, target, &result, &tg, &ctx, depth_threshold] {
+                    parallel_tree_search_cancellable_impl(tg, node->right, target, result, ctx, depth_threshold - 1);
+                });
+            }
+            parallel_tree_search_cancellable_impl(tg, node->left, target, result, ctx, depth_threshold - 1);
+        } else if (node->right) {
+            parallel_tree_search_cancellable_impl(tg, node->right, target, result, ctx, depth_threshold - 1);
+        }
```
aleksei-fedotov: What is the main point we want to make with changes from this patch to the documentation? It is not about cutting the overheads involved, is it? It is rather about guarantees provided for the cases with single or multiple waits and/or work cancellation. In such case, I would try to be more readable and straightforward to readers. So, I am suggesting rewrite this part as:
```suggestion
        } else {
            // Run on left and right subtrees in parallel
            tg.run([node, target, &result, &tg, &ctx, depth_threshold] {
                parallel_tree_search_cancellable_impl(tg, node->left, target, result, ctx,
                                                      depth_threshold - 1);
            });
            tg.run([node, target, &result, &tg, &ctx, depth_threshold] {
                parallel_tree_search_cancellable_impl(tg, node->right, target, result, ctx,
                                                      depth_threshold - 1);
            });
        }
```
Maybe, describing in text somewhere that some optimization can be added to cut the overheads involved in task scheduling, such as adjusting `depth_threshold` and/or spawning less number of tasks if it does not limit parallelism of the entire system.

Opinions?

akukanov: I tend to agree with Aleksei that simplifying the example and structuring it similarly to the sequential tree traversal and the `parallel_invoke` implementation would help with understanding.

Or we can do similar checks for `parallel_invoke` and in the sequential code, and ensure that the sub-tree node passed to a search function is never null.

doc/main/tbb_userguide/examples/task_examples.cpp
vossmjp: Agreed, removed.

doc/main/tbb_userguide/examples/task_examples.cpp
vossmjp: My intention was to show that `task_group` is more flexible than `parallel_invoke`. But perhaps its not necessary and readability should win out. I'll make the suggestion change.

doc/main/tbb_userguide/examples/task_examples.cpp
vossmjp: Ok, will do.

doc/main/tbb_userguide/task_group_cancellation.rst
vossmjp: Yep, I was leaning towards independent pages.

aleksei-fedotov (approved):
Looks good to me!

doc/main/tbb_userguide/examples/task_examples.cpp
```diff
@@ -0,0 +1,316 @@
...
+            sequential_tree_search(node, target, result);
+        } else {
+            // Run on left and right subtrees in parallel
+            tg.run([node, target, &result, &tg, depth_threshold] {
+                parallel_tree_search_impl(tg, node->left, target, result,
+                                          depth_threshold - 1);
+            });
+            tg.run([node, target, &result, &tg, depth_threshold] {
+                parallel_tree_search_impl(tg, node->right, target, result,
+                                          depth_threshold - 1);
```
akukanov:
```suggestion
            tg.run([node, target, &result, &tg, depth_threshold] {
                parallel_tree_search_impl(tg, node->left, target, result, depth_threshold - 1);
            });
            tg.run([node, target, &result, &tg, depth_threshold] {
                parallel_tree_search_impl(tg, node->right, target, result, depth_threshold - 1);
```
There seems to be enough space on the right to fit into the code block width.

doc/main/tbb_userguide/creating_tasks_with_task_group.rst
```diff
@@ -0,0 +1,34 @@
...
+    :language: c++
+    :start-after: /*begin_parallel_search*/
+    :end-before: /*end_parallel_search*/
+
+This example uses recursion to create many tasks. The depth of the parallel recursion is
+limited by the ``depth_threshold`` parameter. After this depth is reached, no new tasks
+are created. The value of ``result`` is periodically checked to see if the value has been
+found by other concurrent tasks, and if so, the search in the current task is terminated.
+
+In this example, tasks that execute within the ``task_group tg`` add additional tasks
```
akukanov:
```suggestion
In this example, tasks that execute within the ``task_group tg`` create additional tasks
```

doc/main/tbb_userguide/creating_tasks_with_task_group.rst
```diff
@@ -0,0 +1,34 @@
...
+by calling ``run`` on the same ``task_group`` object. These calls are thread-safe and the 
+call to ``tg.wait()`` will block until all of these tasks are complete. Although these
+additional tasks might be added from different worker threads, these additions are logically nested
+within the top-most calls to ``tg.run``. There is therefore no race in adding these tasks from
```
akukanov:
```suggestion
call to ``tg.wait()`` will block until all of these tasks are complete. Although
tasks might be added from different worker threads, these additions are logically nested
within the top-most calls to ``tg.run``. There is therefore no race in adding these tasks from
```

doc/main/tbb_userguide/task_group_thread_safety.rst
```diff
@@ -0,0 +1,92 @@
...
+to reason about the concurrent use of a single shared ``task_group`` across threads.
+
+In :ref:`creating_tasks_with_task_group`, tasks that are executing within a ``task_group`` add more tasks
+by calling ``run`` on the same ``task_group`` object. The call to ``task_group::wait`` is then made
+from the single thread that started the recursive parallel algorithm. The calls to ``run`` are logically nested
+within a recursive algorithm and that single call to ``wait`` is guaranteed to wait
+for all of the children tasks, even those added from other worker threads.
+
+In less structured cases, calls to ``task_group::run`` and ``task_group::wait`` on the same object
+from different threads do not offer such simple guarantees. Here is a diagram that shows a single ``task_group``
```
akukanov:
```suggestion
In less structured cases with calls to both ``task_group::run`` and ``task_group::wait`` on the same object
from different threads, the behavior is more sophisticated. Here is a diagram that shows a single ``task_group``
```
Here I rephrased the sentence to avoid "do not offer such simple guarantees" which might be interpreted as a weakness :)

doc/main/tbb_userguide/examples/task_examples.cpp
```diff
@@ -0,0 +1,316 @@
...
+
+/*begin_parallel_search_cancellation*/
+void parallel_tree_search_cancellable_impl(tbb::task_group& tg, TreeNode* node, int target, 
+                                           std::atomic<TreeNode*>& result, 
+                                           tbb::task_group_context& ctx,
+                                           size_t depth_threshold = initial_depth_threshold) {
+    if (node && !ctx.is_group_execution_cancelled()) {
+        if (node->value == target) {
+            result.store(node); // overwrite is ok since any result is valid
+            ctx.cancel_group_execution(); // multiple cancellations are ok due to single wait
```
akukanov: You do not really need to explicitly pass the `task_group_context` for inner cancellation. Actually, you do not even need to use an explicit `task_group_context` at all.
`ctx.is_group_execution_cancelled()` can be replaced with `tbb::is_current_task_group_canceling()`.
`ctx.cancel_group_execution()` can be replaced with `tg.cancel()`.

doc/main/tbb_userguide/task_group_thread_safety.rst
```diff
@@ -0,0 +1,92 @@
...
+``canceled``. However, an intervening call to ``wait`` on another thread, which results in a reset of the
+``task_group_context``, may cause the call to ``wait`` on the thread that executes the `B` tasks to return
+a status other than ``canceled``.
+
+A thread that has canceled a ``task_group`` and then ``runs`` more tasks may see those tasks execute because
+a ``wait`` on another thread completed and reset the ``task_group`` before those tasks were ``run``, effectively
+uncanceling the ``task_group``.
+
+Exceptions that are thrown from a ``task_group`` task cause the cancellation of the ``task_group`` and so
+applications that throw exceptions and have concurrent waits, can have similarly complicated behavior. In
```
akukanov:
```suggestion
Exceptions that are thrown from a ``task_group`` task cause the cancellation of the ``task_group``, so
applications that throw exceptions and have concurrent waits can have similarly complicated behavior. In
```

doc/main/tbb_userguide/examples/task_examples.cpp
vossmjp: Yes, I know. It's not uncommon for people to mix up or conflate task_group and task_group_context. So I intentionally separated them in the code, to try to demonstrate that there's nothing really special about canceling a task_group -- its just canceling the task_group_context.  In the task_group_thread_safety section, it talks about how waiting for a task_group resets the task_group_context. So I just wanted these objects to be very obvious in this example. I could add something that mentions that there are simpler ways to access the context from a task_group.

akukanov: > there's nothing really special about canceling a task_group -- its just canceling the task_group_context. 

The task_group_context is not even needed to cancel a task_group; the latter has the corresponding method. That's what I would use in the documentation example of cancellation.
Anyway, up to you.

vossmjp: Ok, after reviewing the text again, I noticed that I used tg.wait() in one of the figures in task_group_thread_safety, so I was not being even self-consistent. I've updated the example to use the task_group interfaces directly and also updated Cancellation_Without_An_Exception.rst to note the relationship between these functions and the underlying task_group_context. Overall, I think this is better and provides the more preferred usage pattern for cancellation of a task_group while also highlighting that relationship.

PR #1795, opened 2025-07-30
Enable multiple core type constraints and identify additional core type

dnmokhov:
### Description 
Add `set_core_types` API to `tbb::task_arena::constraints`. Identify L3-less cores as a new, least performant, core type.

### Type of change

### Tests
- [ ] not needed

### Documentation
- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

@wangleis


dnmokhov:
> These are changes to the public API. We cannot just introduce those in a code patch - at the very least, we need to decide if these APIs should go into the specification or not, as well as discuss the semantics, naming etc. Will there be an RFC prepared for this purpose?

Good point. I have an RFC (on a branch) for these changes to occur in [master](https://github.com/uxlfoundation/oneTBB/tree/master). This PR was supposed to make the feature available in [onetbb_2021.2_ov](https://github.com/uxlfoundation/oneTBB/tree/onetbb_2021.2_ov) but that does make it public in a way, which is not the intention.

I will reevaluate the options.

include/oneapi/tbb/info.h
```diff
@@ -1,5 +1,5 @@
 /*
-    Copyright (c) 2019-2021 Intel Corporation
+    Copyright (c) 2019-2025 Intel Corporation
```
vossmjp, line 2:
```suggestion
    Copyright (c) 2019-2025 Intel Corporation
    Copyright (c) 2025 UXL Foundation Contributors
```

dnmokhov: updated

akukanov:
> Good point. I have an RFC (on a branch) for these changes to occur in [master](https://github.com/uxlfoundation/oneTBB/tree/master). This PR was supposed to make the feature available in [onetbb_2021.2_ov](https://github.com/uxlfoundation/oneTBB/tree/onetbb_2021.2_ov) but that does make it public in a way, which is not the intention.

I realized soon after commenting that the patch is for the special branch (and this is why I deleted that comment :)). That reduces most of the concern; we should say that this is a preview variation of the functionality, with no guarantees of availability or compatibility beyond this branch.

Yet, an RFC is eventually needed - so it's good that you work on it.

include/oneapi/tbb/info.h
```diff
@@ -57,6 +58,41 @@ struct constraints {
...
         return *this;
     }
+    constraints& set_core_types(const std::vector<core_type_id>& ids) {
+        if (ids.empty()) {
+            core_type = -1;
+            return *this;
+        }
+
+        // Set a marker bit to indicate multiple core type format
+        core_type = (1 << core_type_id_bits);
```
vossmjp: What if ids is only size of 1. Should the new format be used or the old. Seems like it would be slightly faster to just set core_type to ids[0] if the size is 1 and keep the old format. This would also mean that a call to single_core_type() would be more accurate when only a single core type is set using a vector.

vossmjp (commented):
What about get_default_concurrency in tbb_bind.cpp? It takes only a single core_type_index. Will constraints_default_concurrency not work then?

include/oneapi/tbb/info.h
dnmokhov: Good point. Let me look.

dnmokhov:
> What about get_default_concurrency in tbb_bind.cpp? It takes only a single core_type_index. Will constraints_default_concurrency not work then?

Like `binding_handler`, it uses `fill_constraints_affinity_mask`, so it works.

vossmjp:
When RFC is revisited for master branch, we should remember that core_type is public in the constraints structure. So the new format would be exposed directly through that member variable.

src/tbbbind/tbb_bind.cpp
```diff
@@ -191,6 +191,40 @@ class platform_topology {
                 }
             }
         }
+        // On hybrid CPUs, check if there are cores without L3 cache.
+        if (!core_types_parsing_broken && core_types_number > 1) {
+            // The first core type mask (least performant cores)
+            hwloc_cpuset_t& front = core_types_affinity_masks_list.front();
+            hwloc_cpuset_t lp_mask = hwloc_bitmap_dup(front);
```
isaevil, line 198: Could you please add this logic to `common_arena_constraints.h` as well to better test it?

include/oneapi/tbb/info.h
dnmokhov: Switched to the old format when size of `ids` is 1.

src/tbbbind/tbb_bind.cpp
dnmokhov: Added.

dnmokhov: Also added `set_core_types` tests.

src/tbbbind/tbb_bind.cpp
```diff
@@ -266,7 +300,17 @@ class platform_topology {
     void fill_constraints_affinity_mask(affinity_mask input_mask, int numa_node_index, int core_type_index, int max_threads_per_core) {
         __TBB_ASSERT(is_topology_parsed(), "Trying to get access to uninitialized platform_topology");
         __TBB_ASSERT(numa_node_index < (int)numa_affinity_masks_list.size(), "Wrong NUMA node id");
+#if __TBB_PREVIEW_TASK_ARENA_CONSTRAINTS_EXTENSION_PRESENT
```
sunxiaoxia2022, line 298: `__TBB_PREVIEW_TASK_ARENA_CONSTRAINTS_EXTENSION_PRESENT` is not defined in this file. 
I suppose you should add `__TBBBIND_BUILD` in https://github.com/uxlfoundation/oneTBB/blob/dev/dnmokhov/2021.2/include/oneapi/tbb/detail/_config.h#L361.

dnmokhov: Thanks for pointing this out, @sunxiaoxia2022! Added.

sunxiaoxia2022 (approved):

wangleis (approved):

src/tbb/governor.cpp
```diff
@@ -502,7 +503,7 @@ void constraints_assertion(d1::constraints c) {
     int* core_types_begin = system_topology::core_types_indexes;
     int* core_types_end = system_topology::core_types_indexes + system_topology::core_types_count;
     __TBB_ASSERT_RELEASE(c.core_type == system_topology::automatic ||
-        (is_topology_initialized && std::find(core_types_begin, core_types_end, c.core_type) != core_types_end),
+        (is_topology_initialized && ((c.single_core_type() && std::find(core_types_begin, core_types_end, c.core_type) != core_types_end) || !c.single_core_type())),
```
kboyarinov: It seems like we can avoid duplication of `single_core_type` by reversing the order of arguments:

```suggestion
        (is_topology_initialized && (!c.single_core_type() || std::find(core_types_begin, core_types_end, c.core_type) != core_types_end)),
```

src/tbbbind/tbb_bind.cpp
```diff
@@ -191,6 +191,40 @@ class platform_topology {
...
+            int l3_depth = hwloc_get_type_depth(topology, HWLOC_OBJ_L3CACHE);
+            unsigned l3_count = hwloc_get_nbobjs_by_depth(topology, l3_depth);
+
+            // Iterate through all L3 cache objects and remove their cores from lp_mask.
+            for (unsigned i = 0; i < l3_count; ++i) {
+                hwloc_obj_t l3 = hwloc_get_obj_by_depth(topology, l3_depth, i);
+                if (l3) {
+                    hwloc_bitmap_andnot(lp_mask, lp_mask, l3->cpuset);
+                }
+            }
```
isaevil: I wonder wouldn't that be identical implementation:
```suggestion
            hwloc_obj_t l3_package = nullptr;
            while ( (l3_package = hwloc_get_next_obj_by_type(topology, HWLOC_OBJ_L3CACHE, l3_package)) != nullptr ) {
                    hwloc_bitmap_andnot(lp_mask, lp_mask, l3_package->cpuset);
            }
```
Feel free to ignore since it is not a crucial change.

dnmokhov: Let me verify.

dnmokhov: Updated.

It is simpler and achieves the same effect. Thanks!

isaevil (approved):
LGTM

kboyarinov (approved):

include/oneapi/tbb/version.h
```diff
@@ -44,7 +45,7 @@
 // OneAPI oneTBB specification version
 #define ONETBB_SPEC_VERSION "1.0"
 // Full interface version
-#define TBB_INTERFACE_VERSION 12025
+#define TBB_INTERFACE_VERSION 12026
```
omalyshe, line 48: We also need to update copyright year for DLLs, that is change it in *.rc files

PR #1803, opened 2025-08-04
Distributed size for concurrent ordered containers

kboyarinov:
### Description 

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown

### DIFF
--- a/include/oneapi/tbb/concurrent_set.h
+++ b/include/oneapi/tbb/concurrent_set.h
@@ -25,7 +25,7 @@
 
 namespace tbb {
 namespace detail {
-namespace d2 {
+namespace d3 {
 
 template<typename Key, typename KeyCompare, typename RandomGenerator, typename Allocator, bool AllowMultimapping>
 struct set_traits {
@@ -52,8 +52,8 @@ template <typename Key, typename Compare, typename Allocator>
 class concurrent_multiset;
 
 template <typename Key, typename Compare = std::less<Key>, typename Allocator = tbb::tbb_allocator<Key>>
-class concurrent_set : public concurrent_skip_list<set_traits<Key, Compare, concurrent_geometric_level_generator<32>, Allocator, false>> {
-    using base_type = concurrent_skip_list<set_traits<Key, Compare, concurrent_geometric_level_generator<32>, Allocator, false>>;
+class concurrent_set : public concurrent_skip_list<set_traits<Key, Compare, geometric_level_generator<32>, Allocator, false>> {
+    using base_type = concurrent_skip_list<set_traits<Key, Compare, geometric_level_generator<32>, Allocator, false>>;
 public:
     using key_type = Key;
     using value_type = typename base_type::value_type;
@@ -153,8 +153,8 @@ void swap( concurrent_set<Key, Compare, Allocator>& lhs,
 }
 
 template <typename Key, typename Compare = std::less<Key>, typename Allocator = tbb::tbb_allocator<Key>>
-class concurrent_multiset : public concurrent_skip_list<set_traits<Key, Compare, concurrent_geometric_level_generator<32>, Allocator, true>> {
-    using base_type = concurrent_skip_list<set_traits<Key, Compare, concurrent_geometric_level_generator<32>, Allocator, true>>;
+class concurrent_multiset : public concurrent_skip_list<set_traits<Key, Compare, geometric_level_generator<32>, Allocator, true>> {
+    using base_type = concurrent_skip_list<set_traits<Key, Compare, geometric_level_generator<32>, Allocator, true>>;
 public:
     using key_type = Key;
     using value_type = typename base_type::value_type;
@@ -252,13 +252,13 @@ void swap( concurrent_multiset<Key, Compare, Allocator>& lhs,
     lhs.swap(rhs);
 }
 
-} // namespace d2
+} // namespace d3
 } // namespace detail
 
 inline namespace v1 {
 
-using detail::d2::concurrent_set;
-using detail::d2::concurrent_multiset;
+using detail::d3::concurrent_set;
+using detail::d3::concurrent_multiset;
 using detail::split;
 
 } // inline namespace v1
--- a/include/oneapi/tbb/detail/_concurrent_skip_list.h
+++ b/include/oneapi/tbb/detail/_concurrent_skip_list.h
@@ -49,24 +49,26 @@
 
 namespace tbb {
 namespace detail {
-namespace d2 {
+namespace d3 {
 
-template <typename Generator, typename SizeType>
+template <typename LevelGenerator, typename SizeType>
 class skip_list_thread_data {
 public:
-    using generator_type = Generator;
+    using level_generator_type = LevelGenerator;
     using size_type = SizeType;
 
     skip_list_thread_data() : my_local_size(0) {}
 
-    std::size_t random_height() { return my_rng(); }
+    std::size_t random_level() { return my_rng(); }
     std::size_t local_size() const { return my_local_size.load(std::memory_order_relaxed); }
 
-    void set_size(size_type value) { my_local_size = value; }
-    void increment_size() { ++my_local_size; }
+    void set_size(size_type size) { my_local_size.store(size, std::memory_order_relaxed); }
+    void increment_size() {
+        my_local_size.store(local_size() + 1, std::memory_order_relaxed);
+    }
 
 private:
-    generator_type my_rng;
+    level_generator_type my_rng;
     std::atomic<size_type> my_local_size;
 };
 
@@ -295,6 +297,8 @@ protected:
     using const_pointer = typename allocator_traits_type::const_pointer;
 
     using random_level_generator_type = typename container_traits::random_level_generator_type;
+    using thread_data_type = skip_list_thread_data<random_level_generator_type, size_type>;
+    using ets_type = tbb::enumerable_thread_specific<thread_data_type>;
 
     using node_ptr = list_node_type*;
 
@@ -434,7 +438,8 @@ public:
     std::pair<iterator, bool> insert( node_type&& nh ) {
         if (!nh.empty()) {
             auto insert_node = d1::node_handle_accessor::get_node_ptr(nh);
-            std::pair<iterator, bool> insert_result = internal_insert_node(insert_node);
+            thread_data_type& td = my_ets.local();
+            std::pair<iterator, bool> insert_result = internal_insert_node(insert_node, td);
             if (insert_result.second) {
                 d1::node_handle_accessor::deactivate(nh);
             }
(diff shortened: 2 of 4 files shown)


include/oneapi/tbb/detail/_concurrent_skip_list.h
```diff
@@ -1197,10 +1235,9 @@ class concurrent_skip_list {
 
     node_allocator_type my_node_allocator;
     key_compare my_compare;
-    random_level_generator_type my_rng;
     atomic_node_ptr my_head_ptr;
-    std::atomic<size_type> my_size;
```
kboyarinov, line 1202: There is also an option to keep `my_size`, reserve a flag bit in it and use it for optimizing consecutive calls to `size()` by saving the result of combine into `my_size` and marking it as "changed" in the insertion.
I did not investigate potential performance effects from CAS operations on my_size, just wanted to save the idea.

include/oneapi/tbb/detail/_concurrent_skip_list.h
```diff
@@ -49,7 +50,28 @@
...
+    using size_type = SizeType;
+
+    skip_list_thread_data() : my_local_size(0) {}
+
+    std::size_t random_level() { return my_rng(); }
+    std::size_t local_size() const { return my_local_size.load(std::memory_order_relaxed); }
+
+    void set_size(size_type size) { my_local_size.store(size, std::memory_order_relaxed); }
+    void increment_size() {
+        my_local_size.store(local_size() + 1, std::memory_order_relaxed);
```
phprus, line 70: Why not  use atomic `std::atomic<>::fetch_add` instead of `load` and `store` pair?

isaevil: My understanding is that is done to avoid locking the cache line, since only one thread writes to this atomic anyway. I think that it would be good to leave a small comment that explains the motivation.

include/oneapi/tbb/concurrent_set.h
```diff
@@ -1,5 +1,6 @@
 /*
-    Copyright (c) 2019-2021 Intel Corporation
+    Copyright (c) 2019-2025 Intel Corporation
+    Copyright (c) 2025 UXL Foundation Contributors
```
aleksei-fedotov: I believe the copyright is outdated hier and in other files.

include/oneapi/tbb/detail/_concurrent_skip_list.h
```diff
@@ -1197,10 +1243,9 @@ class concurrent_skip_list {
 
     node_allocator_type my_node_allocator;
     key_compare my_compare;
-    random_level_generator_type my_rng;
     atomic_node_ptr my_head_ptr;
-    std::atomic<size_type> my_size;
     std::atomic<size_type> my_max_height;
+    mutable ets_type my_ets;
```
aleksei-fedotov: Is `mutable` needed here because of `ets::combine` method being non-const?

include/oneapi/tbb/detail/_concurrent_skip_list.h
```diff
@@ -276,6 +298,8 @@ class concurrent_skip_list {
     using const_pointer = typename allocator_traits_type::const_pointer;
 
     using random_level_generator_type = typename container_traits::random_level_generator_type;
+    using thread_data_type = skip_list_thread_data<random_level_generator_type, size_type>;
+    using ets_type = tbb::enumerable_thread_specific<thread_data_type>;
```
aleksei-fedotov: Why should not this be the ETS with `ets_suspend_aware`?

include/oneapi/tbb/detail/_concurrent_skip_list.h
```diff
@@ -1179,11 +1222,14 @@ class concurrent_skip_list {
...
-
         swap_atomics_relaxed(my_head_ptr, other.my_head_ptr);
-        swap_atomics_relaxed(my_size, other.my_size);
         swap_atomics_relaxed(my_max_height, other.my_max_height);
+
+        size_type total_size = size();
+        my_ets.clear();
+        my_ets.local().set_size(other.size());
+        other.my_ets.clear();
+        other.my_ets.local().set_size(total_size);
```
aleksei-fedotov, line 1243: Looking at the implementation of the ETS, it seems we can relatively easy re-use its internal_swap member function here. I suggest introducing at least an internal helper function for now (at most consider in the future or even add now public `ets::swap`), which would call that `ets::internal_swap` member function. I am not sure why there is no ets::swap public member function.

include/oneapi/tbb/detail/_concurrent_skip_list.h
```diff
@@ -744,8 +779,9 @@ class concurrent_skip_list {
         my_head_ptr.store(other.my_head_ptr.load(std::memory_order_relaxed), std::memory_order_relaxed);
         other.my_head_ptr.store(nullptr, std::memory_order_relaxed);
 
-        my_size.store(other.my_size.load(std::memory_order_relaxed), std::memory_order_relaxed);
-        other.my_size.store(0, std::memory_order_relaxed);
+        my_ets.clear();
+        my_ets.local().set_size(other.size());
+        other.my_ets.clear();
```
aleksei-fedotov, line 790: The helper swap function can also be reused here.

include/oneapi/tbb/detail/_concurrent_skip_list.h
```diff
@@ -1047,7 +1087,9 @@ class concurrent_skip_list {
                 prev_nodes[level]->set_next(level, erase_node->next(level));
                 erase_node->set_next(level, nullptr);
             }
-            my_size.fetch_sub(1, std::memory_order_relaxed);
+            size_type total_size = size();
+            my_ets.clear();
+            my_ets.local().set_size(total_size - 1);
```
aleksei-fedotov, line 1101: Why cannot we just `decrement_size()`, similarly to what we do in the `internal_insert_node`?

include/oneapi/tbb/detail/_concurrent_skip_list.h
aleksei-fedotov: Definitely, the comment will be useful.

kboyarinov: Added the comment

include/oneapi/tbb/concurrent_set.h
kboyarinov: Fixed

include/oneapi/tbb/detail/_concurrent_skip_list.h
kboyarinov: Yes, added the comment

include/oneapi/tbb/detail/_concurrent_skip_list.h
kboyarinov: Thanks, I agree that it will be more flexible from the user standpoint. 
Applied to check performance effects.

include/oneapi/tbb/detail/_concurrent_skip_list.h
kboyarinov: In theory, we can use `internal_swap`, but the logic above also "shrinks" both `*this` and `other`'s ets. 
I.e., if there were several threads inserting to `*this` and then `swap` or `operator=` was called, `size()` calculates the total size and assigns it to calling thread's `local`. 
Therefore, after the operation the state will be equivalent to inserting all elements from a single thread that simplifies further size calculations.

include/oneapi/tbb/detail/_concurrent_skip_list.h
kboyarinov: Same "shrink" optimization takes place here.

include/oneapi/tbb/detail/_concurrent_skip_list.h
kboyarinov: Same logic here - `decrement_size` will decrement only the calling thread's local size, while the code above "shrinks" the total size to a single thread.
Also, decrementing only the local size will require it to be signed, since the thread that never inserted to the map can erase elements.

include/oneapi/tbb/detail/_concurrent_skip_list.h
aleksei-fedotov: And what if the application does not compute the size() at all? What if the application swaps the containers and the threads continue inserting items in *these?
I mean, in practice, the question is how often these operations are used compared to gained speed for size() calculations, which also incurs memory re-allocations and type construction/destruction in the concurrent_vector underneath, btw.
So, ideally there should be proper analysis done and some heuristics used to understand what path to take. At least, I would suggest placing a comment for now stating explicitly the reasons. Maybe, something like: _Assume gain from optimizing the future size calculations prevails subsequent use of swapped containers_

include/oneapi/tbb/detail/_concurrent_skip_list.h
aleksei-fedotov: Same logic here as well. Should we buy skis or continue renting them? Some comment here would also be helpful stating why the `decrement_size` is not used.

include/oneapi/tbb/detail/_concurrent_skip_list.h
aleksei-fedotov: Here as well.

include/oneapi/tbb/detail/_concurrent_skip_list.h
kboyarinov: I agree that we should investigate it in more details. Added a TODO comment

include/oneapi/tbb/detail/_concurrent_skip_list.h
kboyarinov: Added a comment

include/oneapi/tbb/detail/_concurrent_skip_list.h
kboyarinov: Added a comment

include/oneapi/tbb/detail/_concurrent_skip_list.h
```diff
@@ -1091,6 +1093,9 @@ class concurrent_skip_list {
                 prev_nodes[level]->set_next(level, erase_node->next(level));
                 erase_node->set_next(level, nullptr);
             }
+
+            // Shrink all thread-local sizes into the calling thread's size and decrement
+            // TODO: investigate if performance gain of such and optimization is justified
```
aleksei-fedotov:
```suggestion
            // TODO: investigate if performance gain of this optimization is justified
```

aleksei-fedotov (commented):
I think it is good enough in general.

Found just a small oversight in the comment.

kboyarinov:
Latest performance measurements (LibCuckoo universal benchmark) confirmed the performance improvement on the latest version of the patch. So, I guess it can be merged.

PR #1805, opened 2025-08-05
Specialize get_task for the case of no isolation

akukanov:
### Description 
After introducing in #1779 a couple of additional operations in `get_task` (which is on the hot path in the task scheduler) that are only useful when the scheduler runs in the task isolation mode, I thought that maybe this function deserves two separate implementations, with and without isolation. This PR elaborates on that idea.

Additionally, it unifies the handling of proxy tasks in `get_task` and `steal_task` into a single routine `task_proxy::try_extract_task_from`.

### Type of change
- Refactoring/optimization

### Tests
- [x] not needed - covered by the existing tests

### Documentation
- [x] not needed

- [x] No

### Other information
The PR is developed on top of #1779, and so should either be merged afterwards or significantly changed if the base PR is rejected.

### DIFF
--- a/src/tbb/arena_slot.cpp
+++ b/src/tbb/arena_slot.cpp
@@ -27,7 +27,61 @@ namespace r1 {
 // Arena Slot
 //------------------------------------------------------------------------
 
-d1::task* arena_slot::get_task(execution_data_ext& ed, isolation_type isolation) {
+d1::task* arena_slot::get_task(execution_data_ext& ed, isolation_type isolation, std::false_type) {
+    suppress_unused_warning(isolation);
+    __TBB_ASSERT(isolation == no_isolation, nullptr);
+    __TBB_ASSERT(is_task_pool_published(), nullptr);
+    bool all_tasks_checked = false;
+    // The current task position in the task pool.
+    std::size_t T0 = tail.load(std::memory_order_relaxed);
+    // The bounds of available tasks in the task pool. H0 is only used when the head bound is reached.
+    std::size_t H0 = (std::size_t)-1, T = T0;
+    d1::task* result = nullptr;
+    do {
+        // The full fence is required to sync the store of `tail` with the load of `head` (write-read barrier)
+        T = --tail;
+        // The acquire load of head is required to guarantee consistency of our task pool
+        // when a thief rolls back the head.
+        if ( (std::intptr_t)( head.load(std::memory_order_acquire) ) > (std::intptr_t)T ) {
+            acquire_task_pool();
+            H0 = head.load(std::memory_order_relaxed);
+            if ( (std::intptr_t)H0 > (std::intptr_t)T ) {
+                // The thief has not backed off - nothing to grab.
+                __TBB_ASSERT( H0 == head.load(std::memory_order_relaxed)
+                    && T == tail.load(std::memory_order_relaxed)
+                    && H0 == T + 1, "victim/thief arbitration algorithm failure" );
+                reset_task_pool_and_leave();
+                all_tasks_checked = true;
+                break /*do-while*/;
+            } else if ( H0 == T ) {
+                reset_task_pool_and_leave();
+                all_tasks_checked = true;
+            } else {
+                // Release task pool if there are still some tasks.
+                // After the release, the tail will be less than T, thus a thief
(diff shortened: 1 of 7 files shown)


src/tbb/mailbox.h
```diff
@@ -71,17 +71,37 @@ struct task_proxy : public d1::task {
...
         }
-        // Proxied task has already been claimed from another proxy location.
+        // The proxied task has already been claimed from the other location.
         __TBB_ASSERT( task_and_tag.load(std::memory_order_relaxed) == from_bit, "Empty proxy cannot contain non-zero task pointer" );
         return nullptr;
     }
 
+    //! Checks if a given task is a proxy, then either extracts the real task or frees the proxy.
+    template<bool B = false>
+    static task* try_extract_task_from ( task* t, execution_data_ext& ed, std::integral_constant<bool, B> stolen = {} ) {
```
aleksei-fedotov: The line is too wide. Let's try keeping them not longer than ~100 characters.
```suggestion
    static task* try_extract_task_from ( task* t, execution_data_ext& ed,
                                         std::integral_constant<bool, B> stolen = {} ) {
```

src/tbb/mailbox.h
```diff
@@ -71,17 +71,37 @@ struct task_proxy : public d1::task {
...
+        __TBB_ASSERT(t && !is_poisoned(t), "Not a valid task pointer");
```
aleksei-fedotov, line 87: `try_extract_task_from` seems to be always called from the places where `t != nullptr`. Perhaps, this function could accept a reference on the task instead. Thus, the dereferences would not be needed anywhere in this function, and the `nullptr` check would be naturally integrated into the function contract.

src/tbb/task_dispatcher.h
```diff
@@ -363,7 +366,7 @@ d1::task* task_dispatcher::local_wait_for_all(d1::task* t, Waiter& waiter ) {
                     break;
                 }
                 // Retrieve the task from local task pool
-                if (t || (slot.is_task_pool_published() && (t = slot.get_task(ed, isolation)))) {
+                if (t || (slot.is_task_pool_published() && (t = slot.get_task(ed, isolation, use_isolation)))) {
```
aleksei-fedotov: Probably, make this part a bit more readable?
```suggestion
                if ( !t && slot.is_task_pool_published() ) {
                    t = slot.get_task(ed, isolation, use_isolation);
                }
                if ( t ) {
```

src/tbb/arena_slot.cpp
```diff
@@ -27,7 +27,61 @@ namespace r1 {
 // Arena Slot
 //------------------------------------------------------------------------
 
-d1::task* arena_slot::get_task(execution_data_ext& ed, isolation_type isolation) {
+d1::task* arena_slot::get_task(execution_data_ext& ed, isolation_type isolation, std::false_type) {
```
aleksei-fedotov: Consider adding a comment describing what the last parameter mean for function definitions as well.

src/tbb/mailbox.h
akukanov: It does not feel right to me. It seems like trading two pointer dereferences for getting a task address just as many times.
In general, these parts of the scheduler mostly deal with task pointers as data, not with tasks as objects. One can argue that operations with `task_proxy` need to access the object state and so to dereference the pointer - and that's true but this logic is encapsulated in `task_accessor`, and dereferencing is done when its functions are called. For this function, I prefer the semantics of "take what appears to be a valid task pointer, and return the pointer to the actual task. unwrapped from proxy if necessary".

src/tbb/task_dispatcher.h
akukanov: I instead moved the `is_task_pool_published` check into `get_task`.

src/tbb/arena_slot.cpp
akukanov: Done.

src/tbb/mailbox.h
akukanov: Let me just remove a couple of extra spaces, so that it fits into 120 symbols. Given that in this very file there are lines with more than 140 symbols, and no consistent style for spaces around parentheses, I think that should be enough.

PR #1806, opened 2025-08-06
Send ITT task end notifications with less delay

akukanov:
### Description 
The points where TBB worker threads issue ITT "task end" notifications (on the detection of a new task group context and on the leave of the task dispatch loop) may lead to significant delays with those, sometimes way after completion of the algorithm the "tasks" are supposed to indicate. That leads to skewed results in profiling tools that "consume" the ITT calls.

The patch proposes a change to issue the notification earlier, namely after a few unsuccessful attempts to find a non-local task (via stealing, etc.). The threshold is set to 2 empirically; we can try finding a "better" value later.

### Type of change

### Tests
- [x] not needed

### Documentation
- [x] not needed

- [x] No

### Other information
There is also an idea to set the threshold to 0 for algorithms invoked with a static_partitioner, where the expected number of tasks per thread is 1. However the right way to deliver this information to the scheduler is not yet clear; and it might require layout changes in `task_group_context`. Therefore this PR does not implement that idea.

### DIFF
--- a/src/tbb/scheduler_common.h
+++ b/src/tbb/scheduler_common.h
@@ -1,5 +1,6 @@
 /*
-    Copyright (c) 2005-2024 Intel Corporation
+    Copyright (c) 2005-2025 Intel Corporation
+    Copyright (c) 2025 UXL Foundation Contributors
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
--- a/src/tbb/task_dispatcher.h
+++ b/src/tbb/task_dispatcher.h
@@ -1,5 +1,6 @@
 /*
     Copyright (c) 2020-2025 Intel Corporation
+    Copyright (c) 2025 UXL Foundation Contributors
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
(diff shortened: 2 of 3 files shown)


src/tbb/scheduler_common.h
```diff
@@ -148,6 +149,10 @@ struct task_accessor {
     context execution **/
 template <bool report_tasks>
 class context_guard_helper {
+    // The number of unsuccessful attempts to retrieve a non-local task, after which ITT_TASK_END notification
+    // should be sent. If the notification is postponed for too long, data in profiling tools might get skewed.
+    static constexpr int itt_task_end_threshold = 2;
```
vossmjp, line 154: What is the way to choose the best value for this threshold? The overhead of these calls are only seen if there is an ITT library successfully loaded, right? So we'd only see additional runtime overhead when traces are actively being collected.

tovinkere: That is correct - the extra overhead is only realized when a collector is active. As far as the thresholds go, I have tested with 0, 1, 2 and saw no difference in the test cases I ran. TBB is generally shown as active for longer with higher thresholds, so any analytics built to tune TBB algorithms will start to introduce error in the analysis. However, the differences are negligible from 0 to 2.

vossmjp: If a threshold of 0 is ok, we don't need task_search_count at all.

akukanov: It would need to be tested on some heavily imbalanced workloads and/or producer/consumer patterns where stealing prevails. Without that, I think a low value like 2 might be better,

PR #1808, opened 2025-08-07
Custom assertion handler

### LINKED ISSUE #1258: set_assertion_handler
tbb::set_assertion_handler was removed as from release 2021.1. Is there any other way to set an assertion handler?

tbb/detail/_assert.h still contains:

```
//! Process an assertion failure.
/** Normally called from __TBB_ASSERT macro.
  If assertion handler is null, print message for assertion failure and abort.
  Otherwise call the assertion handler. */
TBB_EXPORT void __TBB_EXPORTED_FUNC assertion_failure(const char* location, int line, const char* expression, const char* comment);
```

dnmokhov:
Add an RFC describing reinstating custom assertion handlers, a missing TBB 2020 feature.

Fixes #1258.

### DIFF
--- a/rfcs/proposed/assertion_handler/README.md
+++ b/rfcs/proposed/assertion_handler/README.md
@@ -1,0 +1,208 @@
+# Custom Assertion Handler
+
+## Introduction
+
+In 2021, with the transition from TBB 2020 to the first release of oneTBB, the custom assertion handler feature
+was removed and has since been requested several times (e.g., https://github.com/uxlfoundation/oneTBB/issues/1258).
+
+OneTBB currently provides assertion functionality for debugging and error reporting purposes through the
+`assertion_failure` function. However, applications using oneTBB may have their own error handling, logging, or
+reporting requirements that differ from the default assertion behavior (which prints to stderr and calls
+`std::abort`).
+
+The motivation for this proposal includes:
+
+- **Improved integration with application error handling**: Applications often have sophisticated error handling,
+  logging frameworks, or crash reporting systems that they want to integrate with oneTBB assertions.
+- **Custom recovery mechanisms**: Some applications may want to attempt recovery or cleanup operations before
+  terminating, rather than immediately calling `std::abort`.
+- **Enhanced debugging capabilities**: Development environments may want to integrate oneTBB assertions with their
+  debugging tools, IDE breakpoints, or custom diagnostic systems.
+- **Production environment requirements**: Production systems may need to log assertion failures to specific
+  locations, send alerts, or perform graceful shutdowns rather than immediate termination.
+
+Currently, when an assertion fails in oneTBB, users have no control over the behavior - the library always prints
+to stderr and calls `std::abort`. This rigid behavior can be problematic for applications with specific
+requirements for error handling and diagnostics.
+
+## Proposal
+
+We propose adding a custom assertion handler mechanism that allows applications to set their own assertion
+handling functions. This proposal is modeled on both the standard library's `std::set_terminate` and
+`std::get_terminate` functions, and the assertion handler API that existed in TBB 2020, providing similar API
+design, functionality, and thread safety guarantees.
+
+The design parallels both:
+- **Standard Library Pattern**: `std::set_terminate` allows setting a custom termination handler and returns the
+  previous handler, with thread-safe atomic operations for handler storage
+- **TBB 2020 Compatibility**: The same `assertion_handler_type` and `set_assertion_handler` function signatures
+  that existed in TBB 2020, enabling seamless migration for applications that previously used this functionality
+
+This approach provides both familiarity for developers accustomed to standard library patterns and direct
+compatibility for applications migrating from TBB 2020 to oneTBB.
+
+### New Public API
+
+The proposal adds the following new functions to the public API, using the same signatures as TBB 2020 for
+compatibility:
+
+```cpp
+namespace tbb {
+    // Type alias for assertion handler function pointer - identical to TBB 2020
+    using assertion_handler_type = void(*)(const char* location, int line,
+                                           const char* expression, const char* comment);
+
+    //! Set assertion handler and return its previous value.
+    //! If new_handler is nullptr, resets to the default handler.
+    //! Uses the same signature as TBB 2020 for migration compatibility.
+    assertion_handler_type set_assertion_handler(assertion_handler_type new_handler) noexcept;
+
+    //! Return the current assertion handler.
+    //! New function not present in TBB 2020, following std::get_terminate pattern.
+    assertion_handler_type get_assertion_handler() noexcept;
+}
+```
+
+Applications that used the custom assertion handler in TBB 2020 can migrate to this proposal with minimal changes.
+
+The only difference is the addition of `get_assertion_handler` for retrieving the current handler, which
+follows the `std::get_terminate` pattern but was not present in TBB 2020.
+
+### Usage Examples
+
+#### Adding Custom Steps
+
+```cpp
+auto default_handler = tbb::get_assertion_handler();
+
+void wrapper_assertion_handler(const char* location, int line,
+                               const char* expression, const char* comment) {
+    // Perform a custom step before calling default handler
+    // ... (e.g., log to a file, notify system, custom cleanup, etc.)
+    default_handler(location, line, expression, comment);
+}
+
+int main() {
+    // Set custom handler (identical to TBB 2020 usage)
+    tbb::set_assertion_handler(wrapper_assertion_handler);
+
+    // Use oneTBB normally - any assertion failures will use custom handler
+    // ...
+
+    // Restore previous handler if needed
+    tbb::set_assertion_handler(default_handler);
+}
+```
+
+#### Integration with Logging Framework
+
+```cpp
+void logging_assertion_handler(const char* location, int line,
+                               const char* expression, const char* comment) {
+    spdlog::critical("TBB Assertion Failed: {} at {}:{} - {}",
+                     expression, location, line, comment ? comment : "");
+
+    // Flush logs before terminating
+    spdlog::shutdown();
+    std::abort();
+}
+```
+
+#### Development/Testing Handler
+
+```cpp
+void test_assertion_handler(const char* location, int line,
+                            const char* expression, const char* comment) {
+    // In testing environments, throw exception instead of aborting
+    // to allow test frameworks to catch and handle assertion failures
+    std::string msg = std::string("TBB Assertion: ") + expression +
+                      " at " + location + ":" + std::to_string(line);
+    if (comment) {
+        msg += " - " + std::string(comment);
+    }
+    throw std::runtime_error(msg);
+}
+```
+
+### Proposed Implementation Strategy
+
+#### Core Implementation Approach
+
+The implementation will follow the same patterns used by `std::set_terminate` and `std::get_terminate`, while
+maintaining compatibility with TBB 2020:
+
+1. **Handler Storage**: Add a global static atomic variable to store the current assertion handler function pointer,
+   ensuring thread-safe access across the library, similar to how the standard library stores the terminate handler.
+
+2. **Handler Invocation**: Modify `assertion_failure` to call the currently set handler instead of directly
+   executing the default logic, mirroring how `std::terminate` calls the registered handler.
+
+3. **TBB 2020 Compatibility**: Use identical function signatures to ensure existing TBB 2020 code works without
+   modification.
+
+#### Thread Safety Design
+
+Following the `std::set_terminate`/`std::get_terminate` model, the implementation will use atomic operations with
+appropriate memory ordering (`memory_order_acq_rel` for exchanges, `memory_order_acquire` for loads), consistent with
+standard library implementations. This approach will ensure multiple threads can safely call `set_assertion_handler`
+and `get_assertion_handler` concurrently while guaranteeing that handler changes are visible across all threads
+without requiring additional synchronization.
+
+Only the default handler will maintain thread-safe assertion execution using the existing `atomic_do_once` pattern
+to ensure assertions are reported exactly once per failure. For custom assertion handlers, it is the user's
+responsibility to ensure that their handler maintains appropriate thread safety if multiple threads might trigger
+the same assertion simultaneously. This behavior is consistent with TBB 2020, where custom handlers were also
+responsible for their own thread safety.
+
+#### Backward Compatibility Preservation
+
+The default behavior will remain unchanged, so existing applications will continue to work exactly as before. There
+will be no changes to existing assertion macros or calling conventions. Applications using custom assertion handlers
+in TBB 2020 will be able to use the same code with little to no modification, ensuring seamless migration to oneTBB.
+
+#### Null Handler Protection
+
+Following both the `std::set_terminate` pattern and TBB 2020 behavior, the implementation will provide protection
+against null handler pointers. When `set_assertion_handler(nullptr)` is called, the function will reset to the default
+handler rather than allowing null dereference. This approach provides a safe way to restore default behavior, similar
+to how `std::set_terminate` handles null pointers in the standard library.
+
+### Platform Considerations
+
+The implementation will add new `set_assertion_handler` and `get_assertion_handler` public entry points to the oneTBB
+shared library across all supported platforms. We will use function pointers instead of `std::function` to avoid
+additional C++ runtime dependencies. Even though `std::function` provides type safety and supports lambda captures,
+it also introduces additional complexity and potential performance overhead. Using function pointers maintains ABI
+stability across different C++ standards and compilers and is consistent with both the standard library approach and
+the TBB 2020 implementation.
+
+### Performance Impact Analysis
+
+The proposed implementation will have minimal performance impact, following the `std::set_terminate` model:
+
+- **Normal Execution**: No overhead on normal execution paths
+- **Assertion Failure Path**: One additional atomic load, which only occurs during actual assertion failures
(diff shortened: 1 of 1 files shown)


dnmokhov:
A prototype implementation based on this RFC is available at [dev/dnmokhov/assert](../tree/dev/dnmokhov/assert).

e4lam:
To be honest, I wanted to stop reading once I got to full TBB 2020 compatibility. 😄 

But to answer the regardless anyways:
1. Re-entrancy: I don't recall what TBB 2020 did anymore. Our custom assertion handlers have always handle re-entrancy themselves.
2. Handler Context: I think this needlessly complicates things. Most custom assertion handlers already have this and `std::stacktrace` can be used starting in C++23. When the assertions trigger, it's not like we're in a signal handler so I don't think it makes sense to try to conjure up something here.
3. Global vs Per-Arena: We've never found the need for per-arena handlers. Per-arena handlers would mean getting some sort of context I think, which again over-complicates things.
4. Experimental Status: Just go stable. This assertion interface has been around for as long as I've used TBB. Our subversion log indicates that the code for using this API was never touched since it was initially committed on June 27, 2012.
5. Thread Safety Documentation: It never hurts to have more documentation but don't let that slow you down from releasing it. A bare minimum of the old TBB documentation on this would suffice. You can always add more docs later. I don't understand the question on migration there. The thread safety design section already said it's the same as TBB 2020.

e4lam:
@dnmokhov Could you explain the what effects doe the new TBBind library has in comparison to TBB 2020? For someone that is using both tbb and tbbmalloc, does this allow assertion failures in tbbmalloc to go to the custom handler?

rfcs/proposed/assertion_handler/README.md
```diff
@@ -0,0 +1,228 @@
...
+
+Currently, when an assertion fails in oneTBB, users have no control over the behavior - the library always prints
+to stderr and calls `std::abort`. This rigid behavior can be problematic for applications with specific
+requirements for error handling and diagnostics.
+
+## Proposal
+
+We propose adding a custom assertion handler mechanism that allows applications to set their own assertion
+handling functions. This proposal is modeled on both the standard library's `std::set_terminate` and
+`std::get_terminate` functions, and the assertion handler API that existed in TBB 2020, providing similar API
```
kboyarinov: I would not say the `std::set_terminate` and `std::get_terminate` are modelling the proposal. I would state that they are similar in semantic, but obviously not identical.

rfcs/proposed/assertion_handler/README.md
```diff
@@ -0,0 +1,228 @@
...
+This approach provides both familiarity for developers accustomed to standard library patterns and direct
+compatibility for applications migrating from TBB 2020 to oneTBB.
+
+### New Public API
+
+The proposal adds the following new functions to the public API, using the same signatures as TBB 2020 for
+compatibility:
+
+```cpp
+namespace tbb {
```
kboyarinov, line 57: We will need to specify in which header these APIs are defined. Or explicitly say that they should be available through any public TBB header.

rfcs/proposed/assertion_handler/README.md
```diff
@@ -0,0 +1,228 @@
...
+    //! Uses the same signature as TBB 2020 for migration compatibility.
+    assertion_handler_type set_assertion_handler(assertion_handler_type new_handler) noexcept;
+
+    //! Return the current assertion handler.
+    //! New function not present in TBB 2020, following std::get_terminate pattern.
+    assertion_handler_type get_assertion_handler() noexcept;
+}
+```
+
+Applications that used the custom assertion handler in TBB 2020 can migrate to this proposal with minimal changes.
```
kboyarinov: As far as I understand, even the minimal changes are not needed for migration since the API available in TBB 2020 is unchanged.

rfcs/proposed/assertion_handler/README.md
```diff
@@ -0,0 +1,228 @@
...
+
+The only difference is the addition of `get_assertion_handler` for retrieving the current handler, which
+follows the `std::get_terminate` pattern but was not present in TBB 2020.
```
kboyarinov: It seems like this paragraph duplicates the comment in the code snippet above.

rfcs/proposed/assertion_handler/README.md
```diff
@@ -0,0 +1,228 @@
...
+
+### Usage Examples
```
kboyarinov: I propose to move the examples section to the end of the document to focus on the proposed semantics first.

rfcs/proposed/assertion_handler/README.md
```diff
@@ -0,0 +1,228 @@
...
+
+#### Thread Safety Design
+
+Following the `std::set_terminate`/`std::get_terminate` model, the implementation will use atomic operations with
+appropriate memory ordering (`memory_order_acq_rel` for exchanges, `memory_order_acquire` for loads), consistent with
+standard library implementations. This approach will ensure multiple threads can safely call `set_assertion_handler`
+and `get_assertion_handler` concurrently while guaranteeing that handler changes are visible across all threads
+without requiring additional synchronization.
+
+Only the default handler will maintain thread-safe assertion execution using the existing `atomic_do_once` pattern
```
kboyarinov:
```suggestion
The default handler will maintain thread-safe assertion execution using the existing `atomic_do_once` pattern
```

rfcs/proposed/assertion_handler/README.md
```diff
@@ -0,0 +1,228 @@
...
+to ensure assertions are reported exactly once per failure. For custom assertion handlers, it is the user's
+responsibility to ensure that their handler maintains appropriate thread safety if multiple threads might trigger
```
kboyarinov: I would also add somewhere that the assertion handler is invoked on each assertion failure, even if it happen concurrently in several threads. And the default handler uses atomic_do_once .

rfcs/proposed/assertion_handler/README.md
```diff
@@ -0,0 +1,228 @@
...
+the same assertion simultaneously. This behavior is consistent with TBB 2020, where custom handlers were also
+responsible for their own thread safety.
+
+#### Backward Compatibility Preservation
+
+The default behavior will remain unchanged, so existing applications will continue to work exactly as before. There
```
kboyarinov: It is unclear what is "default behavior" in this context

rfcs/proposed/assertion_handler/README.md
```diff
@@ -0,0 +1,228 @@
...
+it also introduces additional complexity and potential performance overhead. Using function pointers maintains ABI
+stability across different C++ standards and compilers and is consistent with both the standard library approach and
+the TBB 2020 implementation.
+
+### Performance Impact Analysis
+
+The proposed implementation will have minimal performance impact, following the `std::set_terminate` model:
+
+- **Normal Execution**: No overhead on normal execution paths
+- **Assertion Failure Path**: One additional atomic load, which only occurs during actual assertion failures
```
kboyarinov: Is not it obvious that the assertion failure is happens on the "Assertion Failure Path"?
```suggestion
- **Assertion Failure Path**: One additional atomic load
```

rfcs/proposed/assertion_handler/README.md
```diff
@@ -0,0 +1,228 @@
...
+- **Memory Overhead**: One atomic pointer per process
+- **Binary Size**: Minimal increase due to additional exported functions
+
+## Open Questions
+
+1. **Nested Assertions**: How should the system behave if a custom assertion handler itself triggers an assertion?
+   Should there be protection against infinite recursion, similar to how `std::terminate` handles recursive calls?
+   How did TBB 2020 handle this scenario?
+
+2. **Handler Context**: Would it be valuable to provide additional context to handlers, such as thread ID or stack
```
kboyarinov: Is it really needed? It seems like both thread ID and stack trace are available without additional function arguments through `std::this_thread::get_id()` and `std::stacktrace::current()`.

rfcs/proposed/assertion_handler/README.md
```diff
@@ -0,0 +1,228 @@
...
+
+3. **Global vs Per-Arena**: Should assertion handlers be global or configurable per task arena? The
+   `std::terminate` model is global, and TBB 2020 used a global handler, which argues for consistency, but
+   per-arena might provide more flexibility.
+
+4. **Experimental Status**: Should this feature initially be released as experimental/preview API to gather user
+   feedback before stabilizing the interface, or should it be considered stable from the start given the proven
+   TBB 2020 design?
+
+5. **Thread Safety Documentation**: Should the documentation provide specific guidance and examples for
```
kboyarinov: I think such a documentation will be helpful in any case. More documentation never makes things worse:)

dnmokhov:
> @dnmokhov Could you explain the what effects doe the new TBBind library has in comparison to TBB 2020?

TBBBind enables setting constraints on [hybrid CPU and NUMA](https://uxlfoundation.github.io/oneTBB/GSG/next_steps.html#enable-hybrid-cpu-and-numa-support).

> For someone that is using both tbb and tbbmalloc, does this allow assertion failures in tbbmalloc to go to the custom handler?

I propose that stays the same as in TBB 2020, i.e., tbbmalloc does not use the custom handler. The main reason is tbbmalloc's portability, so we don't want to introduce new dependencies (on tbb, on C++ runtime, etc.) I provided some explanation here: https://github.com/uxlfoundation/oneTBB/tree/dev/dnmokhov/rfc-assertion-handler/rfcs/proposed/assertion_handler#tbbbind-integration

rfcs/proposed/assertion_handler/README.md
dnmokhov: Changed.

rfcs/proposed/assertion_handler/README.md
dnmokhov: Changed.

rfcs/proposed/assertion_handler/README.md
dnmokhov: Removed.

rfcs/proposed/assertion_handler/README.md
dnmokhov: Moved.

rfcs/proposed/assertion_handler/README.md
dnmokhov: Changed.

rfcs/proposed/assertion_handler/README.md
dnmokhov: Added to the beginning of the paragraph.

rfcs/proposed/assertion_handler/README.md
dnmokhov: Changed to clarify.

rfcs/proposed/assertion_handler/README.md
dnmokhov: Removed.

rfcs/proposed/assertion_handler/README.md
dnmokhov: I guess we have enough feedback to confirm that this is not needed. Removed.

rfcs/proposed/assertion_handler/README.md
dnmokhov: Moved from an open question to part of the proposal (last paragraph in the thread safety section).

rfcs/proposed/assertion_handler/README.md
dnmokhov: I like the idea of creating `assert.h`, which would simply include the current `detail/_assert.h`. This would preserve compatibility for the current uses of `__TBB_ASSERT` and would still provide a lightweight export of the handler feature. (The include dependencies of `detail/_assert.h` are minimal: only `detail/_config.h` and `detail/_export.h`.)

How does that sound?

rfcs/proposed/assertion_handler/README.md
```diff
@@ -0,0 +1,233 @@
...
+
+3. **Experimental Status**: Should this feature initially be released as experimental/preview API to gather user
+   feedback before stabilizing the interface, or should it be considered stable from the start given the proven
+   TBB 2020 design?
+
+4. **TBBBind Integration Strategy**: Should TBBBind be modified to use the custom assertion handler through a
+   dependency on oneTBB, or should it be left unchanged like TBBMalloc? While making TBBBind dependent on oneTBB
+   would provide unified assertion handling, it may be acceptable to leave TBBBind as-is to maintain simpler
+   dependencies and follow the same pattern as TBBMalloc. What are the trade-offs between consistency of assertion
+   handling versus architectural simplicity?
```
akukanov: I think this question is now decided in favor of adding support to tbbbind, and can be removed.

rfcs/proposed/assertion_handler/README.md
```diff
@@ -0,0 +1,233 @@
...
+
+## Open Questions
+
+1. **Nested Assertions**: How should the system behave if a custom assertion handler itself triggers an assertion?
+   Should there be protection against infinite recursion, similar to how `std::terminate` handles recursive calls?
+   How did TBB 2020 handle this scenario?
+
+2. **Global vs Per-Arena**: Should assertion handlers be global or configurable per task arena? The
+   `std::terminate` model is global, and TBB 2020 used a global handler, which argues for consistency, but
+   per-arena might provide more flexibility.
```
akukanov: There seems to be no interest in per-arena handlers, and given the noreturn behavior the extra flexibility is hardly ever needed. I vote for closing this question.

rfcs/proposed/assertion_handler/README.md
```diff
@@ -0,0 +1,233 @@
...
+    }
+    throw std::runtime_error(msg);
+}
+```
+
+## Open Questions
+
+1. **Nested Assertions**: How should the system behave if a custom assertion handler itself triggers an assertion?
+   Should there be protection against infinite recursion, similar to how `std::terminate` handles recursive calls?
+   How did TBB 2020 handle this scenario?
```
akukanov: As far as I can tell, an assertion in a custom handler would not go down the TBB assertion path, or would it? Certainly the standard C/C++ assertions should not trigger a recursion (and that can be tested I think). I do not think it is a real problem; but if it is, then there must be an example to review and discuss.

rfcs/proposed/assertion_handler/README.md
```diff
@@ -0,0 +1,233 @@
...
+
+2. **Global vs Per-Arena**: Should assertion handlers be global or configurable per task arena? The
+   `std::terminate` model is global, and TBB 2020 used a global handler, which argues for consistency, but
+   per-arena might provide more flexibility.
+
+3. **Experimental Status**: Should this feature initially be released as experimental/preview API to gather user
+   feedback before stabilizing the interface, or should it be considered stable from the start given the proven
+   TBB 2020 design?
```
akukanov: The proposed API and operation model seem solid from the start and are mostly proven in the past versions of TBB. And seemingly there is no big open questions to explore either, so going experimental does not seem necessary.

A question that sounds more interesting for me is whether we should add this functionality to the oneTBB specification or keep it as an implementation-specific extension (and perhaps designate as such, e.g. by moving to a special namespace).

rfcs/proposed/assertion_handler/README.md
```diff
@@ -0,0 +1,233 @@
...
+to how `std::set_terminate` handles null pointers in the standard library.
+
+#### TBBBind Integration
+
+To ensure consistent assertion handling across oneTBB components, the TBBBind shared library will be made dependent on
+the oneTBB shared library. This approach provides unified assertion handling where assertions originating from TBBBind
+will automatically use the custom assertion handler set by the application, creating a consistent user experience.
+
+This dependency addition is acceptable since TBBBind is exclusively used from within the oneTBB shared library and
+does not affect the public API or application dependencies.
```
akukanov: The way in which tbbbind might/should depend on TBB needs to be clarified. According to our latest internal discussions, we incline towards TBB providing TBBbind with a pointer to the assertion function (which internally calls the active handler) - so it will be a "soft" runtime dependency, as opposed to e.g. a "hard" link-time one.

rfcs/proposed/assertion_handler/README.md
```diff
@@ -0,0 +1,233 @@
...
+
+TBBMalloc, on the other hand, will not be changed. It will continue to include and reuse the assertion handling
+implementation file from the core oneTBB library source code, maintaining its own assertion behavior that does not
```
akukanov:
```suggestion
#### No impact on TBBMalloc

TBBMalloc, on the other hand, will not be changed. It will continue to reuse the assertion handling
implementation from the core oneTBB library source code yet maintain its own assertion behavior that does not
```

rfcs/proposed/assertion_handler/README.md
dnmokhov: Removed.

rfcs/proposed/assertion_handler/README.md
dnmokhov: Removed.

rfcs/proposed/assertion_handler/README.md
dnmokhov: It shouldn't go down the TBB assertion path. `__TBB_ASSERT` is internal. And general thread-safety and call-once recommendations will be added to the documentation. Hence, the above is not a real problem.

Removed.

rfcs/proposed/assertion_handler/README.md
dnmokhov: Removed the old question and added one about extension status.

rfcs/proposed/assertion_handler/README.md
dnmokhov: Updated.

kboyarinov (approved):

PR #1815, opened 2025-08-14
Custom assertion handler

### LINKED ISSUE #1258: set_assertion_handler

dnmokhov:
### Description 
Implements #1808 by 
- adding `set_assertion_handler`/`get_assertion_handler` to the API
- modifying TBBBind to use the custom assertion handler through a dependency on the TBB shared library

Fixes #1258.

### Type of change

### Tests
- [ ] not needed

### Documentation
- [x] needs to be updated
- [ ] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/include/oneapi/tbb/detail/_assert.h
+++ b/include/oneapi/tbb/detail/_assert.h
@@ -1,5 +1,6 @@
 /*
     Copyright (c) 2005-2025 Intel Corporation
+    Copyright (c) 2025 UXL Foundation Contributors
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
@@ -32,11 +33,27 @@ namespace r1 {
   Otherwise call the assertion handler. */
 TBB_EXPORT void __TBB_EXPORTED_FUNC assertion_failure(const char *location, int line,
                                                       const char *expression, const char *comment);
+
+using assertion_handler_type = decltype(&assertion_failure);
+
+//! Set assertion handler and return its previous value.
+TBB_EXPORT assertion_handler_type __TBB_EXPORTED_FUNC set_assertion_handler(assertion_handler_type new_handler) noexcept;
+
+//! Return the current assertion handler.
+TBB_EXPORT assertion_handler_type __TBB_EXPORTED_FUNC get_assertion_handler() noexcept;
+
 #if __TBBMALLOC_BUILD
 }} // namespaces rml::internal
 #else
 } // namespace r1
 } // namespace detail
+
+inline namespace v1 {
+using detail::r1::assertion_handler_type;
+using detail::r1::set_assertion_handler;
+using detail::r1::get_assertion_handler;
+} // namespace v1
+
 } // namespace tbb
 #endif
 
--- a/src/tbb/assert_impl.h
+++ b/src/tbb/assert_impl.h
@@ -1,5 +1,6 @@
 /*
-    Copyright (c) 2005-2021 Intel Corporation
+    Copyright (c) 2005-2025 Intel Corporation
+    Copyright (c) 2025 UXL Foundation Contributors
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
@@ -37,43 +38,55 @@ namespace tbb {
 namespace detail {
 namespace r1 {
 #endif
-// TODO: consider extension for formatted error description string
-static void assertion_failure_impl(const char* location, int line, const char* expression, const char* comment) {
 
-    std::fprintf(stderr, "Assertion %s failed (located in the %s function, line in file: %d)\n",
-        expression, location, line);
-
-    if (comment) {
-        std::fprintf(stderr, "Detailed description: %s\n", comment);
-    }
-#if _MSC_VER && _DEBUG
-    if (1 == _CrtDbgReport(_CRT_ASSERT, location, line, "tbb_debug.dll", "%s\r\n%s", expression, comment?comment:"")) {
-        _CrtDbgBreak();
-    } else
-#endif
-    {
-        std::fflush(stderr);
-        std::abort();
-    }
-}
-
-// Do not move the definition into the assertion_failure function because it will require "magic statics".
-// It will bring a dependency on C++ runtime on some platforms while assert_impl.h is reused in tbbmalloc 
+// Do not move the definition into the assertion_failure_impl function because it will require "magic statics".
+// It will bring a dependency on C++ runtime on some platforms while assert_impl.h is reused in tbbmalloc
 // that should not depend on C++ runtime
 static std::atomic<tbb::detail::do_once_state> assertion_state;
 
-void __TBB_EXPORTED_FUNC assertion_failure(const char* location, int line, const char* expression, const char* comment) {
+// TODO: consider extension for formatted error description string
+static void assertion_failure_impl(const char* location, int line, const char* expression, const char* comment) {
 #if __TBB_MSVC_UNREACHABLE_CODE_IGNORED
     // Workaround for erroneous "unreachable code" during assertion throwing using call_once
     #pragma warning (push)
     #pragma warning (disable: 4702)
 #endif
-    // We cannot use std::call_once because it brings a dependency on C++ runtime on some platforms 
+    // We cannot use std::call_once because it brings a dependency on C++ runtime on some platforms
     // while assert_impl.h is reused in tbbmalloc that should not depend on C++ runtime
-    atomic_do_once([&](){ assertion_failure_impl(location, line, expression, comment); }, assertion_state);
+    atomic_do_once([&](){
+        std::fprintf(stderr, "Assertion %s failed (located in the %s function, line in file: %d)\n",
+            expression, location, line);
+
+        if (comment) {
+            std::fprintf(stderr, "Detailed description: %s\n", comment);
+        }
+#if _MSC_VER && _DEBUG
+        if (1 == _CrtDbgReport(_CRT_ASSERT, location, line, "tbb_debug.dll", "%s\r\n%s", expression, comment?comment:"")) {
+            _CrtDbgBreak();
+        } else
+#endif
+        {
+            std::fflush(stderr);
+            std::abort();
+        }
+    }, assertion_state);
 #if __TBB_MSVC_UNREACHABLE_CODE_IGNORED
     #pragma warning (pop)
 #endif
+}
+
+static std::atomic<assertion_handler_type> assertion_handler{assertion_failure_impl};
+
+assertion_handler_type __TBB_EXPORTED_FUNC set_assertion_handler(assertion_handler_type new_handler) noexcept {
+    return assertion_handler.exchange(new_handler ? new_handler : assertion_failure_impl, std::memory_order_acq_rel);
+}
+
+assertion_handler_type __TBB_EXPORTED_FUNC get_assertion_handler() noexcept {
+    return assertion_handler.load(std::memory_order_acquire);
+}
+
+void __TBB_EXPORTED_FUNC assertion_failure(const char* location, int line, const char* expression, const char* comment) {
+    get_assertion_handler()(location, line, expression, comment);
 }
 
 //! Report a runtime warning.
--- a/src/tbb/def/lin32-tbb.def
+++ b/src/tbb/def/lin32-tbb.def
@@ -19,6 +19,8 @@ global:
 
 /* Assertions (assert.cpp) */
 _ZN3tbb6detail2r117assertion_failureEPKciS3_S3_;
+_ZN3tbb6detail2r121set_assertion_handlerEPFvPKciS3_S3_E;
+_ZN3tbb6detail2r121get_assertion_handlerEv;
 
 /* ITT (profiling.cpp) */
 _ZN3tbb6detail2r112itt_task_endENS0_2d115itt_domain_enumE;
--- a/src/tbb/def/lin64-tbb.def
+++ b/src/tbb/def/lin64-tbb.def
@@ -19,6 +19,8 @@ global:
 
 /* Assertions (assert.cpp) */
 _ZN3tbb6detail2r117assertion_failureEPKciS3_S3_;
+_ZN3tbb6detail2r121set_assertion_handlerEPFvPKciS3_S3_E;
+_ZN3tbb6detail2r121get_assertion_handlerEv;
 
 /* ITT (profiling.cpp) */
 _ZN3tbb6detail2r112itt_task_endENS0_2d115itt_domain_enumE;
--- a/src/tbb/def/mac64-tbb.def
+++ b/src/tbb/def/mac64-tbb.def
@@ -21,6 +21,8 @@
 
 # Assertions (assert.cpp)
 __ZN3tbb6detail2r117assertion_failureEPKciS3_S3_
+__ZN3tbb6detail2r121set_assertion_handlerEPFvPKciS3_S3_E
+__ZN3tbb6detail2r121get_assertion_handlerEv
 
 # ITT (profiling.cpp)
 __ZN3tbb6detail2r112itt_task_endENS0_2d115itt_domain_enumE
--- a/src/tbb/def/win32-tbb.def
+++ b/src/tbb/def/win32-tbb.def
@@ -18,6 +18,8 @@ EXPORTS
 
 ; Assertions (assert.cpp)
 ?assertion_failure@r1@detail@tbb@@YAXPBDH00@Z
+?set_assertion_handler@r1@detail@tbb@@YAP6AXPBDH00@ZP6AX0H00@Z@Z
+?get_assertion_handler@r1@detail@tbb@@YAP6AXPBDH00@ZXZ
 
 ; ITT (tbb_profiling.cpp)
 ?call_itt_notify@r1@detail@tbb@@YAXHPAX@Z
--- a/src/tbb/def/win64-tbb.def
+++ b/src/tbb/def/win64-tbb.def
@@ -18,6 +18,8 @@ EXPORTS
 
 ; Assertions (assert.cpp)
 ?assertion_failure@r1@detail@tbb@@YAXPEBDH00@Z
+?set_assertion_handler@r1@detail@tbb@@YAP6AXPEBDH00@ZP6AX0H00@Z@Z
+?get_assertion_handler@r1@detail@tbb@@YAP6AXPEBDH00@ZXZ
 
 ; ITT (tbb_profiling.cpp)
 ?call_itt_notify@r1@detail@tbb@@YAXHPEAX@Z
--- a/src/tbbbind/CMakeLists.txt
+++ b/src/tbbbind/CMakeLists.txt
@@ -85,6 +85,7 @@ function(tbbbind_build TBBBIND_NAME REQUIRED_HWLOC_TARGET)
     target_link_libraries(${TBBBIND_NAME}
         PUBLIC
         ${REQUIRED_HWLOC_TARGET}
+        TBB::tbb
         PRIVATE
         ${TBB_LIB_LINK_LIBS}
         ${TBB_COMMON_LINK_LIBS}
--- a/src/tbbbind/tbb_bind.cpp
+++ b/src/tbbbind/tbb_bind.cpp
@@ -1,5 +1,6 @@
 /*
-    Copyright (c) 2019-2024 Intel Corporation
+    Copyright (c) 2019-2025 Intel Corporation
+    Copyright (c) 2025 UXL Foundation Contributors
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
@@ -17,7 +18,6 @@
 #include <vector>
 #include <mutex>
 
-#include "../tbb/assert_impl.h" // Out-of-line TBB assertion handling routines are instantiated here.
 #include "oneapi/tbb/detail/_assert.h"
 #include "oneapi/tbb/detail/_config.h"
 


include/oneapi/tbb/detail/_assert.h
```diff
@@ -30,8 +31,17 @@ namespace r1 {
...
+                                                                         const char *expression, const char *comment);
+
+using assertion_handler_type = decltype(&assertion_failure);
+
+//! Set assertion handler and return its previous value.
+TBB_EXPORT assertion_handler_type __TBB_EXPORTED_FUNC set_assertion_handler(assertion_handler_type new_handler) noexcept;
+
+//! Return the current assertion handler.
+TBB_EXPORT assertion_handler_type __TBB_EXPORTED_FUNC get_assertion_handler() noexcept;
+
```
akukanov: Looks like `_assert.h` is reused by tbbmalloc; so maybe these forward references do not belong here.
We could instead add those to the entry point declaration block in `global_control.h`.

src/tbb/assert_impl.h
```diff
@@ -37,45 +38,57 @@ namespace tbb {
...
-
-// Do not move the definition into the assertion_failure function because it will require "magic statics".
-// It will bring a dependency on C++ runtime on some platforms while assert_impl.h is reused in tbbmalloc 
+// Do not move the definition into the assertion_failure_impl function because it will require "magic statics".
+// It will bring a dependency on C++ runtime on some platforms while assert_impl.h is reused in tbbmalloc
 // that should not depend on C++ runtime
 static std::atomic<tbb::detail::do_once_state> assertion_state;
 
-void __TBB_EXPORTED_FUNC assertion_failure(const char* location, int line, const char* expression, const char* comment) {
+// TODO: consider extension for formatted error description string
```
akukanov, line 48: If we want to consider such an extension, we should better do it now. After the public API settles it will be hard if not impossible.

src/tbb/assert_impl.h
```diff
@@ -37,45 +38,57 @@ namespace tbb {
...
+static void assertion_failure_impl(const char* location, int line, const char* expression, const char* comment) {
```
akukanov: Should the default handler be declared `[[noreturn]]`?

include/oneapi/tbb/detail/_assert.h
dnmokhov: The implementation of these (`assert_impl.h`) is also included by tbbmalloc, tbbmalloc_proxy, and tbbbind. (In fact, the latter now uses it to access the custom handler.)

I believe it is better to keep the symmetry. None of the 3 libraries above have public API headers. They also won't have entry points defined for these functions.

src/tbb/assert_impl.h
dnmokhov: This TODO has been there (line 40 formerly) since the TBB revamp, it seems. According to the current TBB 2020 match request, should we probably just drop it?

src/tbb/assert_impl.h
dnmokhov: Added.

test/tbb/test_arena_constraints.cpp
```diff
@@ -208,3 +212,12 @@ TEST_CASE("Testing constraints_threads_per_core() reserved entry point") {
...
     tbb::detail::r1::constraints_threads_per_core(c);
 }
+
+//! Using custom assertion handler to test failure on invalid constraints
+//! \brief \ref interface \ref error_guessing
+TEST_CASE("Using custom assertion handler to test failure on invalid constraints") {
+    tbb::set_assertion_handler(utils::AssertionFailureHandler);
+    TRY_BAD_EXPR(tbb::info::default_concurrency(tbb::task_arena::constraints{}.set_max_threads_per_core(0)),
+        "Wrong max_threads_per_core constraints field value.");
+    tbb::set_assertion_handler(nullptr); // Reset to default handler
```
kboyarinov: I would add the following checks:
* Check that `set_assertion_handler` returns previously installed handler
* Check that `get_assertion_handler` returns correct handler before and after both sets
* Check that `assertion_handler_type` is exactly the type defined in the documentation

include/oneapi/tbb/detail/_assert.h
dnmokhov: Moved.

test/tbb/test_arena_constraints.cpp
dnmokhov: Added.

src/tbb/governor.cpp
```diff
@@ -404,7 +411,8 @@ static const dynamic_link_descriptor TbbBindLinkTable[] = {
     DLD(__TBB_internal_apply_affinity, apply_affinity_ptr),
     DLD(__TBB_internal_restore_affinity, restore_affinity_ptr),
 #endif
-    DLD(__TBB_internal_get_default_concurrency, get_default_concurrency_ptr)
+    DLD(__TBB_internal_get_default_concurrency, get_default_concurrency_ptr),
+    DLD(__TBB_internal_set_tbbbind_assertion_handler, set_assertion_handler_ptr)
```
akukanov: If we would like to make this new symbol "optional", so that the new TBB binaries work with previous tbbbind, I think it has to be processed separate from the rest of the symbols.

src/tbb/assert_impl.h
akukanov: I guess so.

Maybe at the time oneTBB switches to C++23 as the minimal standard version (if ever :)), the API can be modified to work with `string_view` and `std::format`.

src/tbb/assert_impl.h
```diff
@@ -37,45 +38,61 @@ namespace tbb {
...
+// TODO: consider extension for formatted error description string
+/* [[noreturn]] */ static void assertion_failure_impl(const char* location, int line, const char* expression, const char* comment) {
 #if __TBB_MSVC_UNREACHABLE_CODE_IGNORED
     // Workaround for erroneous "unreachable code" during assertion throwing using call_once
     #pragma warning (push)
     #pragma warning (disable: 4702)
 #endif
-    // We cannot use std::call_once because it brings a dependency on C++ runtime on some platforms 
+    // We cannot use std::call_once because it brings a dependency on C++ runtime on some platforms
     // while assert_impl.h is reused in tbbmalloc that should not depend on C++ runtime
```
akukanov: Should this comment be merged with the one at the lines 42-43? These describe two aspects of the same restriction.

src/tbb/governor.cpp
dnmokhov: Changed.

src/tbb/assert_impl.h
dnmokhov: Merged.

src/tbb/assert_impl.h
```diff
@@ -37,45 +38,59 @@ namespace tbb {
...
-    atomic_do_once([&](){ assertion_failure_impl(location, line, expression, comment); }, assertion_state);
+    atomic_do_once([&](){
+        std::fprintf(stderr, "Assertion %s failed (located in the %s function, line in file: %d)\n",
+            expression, location, line);
+
+        if (comment) {
+            std::fprintf(stderr, "Detailed description: %s\n", comment);
+        }
+#if _MSC_VER && _DEBUG
+        if (1 == _CrtDbgReport(_CRT_ASSERT, location, line, "tbb_debug.dll", "%s\r\n%s", expression, comment?comment:"")) {
```
Alexandr-Konovalov: I can't understand why this code works only for tbb_debug.dll. Why can't it be called from tbbmalloc and tbbbind as well?

dnmokhov: It works from all the 4 shared libraries. We probably should change the module name accordingly though.

test/conformance/conformance_global_control.cpp
```diff
@@ -431,3 +432,9 @@ TEST_CASE("terminate_on_exception: enabled") {
     CHECK(terminate_handler_called);
 }
 #endif
+
+TEST_CASE("custom assertion handler type") {
```
Alexandr-Konovalov: [brief_scan](https://github.com/intel-innersource/libraries.threading.infrastructure.onetbb-ci/actions/runs/17268695145/job/49007273481#logs) reports here
`ERROR: conformance_global_control.cpp:435: found TEST_CASE without brief`

src/tbb/assert_impl.h
dnmokhov: I looked more into this and found that specifying `nullptr` instead of `"tbb_debug.dll"` excludes the module from the debug report. However, it is automatically detected and shown as "Program":

image

The DLL name (and path) is current, not hard-coded there. Do you think that would be better?

test/conformance/conformance_global_control.cpp
dnmokhov: I missed that one. Thanks! Added.

include/oneapi/tbb/detail/_assert.h
```diff
@@ -30,8 +31,11 @@ namespace r1 {
 /** Normally called from __TBB_ASSERT macro.
   If assertion handler is null, print message for assertion failure and abort.
   Otherwise call the assertion handler. */
-TBB_EXPORT void __TBB_EXPORTED_FUNC assertion_failure(const char *location, int line,
-                                                      const char *expression, const char *comment);
+/* [[noreturn]] */ TBB_EXPORT void __TBB_EXPORTED_FUNC assertion_failure(const char *location, int line,
```
kboyarinov: What is the purpose of having `[[noreturn]]` commented here?
It seems like we can't add the attribute neither now nor later, because it will lead to backward compatibility break. 
According to the [standard](https://en.cppreference.com/w/cpp/language/attributes/noreturn#:~:text=If%20a%20function%20is%20declared%20with%20%5B%5Bnoreturn%5D%5D%20in%20one%20translation%20unit%2C%20and%20the%20same%20function%20is%20declared%20without%20%5B%5Bnoreturn%5D%5D%20in%20another%20translation%20unit%2C%20the%20program%20is%20ill%2Dformed%3B%20no%20diagnostic%20required.), having a function declared with `[[noreturn]]` in one translation unit and without it in another, the program is ill-formed, no diagnostic required:

include/oneapi/tbb/global_control.h
```diff
@@ -46,6 +47,14 @@ void release_impl(d1::task_scheduler_handle& handle);
 bool finalize_impl(d1::task_scheduler_handle& handle);
 TBB_EXPORT void __TBB_EXPORTED_FUNC get(d1::task_scheduler_handle&);
 TBB_EXPORT bool __TBB_EXPORTED_FUNC finalize(d1::task_scheduler_handle&, std::intptr_t mode);
+
+#if !__TBB_NO_CUSTOM_ASSERTION_HANDLING
```
kboyarinov: Should the declarations of functions in `detail` namespace be guarded with the macro?

include/oneapi/tbb/global_control.h
```diff
@@ -193,6 +202,11 @@ using detail::d1::attach;
 using detail::d1::finalize;
 using detail::d1::task_scheduler_handle;
 using detail::r1::unsafe_wait;
+using detail::r1::assertion_handler_type;
+#if !__TBB_NO_CUSTOM_ASSERTION_HANDLING
```
kboyarinov: I would use the private version of the umbrella macro instead:

```suggestion
#if !__TBB_DISABLE_SPEC_EXTENTIONS
```

src/tbb/assert_impl.h
```diff
@@ -37,45 +38,59 @@ namespace tbb {
...
+            std::abort();
+        }
+    }, assertion_state);
 #if __TBB_MSVC_UNREACHABLE_CODE_IGNORED
     #pragma warning (pop)
 #endif
 }
 
+namespace custom_handler {
+    static std::atomic<assertion_handler_type> handler{assertion_failure_impl}; // initial value is default handler
```
kboyarinov: It seems like this atomic do not always mean the "custom" assertion handler, it can be also the default one.
I would either remove the namespace or rename it accordingly.

test/conformance/conformance_global_control.cpp
```diff
@@ -431,3 +432,11 @@ TEST_CASE("terminate_on_exception: enabled") {
     CHECK(terminate_handler_called);
 }
 #endif
+
+//! Testing that assertion_handler_type is exactly the type defined in the documentation
+//! \brief \ref interface \ref requirement
+TEST_CASE("custom assertion handler type") {
```
kboyarinov: The purpose of the conformance test is to check the direct requirements of the spec. Since `tbb::assertion_handler_type` is not part of the spec - this test belong to `test_global_control.cpp`.

test/tbb/test_arena_constraints.cpp
kboyarinov: May be it is better to arrange the test in the common function in `utils_assert.h` and perform same checks for `global_control` and `arena_constraints`. 
Different "bad" expressions can be accepted as a lambda.

src/tbb/assert_impl.h
dnmokhov: Agreed offline not to change anything here, as this is unrelated to this PR.

test/tbb/test_arena_constraints.cpp
dnmokhov: I moved the before/after tests to `utils_assert.h` but decided to keep the macro-based bad expression stringification from TBB 2020.

include/oneapi/tbb/detail/_assert.h
dnmokhov: > It seems like we can't add the attribute neither now nor later, because it will lead to backward compatibility break.

I guess you are referring to compatibility with custom handlers written for TBB 2020. Good point. I removed this but added a more detailed comment to the public declaration about not returning.

include/oneapi/tbb/global_control.h
dnmokhov: The `detail` namespace is indeed not in the spec, so I removed the macro guard here.

include/oneapi/tbb/global_control.h
dnmokhov: Renamed.

src/tbb/assert_impl.h
dnmokhov: Renamed. Did not remove (to keep the the atomic and its setter/getter together).

test/conformance/conformance_global_control.cpp
dnmokhov: Moved.

src/tbb/assert_impl.h
```diff
@@ -37,45 +38,59 @@ namespace tbb {
...
+#if (__TBB_BUILD || __TBBBIND_BUILD) // only TBB and TBBBind use custom handler
+    static assertion_handler_type set(assertion_handler_type new_handler) noexcept {
+        return handler.exchange(new_handler ? new_handler : assertion_failure_impl, std::memory_order_acq_rel);
+    }
+#endif
+
+    static assertion_handler_type get() noexcept {
+        return handler.load(std::memory_order_acquire);
+    }
+}
```
kboyarinov: As far as I know, we don't add indentation for namespaces:

```suggestion
namespace assertion_handler {
static std::atomic<assertion_handler_type> handler{assertion_failure_impl}; // initial value is default handler

#if (__TBB_BUILD || __TBBBIND_BUILD) // only TBB and TBBBind use custom handler
static assertion_handler_type set(assertion_handler_type new_handler) noexcept {
    return handler.exchange(new_handler ? new_handler : assertion_failure_impl, std::memory_order_acq_rel);
}
#endif

static assertion_handler_type get() noexcept {
    return handler.load(std::memory_order_acquire);
}
} // namespace assertion_handler
```

test/tbb/test_arena_constraints.cpp
```diff
@@ -208,3 +212,10 @@ TEST_CASE("Testing constraints_threads_per_core() reserved entry point") {
     tbb::task_arena::constraints c{};
     tbb::detail::r1::constraints_threads_per_core(c);
 }
+
+//! Using custom assertion handler to test failure on invalid constraints
+//! \brief \ref interface \ref error_guessing
+TEST_CASE("Using custom assertion handler to test failure on invalid constraints") {
+    TRY_BAD_EXPR(tbb::info::default_concurrency(tbb::task_arena::constraints{}.set_max_threads_per_core(0)),
+        "Wrong max_threads_per_core constraints field value.");
```
kboyarinov: I would rename `TRY_BAD_EXPR` to something reflecting the testing of custom assertion handler. Something like `TEST_CUSTOM_ASSERTION_HANDLER_WITH_BAD_EXPR` or simply `TEST_CUSTOM_ASSERTION_HANDLER`

include/oneapi/tbb/detail/_assert.h
kboyarinov: I more thought about the compatibility of existing `assertion_failure` in the older oneTBB releases (without custom handler at all). It is already defined without the attribute, so adding it in the current (or any further) release will lead to undefined behavior.

dnmokhov: Good idea. In fact, we should document that the "no returning" should involve terminating the application. E.g., throwing an exception is not enough.

test/tbb/test_arena_constraints.cpp
dnmokhov: Renamed.

rfcs/supported/assertion_handler/README.md
```diff
@@ -67,11 +69,21 @@ namespace tbb {
...
+#endif
 }
 ```
 
 Applications that used the custom assertion handler in TBB 2020 can migrate to this proposal with no changes.
 
+#### Specification Extension
+
+This API is introduced as an extension to the oneTBB specification, controlled by the
+`__TBB_DISABLE_SPEC_EXTENTIONS` macro. By default (macro undefined or defined as 0), the extension will
```
isaevil: Is this going to be a public macro? If so, let's omit the underscore or add a public wrapping macro.

include/oneapi/tbb/global_control.h
```diff
@@ -193,6 +201,11 @@ using detail::d1::attach;
 using detail::d1::finalize;
 using detail::d1::task_scheduler_handle;
 using detail::r1::unsafe_wait;
+#if !__TBB_DISABLE_SPEC_EXTENTIONS
```
isaevil:
```suggestion
#if !__TBB_DISABLE_SPEC_EXTENSIONS
```

rfcs/supported/assertion_handler/README.md
kboyarinov: As we discussed last week, we cannot introduce the public macro now because we need to wrap and exclude the existing extensions as well. For some of them, we will need to establish testing with extensions on/off. Agreed to introduce the public macro in one of the next releases and wrap the assertion handler in the private version to simplify further work.

isaevil: Okay, thanks!

include/oneapi/tbb/global_control.h
isaevil: Please fix in other places as well.

test/tbb/test_global_control.cpp
```diff
@@ -272,3 +275,18 @@ TEST_CASE("test concurrent task_scheduler_handle destruction") {
     stop = true;
     thr1.join();
 }
+
+//! Testing that assertion_handler_type is exactly the type defined in the documentation
+//! \brief \ref interface \ref requirement
+TEST_CASE("custom assertion handler type") {
```
aleksei-fedotov: Perhaps, simply:
```suggestion
TEST_CASE("Assertion handler type") {
```

test/tbb/test_global_control.cpp
```diff
@@ -272,3 +275,18 @@ TEST_CASE("test concurrent task_scheduler_handle destruction") {
...
+    using documented = void(*)(const char* /* location */, int /* line */,
+                               const char* /* expression */, const char* /* comment */);
+    static_assert(std::is_same<oneapi::tbb::assertion_handler_type, documented>::value, "Incorrect assertion handler type");
```
aleksei-fedotov: As far as I know, we adhere to no longer than 100 column wide lines.
```suggestion
    static_assert(std::is_same<oneapi::tbb::assertion_handler_type, documented>::value,
                  "Incorrect assertion handler type");
```
Could you please fix here and in the other places as well?

src/tbb/misc.cpp
```diff
@@ -151,6 +153,18 @@ void detect_cpu_features(cpu_features_type& cpu_features) {
 #endif
 }
 
+//------------------------------------------------------------------------
+// custom assertion handler
+//------------------------------------------------------------------------
+#if __TBB_BUILD
```
aleksei-fedotov, line 159: Are there situations when this file is not built as part of TBB? In other words, why do we need this macro wrapping the functionality?

rfcs/supported/assertion_handler/README.md
aleksei-fedotov: I would use simpler macro name. Specifically, I suggest skipping `DISABLE/ENABLE` word in the name and having only `TBB_SPEC_EXTENSIONS` as the macro name with possibly two underscores prepended for the internal macro if necessary. Set it to `1` by default.

include/oneapi/tbb/global_control.h
dnmokhov: Changed.

test/common/utils_assert.h
```diff
@@ -49,4 +50,73 @@ void AssertSameType( const T& /*x*/, const T& /*y*/ ) {}
...
+    return default_handler;
+}
+
+void CheckAssertionFailure(int line, std::string expression, bool okay, const char* message, std::string substr) {
+    REQUIRE_MESSAGE(okay, "Line ", line, ", ", expression, " failed to fail");
+    REQUIRE_MESSAGE(message, "Line ", line, ", ", expression, " failed without a message");
+
+    std::string msg_str = message;
+    REQUIRE_MESSAGE(msg_str.find(substr) != std::string::npos,
+        expression, " failed with message '", msg_str, "' missing substring '", substr, "'");
```
aleksei-fedotov, line 113: Why searching for a substring instead of requiring the exact match?

dnmokhov: I copied this over from TBB 2020. My guess is that it can help when, e.g., the assertion message includes a variable parameter.

rfcs/supported/assertion_handler/README.md
dnmokhov: I like the opt-out nature of `DISABLE`. Do you think it is OK that there is no `TBB_SPEC_EXTENSIONS` guard around the existing extensions (which we will leave untouched for now)?

src/tbb/misc.cpp
dnmokhov: I had to add it, because `test_concurrent_monitor.cpp` includes `misc.cpp` for white-box unit testing.

rfcs/supported/assertion_handler/README.md
```diff
@@ -67,11 +69,21 @@ namespace tbb {
...
 
+#### Specification Extension
+
+This API is introduced as an extension to the oneTBB specification, controlled by the
+`__TBB_DISABLE_SPEC_EXTENSIONS` macro. By default (macro undefined or defined as 0), the extension will
+be enabled: `set_assertion_handler` and `get_assertion_handler` will be declared and exported, and
+`assertion_failure` will dispatch to the active handler. Defining `__TBB_DISABLE_SPEC_EXTENSIONS` to a non-zero
+value before including oneTBB headers will disable the extension: these declarations will be excluded from
+the public API, and the library will always use the default assertion behavior.
+
```
akukanov, line 86: I think it should at least be a public documented macro, i.e., no preceding underscores.

Then comes the name. As proposed, it is overarching and applies not only to the custom handler but also to other exceptions already in the library and any future ones. We might want to provide more fine-grained control with specific macros.

Third comes the semantics. The proposed macro opts the extensions out. It is a reasonable semantics for the overarching macro - a single switch to conformant-only mode. But for fine-grained control the opt-in semantics might be better, as it make the use of an extension somewhat visible in the code.

rfcs/supported/assertion_handler/README.md
aleksei-fedotov: No, I think for the current patch we need to settle with an internal macro name and decide about the public one later.

Do you like the opt-out nature of `DISABLE` only for public macro name or for internal one as well?
I just wonder if you find the implementation part to look more clear as I do if it would contain something like the following:
```cpp
#if __TBB_SPEC_EXTENSION
// Extension API and implementation goes here
#endif
```
, which, in my opinion, laconically conveys the message that:
1. The surrounded code is an extension, and
2. It is enabled (since there is the `#if` preprocessor directive here)

rather than:
```cpp
#if !__TBB_DISABLE_SPEC_EXTENSION
// Extension API and implementation goes here
#endif
```
, which, again in my opinion, requires a bit more brain activity while reading the code, as it is a negation of a disabling.

But, I don't insist on anything. We can leave it as you like. I just wanted to clarify your and my preferences.

rfcs/supported/assertion_handler/README.md
kboyarinov: Yes, we definitely need a public macro to disable all of the extensions. But since we did not yet covered (and not tested) excluding the existing extensions, the plan was to wrap the assertion handler extension with the internal version of the macro and add the public one in the next release (covering all of the features).

The proposed semantics  was to opt all of the extensions out and have all of the extensions by default since disabling single extensions does not make much sense - the user should have either the strict spec-compliant version of the library or the version with the extensions.

test/common/utils_assert.h
aleksei-fedotov: I think it does not matter, because the test uses public interfaces for assertion handling, and therefore sets the message content that should be thrown. So, it is up to the test what message to wait for in the fired assertion.

include/oneapi/tbb/detail/_assert.h
```diff
@@ -32,6 +33,9 @@ namespace r1 {
   Otherwise call the assertion handler. */
```
aleksei-fedotov: The comment above looks outdated. Please take a look. Unfortunately, I cannot comment on the untouched lines of the patch due to lack of such feature in GitHub.

Specifically, the line _If assertion handler is null, print message for assertion failure and abort. Otherwise call the assertion handler._
I think it needs to be revised to something like: _If assertion handler is not set, the default one is used, which prints the message for assertion failure and aborts. The setting of an assertion handler is available as part of oneTBB specification extension._

test/common/utils_assert.h
dnmokhov: The test sets `substr`, but the assertion message comes from the library usually, i.e., from executing the expression `x`.

rfcs/supported/assertion_handler/README.md
aleksei-fedotov: How about if we don't mention the name of the macro in this file, but refer to it through indirection like _.. controlled by overarching spec extension macro. That macro enables extensions by default, in which case `set_assertion_handler` and `get_assertion_handler` will be declared and exported, and `assertion_failure` will dispatch to the active handler_?

test/common/utils_assert.h
aleksei-fedotov: Oh, okay. I see now. Thanks for the explanation. The assertion handler tests are sort of whitebox tests that require certain messages in the oneTBB asserts.

test/tbb/test_global_control.cpp
dnmokhov: Fixed,

include/oneapi/tbb/detail/_assert.h
dnmokhov: Updated as suggested.

rfcs/supported/assertion_handler/README.md
vossmjp: I agree w/ @kboyarinov that it would be premature to have a public overarching macro before all extensions are covered by it. It's debatable if public per-feature marcos will be useful or not. Most commonly, I think, would be to be either strictly specification-compliant or not, but not to mix in some subset of extensions.

vossmjp (approved):
Looks good to me.

PR #1821, opened 2025-08-18
Add current_task_ptr entry point

kboyarinov:
### Description 
Add current_task_ptr entry point for task_group dynamic dependencies feature.

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown

### DIFF
--- a/test/tbb/test_task.cpp
+++ b/test/tbb/test_task.cpp
@@ -914,13 +914,13 @@ enum current_task_ptr_submit_type {
     enqueue
 };
 
-template <current_task_ptr_submit_type SubmitType>
 struct current_task_ptr_checking_task : public tbb::detail::d1::task {
-    current_task_ptr_checking_task(std::size_t d, tbb::task_arena& a, tbb::detail::d1::wait_context& wtc, tbb::task_group_context& ctx)
-        : depth(d), arena(a), wait_context(wtc), task_group_ctx(ctx) {}
+    current_task_ptr_checking_task(std::size_t d, tbb::task_arena& a, tbb::detail::d1::wait_context& wtc,
+                                   tbb::task_group_context& ctx, current_task_ptr_submit_type tag)
+        : depth(d), arena(a), wait_context(wtc), task_group_ctx(ctx), submit_tag(tag) {}
 
     void submit_and_wait(current_task_ptr_checking_task& t, tbb::detail::d1::wait_context& wait_ctx) {
-        switch (SubmitType) {
+        switch (submit_tag) {
             case current_task_ptr_submit_type::execute_and_wait: {
                 tbb::detail::d1::execute_and_wait(t, task_group_ctx, wait_ctx, task_group_ctx);
                 break;
@@ -931,7 +931,7 @@ struct current_task_ptr_checking_task : public tbb::detail::d1::task {
                 submit(t, arena, task_group_ctx, true);
                 break;
             } case current_task_ptr_submit_type::slot_spawn: {
-                tbb::detail::d1::slot_id task_slot = tbb::detail::d1::slot_id(tbb::this_task_arena::current_thread_index() + 1 % tbb::this_task_arena::max_concurrency());
+                auto task_slot = tbb::detail::d1::slot_id(tbb::this_task_arena::current_thread_index() + 1 % tbb::this_task_arena::max_concurrency());
                 tbb::detail::d1::spawn(t, task_group_ctx, task_slot);
                 break; 
             } case current_task_ptr_submit_type::enqueue: {
@@ -941,7 +941,7 @@ struct current_task_ptr_checking_task : public tbb::detail::d1::task {
                 CHECK(false);
         }
 
-        if (SubmitType != current_task_ptr_submit_type::execute_and_wait) {
+        if (submit_tag != current_task_ptr_submit_type::execute_and_wait) {
             arena.execute([&] {
                 tbb::detail::r1::wait(wait_ctx, task_group_ctx);
             });
@@ -957,7 +957,7 @@ struct current_task_ptr_checking_task : public tbb::detail::d1::task {
         if (depth < max_current_task_ptr_test_depth) {
             tbb::detail::d1::wait_context next_task_wait_context(1);
 
-            current_task_ptr_checking_task next_task(depth + 1, arena, next_task_wait_context, task_group_ctx);
+            current_task_ptr_checking_task next_task(depth + 1, arena, next_task_wait_context, task_group_ctx, submit_tag);
             submit_and_wait(next_task, next_task_wait_context);
 
             REQUIRE_MESSAGE(tbb::detail::d1::current_task_ptr() == this,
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -19,7 +19,7 @@ on:
     branches: [master]
 
   pull_request:
-    branches: [master, dev/kboyarinov/poc-dynamic-dependencies-part1]
+    branches: [master, dev/kboyarinov/poc-dynamic-dependencies-part1, dev/kboyarinov/poc-dynamic-dependencies-part2]
```
akukanov: Can this patch go directly to master instead?

src/tbb/task_dispatcher.h
```diff
@@ -328,6 +331,7 @@ d1::task* task_dispatcher::local_wait_for_all(d1::task* t, Waiter& waiter ) {
 
                     ITT_CALLEE_ENTER(ITTPossible, t, itt_caller);
 
+                    m_innermost_running_task = t;
```
akukanov: Should not `m_innermost_running_task` be reset to null (or to the previous value?) after the call to `execute`/`cancel` ? (See 5-7 lines down in the code, since GH does not allow to extend comments to unmodified lines)

src/tbb/task.cpp
```diff
@@ -256,6 +256,10 @@ d1::wait_tree_vertex_interface* get_thread_reference_vertex(d1::wait_tree_vertex
     return ref_counter;
 }
 
+d1::task* current_task_ptr() {
+    return governor::get_thread_data()->get_innermost_running_task();
+}
+
```
akukanov: The logic of splitting code between `task.cpp` and `task_dispatcher.cpp` seems unclear to me. Are there some principles, or is it chosen randomly? :) Why does this function go into `task.cpp` and not `task_dispatcher.cpp`?

And also, should or should not the function get marked as `__TBB_EXPORTED_FUNC` in the implementation? There seem to be both cases for the entry points.

src/tbb/task_dispatcher.h
kboyarinov: Do you think it is really needed? It will reset `m_innermost_running_task` only for while searching for another task (stealing loop, get_critical, etc.). If the new task is found, `m_innermost_running_task` is overwritten to point to it.
If the thread is leaving the execution loop, the pointer is reset by `dispatch_loop_guard` to point to the previous task.

akukanov: Let's imagine that some user callback is invoked during the task search. It can be e.g. a new observer enabled.
What should `current_task_ptr` return if called in such a callback? I think that either nullptr or the blocked outer-level task is in any case better than the task that has already been handled, and is likely destroyed.

kboyarinov: Applied

.github/workflows/ci.yml
kboyarinov: Yes, it can. But I will need to rebase all of the other PRs for this one since task group dynamic dependencies PRs depend on these changes and transfer_this_task_completion_to depend on both set_task_order and this PR.
I would also merge this together with all other task_group dependency changes.
BTW, all changes in ci.yml will be removed before merging to master. It is only necessary to enable public CI checks for these changes.

src/tbb/task.cpp
kboyarinov: I have noticed that all functions in `task_dispatcher.cpp` are using `task_dispatcher` instance inside of the body. Since this function do not use it, I have placed it into `task.cpp`.

Regarding `__TBB_EXPORTED_FUNC` attribute, it seems like all our entry points are marked in `_task.h` and both `task.cpp` and `task_dispatcher.cpp` include this header. Hence, `__TBB_EXPORTED_FUNC` is not necessary for definitions in cpp files. Since all of the functions in `task.cpp` are not marked, I did not add the attribute for `current_task_ptr` as well.

.github/workflows/ci.yml
akukanov: I am not concerned about the ci.yml changes, just used it as the place to leave the comment.

But I would prefer the commit history to have a separate commit for `current_task_ptr` vs. the preview extension. If not with rebasing, perhaps this can be achieved by having separate single commits for all the 5 parts, just squashing the intermediate ones in each PR.

src/tbb/task.cpp
akukanov: > I have noticed that all functions in task_dispatcher.cpp are using task_dispatcher instance inside of the body.

But the opposite is not true, since there are many uses of `task_dispatcher` in `task.cpp`, including some of the class methods being implemented there. The split looks like an illogical mess, and perhaps it is such a mess :(

For the attribute - I understand that technically it is not needed here, but it helps reading the code, explicitly marking the entry points. And it'd be good to be consistent as well. Anyway, I just wanted to bring this to attention; feel free to ignore.

src/tbb/task_dispatcher.h
```diff
@@ -331,7 +331,9 @@ d1::task* task_dispatcher::local_wait_for_all(d1::task* t, Waiter& waiter ) {
 
                     ITT_CALLEE_ENTER(ITTPossible, t, itt_caller);
 
+                    d1::task* prev_innermost_running_task = m_innermost_running_task;
                     m_innermost_running_task = t;
+
```
akukanov: Maybe keep this out of `ITT_CALLEE_ENTER/LEAVE` scope? (see the line 332 above)

test/tbb/test_task.cpp
```diff
@@ -902,3 +903,58 @@ TEST_CASE("Force thread limit on per-thread reference_vertex") {
...
+        }
+
+        wait_context.release();
+        return nullptr;
+    }
+
+    tbb::detail::d1::task* cancel(tbb::detail::d1::execution_data&) override {
+        REQUIRE_MESSAGE(false, "current_task_ptr_checking_tasks should not be cancelled");
+        return nullptr;
+    }
```
akukanov: Why should they not be cancelled? It would be appropriate to test `current_task_ptr` in the scope of `cancel()` as wel.

Ideally, the test should also create some slack for stealing, and maybe even use other task distribution mechanisms like affinity & enqueuing. It might seem redundant, but only because the implementation details are known. If you pretend you do not know these, I think it seems logical to use different from-submission-to-execution tracks to make sure that the API works in the same way for all.

src/tbb/task.cpp
kboyarinov: Added the attribute.

src/tbb/task_dispatcher.h
kboyarinov: Done

test/tbb/test_task.cpp
kboyarinov: Extended the test to test cancellation and different submission forms.

.github/workflows/ci.yml
kboyarinov: Retargeted to master

PR #1828, opened 2025-08-26
Add test for custom assertion handler inside TBBbind

Alexandr-Konovalov:
### Description 
Add test for custom assertion handler inside TBBbind

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [X] not needed

- [ ] Yes
- [X] No
- [ ] Unknown

### DIFF
--- a/test/tbb/test_dynamic_link.cpp
+++ b/test/tbb/test_dynamic_link.cpp
@@ -61,6 +61,7 @@ FOO_TYPE dummy_foo2() { return FOO_DUMMY; }
 #ifdef __TBB_SKIP_DEPENDENCY_SIGNATURE_VERIFICATION
 #warning "Signature verification must not be turned off to fully test dynamic link functionality"
 #endif
+// The direct include since we want to test internal functionality
 #include "src/tbb/dynamic_link.cpp"
 
 


test/tbb/test_tbbbind.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+    const char* tbbbind_version = system_topology::load_tbbbind_shared_object();
+    REQUIRE_MESSAGE(tbbbind_version, "TBBbind must be found");
+
+    set_assertion_handler(custom_handler);
+
+#if TBB_USE_ASSERT
+    // Trigger an assertion failure to test the custom handler
+    TRY_BAD_EXPR(deallocate_binding_handler_ptr(nullptr), "overloaded assertion");
+#endif
```
dnmokhov: I suggest setting `utils::AssertionFailureHandler` and then directly checking for the message from `tbb_bind.cpp`: "Trying to deallocate nullptr pointer."

Alexandr-Konovalov: Done.

test/tbb/test_tbbbind.cpp
```diff
@@ -0,0 +1,65 @@
...
+#include "../../src/tbb/load_tbbbind.cpp"
+
+
+// All assertions in TBBbind are available only in TBB_USE_ASSERT mode,
+// and testing can't be done without TBBbind.
+#if TBB_USE_ASSERT && __TBB_HWLOC_VALID_ENVIRONMENT
+static bool canHandlerBeTested() { return true; }
+#else
+static bool canHandlerBeTested() { return false; }
+#endif
```
dnmokhov: Not 100% sure, but I think that for TBBBind which is statically linked with HWLOC, `__TBB_SELF_CONTAINED_TBBBIND` is specified instead of `__TBB_HWLOC_VALID_ENVIRONMENT`.

Alexandr-Konovalov: Thanks, missed that. Added.

test/tbb/test_tbbbind.cpp
```diff
@@ -0,0 +1,80 @@
...
+}
+
+TEST_CASE("Check TBBbind availability" * doctest::skip(!isTbbBindAvailable())) {
+    const char *tbbbind_path = system_topology::load_tbbbind_shared_object();
+    REQUIRE_MESSAGE(tbbbind_path != nullptr, "Failed to load TBBbind");
+}
+
+// The test relies on an assumption that system_topology::load_tbbbind_shared_object() find
+// same instance of TBBbind as TBB uses internally.
+TEST_CASE("Using custom assertion handler inside TBBbind"
```
kboyarinov: Why should we do it as 3 separate test cases? It seems like each test case depends on the execution of the previous one. 
Doctest also allows to run only the specified test case and seems like in this case the test will not work correct. Correct me if I'm wrong

test/tbb/test_tbbbind.cpp
```diff
@@ -0,0 +1,80 @@
+/*
+    Copyright (c) 2019-2025 Intel Corporation
```
kboyarinov: Should be only the current date for the new file
```suggestion
    Copyright (c) 2025 Intel Corporation
```

test/CMakeLists.txt
```diff
@@ -542,6 +542,7 @@ if (TARGET TBB::tbb)
     # HWLOC related test
```
kboyarinov, line 543: UXL copyright is missed for this file

test/tbb/test_tbbbind.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include "../../src/tbb/dynamic_link.cpp"
+
+class binding_handler;
+    // dummy implementation, not used
+class governor {
+    public:
+        static unsigned default_num_threads () { return 0; }
+};
+
+#include "../../src/tbb/load_tbbbind.cpp"
```
kboyarinov: Why we can't directly inject `dynamic_link.cpp` and `governor.cpp` here instead of creating the separate source file? Or perhaps using `test/common/inject_scheduler.h` header

Alexandr-Konovalov: Direct inclusion is probably better, while requiring some manual work.

test/tbb/test_tbbbind.cpp
Alexandr-Konovalov: Yes, that's true. There is possibly no run-time skip, so I created separate test case for successful manual loading of TBBbind. But, as we currently not testing TBBbind per se, it can be eliminated.

test/CMakeLists.txt
Alexandr-Konovalov: Done.

test/tbb/test_tbbbind.cpp
Alexandr-Konovalov: And now I think that creation of isolated component like dynamic_link (with a possibility to just `#include` it) is better, but it's unclear how simple is that.

akukanov:
Admittedly I did not look carefully at all nuances here and can miss something, but this test design seems overly complicated.

If the goal is to test that a custom assertion handler set for TBB is also used by tbbbind, why not just ensure that tbbbind is loaded and then directly call one of its entry points with incorrect parameters that should trigger an assertion?

Alexandr-Konovalov:
> Admittedly I did not look carefully at all nuances here and can miss something, but this test design seems overly complicated.
> 
> If the goal is to test that a custom assertion handler set for TBB is also used by tbbbind, why not just ensure that tbbbind is loaded and then directly call one of its entry points with incorrect parameters that should trigger an assertion?

TBBbind is loaded with `RTLD_LOCAL`, so we need dlopen()/dlsym(). To call dlopen(), we need to know name, I don't want to copy/paste [that logic](https://github.com/uxlfoundation/oneTBB/blob/712ad98443300aab202f5e93a76472d59b79752a/src/tbb/governor.cpp#L431). And to find out right library to load, there is a logic that's again better not to copy-paste.
`for (const auto& tbbbind_version : {TBBBIND_2_5_NAME, TBBBIND_2_0_NAME, TBBBIND_NAME}) `

So probably all that stuff should go to a separate TU just like `dynamic_link.cpp`.

src/tbb/dynamic_link.cpp
```diff
@@ -376,9 +376,18 @@ namespace r1 {
...
         *(backslash+1) = 0;
     #else
+        // There are use cases, when we need to find a library, not just some
+        // shared object providing dynamic_link symbol (it can be shared object
+        // that directly includes dynamic_link.cpp).
+        #if defined(SYMBOL_TO_FIND_LIBRARY)
+        static void *func_from_lib = (void*)&SYMBOL_TO_FIND_LIBRARY;
+        #else
+        static void *func_from_lib = (void*)&dynamic_link;
+        #endif
```
akukanov: Instead of a function name passed in the macro, could not we just use either an internal function (`dynamic_link`) or an ABI  function (`TBB_runtime_version`) in this piece of code? And perhaps it can be decided based on `__TBB_BUILD` (and other similar macros if needed) instead of the special new one?

Alexandr-Konovalov: Yes, for sure, but I have taken into account that this component (dynamic_link.cpp) is shared at least with the OpenMP project, so I wish to not introduce TBB-centric names in the code.

akukanov: As far as I know, it is modified there anyway, so changing/removing TBB specifics as needed should not be a big deal.

Alexandr-Konovalov: Please find the updated version. To me, one is a little less clear, as dynamic_link_test now must know something about `__TBB_BUILD` vs explicit opt-in in the previous version. But this is not very significant.

src/tbb/dynamic_link.cpp
```diff
@@ -19,9 +19,12 @@
 
 #include "oneapi/tbb/detail/_template_helpers.h"
 #include "oneapi/tbb/detail/_utils.h"
+#if !__TBB_BUILD && !_WIN32
+#include "oneapi/tbb/version.h" // to get TBB_runtime_version
+#endif
```
akukanov: `#if DYNAMIC_LINK_FIND_LIB_WITH_TBB_RUNTIME_VERSION` ?

test/tbb/test_tbbbind.cpp
akukanov, line 1: Should it be called `test_tbbbind_custom_assertion`? Or do you think it will be extended over time to cover more use cases?

src/tbb/dynamic_link.cpp
Alexandr-Konovalov: Thanks, missed.

test/tbb/test_tbbbind.cpp
Alexandr-Konovalov: At least this is one possible place to do so. We have right version of TBBbind loaded and all entry points are potentially at hand.

test/tbb/test_tbbbind.cpp
```diff
@@ -0,0 +1,115 @@
...
+        DLD(__TBB_internal_deallocate_binding_handler, deallocate_binding_handler_ptr)
+    };
+
+    for (const auto& tbbbind_version : tbbbind_libraries_list) {
+        if (dynamic_link(tbbbind_version, LinkTable,
+                         sizeof(LinkTable) / sizeof(dynamic_link_descriptor), nullptr,
+                         DYNAMIC_LINK_LOCAL_BINDING)) {
+            return deallocate_binding_handler_ptr;
+        }
+    }
```
akukanov, line 69: We need to somehow make sure that `dynamic_link` here does not attempt to load a tbbbind library but only searches for the symbol if that library is already loaded into the process address space (which is the job of TBB, not the test). Is this already the case? If not yet, let's not forget to improve the test later.

akukanov: A short investigation has shown that at least on Linux the symbols of `libtbbbind` are not visible if the library handle is unknown, so addressing the concern requires some extra work. OK to postpone for later.

akukanov: Though maybe it is sufficient to pass `DYNAMIC_LINK_GLOBAL` as the flag. It limits the search to only the libraries already loaded into the process, but it still `dlopen`s these libraries with `RTLD_GLOBAL` which should make their symbols visible.

Alexandr-Konovalov: Yes, this seems to eliminate the potential issue with loading another library from the test.
Do not load TBBbind, only use already loaded #1847

PR #1829, opened 2025-08-27
Enable TTAS strategy for spin_mutex and queuing_mutex

kboyarinov:
### Description 
Current implementation of `tbb::spin_mutex::lock`, `tbb::spin_mutex::try_lock` and `tbb::queuing_mutex::scoped_lock::try_acquire` implements TAS (test-and-set) strategy to acquire the lock. `spin_mutex::lock` atomically exchanges the Boolean flag in the backoff loop until the previous value is `false`, meaning the current thread acquires the lock. 
Under high contention since each thread locks the cache line during exchange that results in invalidations on other threads and leads to poor performance scalability. 

The common strategy to fix this issue is to change the strategy to [TTAS](https://en.wikipedia.org/wiki/Test_and_test-and-set) (test and test-and-set). It adds extra check in the backoff loop. It only tries to exchange the atomic if its value is false (mutex unlocked). This will lock the cache line only if the current iteration has a chance to acquire the lock.

Pre-revamp TBB version [used to implement TTAS](https://github.com/uxlfoundation/oneTBB/blob/81e4471d1191c75e0a78311c584b64b6591da842/include/tbb/tbb_machine.h#L925C1-L931C2) in `spin_mutex::lock` but it was changed during TBB Revamp.

This PR restores the previous behavior. It also changes `spin_mutex::try_lock` and `queuing_mutex::scoped_lock::try_acquire` since it can result in the same performance issue if the user is trying to acquire the lock in the loop while doing something else between the attempts:

```cpp
while (!mutex.try_lock()) {
    do_something_useful();
}
```

The chart below shows the performance improvement of using TTAS in `spin_mutex` on the simple benchmark that is doing 10000 iterations of `parallel_for` with `static_partitioner` each contains the 1ns critical section:

Screenshot 2025-08-27 191307

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/include/oneapi/tbb/queuing_mutex.h
+++ b/include/oneapi/tbb/queuing_mutex.h
@@ -1,5 +1,6 @@
 /*
-    Copyright (c) 2005-2022 Intel Corporation
+    Copyright (c) 2005-2025 Intel Corporation
+    Copyright (c) 2025 UXL Foundation Contributors
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
--- a/include/oneapi/tbb/spin_mutex.h
+++ b/include/oneapi/tbb/spin_mutex.h
@@ -1,5 +1,6 @@
 /*
-    Copyright (c) 2005-2021 Intel Corporation
+    Copyright (c) 2005-2025 Intel Corporation
+    Copyright (c) 2025 UXL Foundation Contributors
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.


isaevil (approved):
LGTM!

include/oneapi/tbb/queuing_mutex.h
```diff
@@ -107,7 +108,7 @@ class queuing_mutex {
             // The compare_exchange_strong must have release semantics, because we are
             // "sending" the fields initialized above to other processors.
             // x86 compare exchange operation always has a strong fence
-            if (!m.q_tail.compare_exchange_strong(expected, this, std::memory_order_acq_rel))
+            if (m.q_tail.load(std::memory_order_relaxed) != expected || !m.q_tail.compare_exchange_strong(expected, this, std::memory_order_acq_rel))
```
akukanov, line 111: Is it OK for `try_acquire` to have no memory fence in the case of failure? The current code has an acquire fence in that case, issued by `compare_exchange_strong`. What did TBB 2020 use?

include/oneapi/tbb/spin_mutex.h
```diff
@@ -68,14 +69,18 @@ class spin_mutex {
...
+            backoff.pause();
+        }
         call_itt_notify(acquired, this);
     }
 
     //! Try acquiring lock (non-blocking)
     /** Return true if lock acquired; false otherwise. */
     bool try_lock() {
-        bool result = !m_flag.exchange(true);
+        bool result = !m_flag.load(std::memory_order_relaxed) && !m_flag.exchange(true);
```
akukanov, line 82: Same question here.

include/oneapi/tbb/queuing_mutex.h
kboyarinov: As far as I understand the old `atomic.h` and `tbb_machine.h`, the old code used `compare_and_swap<tbb::release>` in this place that resulted in the acquire or a full fence in case of failure. 
On one hand, adding relaxed load changes the fence in case of failure. On the other hand, we do not explicitly document the memory semantics and the [C++ standard Mutex requirements](https://en.cppreference.com/w/cpp/named_req/Mutex#:~:text=If%20try_lock()%20succeeds%2C%20prior%20unlock()%20operations%20on%20the%20same%20object%20synchronize%2Dwith%20this%20operation%20(equivalent%20to%20release%2Dacquire%20std%3A%3Amemory_order).%20lock()%20does%20not%20synchronize%20with%20a%20failed%20try_lock().) allows to have no fence in this case.

So I would say the user should not expect the acquire fence in case of failure, in spite of the current implementation that provides it.

I also did not investigate if changing the memory fence in load to be acquire fence affects the performance.

akukanov (approved):

PR #1835, opened 2025-09-04
Stop enforcing ITT domains to be active

### LINKED ISSUE #1823: [ITT] Domain flag being set by TBB, even if it is being reset in an ITT collector
https://github.com/uxlfoundation/oneTBB/blob/c4f2d38b9426ca896c01660a35e627f2102abacf/src/tbb/profiling.cpp#L59

Currently when a domain is created, the ITT collector can attempt to filter out the domain by setting the domain->flag = 0. However, this is being overwritten to always be active. This requires collectors to implement expensive lookup to ensure TBB domains can be filtered. Is this what is intended in the implementation?

akukanov:
### Description 

We should not override the ITT domain flags, keeping enabling or disabling of the domains under the control of a collector/tool.

Fixes #1823.

### Type of change

### Tests
- [x] not needed - the change only affects analysis by tools and cannot be tested in a "normal" execution environment

### Documentation
- [x] not needed

- [x] No

Co-authored by @tovinkere

### DIFF
--- a/src/tbb/profiling.cpp
+++ b/src/tbb/profiling.cpp
@@ -1,5 +1,6 @@
 /*
-    Copyright (c) 2005-2022 Intel Corporation
+    Copyright (c) 2005-2025 Intel Corporation
+    Copyright (c) 2025 UXL Foundation Contributors
 
     Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


isaevil (approved):
LGTM

PR #1838, opened 2025-09-05
Update status of Custom Assertion Handler feature

Alexandr-Konovalov:
### Description 
Move to a different namespace
Add feature presence flag
Update documentation

### Type of change

- [ ] Yes
- [X] No
- [ ] Unknown

### DIFF
--- a/test/tbb/test_global_control.cpp
+++ b/test/tbb/test_global_control.cpp
@@ -310,5 +310,5 @@ TEST_CASE("Check that namespace injection allows to provide TBB 2020 source comp
     tbb::assertion_handler_type new_handler = [](const char*, int, const char*, const char*) {};
 
     tbb::assertion_handler_type old_handler = tbb::set_assertion_handler(new_handler);
-    REQUIRE(old_handler != nullptr);
+    REQUIRE(old_handler != new_handler);
 }


rfcs/supported/assertion_handler/README.md
```diff
@@ -70,10 +72,11 @@ namespace tbb {
     //! New function not present in TBB 2020, following std::get_terminate pattern.
     assertion_handler_type get_assertion_handler() noexcept;
 #endif
-}
+}}}
 ```
 
-Applications that used the custom assertion handler in TBB 2020 can migrate to this proposal with no changes.
+Applications that used the custom assertion handler in TBB 2020 can migrate to this proposal with no changes
+after manuall injection of namespace `oneapi::tbb::ext` to namespace `tbb`.
```
aleksei-fedotov: A typo:
```suggestion
after manual injection of namespace `oneapi::tbb::ext` to namespace `tbb`.
```

include/oneapi/tbb/global_control.h
```diff
@@ -202,12 +202,19 @@ using detail::d1::attach;
...
 using detail::r1::unsafe_wait;
+} // namespace v1
+
+namespace ext {
+inline namespace v1 {
 #if !__TBB_DISABLE_SPEC_EXTENSIONS
-using detail::r1::assertion_handler_type;
-using detail::r1::set_assertion_handler;
-using detail::r1::get_assertion_handler;
+#define TBB_EXT_CUSTOM_ASSERTION_HANDLER 1
```
kboyarinov: Does it make sense to move this macro to `_config.h`? 
I think it may be useful to conditionally include `global_control.h` only if the extension is available:

```cpp
#include <oneapi/tbb/version.h>
#if TBB_EXT_CUSTOM_ASSERTION_HANDLER
#include <oneapi/tbb/global_control.h>
#endif
```

include/oneapi/tbb/global_control.h
```diff
@@ -202,12 +202,19 @@ using detail::d1::attach;
...
+
+using ::tbb::detail::r1::assertion_handler_type;
+using ::tbb::detail::r1::set_assertion_handler;
+using ::tbb::detail::r1::get_assertion_handler;
 #endif
-} // namespace v1
+} // inline namespace v1
+}
```
kboyarinov: Just for consistency:)

```suggestion
} // namespace ext
```

include/oneapi/tbb/global_control.h
Alexandr-Konovalov: I don't think such code is realistic, but I like the idea to keep all feature macros in one place.

kboyarinov (dismissed):
LGTM

include/oneapi/tbb/detail/_config.h
```diff
@@ -545,4 +546,8 @@
 #define __TBB_PREVIEW_BLOCKED_ND_RANGE_DEDUCTION_GUIDES 1
 #endif
 
+#if !__TBB_DISABLE_SPEC_EXTENSIONS
```
aleksei-fedotov, line 549: User-facing macros are not prefixed with double underscore. Those are usually used for internal ones.

kboyarinov: As far as I remember, we agreed not to introduce the public macro for the upcoming release.

akukanov: Not a public opt-in/opt-out macro, that's correct.

include/oneapi/tbb/detail/_config.h
```diff
@@ -545,4 +546,8 @@
...
+#define TBB_EXT_CUSTOM_ASSERTION_HANDLER 1
```
akukanov: Is the macro defined to a true/false value or a version value? If the former, should there be any value at all?
@vossmjp 

By the way, should we have a test that this macro is defined (to a certain value)?

rfcs/supported/assertion_handler/README.md
```diff
@@ -70,10 +72,11 @@ namespace tbb {
     //! New function not present in TBB 2020, following std::get_terminate pattern.
     assertion_handler_type get_assertion_handler() noexcept;
 #endif
-}
+}}}
 ```
 
-Applications that used the custom assertion handler in TBB 2020 can migrate to this proposal with no changes.
+Applications that used the custom assertion handler in TBB 2020 can migrate to this proposal with no changes
+after manual injection of namespace `oneapi::tbb::ext` to namespace `tbb`.
```
akukanov:
```suggestion
Applications that used the custom assertion handler in TBB 2020 can migrate to this proposal with minimal changes
by adding the `set_assertion_handler` function name to `namespace tbb`:
namespace tbb {
    using oneapi::tbb::ext::set_assertion_handler;
}
```

include/oneapi/tbb/global_control.h
akukanov: We will need to document where the feature test macro is defined. The reasonable options are:

1.  the feature header (`global_control.h`);
2. above + the version/configuration header (`version.h`);
3. any library header.

For example, the C++ standard library uses the option 2, and oneDPL uses the option 3.

include/oneapi/tbb/detail/_config.h
vossmjp: For a non-extension feature, the user can check ONETBB_SPEC_VERSION to determine which version is supported.  Since an EXT is not in the spec, seems like it would be good to have a version value in case it changes later in an incompatible way. We discussed using values like C++ feature test macros, so year and month. For this release, that would be 202510 to align with the upcoming release.

Alexandr-Konovalov: > By the way, should we have a test that this macro is defined (to a certain value)?

Thanks, with a non-binary value it makes some sense.

include/oneapi/tbb/global_control.h
Alexandr-Konovalov: Interesting. Option (1) prevent us from renaming the header for the feature.

Alexandr-Konovalov: I'm for (2), as (3) is too intrusive (we must be sure, that any public library header provides something, that leads to inter-component coupling). And (1) is too tightly couples the header and the feature.

kboyarinov: The option (2) seems preferable for me as well. For (3), we will ideally need to test that the macro is present in each public header. For (2), we can simply extend feature test (`test_global_control` in this case) and `test_version` (which should be introduced in tbb subdirectory to test the extensions)

Alexandr-Konovalov: >and test_version (which should be introduced in tbb subdirectory to test the extensions)

Can it be done just in `test_global_control`? I mean, all we can test for now we may test in `test_global_control`. And new almost empty `tbb/test_version` is an additional overhead, as it's new `.exe`, new compilation etc. Can we add TODO that the special test will be created as soon as 2nd extension added?

kboyarinov: I think it can be done only in `test_global_control` for now. Adding a TODO sounds good as well.

Alexandr-Konovalov: Done.

rfcs/supported/assertion_handler/README.md
```diff
@@ -90,6 +92,10 @@ be enabled: `set_assertion_handler` and `get_assertion_handler` will be declared
 value before including oneTBB headers will disable the extension: these declarations will be excluded from
 the public API, and the library will always use the default assertion behavior.
 
+The availability of the extension can be checked with `TBB_EXT_CUSTOM_ASSERTION_HANDLER` macro. The value of it
+will be increased after adding of incompatibe changes. The macro is available after inclusion of
+`oneapi/tbb/version.h` header.
```
akukanov: I would not limit version increases to incompatible changes. Any modification of the API might deserve an increase of the macro value, so that the code can check if it can use these modifications.

Also, if we follow the C++ model for feature test macros, it should be available via both the feature header and the version header. Sorry if that was not clear in my previous comment.

include/oneapi/tbb/global_control.h
akukanov: > For (3), we will ideally need to test that the macro is present in each public header.

In oneDPL, there is no such extensive testing. In fact, the oneDPL macros are not said to be defined in any specific header - these are just "defined" by the implementation.

The option (2) sounds reasonable as a stated "guarantee".

rfcs/supported/assertion_handler/README.md
```diff
@@ -70,10 +72,16 @@ namespace tbb {
...
 ```
 
-Applications that used the custom assertion handler in TBB 2020 can migrate to this proposal with no changes.
+Applications that used the custom assertion handler in TBB 2020 can migrate to this proposal with minimal changes
+by adding the `set_assertion_handler` function name to `namespace tbb`:
+```cpp
+namespace tbb {
+    using oneapi::tbb::ext::set_assertion_handler;
+}
+```
```
akukanov: Should we add a test for such injection as well?

include/oneapi/tbb/global_control.h
vossmjp: Option 2 sounds good to me too. Seems like the most logical places to expect the macros to be, whereas 3 seems like overkill.

rfcs/supported/assertion_handler/README.md
```diff
@@ -90,6 +92,10 @@ be enabled: `set_assertion_handler` and `get_assertion_handler` will be declared
 value before including oneTBB headers will disable the extension: these declarations will be excluded from
 the public API, and the library will always use the default assertion behavior.
 
+The availability of the extension can be checked with `TBB_EXT_CUSTOM_ASSERTION_HANDLER` macro. The value of it
+will be increased after adding of incompatibe changes. The macro is available after inclusion of
```
vossmjp:
```suggestion
The availability of the extension can be checked with `TBB_EXT_CUSTOM_ASSERTION_HANDLER` macro. The value of the macro
will be increased after adding of incompatibe changes. The macro is available after inclusion of
```

rfcs/supported/assertion_handler/README.md
Alexandr-Konovalov: Thank you, missed.

rfcs/supported/assertion_handler/README.md
```diff
@@ -84,6 +93,10 @@ be enabled: `set_assertion_handler` and `get_assertion_handler` will be declared
 value before including oneTBB headers will disable the extension: these declarations will be excluded from
 the public API, and the library will always use the default assertion behavior.
 
+The availability of the extension can be checked with `TBB_EXT_CUSTOM_ASSERTION_HANDLER` macro. The value of the macro
+will be increased after adding of incompatibe changes. The macro is available after inclusion of
+`oneapi/tbb/version.h` header.
```
akukanov:
```suggestion
The availability of the extension can be checked with the `TBB_EXT_CUSTOM_ASSERTION_HANDLER` macro
after including either the `oneapi/tbb/global_control.h` or the `oneapi/tbb/version.h` header.
The value of the macro should be increased when observable modifications are made to the feature.
```

include/oneapi/tbb/global_control.h
Alexandr-Konovalov: And as we have both `version.h` and `global_control.h`, adding test_tbb_version.cpp became imminent.

test/tbb/test_tbb_version.cpp
```diff
@@ -0,0 +1,29 @@
+/*
+    Copyright (c) 2025 Intel Corporation
+    Copyright (c) 2025 UXL Foundation Contributors
```
kboyarinov: Intel copyright should be omitted for the new files:
```suggestion
    Copyright (c) 2025 UXL Foundation Contributors
```

test/CMakeLists.txt
```diff
@@ -502,6 +502,7 @@ if (TARGET TBB::tbb)
     tbb_add_test(SUBDIR tbb NAME test_environment_whitebox DEPENDENCIES TBB::tbb)
     tbb_add_test(SUBDIR tbb NAME test_hw_concurrency DEPENDENCIES TBB::tbb)
     tbb_add_test(SUBDIR tbb NAME test_eh_thread DEPENDENCIES TBB::tbb)
+    tbb_add_test(SUBDIR tbb NAME test_tbb_version DEPENDENCIES TBB::tbb)
```
kboyarinov, line 506: UXL copyright should be added for this file

test/tbb/test_global_control.cpp
```diff
@@ -293,3 +298,17 @@ TEST_CASE("Using custom assertion handler to test failure on invalid max_allowed
...
+    using oneapi::tbb::ext::assertion_handler_type;
+}
+
+//! Check that namespace injection allows to provide TBB 2020 source compatibility
+//! \brief \ref interface
+TEST_CASE("Check that namespace injection allows to provide TBB 2020 source compatibility") {
+    tbb::assertion_handler_type new_handler = [](const char*, int, const char*, const char*) {};
+
+    tbb::assertion_handler_type old_handler = tbb::set_assertion_handler(new_handler);
+    (void)old_handler;
```
kboyarinov: May be it makes sense to either remove the `old_handler` variable or to add some simple check for it. 
Something like `REQUIRE(old_handler != nullptr)`

test/CMakeLists.txt
Alexandr-Konovalov: Done

test/tbb/test_global_control.cpp
Alexandr-Konovalov: Strange, but RFC doesn't guarantee `old_handler != nullptr`.

PR #1839, opened 2025-09-05
Fix assertion in task_handle_task::finalize

### LINKED ISSUE #1834: Assertion failure in task_handle destruction due to object size mismatch in polymorphic cleanup：m_private_counter >= 0 failed (located in the destroy function, line in file: 141)
# Summary
The task_handle_task base class assumes it knows the size of derived objects during destruction, causing memory pool counter mismatches and assertion failures. When task_handle objects are destroyed through their unique_ptr custom deleter, destruction happens through base class pointer, leading to incorrect object size calculations and race conditions.

# Version
oneTBB 2022.0.0

# Environment
Hardware: Apple Silicon (M4 Pro) 
OS: macOS (all versions)
Compiler: Apple Clang (debug builds)
Build Type: Debug 

# Observed Behavior

[TBB_DESTROY] final_private_counter=-1 (will assert if < 0)
[TBB_DESTROY] *** ASSERTION WILL FAIL ***
[TBB_DESTROY] This indicates a race condition in USD's TBB usage
**Assertion m_private_counter >= 0 failed (located in the destroy function, line in file: 273)**
The issue occurs specifically on macOS debug builds when task_handle objects are destroyed, causing memory pool counter mismatches due to debug build object padding and incorrect object size assumptions during polymorphic destruction.

- USD Test Case:

The issue was discovered while running USD's test suite on macOS debug builds, particularly in tests involving parallel processing with task_group and task_handle objects. The assertion failure occurs during test cleanup when task_handle objects are automatically destroyed.
Detailed Race Condition Logs:
When running ctest -R testUsdCatMask -V with OpenUSD, we get the following logs explaining the race condition:
[TBB_CLEANUP] cleanup_list completed, total_removed=40
[TBB_DESTROY] private_list cleaned, removed=40
[TBB_DESTROY] after private cleanup, counter=1
[TBB_DESTROY] public_list exchanged, list=0x1053dfa00
[TBB_CLEANUP] cleanup_list starting, list=0x1053dfa00 thread=0x200d7df00
[TBB_CLEANUP] cleaning object=0x1053dfa00 next=0x1053ff500 count=1
[TBB_CLEANUP] cleaning object=0x1053ff500 next=0x0 count=2
[TBB_CLEANUP] cleanup_list completed, total_removed=2
[TBB_DESTROY] public_list cleaned, removed=2
[TBB_DESTROY] final_private_counter=-1 (will assert if < 0)
[TBB_DESTROY] *** ASSERTION WILL FAIL ***
[TBB_DESTROY] This indicates a race condition in USD's TBB usage
[TBB_DESTROY] private_removed=40 public_removed=2 total_removed=42
[TBB_DESTROY] Expected objects=43 Actual counter=42 Difference=1 (migrated objects)
[TBB_DESTROY] CONCLUSION: 1 object(s) were migrated from other thread pools to this pool
[TBB_DESTROY] These migrated objects caused the counter mismatch
Assertion m_private_counter >= 0 failed (located in the destroy function, line in file: 273)
Detailed description: Private counter may not be less than 0
The logs show that 1 object was "migrated" between thread pools due to incorrect object size calculations during destruction, causing a counter mismatch of 1 object.

# Expected Behavior
Object destruction should use correct object sizes for memory pool operations, maintaining consistent counter values and avoiding assertion failures, regardless of debug build padding.

# Steps To Reproduce
Build oneTBB in debug mode on macOS
Build USD with debug configuration using oneTBB 2022.0.0
Run ctest -R testUsdCatMask -V with OpenUSD
Observe the detailed race condition logs and assertion failure during test cleanup

# Potential Solution:
One possible approach to address this issue would be to:
Make m_allocator protected in task_handle_task to allow derived classes access，change finalize() to pure virtual in the base class to ensure proper polymorphic destruction, implement finalize() override in function_task derived class with correct object size handling.
This would ensure all destruction paths use the correct object sizes through virtual dispatch
eg：
```
diff --git a/include/oneapi/tbb/detail/_task_handle.h b/include/oneapi/tbb/detail/_task_handle.h
index 26212b46..a4d7e9b1 100644
--- a/include/oneapi/tbb/detail/_task_handle.h
+++ b/include/oneapi/tbb/detail/_task_handle.h
@@ -36,11 +36,8 @@ class task_handle_task : public d1::task {
     std::uint64_t m_version_and_traits{};
     d1::wait_tree_vertex_interface* m_wait_tree_vertex;
     d1::task_group_context& m_ctx;
+protected:
     d1::small_object_allocator m_allocator;
 public:
-    void finalize(const d1::execution_data* ed = nullptr) {
-        if (ed) {
-            m_allocator.delete_object(this, *ed);
-        } else {
-            m_allocator.delete_object(this);
-        }
-    }
+    virtual void finalize(const d1::execution_data* ed = nullptr) = 0;
 
     task_handle_task(d1::wait_tree_vertex_interface* vertex, d1::task_group_context& ctx, d1::small_object_allocator& alloc)
         : m_wait_tree_vertex(vertex)
```
And:
```
diff --git a/include/oneapi/tbb/task_group.h b/include/oneapi/tbb/task_group.h
index 393703f4..71cf6ec6 100644
--- a/include/oneapi/tbb/task_group.h
+++ b/include/oneapi/tbb/task_group.h
@@ -84,6 +84,13 @@ class function_task : public task_handle_task  {
     const F m_func;
 
 private:
+    void finalize(const d1::execution_data* ed = nullptr) override {
+        if (ed) {
+            m_allocator.delete_object(this, *ed);
+        } else {
+            m_allocator.delete_object(this);
+        }
+    }
     d1::task* execute(d1::execution_data& ed) override {
         __TBB_ASSERT(ed.context == &this->ctx(), "The task group context should be used for all tasks");
         task* res = task_ptr_or_nullptr(m_func);
```

While the proposed solution addresses the technical issue, making finalize() pure virtual could be considered a breaking change to the header API. According to oneTBB's versioning rules, such changes might require namespace changes to maintain compatibility guarantees. Could you suggest alternative solutions that maintain API compatibility while still addressing the core issue?

kboyarinov:
### Description 
Use the correct size of the task in task_handle_task::finalize.
Since introducing new virtual functions results in ABI breaks, we need to store a pointer to the instantiation of the finalize function in place of task_handle_task::m_version_and_traits to preserve backward compatibility.

Fixes #1834

### Type of change

### Tests

### Documentation
- [x] not needed

- [x] No

### DIFF
--- a/include/oneapi/tbb/detail/_task_handle.h
+++ b/include/oneapi/tbb/detail/_task_handle.h
@@ -32,10 +32,10 @@ namespace d1 { class task_group_context; class wait_context; struct execution_da
 namespace d2 {
 
 class task_handle;
-class task_handle_task;
 
 #if __TBB_PREVIEW_TASK_GROUP_EXTENSIONS
 
+class task_handle_task;
 class task_dynamic_state;
 
 struct successor_list_node {


include/oneapi/tbb/detail/_task_handle.h
```diff
@@ -136,32 +136,52 @@ class task_dynamic_state {
 };
 #endif // __TBB_PREVIEW_TASK_GROUP_EXTENSIONS
 
+template <typename DerivedType>
+void finalize_base(task_handle_task* p, d1::small_object_allocator& alloc, const d1::execution_data* ed) {
```
akukanov: Can we call the function `destroy_function_task` or `delete_function_task`? That's what it does.

include/oneapi/tbb/detail/_task_handle.h
```diff
@@ -136,32 +136,52 @@ class task_dynamic_state {
...
+    }
+}
+
 class task_handle_task : public d1::task {
-    std::uint64_t m_version_and_traits{};
+    // The pointer to the instantiation of finalize_base with the concrete derived type class for correct finalization
+    // Reuse the first std::uint64_t field (was m_version_and_traits) to preserve backward compatibility
+    using finalize_func_type = void (*)(task_handle_task*, d1::small_object_allocator&, const d1::execution_data*);
+    std::uint64_t m_finalize_func;
+    static_assert(sizeof(finalize_func_type) <= sizeof(m_finalize_func), "Cannot fit finalize pointer into std::uint64_t");
```
akukanov: Why not like this?
```suggestion
    static_assert(sizeof(finalize_func_type) <= sizeof(std::uint64_t), "Cannot fit finalize pointer into std::uint64_t");
    alignas(std::uint64_t) finalize_func_type m_finalize_func;
```

And I would also use destroy/delete in the names here.

include/oneapi/tbb/detail/_task_handle.h
```diff
@@ -136,32 +136,52 @@ class task_dynamic_state {
...
-            m_allocator.delete_object(this, *ed);
+        finalize_func_type finalize_func = reinterpret_cast<finalize_func_type>(m_finalize_func);
+        if (finalize_func != nullptr) {
+            // If the finalize function is set for the current instantiation - use it
+            (*finalize_func)(this, m_allocator, ed);
         } else {
-            m_allocator.delete_object(this);
+            // Otherwise, the object was compiled with the old version of the library
+            // preserve the previous behavior to avoid undefined behavior
+            finalize_base<task_handle_task>(this, m_allocator, ed);
```
akukanov: I wonder if it would be better to just let this memory leak if we cannot deallocate it properly. Deallocating memory with a wrong size scares me more than a leak - unless we know that `small_object_allocator` deals well with that.

include/oneapi/tbb/detail/_task_handle.h

include/oneapi/tbb/detail/_task_handle.h
kboyarinov: Forcing the alignment requirement on the first field alone is not sufficient. If `sizeof(pointer)` is less than `sizeof(std::uint64_t)`, the overall size of structure and the offsets of subsequent fields will be affected.
Consider a 32-bit build:
* `sizeof(pointer)` is `4`.
* `alignof(std::uint64_t)` is `4`.

In this case:
* The pointer to the destroy function occupies offset `0-4`.
* The next field (also a pointer) occupies offset `4-8`.

In the old layout:
* `m_version_and_traits` was of type `std::uint64_t` (size = 8, alignment = 4).
* It occupied offset `0-8`, and the next pointer - the offset `8-12`.

This difference in layout breaks the ABI compatibility.
See [this example](https://godbolt.org/z/EnqPfoc31) for more details.

If we don't want to keep the type of the first field as `std::uint64_t `and use `reinterpret_cast` to access the destroy function, we can manually 
insert padding to preserve the original layout:

```cpp
struct task_handle_task {
    alignas(std::uint64_t) destroy_func_type m_destroy_func;
    char padding[sizeof(std::uint64_t) - sizeof(destroy_func_type)];
    wait_tree_vertex_interface* m_wait_tree_vertex;
};
```

However, I prefer to keep the type as `std::uint64_t` and simply add a comment explainng the rationale.

include/oneapi/tbb/detail/_task_handle.h
kboyarinov: I agree. Replaced with a call to task destructor without deallocating the memory in else branch.

aleksei-fedotov (dismissed):
Looks good to me!

include/oneapi/tbb/detail/_task_handle.h
akukanov: Actually, making the next logical step, I would make it a static member function of `function_task`. This thought should have come earlier :).

include/oneapi/tbb/detail/_task_handle.h
akukanov: Thanks for the explanation; I should have noticed that myself.
You could use an unnamed union of `std::uint64_t` and the pointer - though this would complicate initialization a bit. Anyway, I do not mind `reinterpret_cast` either.

include/oneapi/tbb/detail/_task_handle.h
```diff
@@ -136,32 +136,56 @@ class task_dynamic_state {
...
     d1::small_object_allocator m_allocator;
 #if __TBB_PREVIEW_TASK_GROUP_EXTENSIONS
     std::atomic<task_dynamic_state*> m_dynamic_state;
 #endif
 public:
-    void finalize(const d1::execution_data* ed = nullptr) {
-        if (ed) {
-            m_allocator.delete_object(this, *ed);
+
+    void destroy(const d1::execution_data* ed = nullptr) {
```
akukanov, line 166: `destroy_this`? To make it clear that it's not `execution_data` being destroyed.

include/oneapi/tbb/task_group.h
```diff
@@ -99,7 +99,7 @@ class function_task : public task_handle_task  {
             next_task = successor_task;
         }
 #endif
-        finalize(&ed);
+        destroy(&ed);
```
akukanov: While not technically necessary, calling it as `this->destroy(&ed)` would make more obvious what is destroyed.

include/oneapi/tbb/detail/_task_handle.h
kboyarinov: Great idea! Applied

include/oneapi/tbb/detail/_task_handle.h
```diff
@@ -32,10 +32,10 @@ namespace d1 { class task_group_context; class wait_context; struct execution_da
 namespace d2 {
 
 class task_handle;
+class task_handle_task;
```
akukanov: Seems redundant now.

akukanov (approved):
Looks good.

PR #1845, opened 2025-09-11
Fix missing brief covers for test_tbb_version.cpp

Alexandr-Konovalov:
### Description 
In #1839 test/tbb/test_tbb_version.cpp was added. It in brief it misses "covers" part.

### Type of change

### Tests
- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/test/tbb/test_tbb_version.cpp
+++ b/test/tbb/test_tbb_version.cpp
@@ -14,7 +14,6 @@
     limitations under the License.
 */
 
-//! \file test_tbb_version.cpp
 //! Test for the availability of extensions
 //! \file test_tbb_version.cpp
 //! \brief Test for [configuration.feature_macros] specification


test/tbb/test_tbb_version.cpp
```diff
@@ -15,7 +15,7 @@
 */
 
 //! \file test_tbb_version.cpp
-//! \brief Test for the availability of extensions
+//! \brief Test for the availability of extensions, covers [configuration.feature_macros]
```
Alexandr-Konovalov: I'm not sure this is right section of the spec. Please correct me if I'm wrong.

kboyarinov: I would keep the wording consistent with other test files:

```suggestion
//! Test for the availability of extensions
//! \file test_tbb_version.cpp
//! \brief Test for [configuration.feature_macros] specification
```

kboyarinov: I would keep `[configuration.feature_macros]` for now since we don't have special section for extensions yet.

PR #1849, opened 2025-09-13
Use `dladdr` to keep `libtbbmalloc` loaded on POSIX

emilazy:
### Description 
This fixes issues that arise when `libtbbmalloc` cannot be loaded by relative path, makes the behaviour consistent with the Windows code path, and eliminates the duplication of `MALLOCLIB_NAME`.

### Type of change

### Tests

- [x] not needed

Probably too difficult to test. I confirmed that this stops the test suite crashing on program exit on macOS in Nixpkgs, though.

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown


emilazy:
It seems like `RTLD_NODELETE` may be a better solution to the general problem here, by the way. But I only went for the most direct local fix for the problem we were seeing.

emilazy:
Looks like this is (nondeterministically?) causing problems on Linux, and I don’t know why :(

My guesses (e.g. needs `RTLD_GLOBAL`?) seem like they should apply equally to the previous state of the code.

emilazy:
I guess this was the same thing as:

https://github.com/uxlfoundation/oneTBB/blob/91af3aea3a2be85f889c3bb21a15c667e44564d4/src/tbb/dynamic_link.cpp#L382-L391

Switching to a private function fixed it. Let me know if you want to add an `#if` for this one too.

PR #1853, opened 2025-09-17
Use execution_slot to ensure backwards compatibility

vossmjp:
### Description 
Add to test_task_arena to exercise an exported function that is no longer used in headers.

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown

### DIFF
--- a/test/tbb/test_task_arena.cpp
+++ b/test/tbb/test_task_arena.cpp
@@ -51,8 +51,8 @@
 
 //--------------------------------------------------//
 // Validation function to check that current_thread_index() and execution_slot() return
-// the same value. The exported function tbb::detail::r1::execution_slot is maintained
-// for backwards compatibility only.
+// the same value for a thread running within initialized task_arena. The exported function
+// tbb::detail::r1::execution_slot is maintained for backwards compatibility only.
 void check_slot_compatibility(int expected_idx, const tbb::task_arena& arena) {
     int execution_slot_idx = int(tbb::detail::r1::execution_slot(arena));
     CHECK_MESSAGE(expected_idx == execution_slot_idx,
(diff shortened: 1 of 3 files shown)


test/tbb/test_task_arena.cpp
```diff
@@ -34,6 +34,7 @@
 #include "tbb/spin_mutex.h"
 #include "tbb/spin_rw_mutex.h"
 #include "tbb/task_group.h"
+#include "tbb/task.h"
```
akukanov: Would it make sense to move the entry point declaration to `task_arena.h`, and perhaps add a comment about keeping it for compatibility?

test/tbb/test_task_arena.cpp
```diff
@@ -48,6 +49,16 @@
 //! \file test_task_arena.cpp
 //! \brief Test for [scheduler.task_arena scheduler.task_scheduler_observer] specification
 
+//--------------------------------------------------//
+// Validation function to check that current_thread_index() and execution_slot() return
+// the same value. The exported function tbb::detail::r1::execution_slot is maintained
+// for backwards compatibility only.
```
aleksei-fedotov: Perhaps, to avoid future confusion:
```suggestion
// Validation function to check that current_thread_index() and execution_slot() return
// the same value for a thread running within initialized task_arena. The exported function
// tbb::detail::r1::execution_slot is maintained for backwards compatibility only.
```
That's is because for a non-initialized thread, the `tbb::this_task_arena::current_thread_index()` returns `-2`, while for `execution_slot(task_arena&)` it is either `-1` or something > 0.

vossmjp: Good point.

test/tbb/test_task_arena.cpp
vossmjp: Yes, I think that's a good idea, since now it is most closely related to task_arena.

vossmjp: Move it and removed this include.

PR #1860, opened 2025-09-22
Remove dead function from task_handle class

kboyarinov:
### Description 
Remove `task_handle::release` function since it is not used after merging the dynamic dependencies PR.

Before merging the dynamic dependencies PR, `task_handle::release` was called by `task_handle_accessor::release`.
Since all other functions in `task_handle_accessor` (`ctx_of` and `get_task_dynamic_state`) calls corresponding function from `task_handle::m_handle` field directly, the logic of `release` was also aligned to use `task_handle::m_handle::release` instead of `task_handle::release` (which is doing the same).

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown

### DIFF
--- a/include/oneapi/tbb/detail/_task_handle.h
+++ b/include/oneapi/tbb/detail/_task_handle.h
@@ -267,10 +267,6 @@ private:
 #endif
 
     task_handle(task_handle_task* t) : m_handle {t}{}
-
-    d1::task* release() {
-       return m_handle.release();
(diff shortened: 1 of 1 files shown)


vossmjp (approved):
Looks good to me too.

PR #1862, opened 2025-09-29
RFC for Waiting for Individual Tasks in task_group

kboyarinov:
### Description 
Add an RFC for new functions that allow waiting for Individual Tasks in `task_group`:

```cpp
namespace oneapi {
namespace tbb {
enum class task_status {
   not_complete,
   complete,
   canceled
};

class task_group {
    task_status wait_for(task_completion_handle&);
    task_status run_and_wait_for(task_handle&&);
};

class task_arena {
   task_status wait_for(task_completion_handle&);
};

} // namespace tbb
} // namespace oneapi
```

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown


rfcs/proposed/task_group_wait_single_task/README.md
```diff
@@ -0,0 +1,496 @@
...
+    if (bypass_task && !bypass_allowed) {
+        spawn(*bypass_task);
+        return nullptr;
+    } else {
+        return bypass_task;
+    }
+}
+```
+
+This solution does not restrict the thread that is notified by another thread of the task completion from entering the bypass loop of the currently
```
vossmjp: This line is hard to follow.

rfcs/proposed/task_group_wait_single_task/README.md
```diff
@@ -0,0 +1,496 @@
...
+Ideally, the waiting thread should only execute tasks that contribute to the completion of the awaited task. In such a model, all tasks
+within the same subgraph would need to share a common isolation tag. In theory, a successor could inherit the isolation tag from
+its predecessor. However, multiple predecessors will have different tags.
+
+This would require changes to the current isolation mechanism to support multiple isolation tags, allowing a thread waiting for a successor
+to execute tasks with any of its predecessor's tags. 
+
+As an initial step, it makes sense to isolate the waiting thread so that it only executes tasks related to the same ``task_group`` as the awaited task - similar to the improvement described in the [RFC for another overload of ``task_arena::wait_for``](../task_arena_waiting/task_group_interop.md).
+
+## Exit Criteria and Open Questions
```
vossmjp, line 352: Is the intention to go through an experimental phase, or straight to production?

rfcs/proposed/task_group_wait_single_task/README.md
```diff
@@ -0,0 +1,496 @@
...
+
+The current waiting functions return a ``task_group_status``, which can have three possible values:
+* ``tbb::task_group_status::not_complete`` - The group execution is not cancelled, and not all tasks in the group have completed.
+* ``tbb::task_group_status::complete`` - The group execution is not cancelled, and all tasks in the group have completed.
+* ``tbb::task_group_status::canceled`` - The group execution is cancelled, and the completion status of tasks is unknown.
+
+Since the new waiting functions track the progress of a single task, returning a ``task_group_status`` may be misleading.
+If the group execution is cancelled, the tracked task may still execute, and returning ``canceled`` does not accurately reflect the task's 
+completion status.
+If execution is not cancelled, the function would need to track whether other tasks remain in the group and return ``not_complete` if any are still pending.
```
vossmjp: Will transfer of a task_completion lead to any confusion around cancellation, if for example, the original task has executed but the task that was transfered to is canceled?  The user will see canceled as a status even though the initial task did complete. I think it's fine and will just need to be well documented in transfer.

rfcs/proposed/task_group_wait_single_task/README.md
kboyarinov: Rephrased and added more details.

rfcs/proposed/task_group_wait_single_task/README.md
kboyarinov: Initially to the experimental since the proposed API depend on the experimental `task_completion_handle`.

rfcs/proposed/task_group_wait_single_task/README.md
kboyarinov: I think that if the completion of the task was transferred to another task, the task status after the wait is also transferred. I agree that it should be clearly documented. Added explicit mention in the RFC.

rfcs/proposed/task_group_wait_single_task/README.md
```diff
@@ -0,0 +1,502 @@
...
+
+If the task group execution is canceled and the originally executed task has transferred its completion to another task that was canceled, the
+waiting function will return ``task_status::canceled``.
+
+To address this, a new enum ``task_status`` is proposed to track the status of the awaited task. ``task_status::complete`` indicates that the tracked 
+task was executed and completed, while ``task_status::canceled`` signifies that the task was not executed due to group cancellation.
```
akukanov: I would consider extending the existing enum with one more value, named e.g. ``task_complete``, which would be returned by the single-task waiting functions instead of ``complete``.

rfcs/proposed/task_group_wait_single_task/README.md
```diff
@@ -0,0 +1,502 @@
...
+
+enum class task_status {
+    not_complete,
+    complete,
+    canceled
+};
+
+class task_group {
+    task_status wait_for(task_completion_handle& comp_handle);
+    task_status run_and_wait_for(task_handle&& handle);
```
akukanov: Should there be `run_and_wait_for(function)` as well?

rfcs/proposed/task_group_wait_single_task/README.md
```diff
@@ -0,0 +1,502 @@
...
+* 5 [Implementation Details](#implementation-details)
+* 5.1 [Avoid Bypassing the Successor](#avoid-bypassing-the-successor)
+* 6 [Alternative Implementation Approaches](#alternative-implementation-approaches)
+* 6.1 [Assign Single ``wait_context`` to the Task](#assign-single-wait_context-to-the-task)
+* 6.2 [Avoid ``wait_context`` Usage](#avoid-wait_context-usage)
+
+## Introduction
+The ``tbb::task_group`` class represents a collection of tasks that execute concurrently. Tasks can be dynamically added to the group during executing.
+
+The [Task Group Dynamic Dependencies preview feature](../task_group_dependencies/README.md) extends the ``task_group`` API to support task dependencies, ensuring that a *successor* task begins execution only after all its *predecessors* have completed.
```
akukanov: The link does not work for me.

rfcs/proposed/task_group_wait_single_task/README.md
akukanov: I find `wait_for_task` and especially `run_and_wait_for_task` slightly better names. Or perhaps `wait_task/run_and_wait_task`, if shorter names are preferred.
Being more explicit helps to avoid misinterpretation and/or questions of the difference between `run_and_wait` and `run_and_wait_for`, which look the same except for the name.

That does not apply to the proposed function in `task_arena`.

rfcs/proposed/task_group_wait_single_task/README.md
```diff
@@ -0,0 +1,502 @@
...
+### Member functions of ``task_arena`` class
+
+```cpp
+task_status wait_for(task_completion_handle& comp_handle);
+```
+
+Waits for the completion of the task represented by ``comp_handle``.
+If completion was transferred to another task using ``tbb::task_group::transfer_completion_to``, the function waits for completion of that task.
+
+This is semantically equivalent to: ``execute([&] { tg.wait_for(comp_handle); })``.
```
akukanov: Are we able to get the task group out of a completion handle? If not, are we able to implement the function without requiring a task group to be also provided by the caller?

rfcs/proposed/task_group_wait_single_task/README.md
```diff
@@ -0,0 +1,502 @@
...
+The notified ``wait_context`` is the same context awaited by the calling thread.
+
+If ``end_task`` is bypassed, the calling thread will not exit until all tasks in the graph have completed. If additional tasks follow ``end_task``,
+the thread will only exit after each is bypassed sequentially.
+
+However, since the current thread notifies the ``wait_context`` it is waiting on via the corresponding waiter node, it is more appropriate to avoid
+bypassing the next task. Instead, the task should be spawned, and ``run_and_wait_for`` should exit after executing ``middle_task``.
+
+This can be implemented by storing a pointer to the innermost ``wait_context`` that the current thread is waiting on, within the task
+dispatcher associated with the calling thread.
```
akukanov: At this stage, I would prefer a simpler implementation that just does not bypass if any waiter is found in the list.

rfcs/proposed/task_group_wait_single_task/README.md
```diff
@@ -0,0 +1,502 @@
...
+    }
+}
+```
+
+While the proposed solution avoids task bypassing when the waiting thread executes the task it is explicitly waiting for, bypassing remains
+unrestricted in other scenarios. For example, if the waiting thread executes a task unrelated to the awaited one, such as a ``parallel_for``
+task in the same arena, it may still enter the bypass loop, even if the reference count of the awaited ``wait_context`` is zero.
+
+An alternative approach to address this limitation is to implement a general mechanism within the scheduler that forces the thread to exit the
+bypass loop and spawn the returned task if further execution should not be continued (i.e., ``waiter.continue_execution()`` returns ``false``).
```
akukanov: That makes most sense to me in the long run. Does it seem too complicated or risky to try?

rfcs/proposed/task_group_wait_single_task/README.md
```diff
@@ -0,0 +1,502 @@
...
+
+## Alternative Implementation Approaches
```
akukanov, line 455: What are benefits and downsides of the alternative approaches to the recommended one?

rfcs/proposed/task_group_wait_single_task/README.md
vossmjp: If it's a function, then there cannot be predecessors since no task_completion_handle was created. So you'd execute that function as a task and wait for just that task (which has no dependencies). So why run and wait for it in the task_group at all? Why not just call the function? But then maybe it makes sense if the completion is transferred to work created inside of the task?

rfcs/proposed/task_group_wait_single_task/README.md
kboyarinov: When other waiting functions return `task_group_status::canceled`, they rely on the cancellation of the group (i.e. cancellation of the associated `task_group_context`). Individual tasks can be both executed or canceled. 
For single-task waiting functions, this flag would mean different thing - that the task we are waiting for was canceled.

And even if the task group execution was canceled, the single-task wait can return `task_complete`. And the returned `task_group_status` knows nothing about the status of task group. 

I think separating the statuses would be more obvious for users.

rfcs/proposed/task_group_wait_single_task/README.md
kboyarinov: If the function do not transfer the completion, usage of the function does not have much sense - the function would likely be executed by the calling thread and exit after it.
 
 But if the task transfers the completion, having a function may be useful. E.g. if the function implements a root divide-and-conquer reduce and there are several reductions in the same `task_group`:
 
 ```cpp
 auto function = []() {
     auto left_leaf = tg.defer(...);
     auto right_leaf = tg.defer(...);
     auto join = tg.defer(...);
     
     tbb::task_group::set_task_order(left_leaf, join);
     tbb::task_group::set_task_order(right_leaf, join);
     tbb::task_group::transfer_this_task_completion_to(join);
     
     tg.run(std::move(left_leaf));
     tg.run(std::move(right_leaf));
     tg.run(std::move(join));
 };
 
 tg.run_and_wait_for(function);
 ```
I propose to add it to the future enhancements & open questions sections since I don't know if mixing several reductions (or other similar tasks) in one `task_group` makes more sense than separating into several groups.

rfcs/proposed/task_group_wait_single_task/README.md
akukanov: > For single-task waiting functions, this flag would mean different thing - that the task we are waiting for was canceled.

No, it would mean the same thing - that the task group, which task we are waiting for, was cancelled. There is no way to cancel a single task.

> And even if the task group execution was canceled, the single-task wait can return `task_complete`. And the returned `task_group_status` knows nothing about the status of task group.

Sure, and I see no problem with that. The task was complete, while the status of the whole group is unknown. On the other hand, if the task was cancelled, then the whole group was cancelled.

rfcs/proposed/task_group_wait_single_task/README.md
kboyarinov: With our current implementation, `task_group` is not required for a single-task waiting. All we need is a `task_dynamic_state` and the associated `task_group_context`, both can be obtained from the task object.

In theory, we can implement `task_group::wait_task` (and even `run_and_wait_task` if we omit the "same task group" check) as a static function. But I have proposed member functions for consistency with other waiting functions.

I am not sure if implementing `task_arena::wait_for` wihout the task group is possible for any other TBB implementation. From the perspective of the further inclusion into oneTBB specification, it may make sense to add `task_group` argument into this function and keep it unused in our implementation. What do you think?

rfcs/proposed/task_group_wait_single_task/README.md
akukanov: I am fine to keep it as an open question for the future.

rfcs/proposed/task_group_wait_single_task/README.md
kboyarinov: I would try and prefer this approach in the long run as well. The only side-effect of this approach is potential behavior change in general.
E.g. a thread that is waiting for `parallel_for` to complete is taking a task belonging to other `parallel_for` and enters the bypass loop. This change will allow to avoid execution of the unrelated tasks and exit the first parallel for once it is really completed without the delay. 
Both solution with saving the innermost wait context and avoiding bypass in case of any waiters in the list would not affect the general behavior of TBB master and worker threads.

rfcs/proposed/task_group_wait_single_task/README.md
kboyarinov: The main benefit of the recommended implementation approach is that waiting for completion is guaranteed to be implemented using a single `r1::wait` or `r1::run_and_wait` in case of transferring. 

For both alternative approaches, we will need to switch to another waiting in case of transferring:

```cpp
task_dynamic_state* state = comp_handle.get_dynamic_state();

r1::wait(state->get_wait_context());

while (state->was_transferred()) {
    state = state->get_new_completion_point();
    r1::wait(state->get_wait_context());
}
```

With the recommended approach, in case of transferring the wait context pointer is migrating between tasks, hence we don't need to double check if we completion was transferred. If the wait context was released, the completion is guaranteed to happen (no matter of which "final" task in the transfer chain).

Another benefit is that the wait context is created only when the wait was requested (that is not true for the first alternative approach). 

I will add more details on the benefits and downsides into the RFC.

rfcs/proposed/task_group_wait_single_task/README.md
akukanov: Sure it is possible for any implementation, under the assumption that a `task_completion_handle` (and perhaps also `task_handle`) is always "bound" to a certain task group (that is, can keep a pointer/reference to the group). We just need to be clear about that, as well as at which point the binding happens (I guess that is at task creation, and not at submission). Otherwise, it is a mystery where `tg` comes from in the "equivalent" expression.

rfcs/proposed/task_group_wait_single_task/README.md
akukanov: > E.g. a thread that is waiting for parallel_for to complete is taking a task belonging to other parallel_for and enters the bypass loop. This change will allow to avoid execution of the unrelated tasks and exit the first parallel for once it is really completed without the delay.

Do you think there are cases where it might change user-observable behavior and/or guarantees of the TBB API beyond the usual randomness of task scheduling?

rfcs/proposed/task_group_wait_single_task/README.md
kboyarinov: For `canceled`, I agree, for `run_and_wait_task` it means the awaited task was not executed because the cancellation of the `task_group` execution.
For `complete`, I find it a bit misleading to have a status of `task_group` that knows nothing about the actual status of the group and serves the status of the task instead.

rfcs/proposed/task_group_wait_single_task/README.md
kboyarinov: Added as an open question

rfcs/proposed/task_group_wait_single_task/README.md
kboyarinov: Fixed

rfcs/proposed/task_group_wait_single_task/README.md
kboyarinov: Renamed

rfcs/proposed/task_group_wait_single_task/README.md
kboyarinov: I have added the clarification about what `tg` is in this case.

I agree that this can be approached by binding the handle to the `task_group`. But I think that we should not specify how binding should be implemented (to ensure the validity of our implementation).

rfcs/proposed/task_group_wait_single_task/README.md
kboyarinov: I agree, added this as an option to consider by the initial implementation.

rfcs/proposed/task_group_wait_single_task/README.md
kboyarinov: I don't think that some guarantees are affected since we don't specify which tasks are taken while waiting.
The observable behavior can change in theory - some tasks that were executed while waiting without the change will not be executed after the change., But I think it does not break the guarantees.

rfcs/proposed/task_group_wait_single_task/README.md
kboyarinov: Updated each section with the pros and cons.

rfcs/proposed/task_group_wait_single_task/README.md
vossmjp: Talking with @kboyarinov earlier today, we discussed that for an individual task, a good set of status values would be "executed" and "canceled", instead of complete and canceled. A task might finish before a task group is canceled. Or a long-running task could execute and then discover while executing, by querying  its task_group_context, that its group has been canceled and short-cut its execution. So for a specific task, "executed" simply means that the scheduler executed the task but does not imply anything about completion of the work or other work in the task group. And then "canceled" means the task never started to execute.

akukanov: There can be a semantic difference between a task being executed and a task being complete, if the completion was transferred. For me, "executed" means just "the task body has been run" while "complete" means logical completion, or essentially "task waiters and successors receive a signal". Since `wait_for` has the latter semantics, I really prefer "complete" or better "task_complete". It is also consistent with `transfer_this_task_completion_to`.

> For `complete`, I find it a bit misleading to have a status of task_group that knows nothing about the actual status of the group and serves the status of the task instead.

This is why I suggest to name it `task_group_status::task_complete` to indicate this is about a task and not the whole group.
And I still see no reason to have a separate enum class while an extra enum value seems to work just as good. Yet another reason for the single enum is the consistency of `task_arena::wait_for` overloads.

rfcs/proposed/task_group_wait_single_task/README.md
akukanov: True, we should not specify **how** the binding is to be implemented. When I said a task handle "can keep a pointer/reference", I did not mean it should, just that an implementation that takes this approach is valid and safe.

Specifying **when** binding happens is different, as it affects which task group to wait for, Waiting on a wrong task group could result in a program hang.

rfcs/proposed/task_group_wait_single_task/README.md
```diff
@@ -0,0 +1,489 @@
...
+
+The current ``task_group`` API allows waiting for the completion of all tasks in the group using ``task_group::wait`` or ``task_group::run_and_wait``.
+The waiting thread may participate in executing other tasks within the ``task_arena``, which may not be related to the specific ``task_group``.
+
+Although waiting for all tasks suits many use cases, some workflows would benefit from the ability to wait for individual tasks.
+
+Consider an example using an out-of-order SYCL queue.
+
+```cpp
+sycl::queue q{sycl::property::queue::in_order{false}};
```
akukanov: I do not think the `in_order` property can be constructed as `false`; see https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#api:property-queue-in-order.

rfcs/proposed/task_group_wait_single_task/README.md
```diff
@@ -0,0 +1,489 @@
...
+
+sycl::event task3 = q.submit([&](sycl::handler& h) {
+    h.depends_on(std::vector<sycl::event>{task1, task2});
+
+    // task1_body executes
+    // some_memory_ptr is set before task3_body runs
+    h.single_task(task3_body);
+});
+
+task3.wait(); // wait for event
```
akukanov:
```suggestion
task3.wait(); // wait a single task
```

rfcs/proposed/task_group_wait_single_task/README.md
```diff
@@ -0,0 +1,489 @@
...
+```cpp
+using event = tbb::task_completion_handle;
+
+class queue {
+    template <typename Body>
+    event single_task(Body body) {
+        tbb::task_handle task = m_task_group.defer(body);
+        tbb::task_completion_handle comp_handle = task;
+        m_task_group.run(std::move(task));
+        return {comp_handle};
```
akukanov:
```suggestion
        return comp_handle;
```

rfcs/proposed/task_group_wait_single_task/README.md
```diff
@@ -0,0 +1,489 @@
...
+
+One possible name for such an extension is ``task_arena::enqueue_and_wait_for(task_handle&& task)``.
+This function is semantically close to ``task_arena::execute`` (which may submit the task and wait for its completion), but it guarantees that
+the task is enqueued into the arena.
+
+A similar function could also be added to the ``tbb::this_task_arena`` namespace.
+
+### Work Isolation Issue
+
+Similar to other ``task_arena::wait_for`` overload that take a ``task_group`` argument, it makes sense to also consider work isolation for
```
akukanov:
```suggestion
Similar to another ``task_arena::wait_for`` overload that take a ``task_group`` argument, it makes sense to also consider work isolation for
```

rfcs/proposed/task_group_wait_single_task/README.md
```diff
@@ -0,0 +1,489 @@
...
+the APIs introduced in this proposal.
+If the wait is unrestricted, the waiting thread may execute any task from the arena, including those unrelated to the completion of the awaited
+task or its associated ``task_group``.
+
+Ideally, the waiting thread should only execute tasks that contribute to the completion of the awaited task. In such a model, all tasks
+within the same subgraph would need to share a common isolation tag. In theory, a successor could inherit the isolation tag from
+its predecessor. However, multiple predecessors will have different tags.
```
akukanov:
```suggestion
its predecessor. However, multiple predecessors may have different tags.
```

rfcs/proposed/task_group_wait_single_task/README.md
```diff
@@ -0,0 +1,489 @@
...
+```
+
+``run_and_wait_task`` will exit only after the ``join`` task completes. If several divide-and-conquer tasks are executed in the same ``task_group``, such a function can
+be used to wait for the completion of each individual task.
+
+## Exit Criteria and Open Questions
+
+The following questions should be resolved before promoting the feature out of the ``experimental`` stage.
+
+* Performance targets for this feature should be clearly defined.
```
akukanov:
```suggestion
* Performance targets for this feature should be clearly defined and met.
```

rfcs/proposed/task_group_wait_single_task/README.md
akukanov: I see it now described as a future enhancement, but not yet added to the list of open questions,

rfcs/proposed/task_group_wait_single_task/README.md
```diff
@@ -0,0 +1,489 @@
...
+corresponding ``run_and_wait_task`` call to complete.
+The notified ``wait_context`` is the same context awaited by the calling thread.
+
+If ``end_task`` is bypassed, the calling thread will not exit until all tasks in the graph have completed. If additional tasks follow ``end_task``,
+the thread will only exit after each is bypassed sequentially.
+
+However, since the current thread notifies the ``wait_context`` it is waiting on via the corresponding waiter node, it is more appropriate to avoid
+bypassing the next task. Instead, the task should be spawned, and ``run_and_wait_task`` should exit after executing ``middle_task``.
+
+For the initial implementation, it is proposed to completely avoid bypassing the task returned from the notification list and to spawn it.
```
akukanov:
```suggestion
For the initial implementation, it is proposed to completely avoid bypassing the task returned from the notification list and to spawn it instead.
```

rfcs/proposed/task_group_wait_single_task/README.md
```diff
@@ -0,0 +1,489 @@
...
+
+There are several approaches that can be implemented in the future to improve this approach.
```
akukanov:
```suggestion
There are several ways to improve the implementation in the future.
```

rfcs/proposed/task_group_wait_single_task/README.md
akukanov: So, are the alternative implementation approaches considered and rejected, or is a further discussion necessary?

rfcs/proposed/task_group_wait_single_task/README.md
```diff
@@ -0,0 +1,489 @@
...
+
+Since the example above also uses ``sycl::event::wait`` to wait for the completion of a single task (``task3``), implementing this behavior with
+a ``task_group`` requires support for waiting on a single ``task_completion_handle`` object.
+
+## Proposal
+
+The proposal aims to extend ``tbb::task_group`` and ``tbb::task_arena`` APIs by introducing new waiting functions that accept a
+``task_completion_handle`` argument.
+These functions exit when the task represented by the handle argument is completed or when the group execution is cancelled.
+The completion of other tasks within the ``task_group`` is not guaranteed.
```
vossmjp:
```suggestion
Other tasks within the ``task_group`` may not be complete when these new waiting function return.
```

vossmjp: As written, it sounds like calling these new function may result in other tasks not executing at all.

rfcs/proposed/task_group_wait_single_task/README.md
```diff
@@ -0,0 +1,489 @@
...
+namespace tbb {
+
+enum class task_status {
+    not_complete,
+    complete,
+    canceled
+};
+
+class task_group {
+    task_status wait_task(task_completion_handle& comp_handle);
```
vossmjp: I can't remember if we previously discussed this but `wait_task` seems to me like a getter that returns the wait task. Shouldn't this be something like `wait_for_task`.

rfcs/proposed/task_group_wait_single_task/README.md
```diff
@@ -0,0 +1,489 @@
...
+    task_status run_and_wait_task(task_handle&& handle);
```
vossmjp: Same issue here, why not `run_and_wait_for_task`

rfcs/proposed/task_group_wait_single_task/README.md
vossmjp: I understand your point, that `task_group_status::task_complete` can mean that the task completion handle was signaled because the task or the target of last transfer has finished executing. But how can a user know if the "completed" task has generated a full and valid result? It might have short-circuited due to a canceled task_group. I suppose we can say that `task_complete` only means that the task or transferred task signaled the completion handle and leave it up to the user to determine if that task was successful and finished its work. But I think it still would be better to provide a different name rather than using a form of `complete` which is used by `task_group::wait` to indicate that the work was finished successfully. It doesn't have to be `executed`.

rfcs/proposed/task_group_wait_single_task/README.md
akukanov: See https://github.com/uxlfoundation/oneTBB/pull/1862#discussion_r2391874376
I am OK with `wait_for_task`

kboyarinov: Agree, renamed to `wait_for_task`

rfcs/proposed/task_group_wait_single_task/README.md
kboyarinov: Done

rfcs/proposed/task_group_wait_single_task/README.md
kboyarinov: Fixed

rfcs/proposed/task_group_wait_single_task/README.md
kboyarinov: I believe that the natural support for completion transferring in the proposed approach provides a strong rationale for its implementation over other alternatives. If reviewers see it differently, we can discuss further.

rfcs/proposed/task_group_wait_single_task/README.md
```diff
@@ -0,0 +1,495 @@
...
+}
+
+class task_group {
+    task_group_status wait_for_task(task_completion_handle& comp_handle);
+    task_group_status run_and_wait_for_task(task_handle&& handle);
+};
+
+// Defined in <oneapi/tbb/task_arena.h>
+class task_arena {
+    task_group_status wait_for(task_completion_handle& comp_handle);
```
vossmjp, line 218: Ok so, the `task_arena::wait_for` that receives a task completion handle is an overload of `task_arena::wait_for`  that receives a task group. So a consistent name there. Do we want consistency with names in `task_group`, which now has `task_group::wait_for_task` and `task_group::wait`. Even though I recently suggested `task_group::wait_for_task`, would it be better as `task_group::wait_for`?  And then `task_group::run_and_wait_for`?

rfcs/proposed/task_group_wait_single_task/README.md
kboyarinov: Thanks, added to the open questions as well.

rfcs/proposed/task_group_wait_single_task/README.md
kboyarinov: As Alexey noted in https://github.com/uxlfoundation/oneTBB/pull/1862#discussion_r2391874376, `run_and_wait_for` and the existing `run_and_wait` appear quite similar, so using a more explicit name `run_and_wait_for_task` would help avoid misinterpretation. I agree with this suggestion.

aleksei-fedotov: Thanks, Mike! I also wanted to suggest such consistent naming.
In my opinion that this `_for` suffix is what helps to differentiate. Without that suffix, it `wait`s for all the tasks inside an entity the method is called on - `task_group::wait()` without parameters signifies just that and helps to understand the semantics of `run_and_wait`. While the `_for` suffix indicates that it is the passed argument the method is going to wait for.
Besides becoming inconsistent with the `task_arena`, the structures like `tg.wait_for_task(task)` is too repetitive on the word `task`.

akukanov: > In my opinion that this `_for` suffix is what helps to differentiate.

Maybe for `wait` vs. `wait_for`, which also differ in their arguments, but not quite enough for `run_and_wait` vs. `run_and_wait_for`, which look the same except for the name.

> Besides becoming inconsistent with the task_arena

Names in `task_group` and `task_arena` were never consistent :)
The former class has `run`, `run_and_wait`, and `wait`, while in the latter we have `execute`, `enqueue`, `wait_for`.

> the structures like `tg.wait_for_task(task)` is too repetitive on the word task.

Remember the argument of it is not a task but a completion handle, and so is quite unlikely to be called just `task`; I would expect either some meaningful name or an abbreviation like `tch`.

I guess we will have to just disagree there and find the way to resolve that disagreement. I strongly prefer semantical clarity over consistency with a loosely related method of another class.

rfcs/proposed/task_group_wait_single_task/README.md
akukanov: I would not object to a different status name, but then I think we would need to adjust `task_completion_handle` and the transfer method name accordingly - to me, these all are aligned and speak of the same thing from different angles.

What do you mean by "short-circuited"? If you mean a user action within the task to check the group status and adjust the task behavior, there is no way for `task_group` to know that - for all we know, it was executed and either signaled completion or transferred it to another task. And I do not see it different for the task group `complete` status, really - it also does not mean "a full and valid result", it only means all tasks in the group have been executed - even though they may have short-circuited due to some external condition.

akukanov: What I think makes sense to clarify is when/how the group cancellation status is checked during the task waiting call. For example, can the implementation signal all the pending task waiting calls as "canceled" once the group is cancelled - even if a particular task is in the middle of execution? What if the task transfers its completion, and then the group is canceled and the "continuation" task is not executed?

kboyarinov: I don't believe it is correct for semantics and implementation to "early-exit" the wait call with the `canceled` returned state if the group execution is cancelled. It seems for me that since the state of the particular task (or task tree) is important for the user, the implementation should wait to get cancelation signal from the task, not a group. It may be important for the user to know that the task is completed even if the whole group is canceled. 
The proposed implementation follows this pattern - if task::execute was called for the awaited task, the wait function returns "task_complete", if task::cancel - "canceled". If the task's completion was transferred, it just waits for such a signal from the task, receiving the completion.

vossmjp: I think it is important to distinguish between whether a task (or its transferred completion) has been skipped due to cancellation or has executed. My concern comes from the existence of  `task_group_status::complete` that means the work in a `task_group` is done and was not interrupted by an exception or cancellation that set the task group status to canceled. The status `task_group_status::task_complete` means the task or transferred completion was executed and was not interrupted by an exception. Its logic may have been affected (which I have called short-circuited) by cancellation since it can query cancellation and bail out early. Querying of the `task_group` cancellation status and the performing an early exit is a pattern we endorse. I do agree that we should NOT override the result in the task waiting call by looking at the task group's status after the task has returned. I just think it would be better if we didn't call it `task_complete`. But if I'm the only one who thinks that `task_complete` might be confusing, then I won't press it.

akukanov: There can be three situations:
1. the task group is not canceled at the moment the last task in the transfer chain has executed
  - it can be additionally subdivided, depending on whether the task group is canceled later.
2. the task group has been canceled to the moment the last task in the transfer chain has executed
3. the task group is canceled and at least one task in the transfer chain has not executed

It's only the second case where short-circuiting via task_group status query is in theory possible. And I expect it to be a rare case, comparing to the normal execution. So if we want to indicate a possibility of this case, I think we need a third status value for that, rather than deciding that the case 2 defines the status name also for 1.

I prefer to use `task_complete` for both, but I would also not object to distinct statuses.

rfcs/proposed/task_group_wait_single_task/README.md
```diff
@@ -0,0 +1,581 @@
...
+task or its associated ``task_group``.
+
+Ideally, the waiting thread should only execute tasks that contribute to the completion of the awaited task. In such a model, all tasks
+within the same subgraph would need to share a common isolation tag. In theory, a successor could inherit the isolation tag from
+its predecessor. However, multiple predecessors may have different tags.
+
+This would require changes to the current isolation mechanism to support multiple isolation tags, allowing a thread waiting for a successor
+to execute tasks with any of its predecessor's tags. 
+
+As an initial step, it makes sense to isolate the waiting thread so that it only executes tasks related to the same ``task_group`` as the awaited task - similar to the improvement described in the [RFC for another overload of ``task_arena::wait_for``](../task_arena_waiting/task_group_interop.md).
```
vossmjp: The `task_group_interop.md` file has move to supported. So this link doesn't work now.

rfcs/proposed/task_group_wait_single_task/README.md
```diff
@@ -0,0 +1,581 @@
...
+
+```cpp
+#if TBB_VERSION >= VER // VER is the first TBB version where waiting for individual tasks was released
+```
+
+### Summary
+
+```cpp
+// Defined in <oneapi/tbb/task_group.h>
+#define TBB_TASK_GROUP_WAIT_FOR_INDIVIDUAL_TASKS 202xxx
```
akukanov: Since TBB both reacts to some macros (`TBB_USE_*`, `TBB_PREVIEW_*`) and provides some macros (for version and for extension presence), it makes sense to consider a common naming schema (prefix or suffix) for feature test macros. The closest macro we have today is `TBB_EXT_CUSTOM_ASSERTION_HANDLER`, but clearly `TBB_EXT` is not good for features that are or planned to be specified.

In oneDPL, the schema is `ONEDPL_HAS_*` (https://uxlfoundation.github.io/oneDPL/macros.html#feature-macros), though some colleagues do not like it because of its "Boolean" impression. In the C++ standard, `__cpp_lib_*` macros are used for library features.

My personal preference is `TBB_HAS_*`, similar to oneDPL - I have no problems with reading `#if TBB_HAS_TASK_GROUP_WAIT_FOR_TASK >= 2026xx`  :)

rfcs/proposed/task_group_wait_single_task/README.md
```diff
@@ -0,0 +1,581 @@
...
+An alternative is to use ``TBB_VERSION`` macro for this purpose:
+
+```cpp
+#if TBB_VERSION >= VER // VER is the first TBB version where waiting for individual tasks was released
+```
+
+### Summary
+
+```cpp
+// Defined in <oneapi/tbb/task_group.h>
```
akukanov, line 194: Does it make sense for feature test macros to be also available through `tbb/version.h`? We already do that for `TBB_EXT_CUSTOM_ASSERTION_HANDLER`.

rfcs/proposed/task_group_wait_single_task/README.md
akukanov: Another question is the feature granularity. In this case, is "waiting for single tasks" the feature, or should "support for task dependencies and waiting" be the feature, with different values indicating the stages of enabling?

rfcs/proposed/task_group_wait_single_task/README.md
```diff
@@ -0,0 +1,581 @@
...
+
+* Performance targets for this feature should be clearly defined and met.
+* Should the ``enqueue_and_wait`` API be added to ``task_arena``? Refer to the
+  [Run-and-wait methods for ``task_arena`` section](#run-and-wait-methods-for-task_arena) for more details.
+* Should work isolation constraints be applied while waiting for a task to complete? See the
+  [Work Isolation Issue section](#work-isolation-issue) for more details.
+* Should ``run_and_wait_for_task`` overload accepting the task body be added? Refer to the
+  [``run_and_wait_for_task`` Accepting the Task Body](#run_and_wait_for_task-accepting-the-task-body) for more details.
+* Proper API and naming for functions returning the task status should be defined. Refer to the
+  [``Returning the status of the task``] section for more details.
```
akukanov: This is not rendered as a link at GitHub.

rfcs/proposed/task_group_wait_single_task/README.md
```diff
@@ -0,0 +1,581 @@
...
+* Name and the value for the feature test macro should be defined. Refer to the [``Feature Test Macro``](#feature-test-macro) for more details.
```
akukanov: I think the macro name should be finalized now, so that the feature status (preview vs. experimental) can be checked via the macro values in the code. Changing that name later would complicate checking for feature presence by users.

rfcs/proposed/task_group_wait_single_task/README.md
kboyarinov: While writing the initial proposal, I was inspired by the C++ Standard Library approach.
At the same time, I agree that providing a common prefix for feature-test macros makes sense and is more convenient for users.

I have no objections to using `TBB_HAS` with an associated value (similar to the oneDPL approach).

Regarding the "Boolean" impression: the macro is not Boolean in nature, but it still be used in a Boolean context to check for feature presence

```cpp
#if TBB_HAS_TASK_GROUP_WAIT_FOR_TASK
...
#endif
```

The concrete value becomes relevant only when multiple values are documented and a specific behavior needs to be distinguished.

rfcs/proposed/task_group_wait_single_task/README.md
kboyarinov: Yes, it makes sense

rfcs/proposed/task_group_wait_single_task/README.md
kboyarinov: I don't think it makes sense to mix task dependencies with waiting.
On the one hand, they are related at the implementation level and are enabled by a single macro.
But on the other hand, waiting for individual tasks can be useful even when task dependencies are not involved.
For this reason, I would prefer using separate macros for these two features.

rfcs/proposed/task_group_wait_single_task/README.md
kboyarinov: Fixed

rfcs/proposed/task_group_wait_single_task/README.md
kboyarinov: Fixed

rfcs/proposed/task_group_wait_single_task/README.md
vossmjp: I think "TBB_HAS" is fine too, so lets add that prefix. It is more readable as a feature check, which I like. And I think its nice to have a non-Boolean value so that if the functionality changes we can change the value, so versions can be distinguished between.

vossmjp: I agree that they are different. I think both features could have separate stages of enabling. So it might be easier to track that (as a user) if they are separate. For example, if new run and wait methods are added. So two things like TBB_HAS_TASK_DEPENDENCIES and TBB_HAS_WAIT_FOR_INDIVIDUAL_TASKS seems cleaner to me.

kboyarinov: Renamed to `TBB_HAS_TASK_GROUP_WAIT_FOR_TASK`

rfcs/proposed/task_group_wait_single_task/README.md
kboyarinov: Removed from the list of open questions since we are resolving it now.

rfcs/proposed/task_group_wait_single_task/README.md
akukanov: FWIW, I have no strong preference for this very name.

akukanov: As it comes to the names:
- should we add `TASK_GROUP` or `TG` to these names? It will become clumsier but more accurate. I have no strong preference either way, as long as we are consistent. I use TG in my examples below but only for exposition.
- in the API names we use "order" and "completion", not "dependency". Should the first macro name be aligned, perhaps? E.g., `TBB_HAS_TG_TASK_ORDERING`.
- For the sake of brevity, I have slight preference to replace "individual" with "single" in the second name, e.g. `TBB_HAS_TG_WAIT_FOR_SINGLE_TASKS`

kboyarinov: Regarding the use of `TG` instead of `TASK_GROUP`, I think it makes sense, since `TBB_HAS_TASK_GROUP_TASK_ORDERING/DEPENDENCIES` sound a bit awkward due to the repetition of TASK. But it is not aligned with the `TBB_PREVIEW_TASK_GROUP_EXTENSIONS` macro that enables the feature.

As for the second point, on the one hand, we could align with the method names and use `TBB_HAS_TG_TASK_ORDERING`. On the other hand, 
in our [reference documentation](https://uxlfoundation.github.io/oneTBB/main/reference/task_group_dynamic_dependencies.html) and [Release Notes](https://github.com/uxlfoundation/oneTBB/releases#:~:text=%5BPreview%5D%20Introduced%20API%20for%20setting%20dynamic%20task%20dependencies%20in%20task_group.%20This%20allows%20successor%20tasks%20to%20execute%20only%20after%20all%20their%20predecessors%20have%20completed.) use "task dynamic dependencies" as the feature name. It may therefore make sense to align that terminology, which is what I would prefer.

Regarding the last bullet, I like the idea of using "single" instead of "individual", but I also like the name "TBB_HAS_TG_WAIT_FOR_TASK` that was proposed earlier.

vossmjp: So is it (1) `TBB_HAS_TG_WAIT_FOR_TASK` or (2) `TBB_HAS_TG_WAIT_FOR_SINGLE_TASKS`. And  is it (3) `TBB_HAS_TG_TASK_DEPENDENCIES` or (4) `TBB_HAS_TG_TASK_ORDERING`. I prefer (1) & (3) but am ok with any choice.

kboyarinov: As agreed offline, renamed the feature-test macro to `TBB_HAS_TASK_GROUP_WAIT_FOR_SINGLE_TASK`.
For task_group dependencies feature, the macro is introduced in a separate PR #1993

PR #1863, opened 2025-10-01
Documentation for Task Group Dynamic Dependencies

kboyarinov:
### Description 

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown

### DIFF
--- a/doc/main/reference/task_group_extensions.rst
+++ b/doc/main/reference/task_group_extensions.rst
@@ -13,8 +13,11 @@ task_group extensions
 Description
 ***********
 
-|full_name| implementation extends the `tbb::task_group specification <https://oneapi-spec.uxlfoundation.org/specifications/oneapi/latest/elements/onetbb/source/task_scheduler/task_group/task_group_cls>`_ with the requirements for a user-provided function object.
-   
+|full_name| implementation extends the
+`tbb::task_group specification <https://oneapi-spec.uxlfoundation.org/specifications/oneapi/latest/elements/onetbb/source/task_scheduler/task_group/task_group_cls>`_
+with the following extensions:
+1. Task bypassing support, which expands the requirements for user-provided function objects by allowing it to return another task as a hint for further execution.
+2. Dynamic Dependencies support, which provide API for establishing predecessor-successor relationships between tasks.
 
 API
 ***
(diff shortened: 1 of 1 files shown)


vossmjp:
FYI, you accidentally checked in the try_put_and_wait binary.

doc/main/reference/task_group_bypass_support.rst
```diff
@@ -0,0 +1,94 @@
...
+.. contents::
+    :local:
+    :depth: 2
+
+Description
+***********
+
+The |full_name| implementation extends the requirements for user-provided function object from
+`tbb::task_group specification <https://oneapi-spec.uxlfoundation.org/specifications/oneapi/latest/elements/onetbb/source/task_scheduler/task_group/task_group_cls>`_
+to allow them to return a ``task_handle`` object.
```
vossmjp, line 18: It seems that [task_group specification](https://oneapi-spec.uxlfoundation.org/specifications/oneapi/latest/elements/onetbb/source/task_scheduler/task_group/task_group_cls) is silent on the return type of the user-provided function object. So "extends the requirements" doesn't seem right.

doc/main/reference/task_group_bypass_support.rst
```diff
@@ -0,0 +1,94 @@
...
+
+`Task Bypassing <../tbb_userguide/Task_Scheduler_Bypass.html>`_ allows developers to reduce task scheduling overhead by providing a hint about
+which task should be executed next.
+
+Execution of the hinted task is not guaranteed to occur immediately, nor to be performed by the same thread.
```
vossmjp:
```suggestion
Execution of the deferred task owned by a returned ``task_handle`` is not guaranteed to occur immediately, nor to be performed by the same thread.
```

doc/main/reference/task_group_dynamic_dependencies.rst
```diff
@@ -0,0 +1,362 @@
...
+    :local:
+    :depth: 2
+
+Description
+***********
+
+The |full_name| implementation extends the
+`tbb::task_group specification <https://oneapi-spec.uxlfoundation.org/specifications/oneapi/latest/elements/onetbb/source/task_scheduler/task_group/task_group_cls>`_
+with an API for defining predecessor-successor relationships between tasks
+meaning a successor task can begin execution only after all of its predecessors are completed.
```
vossmjp:
```suggestion
with an API for defining predecessor-successor relationships between tasks,
such that a successor task can begin execution only after all of its predecessors are completed.
```

doc/main/reference/task_group_dynamic_dependencies.rst
```diff
@@ -0,0 +1,362 @@
...
+
+The Dynamic Dependencies API enables developers to define predecessor-successor relationships between tasks,
+meaning a successor task can begin execution only after all of its predecessors are completed.
```
vossmjp:
```suggestion
```

doc/main/reference/task_group_dynamic_dependencies.rst
```diff
@@ -0,0 +1,362 @@
...
+If the associated task is involved in a predecessor-successor relationship, the behavior is undefined.
+
+Member Functions of ``task_group`` Class
+----------------------------------------
+
+.. code:: cpp
+
+    void run(task_handle&& handle);
+
+Schedules the task object associated with ``handle`` for execution, provided its dependencies are satisfied.
```
vossmjp:
```suggestion
Schedules the task object associated with ``handle`` for execution when its dependencies are satisfied.
```

doc/main/reference/task_group_dynamic_dependencies.rst
```diff
@@ -0,0 +1,362 @@
...
+
+It is thread-safe to concurrently add multiple predecessors to a single successor and to register the same predecessor with multiple successors.
+
+It is thread-safe to concurrently add successors to both the task transferring its completion and the task receiving the completion.
+
+The behavior is undefined in the following cases:
+
+* Either ``pred`` or ``succ`` is empty.
+* The tasks referred by ``pred`` and ``succ`` belong to different ``task_group`` instances.
+* The task referred by ``pred`` was destroyed before begin submitted for execution.
```
vossmjp:
```suggestion
* The task referred by ``pred`` was destroyed before being submitted for execution.
```

doc/main/reference/task_group_dynamic_dependencies.rst
```diff
@@ -0,0 +1,362 @@
...
+
+.. code:: cpp
+
+    static void set_task_order(task_completion_handle& pred, task_handle& succ);
+
+Registers the task referenced by ``pred`` as a predecessor that must complete before the task associated with ``succ`` can begin execution.
+
+It is thread-safe to concurrently add multiple predecessors to a single successor and to register the same predecessor with multiple successors.
+
+It is thread-safe to concurrently add successors to both the task transferring its completion and the task receiving the completion.
```
vossmjp, line 302:
```suggestion
It is thread-safe to concurrently add successors to both the task transferring its completion and the task receiving the completion.

It is thread-safe to concurrently add a predecessor while the ``task_handle`` associated with the same predecessor task is being run.
```

doc/main/reference/examples/try_put_and_wait
akukanov, line 1: What is this file for?

doc/main/reference/examples/task_group_extensions_bypassing.cpp
```diff
@@ -0,0 +1,82 @@
...
+        tbb::task_handle next_task;
+        std::size_t size = end - begin;
+        if (size < serial_threshold) {
+            // Execute the work serially
+            foo(begin, end);
+        } else {
+            // Processed range to big - split it
+            std::size_t middle = begin + size / 2;
+            tbb::task_handle left_subtask = tg.defer(for_task{begin, middle, tg});
+            tbb::task_handle right_subtask = tg.defer(for_task{middle, end, tg});
```
akukanov:
```suggestion
            // Enough work to split the range
            std::size_t middle = begin + size / 2;
```
Subtasks are not used in the code.

doc/main/reference/examples/task_group_extensions_bypassing.cpp
```diff
@@ -0,0 +1,82 @@
...
+    std::size_t begin;
+    std::size_t end;
+    tbb::task_group& tg;
+}; // struct for_task
+
+void calculate_parallel_for(std::vector<std::size_t>& vec) {
+    tbb::task_group tg;
+    // Run the root task
+    tg.run_and_wait(for_task{0, vec.size(), tg});
+}
```
akukanov: This function looks strange. It takes a vector but does nothing with that, except for `vec.size()`. There certainly are more efficient ways to specify the number of iterations.

doc/main/reference/examples/task_group_extensions_reduction.cpp
```diff
@@ -0,0 +1,100 @@
...
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+
+#include <cstdint>
+
+/*begin_task_group_extensions_reduction_example*/
+
+#define TBB_PREVIEW_TASK_GROUP_EXTENSIONS 1
+#include "oneapi/tbb/task_group.h"
```
akukanov, line 24:
```suggestion
/*begin_task_group_extensions_reduction_example*/
#define TBB_PREVIEW_TASK_GROUP_EXTENSIONS 1
#include "oneapi/tbb/task_group.h"
```

doc/main/reference/examples/task_group_extensions_reduction.cpp
```diff
@@ -0,0 +1,100 @@
...
+    tbb::task_group tg;
+
+    std::shared_ptr<std::size_t> reduce_result = std::make_shared<std::size_t>(0);
+    reduce_task root_reduce_task{begin, end, reduce_result, tg};
+    tg.run_and_wait(root_reduce_task);
+
+    return *reduce_result;
+}
+
+/*end_task_group_extensions_reduction_example*/
```
akukanov, line 90:
```suggestion
}
/*end_task_group_extensions_reduction_example*/
```

doc/main/reference/examples/task_group_extensions_reduction.cpp
```diff
@@ -0,0 +1,100 @@
...
+            // Divide it into smaller segments for parallel execution
+            std::size_t middle = begin + size / 2;
+
+            std::shared_ptr<std::size_t> left_result = std::make_shared<std::size_t>(0);
+            tbb::task_handle left_leaf = tg.defer(reduce_task{begin, middle, left_result, tg});
+
+            std::shared_ptr<std::size_t> right_result = std::make_shared<std::size_t>(0);
+            tbb::task_handle right_leaf = tg.defer(reduce_task{middle, end, right_result, tg});
+
+            tbb::task_handle join = tg.defer(join_task{left_result, right_result, result});
```
akukanov: Do we use `shared_ptr` in the example because we recommend it to share data across tasks?
My concern is that it is quite inefficient due to allocations and reference counting, neither of which is needed because task dependencies can both ensure sufficient lifetime and prevent races.
Instead, I would probably use `std::ref` for "forwarding" the result storage to `join_task`, while storing `left` and `right` as data members of `join_task`.

doc/main/reference/examples/task_group_extensions_reduction.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+#include <cstdint>
+
+/*begin_task_group_extensions_reduction_example*/
+
+#define TBB_PREVIEW_TASK_GROUP_EXTENSIONS 1
+#include "oneapi/tbb/task_group.h"
+
+struct reduce_task {
+    static constexpr std::size_t serial_threshold = 16;
```
akukanov: Is the threshold justified anyhow, or chosen rather arbitrarily?

doc/main/reference/examples/try_put_and_wait
kboyarinov: Removed

doc/main/reference/examples/task_group_extensions_reduction.cpp
kboyarinov: As far as I understand, the proposal is to store `left` and `right` placeholders inside of the `join_task` function object and refer to these placeholders from `left_leaf` and `right_leaf`.
This will not work because `task_group::defer` copies the function object and there is no API in `task_handle` to access the real function object that will be used to execute the task. Therefore, we don't have access to the actual `left` and `right` placeholder to save them into the leaf tasks.

doc/main/reference/examples/task_group_extensions_reduction.cpp
kboyarinov: It was chosen arbitrarily, just to illustrate the example. 
I have moved it into "invisible" section to exclude from the documentation.

doc/main/reference/examples/task_group_extensions_bypassing.cpp
kboyarinov: I have rewritten it to illustrate `for_each` implementation

doc/main/reference/task_group_dynamic_dependencies.rst
kboyarinov: I think "when" suggests that `run` will schedule the task when the dependencies are satisfies. It can be interpreted as waiting for the dependencies to be satisfied and then schedule the task. 
Or you believe that together with the following note the behavior is obvious?

doc/main/reference/task_group_bypass_support.rst
kboyarinov: As far as I understand the idea of named requirement and the return type is that if the specification says nothing about the return type means that any type can be returned, but will be ignored. So "extending" the named requirement means that the implementation will react if the handle is returned.

doc/main/reference/task_group_dynamic_dependencies.rst
vossmjp: I think the ", provided" version could be read as though the "run" function does nothing if the task object dependencies are not satisfied at the time of the call to run. But of course calling run does have an effect. It makes the task available for dependency tracking, and ultimately execution once its dependencies are satisfied.

doc/main/reference/task_group_dynamic_dependencies.rst
```diff
@@ -0,0 +1,362 @@
...
+If the associated task is involved in a predecessor-successor relationship, the behavior is undefined.
+
+Member Functions of ``task_group`` Class
+----------------------------------------
+
+.. code:: cpp
+
+    void run(task_handle&& handle);
+
+Schedules the task object associated with ``handle`` for execution, provided its dependencies are satisfied.
```
vossmjp:
```suggestion
Makes the task object associated with ``handle`` available for execution once its dependencies are satisfied. A task associated with a ``task_handle`` will never execute unless it is run.
```

doc/main/reference/examples/task_group_extensions_reduction.cpp
akukanov: Looks like you did it only for one example of the two.

doc/main/reference/examples/task_group_extensions_reduction.cpp
akukanov: I see. Thanks for the explanation.

doc/main/reference/examples/task_group_extensions_reduction.cpp
kboyarinov: Sorry, missed the second file. Fixed

doc/main/reference/task_group_bypass_support.rst
```diff
@@ -0,0 +1,94 @@
...
+***********
+
+The |full_name| implementation extends the requirements for user-provided function object from
+`tbb::task_group specification <https://oneapi-spec.uxlfoundation.org/specifications/oneapi/latest/elements/onetbb/source/task_scheduler/task_group/task_group_cls>`_
+to allow them to return a ``task_handle`` object.
+
+`Task Bypassing <../tbb_userguide/Task_Scheduler_Bypass.html>`_ allows developers to reduce task scheduling overhead by providing a hint about
+which task should be executed next.
+
+Execution of the deferred task owned by a returned ``task_handle`` is not guaranteed to occur immediately, nor to be performed by the same thread.
```
akukanov, line 23: I think we need to say something about the returned task being effectively scheduled for execution, so it **must not** be explicitly submitted. This is also important because of the difference with task dependencies, where a task **must** be explicitly submitted (or bypassed).

doc/main/reference/task_group_bypass_support.rst
```diff
@@ -0,0 +1,94 @@
...
+
+The function object ``F`` may return a ``task_handle`` object. If the returned handle is non-empty and owns a task without dependencies, it serves as an optimization
+hint for a task that could be executed next.
+
+If the returned handle was created by a ``task_group`` other than ``*this``, the behavior is undefined.
+
+Example
+-------
+
+The example below demonstrates how to implement a parallel for loop using ``task_group`` and divide-and-conquer pattern.
```
akukanov:
```suggestion
The example below demonstrates how to process a sequence in parallel using ``task_group`` and the divide-and-conquer pattern.
```

doc/main/reference/examples/task_group_extensions_bypassing.cpp
```diff
@@ -0,0 +1,81 @@
...
+
+    Iterator begin;
+    Iterator end;
+    Function f;
+    tbb::task_group& tg;
+}; // struct for_task
+
+// Function accepts std::iterator_traits<RandomAccessIterator>::reference argument
+template <typename RandomAccessIterator, typename Function>
+void parallel_for(RandomAccessIterator begin, RandomAccessIterator end, Function f) {
```
akukanov:
```suggestion
template <typename RandomAccessIterator, typename Function>
void for_each(RandomAccessIterator begin, RandomAccessIterator end, Function f) {
```

doc/main/reference/examples/task_group_extensions_bypassing.cpp
```diff
@@ -0,0 +1,81 @@
...
+        return next_task;
+    }
+
+    Iterator begin;
+    Iterator end;
+    Function f;
+    tbb::task_group& tg;
+}; // struct for_task
+
+// Function accepts std::iterator_traits<RandomAccessIterator>::reference argument
```
akukanov: I do not understand this comment; it seems to be incorrect.

doc/main/reference/examples/task_group_extensions_bypassing.cpp
```diff
@@ -0,0 +1,81 @@
...
+/*begin_task_group_extensions_bypassing_example*/
+#define TBB_PREVIEW_TASK_GROUP_EXTENSIONS 1
+#include "oneapi/tbb/task_group.h"
+
+template <typename Iterator, typename Function>
+struct for_task {
+    tbb::task_handle operator()() const {
+        tbb::task_handle next_task;
+
+        auto size = std::distance(begin, end);
```
akukanov:
```suggestion
        auto size = end - begin;
```
You assume anyway that the iterators are random access, e.g. when computing `middle`.

doc/main/reference/task_group_dynamic_dependencies.rst
```diff
@@ -0,0 +1,362 @@
...
+
+The |full_name| implementation extends the
+`tbb::task_group specification <https://oneapi-spec.uxlfoundation.org/specifications/oneapi/latest/elements/onetbb/source/task_scheduler/task_group/task_group_cls>`_
+with an API for defining predecessor-successor relationships between tasks,
+such that a successor task can begin execution only after all of its predecessors are completed.
+
+
+Tasks in any state (``created``, ``submitted``, ``executing``, or ``completed``) can serve as predecessors, but only tasks in the ``created`` state may be used as successors.
+
+A ``tbb::task_handle`` represents a task in the ``created`` state, while a ``task_completion_handle`` can represent a task in any state.
```
akukanov: The task states have not yet been properly introduced. In fact, these are not formally introduced at all, just illustrated in the code example below. Moreover, the state is only mentioned in these two sentences and the example.

Can we describe the same idea without the notion of states? Focusing on the practical differences in the semantics and use of `task_handle` vs. `task_completion_handle` could be more helpful.

doc/main/reference/task_group_dynamic_dependencies.rst
```diff
@@ -0,0 +1,362 @@
...
+The ``tbb::task_group::set_task_order(pred, succ)`` function establishes a dependency such that ``succ`` cannot begin execution until ``pred`` has completed.
+
+.. code:: cpp
+
+    tbb::task_handle predecessor = tg.defer(pred_body);
+    tbb::task_handle successor = tg.defer(succ_body);
+
+    tbb::task_group::set_task_order(predecessor, successor);
+
+The feature also allows transferring the completion of the currently executing task to another ``created`` task using ``tbb::task_group::transfer_this_task_completion_to``.
```
akukanov:
```suggestion
The ``tbb::task_group::transfer_this_task_completion_to`` function allows transferring the completion of the currently executing task to another task.
```

doc/main/reference/task_group_dynamic_dependencies.rst
```diff
@@ -0,0 +1,362 @@
...
+
+    bool operator!=(const task_completion_handle& lhs, const task_completion_handle& rhs) noexcept;
+
+Equivalent to ``!(lhs == rhs)``.
+
+.. code:: cpp
+
+    bool operator==(const task_completion_handle& t, std::nullptr_t) noexcept;
+
+*Returns*: ``true`` if ``t`` does not reference any task; otherwise, ``false``.
```
akukanov, line 250: Maybe combine with the other `operator==` that only differs by the order of parameters?

doc/main/reference/task_group_dynamic_dependencies.rst
```diff
@@ -0,0 +1,362 @@
...
+
+.. code:: cpp
+
+    bool operator!=(const task_completion_handle& t, std::nullptr_t) noexcept;
+
+Equivalent to ``!(t == nullptr)``.
```
akukanov: Ditto.

doc/main/reference/task_group_dynamic_dependencies.rst
```diff
@@ -0,0 +1,362 @@
...
+
+    task_completion_handle(const task_completion_handle& other);
+
+Copies ``other`` into ``*this``. After the copy, both ``*this`` and ``other`` refer to the same task.
+
+.. code:: cpp
+
+    task_completion_handle(task_completion_handle&& other);
+
+Moves ``other`` into ``*this``. After the move, ``*this`` refers to the task previously referenced by ``other``, which is left in an empty state.
```
akukanov:
```suggestion
Moves ``other`` into ``*this``. After the move, ``*this`` refers to the task previously referenced by ``other``, which is left empty.
```
For consistency.

doc/main/reference/task_group_dynamic_dependencies.rst
```diff
@@ -0,0 +1,362 @@
...
+
+Member Functions of ``task_handle`` Class
+-----------------------------------------
+
+.. code:: cpp
+
+    ~task_handle();
+
+Destroys the ``task_handle`` object and its associated task, if any.
+If the associated task is involved in a predecessor-successor relationship, the behavior is undefined.
```
akukanov: For the existing API, would it make sense to take the exact description from the spec in full, the modify and highlight the differences?

doc/main/reference/task_group_bypass_support.rst
kboyarinov: I have added a note at the end of description.

doc/main/reference/examples/task_group_extensions_bypassing.cpp
kboyarinov: It was part of the original example. Removed.

doc/main/reference/task_group_dynamic_dependencies.rst
kboyarinov: Rewritten using submitted and unsubmitted tasks instead.

doc/main/reference/task_group_dynamic_dependencies.rst
kboyarinov: Done

doc/main/reference/task_group_dynamic_dependencies.rst
kboyarinov: Done

doc/main/reference/task_group_dynamic_dependencies.rst
kboyarinov: I have copied the exact description from the spec and highlighted the change using ``admonition``. 
It looks like this in the rendered reference:

destructor

doc/main/reference/examples/task_group_extensions_bypassing.cpp
kboyarinov: Renamed to `parallel_for_each` since with the name `for_each` it is challenging to test the function because of the ambiguity with `std::for_each`.

akukanov: Do you think that ambiguity with `tbb::parallel_for_each` is better? :)
Do not do `using namespace std;`, and the testing should be good, no? Or is the ambiguity caused by ADL?

doc/main/reference/task_group_dynamic_dependencies.rst
akukanov: Looks good, though I think the heading like "Extension" would be better.
Now the same approach could be applied to other existing functions.

doc/main/reference/examples/task_group_extensions_bypassing.cpp
kboyarinov: `tbb::parallel_for_each` do not create an issue here. `using namespace std` is not used. The ambiguity is because the test `for_each` is defined in the global namespace. The arguments to for each are iterators of `std::vector` that are defined in the namespace `std`. Because of the ADL, `std::for_each` becomes visible and two versions of the function creates an ambiguity. 
While explaining this, I realized that it should be possible to solve this by using raw pointers instead of vector iterators in testing.

doc/main/reference/task_group_dynamic_dependencies.rst
kboyarinov: For other existing functions, we have completely changed the description, not extended it. 
I don't think it makes sense to keep the original wording from the spec.

akukanov: Well, let's compare it for `task_group::run`.

Specification:
> void run(task_handle &&h)
> Schedules the task object pointed by the h for the execution.
> Note:
> The failure to satisfy the following conditions leads to undefined behavior:
>  - h is not empty.
>  - *this is the same task_group that h is created with.

The doc here:
> void run(task_handle&& handle);
> Makes the task object associated with handle available for execution once its dependencies are satisfied. A task associated with a task_handle will never execute unless it is run, enqueued or bypassed.
> Note
> If the task associated with handle has incomplete predecessors, it will be scheduled for execution once all of them have completed. The run function does not wait for predecessors to complete.

Analysis:
- The first sentence has almost the same meaning as in the spec, with the additional claim about dependencies
- The sentence "A task associated ... will never execute unless..." is misplaced. It's not about the method `run` but either about the `task_handle` or the method `defer` that creates the handle. And it is applicable without the extension (except for "bypassed").
- The note from the specification is omitted - but it remains valid!
- The new note in the PR a) partially rephrases what is already said in the description, b) contains the information that does not need to be marked as a note.

How I would do it:

> void run(task_handle &&h)
> Schedules the task object pointed by h for the execution.
> Extension:
> If the task associated with h has predecessors, scheduling the task execution is postponed until all of the predecessor have completed, while the function returns immediately.
> Note: [the existing note from the specification]

akukanov: Meanwhile, I have prototyped another possible way of highlighting differences, check it out: https://github.com/uxlfoundation/oneTBB/pull/1879

vossmjp: I like the "Extension" admonition as the way to highlight the differences -- where it can be applied. Highlighting might be useful where the changes must be interspersed in an existing paragraph., but otherwise the separate Extension box looks cleaner to me. But that's just a personal preference.

doc/main/reference/examples/task_group_extensions_bypassing.cpp
akukanov: I think though that the conflict with `std::for_each` due to ADL should not be ignored - someone can try the example with std::vector. And I guess there is a risk of the same problem with `tbb::parallel_for_each` as well if the function is applied to e.g. `tbb::concurrent_vector`.

So maybe a different name is worth choosing, or a custom namespace can be introduced.

doc/main/reference/task_group_extensions.rst
```diff
@@ -4,83 +4,14 @@ task_group extensions
 =====================
 
 .. note::
-    To enable these extensions, set the ``TBB_PREVIEW_TASK_GROUP_EXTENSIONS`` macro to 1.
+    To enable these extensions, define the ``TBB_PREVIEW_TASK_GROUP_EXTENSIONS`` macro with a value of ``1``.
```
akukanov, line 9: I would add a generic sentence about extending `task_group` for advanced use cases, or something like that.

doc/main/reference/task_group_bypass_support.rst
```diff
@@ -0,0 +1,98 @@
...
+    template <typename F>
+    task_group_status run_and_wait(const F& f);
+
+    template <typename F>
+    void run(F&& f);
+
+The function object ``F`` may return a ``task_handle`` object. If the returned handle is non-empty and owns a task without dependencies, it serves as an optimization
+hint for a task that could be executed next.
+
+If the returned handle was created by a ``task_group`` other than ``*this``, the behavior is undefined.
```
akukanov: Use the extension admonition for this as well?

We may consider adding the following wording from the specification:
> The `F` type must meet the _Function Objects_ requirements described in the [function.objects] section of the ISO C++ standard.

doc/main/reference/task_group_bypass_support.rst
```diff
@@ -0,0 +1,98 @@
...
+.. code:: cpp
+
+    tbb::task_handle task_body() {
+        tbb::task_handle next_task = group.defer(next_task_body);
+        return next_task;
+    }
+
+.. note::
+    
+    To allow a task to be bypassed, the ``task_handle`` that owns it must **not** be explicitly submitted to ``task_group::run`` or any other submission function.
```
akukanov: Should this be a part of the "formal wording" rather than a description note? Perhaps with slight changes to avoid "bypassed":
> The returned `task_handle` must not be explicitly submitted with `task_group::run` or another submission function.

doc/main/reference/task_group_dynamic_dependencies.rst
```diff
@@ -0,0 +1,378 @@
...
+***********
+
+The |full_name| implementation extends the
+`tbb::task_group specification <https://oneapi-spec.uxlfoundation.org/specifications/oneapi/latest/elements/onetbb/source/task_scheduler/task_group/task_group_cls>`_
+with an API for defining predecessor-successor relationships between tasks,
+such that a successor task can begin execution only after all of its predecessors are completed.
+
+An **unsubmitted** task is one that has not been explicitly submitted for execution, such as by passing a ``task_handle`` to ``task_group::run``.
+
+A **submitted** task is one that has been explicitly submitted for execution.
```
akukanov: Is the word "explicitly" important here? Is there a third category of tasks, submitted implicitly?

doc/main/reference/task_group_dynamic_dependencies.rst
```diff
@@ -0,0 +1,378 @@
...
+
+A non-empty ``task_handle`` object represents an unsubmitted task, while a ``task_completion_handle`` can represent either submitted or unsubmitted tasks.
```
akukanov:
```suggestion
A non-empty ``task_handle`` object represents an unsubmitted task, while a ``task_completion_handle`` can represent any task.
```

doc/main/reference/task_group_dynamic_dependencies.rst
```diff
@@ -0,0 +1,378 @@
...
+
+The ``tbb::task_group::transfer_this_task_completion_to`` function allows transferring the completion of the currently executing task to another task.
+This function must be invoked from within the task body. All successors of the currently executing task will execute only after the task receiving the completion has finished.
+
+.. code:: cpp
+
+    tbb::task_handle t = tg.defer([] {
+        tbb::task_group comp_receiver = tg.defer(receiver_body);
+        tbb::task_group::transfer_this_task_completion_to(comp_receiver);
+    });
```
akukanov: There seems to be a typo: the result of `defer` is assigned to `task_group`.
Also note the "receiver" task remains unsubmitted.

doc/main/reference/task_group_dynamic_dependencies.rst
```diff
@@ -0,0 +1,378 @@
...
+.. admonition:: Extension
+
+    If the task associated with ``h`` has predecessors, scheduling the task execution is postponed until all
+    of the predecessors have completed, while the function returns immediately.
+
+.. note::
+
+    The failure to satisfy the following conditions leads to undefined behavior:
+    * ``h`` is not empty.
+    * ``*this`` is the same ``task_group`` that ``h`` is created with.
```
akukanov, line 291:
```suggestion
    The failure to satisfy the following conditions leads to undefined behavior:
    
    * ``h`` is not empty.
    * ``*this`` is the same ``task_group`` that ``h`` is created with.
```
For the proper rendering of the list.

doc/main/reference/task_group_dynamic_dependencies.rst
```diff
@@ -0,0 +1,378 @@
...
+    ``h`` should not be empty to avoid an undefined behavior.
+
+Enqueues the task object associated with ``handle`` into ``task_arena`` for processing once its dependencies are satisfied.
+A task associated with a ``task_handle`` will never execute unless it is run, enqueued or bypassed.
+
+.. note::
+
+    If the task associated with ``handle`` has incomplete predecessors, it will be scheduled for execution once all of them have completed.
+    The ``enqueue`` function does not wait for predecessors to complete.
+
```
akukanov: Which API this refers to?

doc/main/reference/task_group_dynamic_dependencies.rst
```diff
@@ -0,0 +1,378 @@
...
+After the transfer, the successors of the currently executing task will be reassigned to the task associated with ``handle``.
+
+It is thread-safe to transfer successors to the task while concurrently adding successors to it or to the currently executing task.
+
+The behavior is undefined in the following cases:
+
+* ``handle`` is empty.
+* The function is called outside the body of a ``task_group`` task.
+* The function is called for the task whose completion has already been transferred.
+* The currently executing task and the task associated with ``handle`` belong to different ``task_group`` instances.
```
akukanov, line 327: Tasks from different groups are not mentioned in the UB list for `set_task_order`. Should they?

doc/main/reference/task_group_dynamic_dependencies.rst
```diff
@@ -0,0 +1,378 @@
...
+It is thread-safe to concurrently add successors to both the task transferring its completion and the task receiving the completion.
+
+The behavior is undefined in the following cases:
+
+* Either ``pred`` or ``succ`` is an empty.
+* The tasks referenced by ``pred`` and ``succ`` belong to different ``task_group`` instances.
+
+.. code:: cpp
+
+    static void set_task_order(task_completion_handle& pred, task_handle& succ);
```
akukanov, line 296: There seems to be a lot in common between the descriptions of `set_task_order` overloads. Would it make sense to combine into a single one?

doc/main/reference/examples/task_group_extensions_reduction.cpp
```diff
@@ -0,0 +1,98 @@
...
+            // Divide it into smaller segments for parallel execution
+            std::size_t middle = begin + size / 2;
+
+            std::shared_ptr<std::size_t> left_result = std::make_shared<std::size_t>(0);
+            tbb::task_handle left_leaf = tg.defer(reduce_task{begin, middle, left_result, tg});
+
+            std::shared_ptr<std::size_t> right_result = std::make_shared<std::size_t>(0);
+            tbb::task_handle right_leaf = tg.defer(reduce_task{middle, end, right_result, tg});
+
+            tbb::task_handle join = tg.defer(join_task{left_result, right_result, result});
```
akukanov: I would at least move the shared pointers to the join task.

doc/main/reference/task_group_bypass_support.rst
kboyarinov: Done

doc/main/reference/task_group_bypass_support.rst
kboyarinov: Applied

doc/main/reference/task_group_dynamic_dependencies.rst
kboyarinov: No, I don't think it's important. Removed

doc/main/reference/task_group_dynamic_dependencies.rst
kboyarinov: Fixed

doc/main/reference/task_group_dynamic_dependencies.rst
kboyarinov: It was left from the old enqueue description. Removed

doc/main/reference/task_group_dynamic_dependencies.rst
kboyarinov: They are mentioned on the second bullet:

`* The tasks referred by ``pred`` and ``succ`` belong to different ``task_group`` instances.`

doc/main/reference/task_group_dynamic_dependencies.rst
kboyarinov: I have tried to combine them together.

doc/main/reference/examples/task_group_extensions_reduction.cpp
kboyarinov: Replaced with `unique_ptr`s in the join task

doc/main/reference/task_group_extensions.rst
kboyarinov: Do you mean adding it to the description or as a separate note before the description?

akukanov: I meant a short description prior to the note.

doc/main/reference/task_group_bypass_support.rst
```diff
@@ -0,0 +1,102 @@
...
+.. admonition:: Extension
+
+    ``F`` may return a ``task_handle`` object. If the returned handle is non-empty and owns a task without dependencies, it serves as an optimization hint
+    for a task that could be executed next.
+
+    If the returned handle was created by a ``task_group`` other than ``*this``, the behavior is undefined.     
+
+.. note::
+    
+    The returned ``task_handle`` must not be explicitly submitted with ``task_group::run`` or another submission function.
```
akukanov:
```suggestion
    The returned ``task_handle`` must not be explicitly submitted with ``task_group::run`` or another submission function, otherwise, the behavior is undefined.

    If the returned handle was created by a ``task_group`` other than ``*this``, the behavior is undefined.     
```
Let's make it normative, rather than informative.

akukanov (commented):
The documentation looks sufficiently good to me now.

doc/main/reference/task_group_extensions.rst
kboyarinov: Added

doc/main/reference/task_group_dynamic_dependencies.rst
```diff
@@ -0,0 +1,379 @@
...
+Both submitted and unsubmitted tasks can serve as predecessors. However, only unsubmitted tasks may be used as successors.
+
+.. code:: cpp
+
+    tbb::task_handle task = tg.defer(task_body);
+    // task is unsubmitted
+
+    tbb::task_completion_handle comp_handle = task;
+    // task is unsubmitted
+    // both task_handle and task_completion_handle represents the task
```
isaevil:
```suggestion
    // both task_handle and task_completion_handle represent the task
```

doc/main/reference/task_group_dynamic_dependencies.rst
```diff
@@ -0,0 +1,379 @@
...
+
+    tg.run(std::move(task));
+    // task is submitted
+    // task_handle is empty, task_completion_handle represents a task
```
isaevil:
```suggestion
    // task_handle is empty, task_completion_handle represents the task
```

isaevil (dismissed):
Looks good to me. Just a bit of nitpicking wrt to comments in code snippets.

PR #1864, opened 2025-10-01
Move task_arena_waiting/task_group_interop to supported RFCs

akukanov:
The proposal of an API for convenient co-use of task arenas and task groups is implemented.

The PR moves the respective document to `supported/task_arena_waiting`, adds a new umbrella doc there for sufficient context (copying much from `propoosed`), and modifies the umbrella in `proposed` to reflect these changes.

The main question is the amount of content duplicated in the umbrella docs -does it seem right?

### DIFF
--- a/rfcs/supported/task_arena_waiting/readme.md
+++ b/rfcs/supported/task_arena_waiting/readme.md
@@ -62,12 +62,11 @@ In this case, the task group "registers" a deferred task to run `foo()`, which i
 to the task arena. The task is added by the calling thread, so we can be sure that `tg.wait` will not
 return until the task completes.
 
-## Proposal
-
-### 1. Simplify the use of a task group
+## Implemented improvements
 
 To address extra complexity and verbosity of using together `task_arena` and `task_group`, the `enqueue` method
-of `task_arena` could take `task_group` as the second argument, and a new method could wait for a task group:
+of `task_arena` is overloaded to take `task_group` as the second argument, and a new method is added to wait
+for a task group:
 ```cpp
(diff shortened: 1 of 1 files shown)


rfcs/supported/task_arena_waiting/readme.md
```diff
@@ -0,0 +1,93 @@
...
+tbb::task_group tg;
+ta.enqueue(tg.defer([]{ foo(); }));
+bar();
+ta.execute([&tg]{ tg.wait(); });
+```
+In this case, the task group "registers" a deferred task to run `foo()`, which is then enqueued
+to the task arena. The task is added by the calling thread, so we can be sure that `tg.wait` will not
+return until the task completes.
+
+## Proposal
```
vossmjp:
```suggestion
## API
```

rfcs/supported/task_arena_waiting/readme.md
```diff
@@ -0,0 +1,93 @@
...
+
+### 1. Simplify the use of a task group
```
vossmjp:
```suggestion
```

rfcs/supported/task_arena_waiting/readme.md
vossmjp: Since its is supported, it might be strange to call it a proposal.

rfcs/supported/task_arena_waiting/readme.md
vossmjp: Only 1 item, so remove?

akukanov: Removed

rfcs/supported/task_arena_waiting/readme.md
akukanov: Renamed.

PR #1866, opened 2025-10-02
[Documentation] Use task_arena interoperability functionality with task_group

isaevil:
### Description 
Rewrite example in "Set NUMA Node" section to use correct pattern of usage that won't lead to undersubscription and use `task_arena.wait_for(task_group)` and `task_arena.enqueue(func, task_group)`.

### Type of change

### Tests

- [x] not needed

### Documentation

- [x] updated in # - _add PR number_
- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/doc/main/tbb_userguide/examples/guiding_task_scheduler_execution.cpp
+++ b/doc/main/tbb_userguide/examples/guiding_task_scheduler_execution.cpp
@@ -21,7 +21,7 @@
 #include <oneapi/tbb/task_group.h>
 
 void set_numa_node_example() {
-/* begin_set_numa_node_example */
+/*begin_set_numa_node_example*/
     std::vector<tbb::numa_node_id> numa_nodes = tbb::info::numa_nodes();
     std::vector<tbb::task_arena> arenas(numa_nodes.size());
     std::vector<tbb::task_group> task_groups(numa_nodes.size());
@@ -43,7 +43,7 @@ void set_numa_node_example() {
     for(unsigned j = 0; j < numa_nodes.size(); j++) {
         arenas[j].wait_for(task_groups[j]);
     }
-/* end_set_numa_node_example */
+/*end_set_numa_node_example*/
 }
 
 int main() {


doc/main/tbb_userguide/Guiding_Task_Scheduler_Execution.rst
```diff
@@ -44,19 +44,24 @@ assign a NUMA node identifier to the ``task_arena::constraints::numa_id`` field.
...
+    std::vector<tbb::task_arena> arenas(numa_nodes.size());
+    std::vector<tbb::task_group> task_groups(numa_nodes.size());
+
+    for(unsigned j = 0; j < numa_nodes.size(); j++) {
+        arenas[j].initialize(tbb::task_arena::constraints(numa_nodes[j]), /*reserved_slots*/ 0);
+    }
+
+    for(unsigned j = 1; j < numa_nodes.size(); j++) {
+        arenas[j].enqueue([](){/*some parallel work*/}, task_groups[j]);
     }
```
kboyarinov: I think it is not critical to initialize all the arenas before the enqueue:

```suggestion
    for(unsigned j = 0; j < numa_nodes.size(); j++) {
        arenas[j].initialize(tbb::task_arena::constraints(numa_nodes[j]), /*reserved_slots*/ 0);
        if (j != 0) arenas[j].enqueue([](){/*some parallel work*/}, task_groups[j]);
    }
```

doc/main/tbb_userguide/Guiding_Task_Scheduler_Execution.rst
```diff
@@ -44,19 +44,24 @@ assign a NUMA node identifier to the ``task_arena::constraints::numa_id`` field.
...
-    std::vector<tbb::numa_node_id> numa_indexes = tbb::info::numa_nodes();
-    std::vector<tbb::task_arena> arenas(numa_indexes.size());
-    std::vector<tbb::task_group> task_groups(numa_indexes.size());
-
-    for(unsigned j = 0; j < numa_indexes.size(); j++) {
-        arenas[j].initialize(tbb::task_arena::constraints(numa_indexes[j]));
-        arenas[j].execute([&task_groups, &j](){ 
-            task_groups[j].run([](){/*some parallel stuff*/});
-        });
+    std::vector<tbb::numa_node_id> numa_nodes = tbb::info::numa_nodes();
```
kboyarinov: Can we also separate this example into the .cpp file in `examples` folder to make it testable?

vossmjp: We should do this any time we update an example.

doc/main/tbb_userguide/Guiding_Task_Scheduler_Execution.rst
vossmjp: Why set reserved slots to 0 for arenas[0]? I'd suggest keeping the original loop bounds and then initialize arenas[0] before the call to execute.  And add some explanation in the text for why setting the reserved slots to 0 for arenas 1 through N is useful. Maybe something like, "Since the library creates one less worker thread than the number of total cores, all but one of the arenas is created without reserving a slot for a master thread, allowing them to be fully populated. The main thread then executes in the arena with the reserved slot."

isaevil: Updated the code sample.

doc/main/tbb_userguide/Guiding_Task_Scheduler_Execution.rst
isaevil: Moved into the .cpp file. I haven't touched the other samples from this document though.

kboyarinov (dismissed):
LGTM

PR #1867, opened 2025-10-03
Fix returning an empty task_handle from task_group task

kboyarinov:
### Description 
Task Group Dynamic Dependencies broke the case when an empty `task_handle` is bypassed.
Current documentation do not explicitly allow it, but it is needed to implement constructions like below:

```cpp
class group_task {
    tbb::task_handle operator()() const {
        if (range-too-small) {
            // Do serial work
            return tbb::task_handle{}; // nothing to bypass, but the operator signature forces to return something
        } else {
            // Split work and run in parallel
            tbb::task_handle left = tg.defer(left-subtask);
            tbb::task_handle right = tg.defer(right-subtask);
            
            // Submit the right part
            tg.run(std::move(right);
            
            // Bypass the left part
            return left;
        }
    }
};
```

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown

### DIFF
--- a/include/oneapi/tbb/task_group.h
+++ b/include/oneapi/tbb/task_group.h
@@ -137,7 +137,7 @@ namespace {
         task_handle th = std::forward<F>(f)();
         task_handle_task* task_ptr = task_handle_accessor::release(th);
         // If task has unresolved dependencies, it can't be bypassed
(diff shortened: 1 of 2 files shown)


vossmjp (approved):
Looks good to me

PR #1868, opened 2025-10-03
Added abi table to VERSIONING.md

vossmjp:
### Description 
Added a table to VERSIONING.md to track abi changes by release.

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown


icfaust:
It would be cool if you guys added some sort of ABI check like we do in oneDAL such that the VERSIONING.md must be modified upon ABI changes.

vossmjp:
@icfaust That's a good idea, we'll look into that.

VERSIONING.md
```diff
@@ -107,3 +107,26 @@ Rules for the `d` namespace include:
...
+| 2021.8.0 | 12.8 | Feb 2023 | [Release Notes](https://www.intel.com/content/www/us/en/developer/articles/release-notes/intel-oneapi-threading-building-blocks-release-notes.html#inpage-nav-10) | - | - |
+| 2021.7.0 | 12.7 | Oct 2022 | [Release Notes](https://www.intel.com/content/www/us/en/developer/articles/release-notes/intel-oneapi-threading-building-blocks-release-notes.html#inpage-nav-11) | - | - |
+| 2021.6.0 | 12.6 | Sep 2022 | [Release Notes](https://www.intel.com/content/www/us/en/developer/articles/release-notes/intel-oneapi-threading-building-blocks-release-notes.html#inpage-nav-12) | - | - |
+| 2021.5.0 | 12.5 | Dec 2021 | [Release Notes](https://www.intel.com/content/www/us/en/developer/articles/release-notes/intel-oneapi-threading-building-blocks-release-notes.html#inpage-nav-13) | - | - |
+| 2021.4.0 | 12.4 | Oct 2021 | [Release Notes](https://www.intel.com/content/www/us/en/developer/articles/release-notes/intel-oneapi-threading-building-blocks-release-notes.html#inpage-nav-14) | notify_waiters | - |
```
omalyshe: For those versions where only the RN link is provided - does it mean there were NO API changes at all? If it is not the case, then we need to list API changes for ALL versions. 
I see that for the latest versions API changes are listed.

VERSIONING.md
```diff
@@ -107,3 +107,26 @@ Rules for the `d` namespace include:
 - If the layout of public class X is changed incompatibly, the “d” namespace number should be incremented.
 - If there are existing entry points using the current version of class X, the old version of X should 
 be kept in the previous “d” namespace.
+
+## API and ABI changes by release
+
+| Library Version | Binary Version | Date | Release Notes | API Changes | ABI Changes | Notes |
```
omalyshe: I suggest clarifying that "Binary Version (Linux)".

akukanov: Or "runtime interface version".

akukanov:
> I suggest rewriting it using different format. How about the list of lists?

Markdown is knowingly bad for tables, except simplest ones. But the list of lists would be almost like the regular changelog, just for a subset of changes.

I would suggest moving the table out of Versioning.md into a separate file, perhaps in RST.
Another option is to make an "embedded" HTML table.

VERSIONING.md
```diff
@@ -107,3 +107,26 @@ Rules for the `d` namespace include:
...
+|---------|--------|------|---------------|-------------|-------------|-------|
+| 2022.3.0 | 12.17 | Oct 2025 | *Release Notes TBD* | [task_arena enqueue and wait_for specific task_group](https://github.com/uxlfoundation/oneTBB/tree/master/rfcs/proposed/task_arena_waiting), [custom assertion handler support](https://github.com/uxlfoundation/oneTBB/tree/master/rfcs/supported/assertion_handler), [preview of dynamic task graph](https://github.com/uxlfoundation/oneTBB/tree/master/rfcs/proposed/task_group_dynamic_dependencies) | set/get_assertion_handler, current_task_ptr | set/get_assertion_handler symbols are used by custom assertion handler support, current_task_ptr is used by preview of task_group dependencies |
```
m-fila: Links to the release notes are being de-synchronized when a new release is added since they are counted from top. Perhaps the link could go to something stable, like github markdown release notes where release number is actually part of the link https://github.com/uxlfoundation/oneTBB/blob/master/RELEASE_NOTES.md#onetbb-20222-release-notes

Also 2022.3.0 now has the release notes.

PR #1889, opened 2025-10-29
Fix a set of AddressSanitizer and ThreadSanitizer issues

isaevil:
### Description 

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/src/tbb/queuing_rw_mutex.cpp
+++ b/src/tbb/queuing_rw_mutex.cpp
@@ -348,6 +348,8 @@ struct queuing_rw_mutex_impl {
                     // Failed to acquire the lock on predecessor. The predecessor either unlinks or upgrades.
                     // In the second case, it could or could not know my "in use" flag - need to check
                     // Responsibility transition, the one who reads uncorrupted my_prev will do release.
+                    // Publish the restored predecessor with release, acquire ensures synchronization
+                    // with predecessor if they cleared the flag and wait for us to release the lock.
                     tmp = tricky_pointer::compare_exchange_strong(s.my_prev, tricky_pointer(predecessor) | FLAG, predecessor, std::memory_order_acq_rel);
                     if( !(tricky_pointer(tmp) & FLAG) ) {
                         __TBB_ASSERT(tricky_pointer::load(s.my_prev, std::memory_order_relaxed) != (tricky_pointer(predecessor) | FLAG), nullptr);
@@ -399,6 +401,8 @@ struct queuing_rw_mutex_impl {
                 }
                 next->my_going.store(2U, std::memory_order_relaxed);
                 // Responsibility transition, the one who reads uncorrupted my_prev will do release.
+                // Clear my_prev with release, acquire to synchronize successor's memory accesses to
+                // us before the release of the lock.
                 tmp = tricky_pointer::exchange(next->my_prev, nullptr, std::memory_order_acq_rel);
                 next->my_going.store(1U, std::memory_order_release);
             }
@@ -406,6 +410,7 @@ struct queuing_rw_mutex_impl {
             unblock_or_wait_on_internal_lock(s, get_flag(tmp));
         }
     done:
+        // Ensure that predecessor's writes are visible before node destruction.
         spin_wait_while_eq( s.my_going, 2U, std::memory_order_acquire );
 
         s.initialize();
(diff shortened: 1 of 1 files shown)


src/tbbmalloc/CMakeLists.txt
```diff
@@ -27,12 +28,14 @@ endif()
...
-                           $<$<CONFIG:DEBUG>:TBB_USE_DEBUG>
-                           PRIVATE
-                           __TBBMALLOC_BUILD
-                           $<$<NOT:$<BOOL:${BUILD_SHARED_LIBS}>>:__TBB_DYNAMIC_LOAD_ENABLED=0>
-                           $<$<NOT:$<BOOL:${BUILD_SHARED_LIBS}>>:__TBB_SOURCE_DIRECTLY_INCLUDED=1>)
+    PUBLIC
+    $<$<CONFIG:DEBUG>:TBB_USE_DEBUG>
+    PRIVATE
+    # Disable implicit assertions in C++ headers to avoid dependency on libstdc++
+    $<$<CONFIG:DEBUG>:_GLIBCXX_NO_ASSERTIONS>
```
isaevil, line 35:
```suggestion
```

src/tbb/queuing_rw_mutex.cpp
```diff
@@ -398,15 +399,14 @@ struct queuing_rw_mutex_impl {
...
-                tmp = tricky_pointer::exchange(next->my_prev, nullptr, std::memory_order_release);
+                tmp = tricky_pointer::exchange(next->my_prev, nullptr, std::memory_order_acq_rel);
                 next->my_going.store(1U, std::memory_order_release);
             }
     unlock_self:
             unblock_or_wait_on_internal_lock(s, get_flag(tmp));
         }
     done:
-        // Lifetime synchronization, no need to build happens-before relation
-        spin_wait_while_eq( s.my_going, 2U, std::memory_order_relaxed );
```
Alexandr-Konovalov, line 409: This change probably can have significant performance impact. Are we having a counterexample showing that change from relaxed to acquire is really needed?

isaevil: To be fair, it is the change I am the least confident with. Acquire memory order [was there before](https://github.com/uxlfoundation/oneTBB/blob/478de5b1887c928e52f029d706af6ea640a877be/src/tbb/queuing_rw_mutex.cpp#L378) but later it was changed. The change allowed me to silence one weird TSan's warning about data race but now I can't reproduce it for some reason.
Note that for x86 this change shouldn't have any performance impact: acquire memory order (nor acq_rel) won't introduce any new CPU instructions or change existing ones.

Alexandr-Konovalov: > Note that for x86 this change shouldn't have any performance impact: acquire memory order (nor acq_rel) won't introduce any new CPU instructions or change existing ones.

That is worrying me here, probably we may worsen the life on machines with relaxed memory model, while can't estimate the performance impact, as don't have such machines in the benchmarking pool.

In an ideal world, we may want to have a description what's going on here in term of fences and why acquire is required.

src/tbb/queuing_rw_mutex.cpp
```diff
@@ -1,5 +1,6 @@
 /*
-    Copyright (c) 2005-2024 Intel Corporation
+    Copyright (c) 2005-2025 Intel Corporation
```
isaevil:
```suggestion
    Copyright (c) 2005-2024 Intel Corporation
```

cmake/sanitize.cmake
```diff
@@ -1,4 +1,5 @@
-# Copyright (c) 2020-2022 Intel Corporation
+# Copyright (c) 2020-2025 Intel Corporation
```
isaevil:
```suggestion
# Copyright (c) 2020-2022 Intel Corporation
```

src/tbbmalloc/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
-# Copyright (c) 2020-2024 Intel Corporation
+# Copyright (c) 2020-2025 Intel Corporation
```
isaevil:
```suggestion
# Copyright (c) 2020-2024 Intel Corporation
```

src/tbb/queuing_rw_mutex.cpp
isaevil: > That is worrying me here, probably we may worsen the life on machines with relaxed memory model, while can't estimate the performance impact, as don't have such machines in the benchmarking pool.

Well, considering TSan's reports, it could be possible, that on systems with relaxed memory model these are real races and potential slowdowns are justified to improve correctness.

> In an ideal world, we may want to have a description what's going on here in term of fences and why acquire is required.

Right, but I am not very familiar with this implementation and it seems too complicated to find an explaining counterexample, to be honest.

vossmjp: Not very familiar myself, but this is a `try_release` and this line falls inside of the release for read (line 355), so there can be multiple threads in this area of code at the same, releasing read locks. So, the store of 1 to `my_going` at line 402 by a thread A could allow another thread B spinning at line 409 to proceed, which then calls `s.initialize()`. `initialize` writes to `prev` (as an example) which is also written to at line 401. So I think its possible for either A's or B's write to be the last one.

PR #1902, opened 2025-11-05
Implement API to create NUMA-bound task arenas

### LINKED ISSUE #1899: Add the function to create a set of NUMA bound arenas
Implement the API proposed in https://github.com/uxlfoundation/oneTBB/blob/master/rfcs/proposed/numa_support/create-numa-arenas.md and added to the oneTBB specification.

isaevil:
### Description 
For details see [the corresponding RFC](https://github.com/uxlfoundation/oneTBB/blob/de0daa2b2ff0a61c427a00bf71769f11ded9473e/rfcs/proposed/numa_support/create-numa-arenas.md).

Fixes #1899 

### Type of change

### Tests

- [ ] not needed

### Documentation

- [x] needs to be updated
- [ ] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/rfcs/supported/numa_support/create-numa-arenas.md
+++ b/rfcs/supported/numa_support/create-numa-arenas.md
@@ -48,7 +48,7 @@ The rest of the code in that example might be rewritten with the API for
   // enqueue work to all but the first arena, using the task groups to track work
   for (int i = 1; i < arenas.size(); i++)
     arenas[i].enqueue(
-      [] { tbb::parallel_for(0, N, [](int j) { f(w); }); }, 
+      [] { tbb::parallel_for(0, N, [](int j) { f(w); }); },
       task_groups[i]
     );
 
@@ -78,8 +78,11 @@ namespace tbb {
 ```
 
 It optionally takes a `constraints` argument to change default arena settings such as maximal concurrency
-(the upper limit on the number of threads), core type etc.; the `numa_id` value in `constraints_`
-is ignored. The second optional argument allows to override the number of reserved slots, which by default
+(the upper limit on the number of threads), core type etc., except for the `numa_id` value.
+The main goal of this API is to achieve simplicity in the creation of NUMA-bound task arenas,
+so passing `constraints` with an explicitly set `numa_id` won't result in an error, and there is no clear
+advantage to providing an error indication. Therefore, the `numa_id` value of `constraints` is ignored.
+The second optional argument allows to override the number of reserved slots, which by default
 is 0 (unlike the `task_arena` construction default of 1) for the reasons described in the introduction.
 
 These arena parameters were selected for pre-setting because there appear to be practical use cases to modify


test/conformance/conformance_arena_constraints.cpp
```diff
@@ -18,6 +18,8 @@
 //! \brief Test for [info_namespace scheduler.task_arena] specifications
 
 #include "common/common_arena_constraints.h"
+#include "common/test.h"
+#include "oneapi/tbb/info.h"
```
isaevil:
```suggestion
```

rfcs/supported/numa_support/create-numa-arenas.md
```diff
@@ -1,10 +1,10 @@
 # API to simplify creation of task arenas constrained to NUMA nodes
 
-This sub-RFC proposes an API to ease creation of a one-per-NUMA-node set of task arenas.
+This describes describes an API to ease creation of a one-per-NUMA-node set of task arenas.
```
aleksei-fedotov:
```suggestion
This RFC describes an API to ease creation of a one-per-NUMA-node set of task arenas.
```

test/conformance/conformance_arena_constraints.cpp
```diff
@@ -72,13 +72,35 @@ TEST_CASE("Test core types topology traversal correctness") {
 }
 #endif /*__HYBRID_CPUS_TESTING*/
 
+//! Testing create_numa_task_arenas helper function correctness
+//! \brief \ref interface \ref requirement
+TEST_CASE("Test create_numa_task_arenas conformance correctness") {
+  system_info::initialize();
+  auto numa_indices = oneapi::tbb::info::numa_nodes();
+  auto numa_task_arenas = oneapi::tbb::create_numa_task_arenas();
```
aleksei-fedotov: Let's take advantage of C++ type checking functionality and test what is guaranteed:
```suggestion
  using return_type = decltype(oneapi::tbb::create_numa_task_arenas());
  static_assert(
      std::is_same<std::vector<oneapi::tbb::task_arena>, return_type>::value, 
      "Return type of oneapi::tbb::create_numa_task_arenas() does not match the type guaranteed by oneAPI Specification"
  );
  return_type numa_task_arenas = oneapi::tbb::create_numa_task_arenas();

```

test/tbb/test_arena_constraints.cpp
```diff
@@ -92,6 +93,75 @@ TEST_CASE("Test constraints propagation during arenas copy construction") {
...
+    REQUIRE(!is_worker);
+  }
+};
+
+//! Test that reserved slots parameter makes expected effect on task_arena objects
+//! \brief \ref interface \ref error_guessing
+TEST_CASE("Test reserved slots argument in create_numa_task_arenas") {
+  system_info::initialize();
+  std::vector<index_info> numa_nodes_info = system_info::get_numa_nodes_info();
+  int expected_numa_concurrency = numa_nodes_info[0].concurrency;
```
aleksei-fedotov: Move further to the place of first use?

test/tbb/test_arena_constraints.cpp
```diff
@@ -92,6 +93,75 @@ TEST_CASE("Test constraints propagation during arenas copy construction") {
...
+    }
+  }
+
+  // Some Windows versions may map 1:1 NUMA node to Processor Group,
+  // so lets filter those configurations out
+  bool is_equal_numa_concurrency =
+      std::all_of(numa_nodes_info.begin(), numa_nodes_info.end(),
+                  [expected_numa_concurrency](const index_info &info) {
+                    return info.concurrency == expected_numa_concurrency;
+                  });
```
aleksei-fedotov: Does such versions of Windows may have different concurrencies in each two NUMA nodes?
Not sure I understand why do we need such filtering...

test/tbb/test_arena_constraints.cpp
```diff
@@ -92,6 +93,75 @@ TEST_CASE("Test constraints propagation during arenas copy construction") {
...
+      workerless_arena_observer observer{ta};
+
+      for (int i = 0; i < 100 * ta.max_concurrency(); ++i) {
+        ta.execute([&tg] { tg.run([] {}); });
+      };
+
+      utils::NativeParallelFor(ta.max_concurrency(),
+                               [&ta, &tg] (int) { ta.wait_for(tg); });
+    }
+  }
```
aleksei-fedotov: I think this part does not actually test anything. The problem is that application workers are allowed to join arena slots beyond `reserved_slots` number. Therefore, I would instead use the approach from the previous loop to check that library worker threads may join up to `max_concurrency - reserved_slots` arena slots.

test/tbb/test_arena_constraints.cpp
```diff
@@ -92,6 +93,75 @@ TEST_CASE("Test constraints propagation during arenas copy construction") {
...
+
+//! Test that constraints parameter is propagated to the task_arena construction while ignoring numa_id
+//! \brief \ref interface
+TEST_CASE("Test constraints argument in create_numa_task_arenas") {
+  system_info::initialize();
+  auto numa_indices = tbb::info::numa_nodes();
+  for (const auto& constraints : generate_constraints_variety()) {
+    auto numa_task_arenas = tbb::create_numa_task_arenas(constraints);
+    for (std::size_t i = 0; i < numa_indices.size(); ++i) {
+      tbb::task_arena::constraints expected_constraint = constraints;
```
aleksei-fedotov: This one can be simplified:
```suggestion
    tbb::task_arena::constraints expected_constraint = constraints;
    for (std::size_t i = 0; i < numa_indices.size(); ++i) {

```

test/conformance/conformance_arena_constraints.cpp
aleksei-fedotov, line 1: Do I understand correctly that the reason why some of the statements made in the spec are checked in non-conformance part of the test is because conformance part of the tests do not have necessary parts of testing framework available? And, if yes, should we made them available in conformance part?

rfcs/supported/numa_support/create-numa-arenas.md
```diff
@@ -215,10 +215,10 @@ extra complexity. As of now, we know just a few arena set patterns with potentia
 besides NUMA arenas, we can think of splitting CPU resources by core type and of creating a set of
 arenas with different priorities. We do not however recall any requests to simplify these use cases.
 
-## Open questions
+## Future Questions
 - Instead of a free-standing function in namespace `tbb`, should we consider
   a static member function in class `task_arena`?
```
aleksei-fedotov: I think this one should be deleted since we are introducing free-standing function already.
And the other about being an experimental feature as well.

include/oneapi/tbb/task_arena.h
```diff
@@ -28,6 +28,7 @@
 #include "detail/_task.h"
 
 #include "detail/_task_handle.h"
+#include <vector>
```
aleksei-fedotov: Move this include below to have it after all library includes. Similarly to what we adhere to in other files.

test/tbb/test_arena_constraints.cpp
isaevil: Rewrote the test.

test/conformance/conformance_arena_constraints.cpp
isaevil: Moved reserved_slots test into conformance tests. Since the spec doesn't guarantee hard thread affinity, I left the corresponding test in test_arena_constraints.

rfcs/supported/numa_support/create-numa-arenas.md
isaevil: Done.

include/oneapi/tbb/task_arena.h

src/tbb/arena.cpp
```diff
@@ -255,7 +255,7 @@ arena::arena(threading_control* control, unsigned num_slots, unsigned num_reserv
     my_limit = 1;
     // Two slots are mandatory: for the external thread, and for 1 worker (required to support starvation resistant tasks).
     my_num_slots = num_arena_slots(num_slots, num_reserved_slots);
-    my_num_reserved_slots = num_reserved_slots;
+    my_num_reserved_slots = std::min(num_reserved_slots, my_num_slots);
```
aleksei-fedotov: Consider using own function template as in `num_arena_slots`:
```suggestion
    my_num_reserved_slots = min(num_reserved_slots, my_num_slots);
```

test/conformance/conformance_arena_constraints.cpp
```diff
@@ -94,6 +100,67 @@ TEST_CASE("Test create_numa_task_arenas conformance correctness") {
...
+        int max_num_workers = ta_concurrency - std::min(ta_concurrency, reserved_slots);
+        int max_num_external_threads = reserved_slots == 0 && numa_task_arenas.size() == 1 ? 1 : reserved_slots;
+
+        join_arena_observer observer {ta, max_num_workers, max_num_external_threads};
+        utils::SpinBarrier barrier{(std::size_t)ta_concurrency};
+        for (int w = 0; w < ta_concurrency; ++w) {
+          ta.enqueue([&barrier] { barrier.wait(); }, tg);
+        }
+
+        std::this_thread::sleep_for(std::chrono::milliseconds{1});
```
aleksei-fedotov: I would also add a comment describing the intention. Perhaps, something like the following would suffice:
```suggestion
        // Waiting a bit to give workers an opportunity to occupy more arena slots than
        // are dedicated to workers. Thus, stressing the expectation that workers cannot occupy
        // reserved slots.
        if (reserved_slots > 0)
            std::this_thread::sleep_for(std::chrono::milliseconds{1});

```

test/conformance/conformance_arena_constraints.cpp
```diff
@@ -94,6 +100,67 @@ TEST_CASE("Test create_numa_task_arenas conformance correctness") {
...
+        utils::NativeParallelFor(reserved_slots,
+            [&ta, &tg] (int) { ta.wait_for(tg); });
+
+        if (reserved_slots == 0 && numa_task_arenas.size() == 1) {
+          // No extra worker is going to join the arena
+          observer.num_workers.fetch_add(1, std::memory_order_relaxed);
+          ta.wait_for(tg);
+        }
```
aleksei-fedotov: Understanding why this condition is needed takes me too much time. Could you please explain why it is here? Better not in the GitHub comment here but rather in the code.

include/oneapi/tbb/task_arena.h
```diff
@@ -640,6 +642,18 @@ inline void end_parallel_phase(bool with_fast_leave) {
...
 #endif
 
+inline std::vector<d1::task_arena> create_numa_task_arenas(d1::constraints c = {},
+                                                           unsigned reserved_slots = 0)
+{
+    static std::vector<numa_node_id> node_indices = d1::numa_nodes();
+    std::vector<d1::task_arena> numa_arenas;
+    numa_arenas.reserve(node_indices.size());
+    for (auto numa_id : node_indices) {
+      numa_arenas.emplace_back(c.set_numa_id(numa_id), reserved_slots);
```
aleksei-fedotov: As far as I can see, the indentation in this file adheres to four spaces where applicable.
```suggestion
        numa_arenas.emplace_back(c.set_numa_id(numa_id), reserved_slots);
```

test/conformance/conformance_arena_constraints.cpp
aleksei-fedotov: The last sentence in the description for [the linked constructor of task_arena](https://oneapi-spec.uxlfoundation.org/specifications/oneapi/latest/elements/onetbb/source/task_scheduler/task_arena/task_arena_cls#_CPPv410task_arena11constraintsj8priority), says:
> If `constraints::numa_node` is specified, then all threads that enter the arena are automatically pinned to corresponding NUMA node.

Does not this mean that each thread that joins such an arena should have a mask bound to a corresponding NUMA node set?

isaevil: Does it mean they will be hard pinned or soft pinned (for example using something like `SetThreadSelectedCpuSets` on Windows)? I don't think that we should assume that in conformance testing.

test/conformance/conformance_arena_constraints.cpp
isaevil: Done.

test/conformance/conformance_arena_constraints.cpp
aleksei-fedotov: I believe in conformance tests we should test exactly what is specified. The spec says _automatically pinned to a corresponding NUMA node_. So, I believe the test should check that the threads are automatically _pinned_, whatever is meant by this word in a particular implementation of the library.
However, does the spec or oneTBB implementation define the meaning of pinning? I was only able to find the indirect association of this word with `set_thread_affinity` and `restore_thread_affinity` functions whose use is illustrated in [this example](https://oneapi-spec.uxlfoundation.org/specifications/oneapi/latest/elements/onetbb/source/task_scheduler/task_arena/task_scheduler_observer_cls#example), about whose implementation one may only guess, not to mention that this is only a sketch example. So, strictly speaking the meaning of the phrase "pinning a thread to a NUMA node" seems to be undefined by the spec. In arbitrary implementation, the pinning maybe implemented by assigning a dedicated property of a thread to be equal to an ID of the NUMA node. Therefore, my opinion is that the test on pinning is kind of implementation-defined. Thus, can be left in non-conformance part.

test/conformance/conformance_arena_constraints.cpp
```diff
@@ -72,13 +72,109 @@ TEST_CASE("Test core types topology traversal correctness") {
...
+  std::atomic_int num_workers{};
+  std::atomic_int num_external_threads{};
+};
+
+//! Test that reserved slots parameter makes expected effect on task_arena objects
+//! \brief \ref interface \ref error_guessing
+TEST_CASE("Test reserved slots argument in create_numa_task_arenas") {
+  system_info::initialize();
+  std::vector<index_info> numa_nodes_info = system_info::get_numa_nodes_info();
+  int expected_numa_concurrency = numa_nodes_info[0].concurrency;
```
aleksei-fedotov: Shall the test use the concurrency of the first NUMA node or the one with the maximum/minimum value of concurrency?

test/conformance/conformance_arena_constraints.cpp
```diff
@@ -72,13 +72,109 @@ TEST_CASE("Test core types topology traversal correctness") {
...
+        if (reserved_slots == 0 && numa_task_arenas.size() == 1) {
+          // No extra worker is going to join the arena because having only NUMA node means that
+          // the default total number of workers is equal to concurrnecy of the single NUMA - 1.
+          // Thus, the external thread must join to complete work.
+          observer.num_workers.fetch_add(1, std::memory_order_relaxed);
+          ta.wait_for(tg);
+        }
+
+        REQUIRE(observer.num_workers.load(std::memory_order_relaxed) == max_num_workers);
+        REQUIRE(observer.num_external_threads.load(std::memory_order_relaxed) == max_num_external_threads);
```
aleksei-fedotov: The test may fail sporadically in the case of absence of reserved slots and multiple NUMA arenas as the application thread requires the work to be completed while nobody has waited on the task group within the arena.

Therefore, I suggest rewriting it to call the waiting unconditionally:
```suggestion
        utils::NativeParallelFor(reserved_slots - 1,
            [&ta, &tg] (int) { ta.wait_for(tg); });

        ta.wait_for(tg);

        REQUIRE(observer.num_workers.load(std::memory_order_relaxed) == max_num_workers);
        REQUIRE(observer.num_external_threads.load(std::memory_order_relaxed) == max_num_external_threads);
```
with corresponding adjustment of `max_num_external_threads` to be:
```cpp
int max_num_external_threads = std::max(1, reserved_slots);
```

What do you think?

test/conformance/conformance_arena_constraints.cpp
```diff
@@ -72,13 +72,109 @@ TEST_CASE("Test core types topology traversal correctness") {
...
+        join_arena_observer observer {ta, max_num_workers, max_num_external_threads};
+        utils::SpinBarrier barrier{(std::size_t)ta_concurrency};
+        for (int w = 0; w < ta_concurrency; ++w) {
+          ta.enqueue([&barrier] { barrier.wait(); }, tg);
+        }
+
+        // Waiting a bit to give workers an opportunity to occupy more arena slots than
+        // are dedicated to workers. Thus, stressing the expectation that workers cannot occupy
+        // reserved slots.
+        if (reserved_slots > 0)
```
aleksei-fedotov: No need to wait if this is worker-less arena.
```suggestion
        if (reserved_slots > 0 && max_num_workers > 0)
```

test/conformance/conformance_arena_constraints.cpp
```diff
@@ -72,13 +72,109 @@ TEST_CASE("Test core types topology traversal correctness") {
...
+  for (int reserved_slots = 0; reserved_slots < expected_numa_concurrency; ++reserved_slots) {
+      auto numa_task_arenas = tbb::create_numa_task_arenas({}, reserved_slots);
+      tbb::task_group tg{};
+      for (auto& ta : numa_task_arenas) {
+        int ta_concurrency = ta.max_concurrency();
+        int max_num_workers = ta_concurrency - std::min(ta_concurrency, reserved_slots);
+        int max_num_external_threads = reserved_slots == 0 && numa_task_arenas.size() == 1 ? 1 : reserved_slots;
+
+        join_arena_observer observer {ta, max_num_workers, max_num_external_threads};
+        utils::SpinBarrier barrier{(std::size_t)ta_concurrency};
```
aleksei-fedotov: Adjust barrier counter accordingly? See my proposal wrt test robustness.

Something like:
```suggestion
        utils::SpinBarrier barrier{(std::size_t)ta_concurrency + int(!reserved_slots)};
```

test/conformance/conformance_arena_constraints.cpp
isaevil: I think using the maximum value is better, so we can be sure that number of reserved slots won't be less than arena's maximum concurrency for corresponding test case.

isaevil: Updated.

test/conformance/conformance_arena_constraints.cpp
isaevil: Updated the test.

test/conformance/conformance_arena_constraints.cpp
```diff
@@ -75,97 +75,132 @@ TEST_CASE("Test core types topology traversal correctness") {
...
-  void on_scheduler_entry(bool is_worker) override {
-    int prev;
-    if (is_worker) {
-      prev = num_workers.fetch_add(1, std::memory_order_relaxed);
-      REQUIRE_MESSAGE(prev+1 <= max_num_workers, "More than expected worker threads has joined arena");
-    } else {
-      prev = num_external_threads.fetch_add(1, std::memory_order_relaxed);
-      REQUIRE_MESSAGE(prev+1 <= max_num_external_threads, "More than expected external threads has joined arena");
+    join_arena_observer(tbb::task_arena &ta, int max_workers, int max_external_threads)
+        : tbb::task_scheduler_observer(ta), max_num_workers(max_workers), max_num_external_threads(max_external_threads)
```
aleksei-fedotov: 112 characters long. Let's try sticking to 100?
```suggestion
        : tbb::task_scheduler_observer(ta)
        , max_num_workers(max_workers), max_num_external_threads(max_external_threads)
```

test/conformance/conformance_arena_constraints.cpp
```diff
@@ -75,97 +75,132 @@ TEST_CASE("Test core types topology traversal correctness") {
...
-    auto constraint_concurrency = oneapi::tbb::info::default_concurrency(c);
-    REQUIRE_MESSAGE(constraint_concurrency == numa_task_arenas[numa_i].max_concurrency(),
-        "Maximum concurrency level of task_arena should be the same as for constraints");
-  }
+    system_info::initialize();
+    auto numa_indices = oneapi::tbb::info::numa_nodes();
+    using return_type = decltype(oneapi::tbb::create_numa_task_arenas());
+    static_assert(
+        std::is_same<std::vector<oneapi::tbb::task_arena>, return_type>::value,
+        "Return type of oneapi::tbb::create_numa_task_arenas() does not match the type guaranteed by oneAPI Specification"
```
aleksei-fedotov: Likewise, make it a bit shorter to be within 100 characters, but also to be more future-proof?
```suggestion
        "Return type of oneapi::tbb::create_numa_task_arenas() does not match the expected type"
```

test/conformance/conformance_arena_constraints.cpp
```diff
@@ -75,97 +75,132 @@ TEST_CASE("Test core types topology traversal correctness") {
...
 
-  int max_num_workers;
-  int max_num_external_threads;
-  std::atomic_int num_workers{};
-  std::atomic_int num_external_threads{};
+    void on_scheduler_entry(bool is_worker) override {
+        int current;
+        int expected_peak;
+        if (is_worker) {
+            current = num_workers.fetch_add(1, std::memory_order_relaxed) + 1;
```
aleksei-fedotov, line 116: If you don't want to reuse `ConcurrencyTracker` in this patch, please put a TODO comment at least not to forget about it, saying something like:
```suggestion
            // TODO: Adapt utils::ConcurrencyTracker for its reuse here and in the else branch below
            current = num_workers.fetch_add(1, std::memory_order_relaxed) + 1;
```

test/conformance/conformance_arena_constraints.cpp
```diff
@@ -75,97 +75,132 @@ TEST_CASE("Test core types topology traversal correctness") {
...
+    void on_scheduler_exit(bool is_worker) override {
+      if (is_worker) {
+        num_workers.fetch_sub(1, std::memory_order_relaxed);
+      } else {
+        num_external_threads.fetch_sub(1, std::memory_order_relaxed);
+      }
+    }
+
+    int max_num_workers;
+    int max_num_external_threads;
```
aleksei-fedotov: Make them constants to avoid accidental change?
```suggestion
    const int max_num_workers;
    const int max_num_external_threads;
```

test/conformance/conformance_arena_constraints.cpp
```diff
@@ -75,97 +75,132 @@ TEST_CASE("Test core types topology traversal correctness") {
...
+        auto numa_task_arenas = tbb::create_numa_task_arenas({}, reserved_slots);
+        tbb::task_group tg{};
+        // Having only NUMA node means that the default total number of workers is equal to
+        // concurrnecy of the single NUMA - 1. This means that for task_arena with reserved_slots=0
+        // worker threads won't be able to fully saturate the arena.
+        // This flag is set to adjust test expectations accordingly.
+        bool workers_cannot_fully_occupy_arena = numa_nodes_info.size() == 1 && reserved_slots == 0;
+        for (auto& ta : numa_task_arenas) {
+            int ta_concurrency;
+            ta.enqueue([&ta_concurrency] { ta_concurrency = tbb::this_task_arena::max_concurrency(); }, tg);
```
aleksei-fedotov, line 174: Would it be good if we described in the comment the reason why a call to `ta.max_concurrency()` outside the task arena is not enough?

src/tbb/arena.cpp
```diff
@@ -255,8 +255,8 @@ arena::arena(threading_control* control, unsigned num_slots, unsigned num_reserv
     my_limit = 1;
     // Two slots are mandatory: for the external thread, and for 1 worker (required to support starvation resistant tasks).
     my_num_slots = num_arena_slots(num_slots, num_reserved_slots);
-    my_num_reserved_slots = num_reserved_slots;
-    my_max_num_workers = num_slots-num_reserved_slots;
+    my_num_reserved_slots = min(num_reserved_slots, my_num_slots);
+    my_max_num_workers = num_slots-my_num_reserved_slots;
```
aleksei-fedotov: Since this behavior is now expected in this patch, please add to this patch _separate_ regression test that will instantiate `task_arena` with the number of `reserved_slots` greater than task arena concurrency.

aleksei-fedotov (commented):
Almost complete! A few nitpickings from my side with one request for a new test case.

rfcs/supported/numa_support/create-numa-arenas.md
```diff
@@ -215,12 +215,9 @@ extra complexity. As of now, we know just a few arena set patterns with potentia
...
-## Open questions
-- Instead of a free-standing function in namespace `tbb`, should we consider
-  a static member function in class `task_arena`?
-- The proposal does not consider arena priority, simply keeping the default `priority::normal`.
+## Future Questions
+- At this point arena priority parameter is not considered, simply keeping the default `priority::normal`.
   Are there use cases for pre-setting priorities? Similarly for the experimental thread leave policy.
 - Are there more practical use cases which could justify the universal function approach?
 - Need to consider alternatives to silently ignoring `numa_id` in constraints, such as an exception
   or undefined behavior.
```
akukanov: I think we have discussed all these questions and decided in favor of the proposed and implemented solution. If some of these notes are considered as possible future extensions, the wording should change correspondingly. The notes that are no more relevant should be removed (like the one about function location, thanks). Specifically I count the last note about exceptions/UB irrelevant, as such a change would break compatibility.

isaevil: Updated.

rfcs/supported/numa_support/create-numa-arenas.md
```diff
@@ -215,12 +215,6 @@ extra complexity. As of now, we know just a few arena set patterns with potentia
...
-  a static member function in class `task_arena`?
-- The proposal does not consider arena priority, simply keeping the default `priority::normal`.
-  Are there use cases for pre-setting priorities? Similarly for the experimental thread leave policy.
-- Are there more practical use cases which could justify the universal function approach?
-- Need to consider alternatives to silently ignoring `numa_id` in constraints, such as an exception
-  or undefined behavior.
-- Are there any reasons for the API to first go out as an experimental feature?
+## Future Extensions
+- Consider extending API to support pre-setting arena priority and experimental worker thread leave policy.
+- Revisit the universal function approach if more use cases arise that would benefit from such flexibility.
```
aleksei-fedotov, line 223: I propose adding the results of discussions on these questions, describing why a particular decision was made. This could help future readers in better understanding of the forces that led to these decisions.

isaevil: Done.

PR #1914, opened 2025-11-19
Use compiler's modern name, document cmake controls.

Alexandr-Konovalov:
### Description 

icpc is classic compiler, modern names are icpx/icx. Also add missing controls to README.

### Type of change

### Tests

- [X] not needed

### Documentation

- [ ] needs to be updated
- [X] not needed

- [ ] Yes
- [X] No
- [ ] Unknown

### DIFF
--- a/cmake/README.md
+++ b/cmake/README.md
@@ -156,8 +156,8 @@ cmake -DTBB_WINDOWS_DRIVER=ON -DCMAKE_MSVC_RUNTIME_LIBRARY=MultiThreaded ..
 #### Example
 
 ```bash
-cmake -DCMAKE_CXX_COMPILER=icpx -DCMAKE_C_COMPILER=icx -DTBB_TEST=off -DCMAKE_HWLOC_1_11_LIBRARY_PATH=<path_to_hwloc_library_file>/libhwloc.so.15
--DCMAKE_HWLOC_1_11_INCLUDE_PATH=<path_to_hwloc_header_directory> -DCMAKE_INSTALL_PREFIX=<path_to_install_oneTBB>/oneTBB_install ..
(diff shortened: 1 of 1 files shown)


python/setup.py
dnmokhov: Isn't icx-cl also possible on Windows?

Also, for lines 47-54 below, I think that icx-cc uses Linux-like options: `-qtbb`, etc.

Alexandr-Konovalov: Thanks, you are right, on Windows we should use only icx-cl and not icx-cc, as only icx-cl accepts Windows-like options.

But for me distutils always calls cl.exe somehow. I have to understand why.

Alexandr-Konovalov: On my machine python 3.14.0 uses setuptools 80.9.0, and in setuptools I see that compiler customization via environment variables [is possible](https://github.com/pypa/setuptools/blob/d198e86f57231e83de87975c5c82bc40c196da79/setuptools/_distutils/sysconfig.py#L301) only under Unix, cygwin or MinGW. I may miss something.

So, can we add `icx`/`icpx` to cover Linux and ignore Windows?

python/setup.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2016-2023 Intel Corporation
+# Copyright (c) 2016-2025 Intel Corporation
 #
```
omalyshe, line 3: Also need to add UXL copyright like in https://github.com/uxlfoundation/oneTBB/commit/cfd170368ec6134696b5794b10de76e90f41abd6#diff-93a059393f5350e47303a2426e2c5390e8c587b8a7533eae2c58cd6b6ed2e9d5R2

Alexandr-Konovalov: Done.

PR #1917, opened 2025-11-24
[RFC] Advanced Core Type Selection

dnmokhov:
Add an RFC describing setting multiple core types in task arena constraints.

Reference implementation: [dev/dnmokhov/core-types](../tree/dev/dnmokhov/core-types)

### DIFF
--- a/rfcs/proposed/core_types/README.md
+++ b/rfcs/proposed/core_types/README.md
@@ -73,7 +73,7 @@ Similarly, we can create/initialize an arena for core type(s) selected by a user
 #### Header
 
 ```cpp
-#define TBB_PREVIEW_AFFINITY_SELECTOR 1
+#define TBB_PREVIEW_TASK_ARENA_CORE_TYPE_SELECTOR 1
 #include <oneapi/tbb/task_arena.h>
 #include <oneapi/tbb/info.h>
 ```
@@ -199,7 +199,9 @@ patterns from 1 to 2<sup>n</sup>-1, and mapping each pattern to a core type comb
    - Alternative: leave the arena uninitialized
 2. Should the selector API support NUMA node selection in addition to core types?
    - Current proposal: no. The selector is applied only to core types; NUMA node constraints continue
-     to use the existing `constraints::numa_id` field.
+     to use the existing `constraints::numa_id` field. Core types have unique performance characteristics worth
+     ranking, whereas NUMA nodes typically have equivalent capabilities differing only by ID, making function-style
+     selection inconvenient without dynamic information (e.g., utilization).
    - Alternative: allow the selector to also rank NUMA nodes. (This should become a sub-RFC of the existing
      [umbrella RFC about improving NUMA support](https://github.com/uxlfoundation/oneTBB/tree/master/rfcs/proposed/numa_support).)
      - This raises the question of how to distinguish which constraint dimension the selector applies to, because
@@ -496,7 +498,7 @@ tbb::task_arena throughput_driven({
     tbb::task_arena::constraints{.core_type = core_types[2]}
 });
 
-tbb::task_arena background_work(}
+tbb::task_arena background_work({
     tbb::task_arena::constraints{.core_type = core_types[0]},
     tbb::task_arena::constraints{.core_type = core_types[1]}
 });


dnmokhov:
@wangleis @sunxiaoxia2022, this is an RFC for adding multiple core type selection to the master branch. Feel free to provide feedback. Thanks!

rfcs/proposed/core_types/README.md
```diff
@@ -0,0 +1,285 @@
...
+```
+
+This forces applications to choose one of these suboptimal strategies:
+
+| Strategy | Pros | Cons |
+|----------|------|------|
+| **P-cores only** | Maximum single-threaded performance | Leaves E-cores idle; limited parallelism; higher power |
+| **E-cores only** | Good for parallel workloads | Doesn't utilize P-core performance; excludes LP E-cores |
+| **LP E-cores only** | Minimal power consumption | Severe performance impact for most workloads |
+| **No constraint** | Maximum flexibility | May schedule on inappropriate cores (e.g., LP E-cores for compute) |
```
wangleis: P-cores + E-cores are required, but for latency mode with shared L3 cache, LPE cores should be avoided.

dnmokhov: Thanks for confirming. As it says in the following paragraph,
https://github.com/uxlfoundation/oneTBB/blob/d9af5f4eb479f12c98e277de890896edd1c843dc/rfcs/proposed/core_types/README.md?plain=1#L57

rfcs/proposed/core_types/README.md
```diff
@@ -0,0 +1,285 @@
...
+constraints& add_core_type(core_type_id id);
+constraints& remove_core_type(core_type_id id);
+void clear_core_types();
+```
+
+5. **Core Types Data member**: Should we store core types in a new container data member instead of bit-packing?
+   - Pros: Simpler logic, easier to extend
+   - Cons: Increases struct size, breaks ABI compatibility
+
+6. **Info API**: Should `info::core_types()` be extended to return a count instead of/in addition to a vector?
```
aleksei-fedotov: `info::core_types()` returns `std::vector<core_type_id>`. So, if I understand the sixth question correctly, the count can be retrieved via `ct = info::core_types(); ct.size()`.

rfcs/proposed/core_types/README.md
```diff
@@ -0,0 +1,285 @@
...
+
+## Proposal
+
+We propose extending the `task_arena::constraints` API to support specifying multiple acceptable core types, enabling
+applications to define flexible core type policies that adapt to workload requirements and available hardware
+resources.
+
+### New API
+
+Add the following methods to `tbb::task_arena::constraints`:
```
aleksei-fedotov: Another possibility can be to leave `constraints` struct unchanged, but introduce, for example, a new constructor for the `task_arena` class that would accept vector/array of `constraints` instances, each bound to certain NUMA, core type and threads per core constraints. The task arena constructor would make a union of the masks resulting from each of `constraints` instance and use that union as its constraint.
At first glance, this design looks more flexible to me as it scales better, allowing users not only specifying more than one core type, but also more than one NUMA node. Essentially, users can specify multiple portions of the platform, whose united constraint is desired to be set for a single `task_arena` instance.

rfcs/proposed/core_types/README.md
dnmokhov: I guess I only meant "instead", the main idea being that

`info::core_types()` → {0, 1, ..., n-1}

is no more useful than something like

`info::num_core_types()` → n

rfcs/proposed/core_types/README.md
dnmokhov: Sounds like TCM. 😉 I will try to add this as an alternative.

rfcs/proposed/core_types/README.md
```diff
@@ -0,0 +1,285 @@
+# Advanced Core Type Selection
+
+## Introduction
+
+### Motivation
+
+The current oneTBB API allows users to constrain task execution to a single core type using
```
vossmjp, line 8:
```suggestion
By default, oneTBB includes all available core types in a task arena unless explicitly constrained. 
The current oneTBB API allows users to constrain task execution to a single core type using
```

rfcs/proposed/core_types/README.md
```diff
@@ -0,0 +1,285 @@
...
+`task_arena::constraints::set_core_type(core_type_id)`. While this provides control, it creates limitations for
+real-world applications running on processors with more than two core types (e.g., on a system with performance (P),
+efficient (E), and low power efficient (LP E) cores):
+
+#### 1. **Flexibility and Resource Utilization**
+
+Many parallel workloads can execute efficiently on multiple core types. For example:
```
vossmjp:
```suggestion
While it is often best to allow the OS to use all core types and flexibly schedule threads, some advanced users may find it necessary to constrain scheduling. 
When there are more than two core types, it may be desired to constrain execution to not just a single core type.
Many parallel workloads can execute efficiently on multiple core types that make up a subset of the available core types. For example:
```

rfcs/proposed/core_types/README.md
```diff
@@ -0,0 +1,285 @@
...
+#### 2. **Workload Classification Challenges**
+
+Applications often have workloads that don't fit neatly into a single core type category:
+- **Moderate priority tasks**: Not critical enough to demand P-cores exclusively, but shouldn't use LP E-cores
+- **Adaptive workloads**: Performance requirements that vary based on input size or system state
+- **Mixed computation phases**: Algorithms that alternate between compute-intensive and throughput-oriented phases
+
+#### 3. **Avoiding Inappropriate Core Selection**
+
+Without the ability to specify "P-cores OR E-cores (but not LP E-cores)", applications face a dilemma:
```
vossmjp:
```suggestion
Without the ability to specify "P-cores OR E-cores (but not LP E-cores)" or 
"LP E-cores and E-cores but not P-cores" applications face dilemmas.
For example, without being able to specify "P-cores OR E-cores (but not LP E-cores)":
```

rfcs/proposed/core_types/README.md
```diff
@@ -0,0 +1,285 @@
...
+);
+```
+
+This forces applications to choose one of these suboptimal strategies:
+
+| Strategy | Pros | Cons |
+|----------|------|------|
+| **P-cores only** | Maximum single-threaded performance | Leaves E-cores idle; limited parallelism; higher power |
+| **E-cores only** | Good for parallel workloads | Doesn't utilize P-core performance; excludes LP E-cores |
+| **LP E-cores only** | Minimal power consumption | Severe performance impact for most workloads |
```
vossmjp:
```suggestion
| **LP E-cores only** | Minimal power consumption | Severe performance impact for some workloads that require large, shared caches. |
```

rfcs/proposed/core_types/README.md
vossmjp: Along this same line of thinking, if there is a ``set_core_types`` there should likely, eventually, be a ``set_num_ids`` function. We should consider which is easier to reason about, a combination created from a vector of constraints, or these specific functions.

dnmokhov: Added a section on the above.

dnmokhov:
> In general this RFC should not read like guidance on how to select which core type(s) to use based on application characteristics. The relative capabilities of core types may differ based on the HW platform and benefits will be highly application dependent. Instead, it should describe use cases with appropriate caveats that its generally better to the let the OS decide and that constraints should be applied carefully.

Added several clarifications w.r.t. this topic.

rfcs/proposed/core_types/README.md
```diff
@@ -0,0 +1,351 @@
...
+**Future Extensibility Consideration:** This approach naturally extends to other constraint types—if `set_core_types`
+is added, a corresponding `set_numa_ids` function would likely follow. The choice between a vector of `constraints`
+instances versus dedicated multi-value setters affects API consistency and usability: the former provides a unified
+pattern for combining any constraints, while the latter offers more intuitive, type-specific methods.
+
+## Open Questions
+
+1. **API Naming**: Is `set_core_types` (plural) sufficiently distinct from `set_core_type` (singular)?
+   - Alternative: overload the existing `set_core_type` to accept `vector<core_type_id>`
+   - Alternative: `set_acceptable_core_types` or `allow_core_types`
```
isaevil: Is it necessary to differentiate it by name? I think that the overload alternative is quite good and user won't have to choose constantly between "singular" and "plural" functions. Even if the vector with a single element is passed, we can forward it to the old `set_core_type(core_type_id)` function.

rfcs/proposed/core_types/README.md
```diff
@@ -253,6 +284,40 @@ The test infrastructure could generate all possible core type combinations using
...
+**Pros:**
+- More scalable: can extend to any other constraint type and specify multiple platform portions as a unified constraint
+- Reuses existing `constraints` struct without modification
+- Avoids bit-packing, format markers, and special value handling
+- No risk of misinterpretation of existing single core type constraints
+
+**Cons:**
+- Requires creating multiple `constraints` objects for simple core type combinations
+- Vector of `constraints` instances vs. single integer field with bit-packing creates memory overhead
+- Unclear how to handle conflicting `max_concurrency` or `max_threads_per_core` across instances
```
aleksei-fedotov: A possible way of handling conflicting `max_concurrency` or `max_threads_per_core` can be interpreting each constraint instance as an independent constraint with its own `max_concurrency` and `max_threads_per_core`. Once affinity mask is determined for each constraint, the union of them is created, so the sum of `max_concurrency`-ies, which can be used the overall `max_concurrency` for result `task_arena` instance.

`max_threads_per_core` is not needed to be handled anyhow specifically as it is directly reflected in the mask of the constraint.

rfcs/proposed/core_types/README.md
```diff
@@ -253,6 +284,40 @@ The test infrastructure could generate all possible core type combinations using
...
+- Library entry points `constraints_default_concurrency()` and `constraints_threads_per_core()` accept single
+  `constraints`; would require new overloads or replacement APIs, affecting ABI
```
aleksei-fedotov: I think this can be handled on the caller side by calling existing API multiple times, each time for a separate instance of constraints.

rfcs/proposed/core_types/README.md
vossmjp: I asked a similar question in [PR1926](https://github.com/uxlfoundation/oneTBB/pull/1926) but it applies here too. Assume there is a system with three core types, indexes 0, 1 and 2 and there are 4 of each kind. What if the user wants 6 slots and to use the most performant cores. From what you describe, I think they would create a constraint for index 2 and max_concurrency of 4 slots. And then another for index 1 and max_concurrency of 2?  Is that right?  The result would be a max_concurrency of six and a mask that includes both index 2 and index 1 core types?  So if all core types are idle, the arena may populate with 4 index 1 types and 2 index 2 types, or 2 index 1 types and 4 index 2 types, right?  Would one be expected over the other?

dnmokhov: I have removed this from the cons. If we adopt this alternative, the new behavior will be part of the proposal.

rfcs/proposed/core_types/README.md
dnmokhov: I have removed this from the cons. If we adopt this alternative, the new usage mode will be part of the proposal.

rfcs/proposed/core_types/README.md
aleksei-fedotov: @vossmjp I think in this case the only possible way to populate arena would be with 4 index 2 and 2 index 1 threads, since user has specified such constraints that result in the mask that maps threads joining the arena to 4 P-cores and 2 E-cores.

vossmjp: But you would need to pick which e-cores are in the mask.

aleksei-fedotov: Yeah, I understand your question now. It seems we can try applying the same approach we are using right now, which is "Let the OS decide". Currently, if the user specifies 2 threads as `max_concurrency` for the arena constrained to E-cores on the platform which has, let's say, 10 of them, the mask will be created for all these 10 e-cores, right? But only two threads will join the arena and has their affinity set to the mask.
I guess with multiple constraints, arena should apply the mask per constraint. For the example you described above, first four threads joining the arena will have affinity of the P-cores, while the last two threads will be assigned E-cores mask. 

Or simply extend the current logic - do not make any preference for what cores to populate first. That is, use the united mask that will include all P-cores and all E-cores, and let the system decide on the threads migration.

akukanov: > I guess with multiple constraints, arena should apply the mask per constraint. For the example you described above, first four threads joining the arena will have affinity of the P-cores, while the last two threads will be assigned E-cores mask.

That makes most sense for the "multiple constraints" proposal, but it would complicate the implementation - you would need to keep track of which thread contributes to which constraint, choose which mask to apply to a joining thread, possibly decide which threads should leave the arena first if its allotment is reduced, etc.

> Or simply extend the current logic - do not make any preference for what cores to populate first. That is, use the united mask that will include all P-cores and all E-cores, and let the system decide on the threads migration.

That might be confusing / unexpected for users - what appears to be two independent constraints in the API would be de-facto a single aggregated one, with concurrencies summed up and masks combined.

rfcs/proposed/core_types/README.md
```diff
@@ -0,0 +1,653 @@
...
+- **Mixed computation phases**: Algorithms that alternate between compute-intensive and throughput-oriented phases
+
+#### 3. **Avoiding Inappropriate Core Selection**
+
+Without the ability to specify "P-cores OR E-cores (but not LP E-cores)" or 
+"LP E-cores OR E-cores but not P-cores" applications face dilemmas.
+For example, without being able to specify "P-cores OR E-cores (but not LP E-cores)":
+- **No constraint**: Work might be scheduled on LP E-cores, causing significant performance degradation
+- **P-cores only**: Leaves E-cores idle, reducing parallelism
+- **E-cores only**: Misses opportunities to use faster P-cores when available
```
vossmjp:
```suggestion
Let's assume that a specific workload is known to perform well on both E-cores and P-cores, but very poorly on LP E-cores. 
Without the ability to specify "P-cores OR E-cores (but not LP E-cores)", applications face dilemmas:
- **No constraint**: Work might be scheduled on LP E-cores, causing significant performance degradation
- **P-cores only**: Unnecessarily leaves E-cores idle, reducing parallelism
- **E-cores only**: Misses opportunities to use faster P-cores when available
```

rfcs/proposed/core_types/README.md
```diff
@@ -0,0 +1,653 @@
...
+Restricting to a single core type may leave available cores idle, reducing overall system throughput.
+
+#### 2. **Workload Classification Challenges**
+
+Applications often have workloads that don't fit neatly into a single core type category:
+- **Moderate priority tasks**: Not critical enough to demand P-cores exclusively, but shouldn't use LP E-cores
+- **Adaptive workloads**: Performance requirements that vary based on input size or system state
+- **Mixed computation phases**: Algorithms that alternate between compute-intensive and throughput-oriented phases
+
+#### 3. **Avoiding Inappropriate Core Selection**
```
vossmjp:
```suggestion
#### 2. **Avoiding Inappropriate Core Selection**
```

vossmjp: I think we can shorten this whole section, while also avoiding the appearance of given classification advice that may not be applicable across different hardware platforms.

rfcs/proposed/core_types/README.md
vossmjp: I added the assumption, so it doesn't appear like we're giving general advice.

rfcs/proposed/core_types/README.md
```diff
@@ -0,0 +1,645 @@
...
+```
+
+### Example 3: Code that does not depend on oneTBB binary versions
+
+This example adjusts the code in Example 1 to illustrates how an application could be written so that
+it uses the new arena capabilities when present, otherwise falls back to the old arena construction.
+
+#### Proposed API
+
+The runtime version check can be hidden within the implementation, and the type with the highest score takes
```
wangleis, line 533: It may be preferable to expose a query interface that allows users to inspect runtime behavior dynamically. With this information, users can determine the most appropriate execution configuration for their specific environment.

For example, consider a platform with 2 P-cores, 8 E-cores, and 4 LP E-cores. On a newer runtime, the optimal configuration might be to utilize only the 2 P-cores and 8 E-cores. However, when running on an older runtime, this same configuration may effectively use only the 2 P-cores due to limited support. In contrast, the optimal configuration on the older runtime might be to enable all available cores.

Providing a runtime query function would allow users to adapt their configuration accordingly and achieve better performance across different runtime versions.

akukanov: The TBB runtime version query is available via `TBB_runtime_interface_version()`; see the last example in the document where it is used.

You raised a good point however - what if the desired fallback approach is not to take the best scored constraint but have no constraints at all? That's another open question I think.

One possible way to address is to use the score of 0 as a "mild exclusion" - that is, exclude this core type from the arena if possible, but if not, keep it. In that case, the selector function may return something like {0, 1, 2} for LPE-, E-, and P-cores correspondingly, and the implementation would exclude LPE based on the score of 0 if advanced constraints are available, and exclude nothing otherwise.

dnmokhov: Added to open questions.

rfcs/proposed/core_types/README.md
```diff
@@ -0,0 +1,650 @@
...
+### Compatibility Requirements
+
+This proposal must maintain compatibility with previous oneTBB library versions:
+- **API and Backward Compatibility (Old Application + New Library)**: Existing code using the current
+  `set_core_type(core_type_id)` API must compile and behave identically with newer oneTBB binaries.
+- **Binary Compatibility (ABI)**: The `task_arena::constraints` struct layout must remain unchanged.
+- **Forward Compatibility (New Application + Old Library)**: Applications compiled with the proposed new functionality
+  must be able to handle execution against older oneTBB binaries gracefully, without crashes or undefined behavior.
+  This requirement is mandated by the necessity for the key customer, OpenVINO, to support their users
+  that might not be ready to upgrade to the latest version of oneTBB.
```
vossmjp:
```suggestion
- **Forward Compatibility (New Application + Old Library)**: In general, oneTBB does **NOT** support forward compatibility.  
  However for limited use cases where only specific oneTBB functionality is required, it may be possible to compile using a subset of new headers and load an old binary library.
 Considering this feature in isolation, applications compiled with the proposed new functionality
  must be able to handle execution against older oneTBB binaries gracefully, without crashes or undefined behavior.
  This requirement is mandated by the necessity for the key customer, OpenVINO, to support their users
  that might not be ready to upgrade to the latest version of oneTBB.
  This goal does not imply any general form of forward compatibility.
```

rfcs/proposed/core_types/README.md
```diff
@@ -0,0 +1,650 @@
...
+
+## Proposal
+
+This proposal takes a different approach to the API, motivated by
+[SYCL device selectors](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:device-selection).
+In SYCL, "a device selector which is a ranking function that will give an integer ranking value to all the devices
```
vossmjp:
```suggestion
In SYCL, "a device selector ... is a ranking function that will give an integer ranking value to all the devices
```

rfcs/proposed/core_types/README.md
```diff
@@ -0,0 +1,650 @@
...
+on the system". It takes a device as an argument and returns a score for that device, according to user's criteria.
+The SYCL implementation calls that function on each device, and then selects one with the highest score.
+
+Similarly, we can create/initialize an arena for core type(s) or NUMA node(s) selected by a user-provided function.
+
+This approach has several advantages over Alternative 2 ("Extend Constraints API In-Place") below:
```
vossmjp: This is now strange placement, since it talks about an alternative that has not yet been described. It made more sense when the alternative was the preferred choice. But since that has changed, I think this whole bulleted section can be safely removed.

rfcs/proposed/core_types/README.md
```diff
@@ -0,0 +1,650 @@
...
+A negative score would indicate that the corresponding core type or NUMA node should be excluded from use
+by the task arena. A positive score would indicate that the corresponding core type or NUMA node is good to use
+by the task arena, and the bigger the score the better the "resource" is from the user's viewpoint.
+If for whatever reason (some described below) the implementation can only use a single core type or NUMA node,
+it should take the one with the biggest score.
+
+### Implementation aspects
+
+The key implementation problem is how to pass the additional information about multiple core types (or NUMA nodes)
+to the TBB library functions for arena creation without violating the forward compatibility requirement.
```
vossmjp:
```suggestion
to the TBB library functions for arena creation without violating the forward compatibility requirement for this specific feature.
```

rfcs/proposed/core_types/README.md
dnmokhov: Done

dnmokhov:
> I think you should state the specific choice for some of the options as this should move directly into implementation. So, for example, state whether both `std::tuple` or `std::vector` will be supported, or both. And state how selector will distinguish between code type or NUMA nodes. Since this is targeting 2023.0, the choice should be stated. And non-selected options can be kept to be possibly be revisited but as part of open-questions. For example, choose `std::tuple` only and as an open-question ask whether `std::vector` is needed.

Done

rfcs/proposed/core_types/README.md
```diff
@@ -0,0 +1,597 @@
...
+`task_arena::constraints::set_core_type(core_type_id)`. While this provides control, it creates limitations for
+real-world applications running on processors with more than two core types (e.g., on a system with performance (P),
+efficient (E), and low power efficient (LP E) cores):
+
+#### 1. **Flexibility and Resource Utilization**
+
+While it is often best to allow the OS to use all core types and flexibly schedule threads, some advanced users may find it necessary to constrain scheduling.
+When there are more than two core types, it may be desired to constrain execution to not just a single core type.
+Many parallel workloads can execute efficiently on multiple core types that make up a subset of the available core types. For example:
+- A parallel algorithm with good scalability works well on both P-cores and E-cores
```
vossmjp:
```suggestion
- A parallel algorithm with good scalability or with a mix of requirements works well on both P-cores and E-cores
```

rfcs/proposed/core_types/README.md
```diff
@@ -0,0 +1,597 @@
...
+- Background processing can run on E-cores or LP E-cores depending on availability
+- Mixed workloads benefit from utilizing any available performance-class cores (P or E)
```
vossmjp:
```suggestion
```

rfcs/proposed/core_types/README.md
```diff
@@ -0,0 +1,597 @@
...
+```cpp
+auto core_types = tbb::info::core_types();
+// Assume: [0] = LP E-core, [1] = E-core, [2] = P-core
+
+tbb::task_arena arena(
+    tbb::task_arena::constraints{}.set_core_type(core_types[2])  // Only P-cores
+);
+```
+
+This forces applications to choose one of these suboptimal strategies:
```
vossmjp:
```suggestion
This may force an application to choose a suboptimal strategy that
uses a single core type when multiple core types would be better.
```

rfcs/proposed/core_types/README.md
```diff
@@ -0,0 +1,597 @@
...
+
+| Strategy | Pros | Cons |
+|----------|------|------|
+| **P-cores only** | Maximum single-threaded performance | Leaves E-cores idle; limited parallelism; higher power |
+| **E-cores only** | Good for parallel workloads | Doesn't utilize P-core performance; excludes LP E-cores |
+| **LP E-cores only** | Minimal power consumption | Severe performance impact for some workloads that require large, shared caches. |
+| **No constraint** | Maximum flexibility | May schedule on inappropriate cores (e.g., LP E-cores for compute) |
```
vossmjp:
```suggestion
```

rfcs/proposed/core_types/README.md
```diff
@@ -0,0 +1,597 @@
...
+
+None of these options provide the desired behavior: **"Use P-cores or E-cores, but avoid LP E-cores"** or **"Use any
+efficiency cores (E-core or LP E-core)"**.
```
vossmjp:
```suggestion
```

rfcs/proposed/core_types/README.md
```diff
@@ -0,0 +1,597 @@
...
+best-scored single type). For example, `{0, 1, 2}` would exclude LP E-cores on a newer runtime but impose no
+constraint on an older one.
+
+The new `default_concurrency` template returns the number of threads available for the given constraints and selector.
+This allows querying the effective concurrency without creating a task arena.
+
+### Implementation aspects
+
+The key implementation problem is how to pass the additional information about multiple core types
+to the TBB library functions for arena creation without violating the forward compatibility requirement for this specific feature.
```
vossmjp:
```suggestion
The key implementation problem is how to pass the additional information about multiple core types
to the TBB library functions for arena creation while meeting compatibility requirements.
```

rfcs/proposed/core_types/README.md
```diff
@@ -0,0 +1,597 @@
...
+constructor for `task_arena`, though we need to ensure that the internal state stores all the information needed
+to create another instance.
+
+### Testing Strategy
+
+Tests should cover:
+
+- **Correctness**: Verify that core type combinations are accurately stored and retrieved
+- **Backward compatibility**: Ensure existing single core type constraints work identically
+- **Forward compatibility**: Verify highest-scored core type fallback with older runtimes
```
vossmjp:
```suggestion
- **Forward compatibility**: Testing is constrained to the limited scenarios required by specific customers
```

rfcs/proposed/core_types/README.md
```diff
@@ -0,0 +1,597 @@
...
+- **Comprehensive combination testing**: Test all possible core type combinations on the target hardware
+
+#### Selector Testing
+
+Tests should verify that selectors receive correct core type information. When running against an older oneTBB runtime,
+we need to validate that the highest-scored core type is selected and that version threshold detection works correctly.
```
vossmjp:
```suggestion
Tests should verify that selectors receive correct core type information. When running against an older oneTBB
runtime, we need to validate that appropriate core type(s) are selected and that the version threshold detection works
correctly.
```

rfcs/proposed/core_types/README.md
```diff
@@ -0,0 +1,597 @@
...
+corresponding bit is set, returning them as a vector.
+
+**3. Affinity Mask Handling in TBBBind:**
+
+The system topology binding layer (TBBBind) would combine affinity masks for multiple core types by performing a
+logical OR operation across the hardware affinity masks of all specified core types. This combined mask would then be
+intersected with other constraint masks (NUMA node, threads-per-core) to produce the final thread affinity constraint,
+ensuring threads can be scheduled on any of the specified core types while respecting all other constraints.
+
+### Backward and Forward Compatibility
```
vossmjp:
```suggestion
### Compatibility
```

rfcs/proposed/core_types/README.md
```diff
@@ -0,0 +1,597 @@
...
+
+With the `constraints` API being header-only, the unmodified ABI, and no new library entry points, applications
+compiled with the proposed new functionality can handle execution against older oneTBB binaries through runtime
+detection and fallback mechanisms. Runtime detection is achieved using `TBB_runtime_interface_version()`, which allows
+applications to verify that the loaded oneTBB binary supports the new API before attempting to use it. When the runtime
+check indicates an older library version, applications can gracefully fall back to alternative strategies: either using
+all available core types (no constraint) or constraining to a single core type using the existing `set_core_type()`
+API. This approach satisfies the forward compatibility requirement stated in the "Compatibility Requirements" section.
```
vossmjp:
```suggestion
API. This approach satisfies forward compatibility for this feature in isolation, but does not ensure that oneTBB in general
supports forward compatibility. General forward compatibility is never guaranteed.
```

vossmjp (dismissed):
Looks good to me.

rfcs/proposed/core_types/README.md
```diff
@@ -0,0 +1,589 @@
...
+ways.
+
+Additionally, this API might have sensible semantics even if it has to operate with an older version
+of the runtime library. Such library can only use a single core type in the constraints, and the API
+implementation in the headers can adjust to that by using the core type with the highest score.
+If the API semantics is defined to allow such behavior, the API can do the oneTBB runtime version check
+internally (instead of users doing it in their code).
+
+One disadvantage of this approach is that neither single nor several instances of `constraints` represent
+the limitations of the created arena. It is still possible to create another arena with the same limitations,
```
aleksei-fedotov: Seems like this word is asking to be included:
```suggestion
the limitations of the created arena. Although, it is still possible to create another arena with the same limitations,
```

rfcs/proposed/core_types/README.md
```diff
@@ -0,0 +1,589 @@
...
+runtime, we need to validate that appropriate core type(s) are selected and that the version threshold detection works
+correctly.
+Tests should also cover selectors returning various score combinations, including positive, negative, zero, and
+all-negative scores.
+
+#### Core Type Combination Generation
+
+The test infrastructure could generate all possible core type combinations using a **power set approach**, producing
+2<sup>n</sup>-1 combinations for *n* core types by discovering available core types from the system, enumerating bit
+patterns from 1 to 2<sup>n</sup>-1, and mapping each pattern to a core type combination.
```
aleksei-fedotov, line 192: Why it is $2^n-1$ instead of $2^n$? I think empty set also needs to be used, to check the code works correctly if core type detection is completely broken. In this case, the vector of core types should include single `automatic` value, as far as I remember.

vossmjp (approved):
LGTM

rfcs/proposed/core_types/README.md
```diff
@@ -0,0 +1,589 @@
...
+
+tbb::task_arena throughput_driven(
+    tbb::task_arena::constraints{.core_type = core_types[1]},
+    tbb::task_arena::constraints{.core_type = core_types[2]}
+);
+
+tbb::task_arena background_work(
+    tbb::task_arena::constraints{.core_type = core_types[0]},
+    tbb::task_arena::constraints{.core_type = core_types[1]}
+);
```
aleksei-fedotov: Shouldn't these be within embracing curly braces to form a vector/array of constraints?

```suggestion
tbb::task_arena throughput_driven({
    tbb::task_arena::constraints{.core_type = core_types[1]},
    tbb::task_arena::constraints{.core_type = core_types[2]}
});

tbb::task_arena background_work(}
    tbb::task_arena::constraints{.core_type = core_types[0]},
    tbb::task_arena::constraints{.core_type = core_types[1]}
});
```

rfcs/proposed/core_types/README.md
```diff
@@ -0,0 +1,589 @@
...
+for (int index = core_types.size() - 2; index > 0; --index) {
+    avoid_LPE_cores.push_back(tbb::task_arena::constraints{.core_type = core_types[index]});
+}
+
+tbb::task_arena arena(avoid_LPE_cores);
+```
+
+#### Alternative 2
+
+The new API cannot be used with the old binaries which do not recognized encoded constraints.
```
aleksei-fedotov: A typo?
```suggestion
The new API cannot be used with the old binaries which do not recognize encoded constraints.
```

rfcs/proposed/core_types/README.md
aleksei-fedotov: Have not found that in the open questions. I believe, however, that the description of this exists in the _Compatibility_ section above (lines 378 - 388), where it covers users possibility to do the check on their own before using corresponding library API.

rfcs/proposed/core_types/README.md
```diff
@@ -0,0 +1,589 @@
...
+2<sup>n</sup>-1 combinations for *n* core types by discovering available core types from the system, enumerating bit
+patterns from 1 to 2<sup>n</sup>-1, and mapping each pattern to a core type combination.
+
+### Open questions
+
+1. What happens if all scores are negative?
+   - Current proposal: switch the parameter to `automatic`
+   - Alternative: return an error (exception)
+   - Alternative: leave the arena uninitialized
+2. Should the selector API support NUMA node selection in addition to core types?
```
akukanov, line 200: For the sake of discussion, currently I think that we should not extend selectors to the `numa_id` constraint.

While each core type at a given system has its own unique performance characteristics, and so ranking its applicability to a certain application task makes sense, NUMA nodes on a typical platform have equivalent capabilities and differ only by their IDs. Unless TBB provides some dynamic information about NUMA domains (for example, utilization), function-style selection would be inconvenient comparing to just taking a list of IDs.

rfcs/proposed/core_types/README.md
dnmokhov: You are right. The proposal above describes the additional constraints combinations to be added on top of the already existing `automatic` and single core type checks.

rfcs/proposed/core_types/README.md
dnmokhov: Added this justification for "current proposal: no".

rfcs/proposed/core_types/README.md
dnmokhov: That's correct. And the (now removed) open question referenced above, was w.r.t. a score of zero, which is now described in the "API description" section.

PR #1923, opened 2025-12-08
[RFC] Global Control Parameter for Worker Fast Leave Behavior

dnmokhov:
An RFC describing a new `global_control` parameter to toggle worker fast leave behavior.

Reference implementation: [dev/dnmokhov/fast-leave](https://github.com/uxlfoundation/oneTBB/tree/dev/dnmokhov/fast-leave)

### DIFF
--- a/rfcs/proposed/global_control_fast_leave/README.md
+++ b/rfcs/proposed/global_control_fast_leave/README.md
@@ -52,12 +52,15 @@ The proposal adds a new enumeration value to the existing `global_control::param
 #### Header
 
 ```cpp
+#define TBB_PREVIEW_PARALLEL_PHASE 1
 #include <oneapi/tbb/global_control.h>
 ```
 
 #### Syntax
 
 ```cpp
+#define TBB_HAS_PARALLEL_PHASE 202xxx
+
 namespace oneapi {
 namespace tbb {
 
@@ -226,19 +229,33 @@ int main() {
 }
 ```
 
-#### Scoped Fast Leave
+#### Global Control Scope, Initialization Order, and Disjunction
 
 ```cpp
-void process_interactive_workload() {
-    // Enable fast leave only during interactive phase
-    tbb::global_control gc(tbb::global_control::leave_policy, tbb::task_arena::leave_policy::fast);
+// No global_control is active yet; this call lazily initializes the implicit arena with automatic (the default).
+// Once initialized, the implicit arena's leave policy is fixed and not affected by later global_control changes.
+tbb::parallel_for(0, 1000000, [](int i) { do_parallel_work(i); });
 
-    while (user_is_interacting()) {
-        render_frame();  // Mix of serial and parallel work
+{
+    tbb::global_control gc1(tbb::global_control::leave_policy, tbb::task_arena::leave_policy::fast);
+    tbb::task_arena arena1;
+    // arena1 is lazily initialized on first use; since gc1 is active, it is initialized with fast leave.
+    arena1.execute([&] { /* ... */ });
+
+    // The implicit arena was already initialized with automatic above; gc1 does not affect it retroactively.
+    tbb::parallel_for(0, 1000000, [](int i) { do_parallel_work(i); });
+
+    {
+        tbb::global_control gc2(tbb::global_control::leave_policy, tbb::task_arena::leave_policy::automatic);
+        tbb::task_arena arena2;
+        // Both gc1 (fast) and gc2 (automatic) are active. Disjunction rule: any fast value present => fast wins.
+        arena2.execute([&] { /* ... */ });
     }
 }
-// New arenas initialized after gc goes out of scope use the default policy;
-// already-initialized arenas retain their initial state
+
+tbb::task_arena arena3;
+// Both gc1 and gc2 are now destroyed, so arena3 is initialized with automatic (the default).
+arena3.execute([&] { /* ... */ });
 ```
 
 #### Combining with Parallel Phase API
@@ -309,8 +326,7 @@ The default behavior could be changed to fast leave, making delayed leave opt-in
 - Simpler mental model
 
 **Cons:**
-- Breaking change for users who benefit from delayed leave
-- Will cause performance regression for other workloads
+- Fixes performance regression for some customers while causing it for others
 
 ## Open Questions
 
@@ -325,10 +341,6 @@ The default behavior could be changed to fast leave, making delayed leave opt-in
 
 3. **Composition**: Should the global control be a logical disjunction, first-registered wins, last-set wins, etc.?
    - Current proposal: disjunction
-
-4. **Feature Macro Dependency**: Should this feature require `TBB_PREVIEW_PARALLEL_PHASE` or have
-   its own independent macro?
-   - Current proposal: `TBB_PREVIEW_PARALLEL_PHASE` since it is closely related to the parallel phase API
 
 ## Exit Criteria
 


dnmokhov:
@tristanbarback,  this is an RFC for a new global control parameter to toggle worker fast leave behavior. Feel free to provide feedback. Thanks!

rfcs/proposed/global_control_fast_leave/README.md
```diff
@@ -0,0 +1,357 @@
...
+
+class global_control {
+public:
+    enum parameter {
+        max_allowed_parallelism,
+        thread_stack_size,
+        terminate_on_exception,
+        scheduler_handle,  // not a public parameter
+#if  TBB_PREVIEW_PARALLEL_PHASE
+        fast_leave,        // NEW: Controls worker fast leave behavior
```
isaevil: I suggest to not restrict this parameter to simply "Fast leave or not fast leave" and rename it to either `workers_leave_policy` or just `leave_policy` like in `task_arena`.

rfcs/proposed/global_control_fast_leave/README.md
```diff
@@ -0,0 +1,357 @@
...
+
+#### Semantics
+
+The `fast_leave` parameter would control whether worker threads leave arenas immediately after
+completing work:
+
+| Value | Behavior |
+|-------|----------|
+| `0` (default) | Workers follow the default system-specific policy (may spin before leaving) |
+| `1` | Workers leave immediately (fast leave enabled) |
```
isaevil: I think these values should be effectively the same values as those in`task_arena::leave_policy` enum. In that case, the invocation would look like:
```c++
tbb::global_control(tbb::global_control::workers_leave_policy, tbb::task_arena::leave_policy::fast_leave);
```
Perhaps moving `leave_policy` enum  out of arena could be done to shorten it to:
```c++
tbb::global_control(tbb::global_control::workers_leave_policy, tbb::leave_policy::fast_leave);
```

And I guess, the specifying, that the underlying type of enum is `size_t`, may be required.

rfcs/proposed/global_control_fast_leave/README.md
```diff
@@ -0,0 +1,357 @@
...
+### Proposed Implementation Strategy
+
+The proposed implementation would modify the `thread_leave_manager` to check the
+`global_control::fast_leave` parameter. When a worker thread queries whether it should remain in an
+arena, the method would first read the current state as usual. If the current state indicates
+delayed leave behavior and the global `fast_leave` control is set to enabled, the method would
+treat the state as if it were fast leave instead, causing the worker to exit immediately.
+
+```mermaid
+flowchart TD
```
isaevil, line 109: Will that work? It should straighten the lines and make it a bit more readable (in my opinion).
```suggestion
```mermaid
%%{init: {"layout": "elk"}}%%
flowchart TD
```

rfcs/proposed/global_control_fast_leave/README.md
```diff
@@ -0,0 +1,357 @@
...
+
+1. **Naming**: Should the parameter be named `fast_leave`, `worker_fast_leave`, or something else
+   like `disable_delayed_leave`?
+
+2. **Selection Rule**: Should the active value be the logical disjunction (as proposed) or
+   conjunction of all active controls? Disjunction means "any fast_leave enables it globally" while
+   conjunction means "all must agree."
+
+3. **Interaction Priority**: Should explicit `leave_policy::automatic` arenas respect the global
+   setting, or should only truly "default" arenas be affected?
```
isaevil: I think that the effect should be similar to what `max_allowed_parallelism` control does. Both wrt how arenas are affected and what is considered an active value. If combined with my comments above, the control would behave like this:
* Control allows setting global leave policy for workers and affects all arenas
* Last set leave policy is considered active.

rfcs/proposed/global_control_fast_leave/README.md
akukanov: I agree, the names should be aligned, and so should the values.

rfcs/proposed/global_control_fast_leave/README.md
akukanov: If I remember correctly, the arena leave policy API is still experimental. We can keep the names as-is for now, and decide later if we want to change anything.

rfcs/proposed/global_control_fast_leave/README.md
akukanov: I think the behavior could be much simpler than what is proposed - essentially, the API should just set the default value of `leave_policy` when an arena is created, essentially enforcing `automatic` to behave like `fast`. With the current implementation, that would result in the same effect without extra per-thread checks, lasting over the arena lifetime.

A possible problem with that approach though is limited control over the initialization order of global variables - i.e., some arenas or threads might get initialized before the global control takes effect. For explicit arenas it is mitigated by deferred initialization, while for implicit arenas the issue is arguably rather unlikely, as it means starting TBB parallel work during construction of global variables. We might as well consider going over all existing (or only over implicitly created) arenas and forcefully changing their leave policy.

akukanov: > Last set leave policy is considered active.

This selection rule is not reliable in case of multiple threads creating instances of `global_control`, effectively making the behavior unpredictable. Also, it's not the rule for `max_allowed_parallelism` which uses the minimum.

isaevil: > Also, it's not the rule for `max_allowed_parallelism` which uses the minimum.

You are right, seems like I overlooked that.

> the API should just set the default value of leave_policy when an arena is created, essentially enforcing automatic to behave like fast. With the current implementation, that would result in the same effect without extra per-thread checks and lasting over the arena lifetime.

Do I understand it correctly, that for this suggestion leave policy of the arena will stay overridden even after the control object destroys? I think it kind of aligns with current "static" behavior of `task_arena::leave_policy` (unless `parallel_phase` is used).

rfcs/proposed/global_control_fast_leave/README.md
```diff
@@ -0,0 +1,357 @@
...
+completing work:
+
+| Value | Behavior |
+|-------|----------|
+| `0` (default) | Workers follow the default system-specific policy (may spin before leaving) |
+| `1` | Workers leave immediately (fast leave enabled) |
+
+When multiple `global_control` objects exist for `fast_leave`, their logical disjunction would be
+used (consistent with the `terminate_on_exception` parameter). This means if any
+`global_control(fast_leave, 1)` is active, fast leave would be enabled globally.
```
akukanov: What if in the future there are at least two non-default values of `leave_policy`, and we want to allow enforcing any of these with `global_control`? Let's imagine there are both `fast` and `slow` policies, with `automatic` being in-between with regard to spin wait duration. In that case, what would be the selection rule for multiple instances?

Even though we can define strict ordering for the values and in theory use minimum or maximum as the rule, I see no way to choose between preferring fast leave to slow leave or vice versa. One could think that the new value should override previous ones but, as I noted in [another comment](https://github.com/uxlfoundation/oneTBB/pull/1923/files#r2598517126), it makes the behavior unpredictable.

It seems the rule should be "You snooze, you lose" :) - i.e. the first registered value remains active for its lifetime, while latecomers make no impact and maybe are not registered at all. The effective setting can then be tested with `active_value`. A question though is how to indicate which specific `global_control` instance really keeps control. For that, we can either throw an exception if the instance is not registered or - and that would be my preference - add a method to check if the instance was successfully registered.

rfcs/proposed/global_control_fast_leave/README.md
akukanov: > Do I understand it correctly, that for this suggestion leave policy of the arena will stay overridden even after the control object destroys?

Yes, exactly. That's similar to how `global_control::thread_stack_size` works.

vossmjp: I think we are assuming that the expected use case for this new value for global control is not to create changes that are limited to some phase of the application. That would still be the role of `start_parallel_phase`/`end_parallel_phase`, which also is arena specific. But if we do see that this could be useful during limited phases and globally, then changing only the default value at arena creation won't be sufficient. Imagine there are long-lived arenas, and some phase during which `fast_leave` is desired for all of these long-lived arenas. Without dynamic checks or iterating over arenas to change the values, developers would be forced to create new arenas or use the `start_parallel_phase`/`end_parallel_phase` for each arena. I think one of the motivations for this global control extension is to limit required code changes to get the fast leave behavior, so adding `start_parallel_phase`/`end_parallel_phase` might not be attractive. Again though, if we think the dominant or sole use case is just to override the default value once and for all, then changing the value at construction might be fine.

akukanov: >  But if we do see that this could be useful during limited phases and globally, then changing only the default value at arena creation won't be sufficient.

Sure, However I think such changes will be more likely needed per-arena rather than globally for all arenas - and yet we decided back then that we do not want to support that usage until it is clearly necessary. I would apply the same logic for the global setting.

> Again though, if we think the dominant or sole use case is just to override the default value once and for all, then changing the value at construction might be fine.

In my opinion, the for-lifetime arena leave policy is the dominant case in general, and most likely the sole case when applied to all the arenas.

rfcs/proposed/global_control_fast_leave/README.md
isaevil: Seems like GitHub doesn't support this because it is still rendered with default layout.

rfcs/proposed/global_control_fast_leave/README.md
```diff
@@ -0,0 +1,358 @@
...
+
+### Proposed Implementation Strategy
+
+The proposed implementation would modify the `thread_leave_manager` to check the
+`global_control::fast_leave` parameter. When a worker thread queries whether it should remain in an
+arena, the method would first read the current state as usual. If the current state indicates
+delayed leave behavior and the global `fast_leave` control is set to enabled, the method would
```
aleksei-fedotov: Since global setting should not change the explicit per arena setting, here it would be better to write something like:
_If the current state indicates the default behavior and the global control changes it, the method would treat the state as if it was specified per arena setting, causing the worker to behave correspondingly._

rfcs/proposed/global_control_fast_leave/README.md
aleksei-fedotov: I would name the global setting more accurately then. It feels like in this case the setting should include the words like "override default" or "override automatic" workers leave policy to indicate that creation of such a control will not affect workers leave policy setting for already initialized arenas.

rfcs/proposed/global_control_fast_leave/README.md
dnmokhov: Renamed to `leave_policy`.

rfcs/proposed/global_control_fast_leave/README.md
dnmokhov: Changed to be the same as in `task_arena::leave_policy`.

rfcs/proposed/global_control_fast_leave/README.md
dnmokhov: Changed.

rfcs/proposed/global_control_fast_leave/README.md
```diff
@@ -0,0 +1,339 @@
...
+
+        // Parallel work - workers leave immediately after completion
+        tbb::parallel_for(0, 1000000, [](int j) {
+            do_parallel_work(j);
+        });
+    }
+}
+```
+
+#### Scoped Fast Leave
```
akukanov: This example seems strange to me - there is no arena creation, so it is unclear what it illustrates. Perhaps it was relevant for the initial proposal with dynamic control.

rfcs/proposed/global_control_fast_leave/README.md
```diff
@@ -0,0 +1,339 @@
...
+
+The default behavior could be changed to fast leave, making delayed leave opt-in.
+
+**Pros:**
+- Addresses the regression for all affected users automatically
+- Simpler mental model
+
+**Cons:**
+- Breaking change for users who benefit from delayed leave
+- Will cause performance regression for other workloads
```
akukanov: I would not call it a "breaking change" but rather "negative impact".
And actually these two bullets tell about the same problem - we would automatically fix performance regression for some customers while causing performance regression for others.

rfcs/proposed/global_control_fast_leave/README.md
```diff
@@ -0,0 +1,339 @@
...
+   - Current proposal: only affect arenas initialized after the `global_control` is set
+   - Alternative: allow retroactive effect on existing arenas
+   - Alternative: allow per worker control (e.g., via observers) to enable/disable fast leave on a per-thread basis
+
+3. **Composition**: Should the global control be a logical disjunction, first-registered wins, last-set wins, etc.?
+   - Current proposal: disjunction
+
+4. **Feature Macro Dependency**: Should this feature require `TBB_PREVIEW_PARALLEL_PHASE` or have
+   its own independent macro?
+   - Current proposal: `TBB_PREVIEW_PARALLEL_PHASE` since it is closely related to the parallel phase API
```
akukanov: Is it really an open question? If nobody has proposed a separate macro or objected to the current proposal, let's just go with that.

rfcs/proposed/global_control_fast_leave/README.md
```diff
@@ -0,0 +1,339 @@
...
+
+The proposal adds a new enumeration value to the existing `global_control::parameter` enum:
+
+#### Header
+
+```cpp
+#include <oneapi/tbb/global_control.h>
+```
+
+#### Syntax
```
vossmjp, line 59: Seems like we should introduce a feature-test macro. `TBB_HAS_GC_LEAVE_POLICY`.

vossmjp: Or maybe `TBB_HAS_PARALLEL_PHASE`?

rfcs/proposed/global_control_fast_leave/README.md
```diff
@@ -0,0 +1,339 @@
...
+        });
+    }
+}
+```
+
+#### Scoped Fast Leave
+
+```cpp
+void process_interactive_workload() {
+    // Enable fast leave only during interactive phase
```
vossmjp: Unless there is another GC already active that has won the race to set the value.

vossmjp: Maybe it would be a good idea to show a nested example that has (1) an outer GC, (2) and inner GC that is ignored and (3) a parallel_phase that overrides the outermost GC.

rfcs/proposed/global_control_fast_leave/README.md
dnmokhov: added `TBB_HAS_PARALLEL_PHASE`

rfcs/proposed/global_control_fast_leave/README.md
dnmokhov: updated

rfcs/proposed/global_control_fast_leave/README.md
dnmokhov: removed

rfcs/proposed/global_control_fast_leave/README.md
dnmokhov: rewritten

rfcs/proposed/global_control_fast_leave/README.md
dnmokhov: rewritten

PR #1926, opened 2025-12-09
An alternative proposal for "Advanced Core Type Selection"

akukanov:
This PR attempts to complement the RFC #1917 with an alternative API and design approach.

### DIFF
--- a/rfcs/proposed/core_types/README.md
+++ b/rfcs/proposed/core_types/README.md
@@ -446,6 +446,8 @@ additional exploration.
 4. For a hypothetical platform having both a few NUMA nodes and different core types, would we allow
    to select both at the same time, and if yes - how would that work?
 5. Which other usability names would be useful? For example, a named negative score constant?
+6. How should `max_concurrency` interact with scoring? Users might think that scoring also indicates scheduling
+   preference, for example if `max_concurrency` is less than the HW concurrency for selected core types.
 
 ## Usage Examples
 


rfcs/proposed/core_types/README.md
```diff
@@ -70,6 +70,8 @@ This proposal must maintain compatibility with previous oneTBB library versions:
 - **Binary Compatibility (ABI)**: The `task_arena::constraints` struct layout must remain unchanged.
 - **Forward Compatibility (New Application + Old Library)**: Applications compiled with the proposed new functionality
   must be able to handle execution against older oneTBB binaries gracefully, without crashes or undefined behavior.
+  This requirement is mandated by the necessity for the key customer, OpenVINO, to support their users
+  that might not be ready to upgrade to the latest version of oneTBB.
```
vossmjp, line 74: In general oneTBB does not promise forward compatibility, so what does it mean for one specific feature to guarantee forward compatibility. Alone, this wouldn't guarantee compatibility.

akukanov: Correct, but nevertheless this is a design requirement we seemingly adhere to, for this very feature. I simply added the explanation why we have that requirement.

rfcs/proposed/core_types/README.md
```diff
@@ -318,34 +300,370 @@ is added, a corresponding `set_numa_ids` function would likely follow. The choic
...
+
+In principle, it could be possible for the implementation to support both variations, distinguished at compile time.
+For each ID value, the selection function should return a signed integral number as the score for that ID. In case (1),
+the function should therefore return a vector of scores, one for each element of its input vector.
+
+A negative score would indicate that the corresponding core type or NUMA node should be excluded from use
+by the task arena. A positive score would indicate that the corresponding core type or NUMA node is good to use
+by the task arena, and the bigger the score the better the "resource" is from the user's viewpoint.
+If for whatever reason (some described below) the implementation can only use a single core type or NUMA node,
+it should take the one with the biggest score.
```
vossmjp, line 385: How should `max_concurrency` interact with scoring? For example, assume there is a system with three core types, indexes 0, 1 and 2 and there are 4 of each kind. What if an arena is created with 6 slots, and the cores are scored by the index. Should the arena have six slots and a preference with first filling these with core types 2, followed by core type 1?  Or, does the arena ignore the rankings (except for -1) and have 6 slots but a full cpu mask? I doubt we will want to limit what can fill the slots, but perhaps the ability to score will lead users to think there should be a preference if max_concurrency is limited and yet all core types are idle.

rfcs/proposed/core_types/README.md
```diff
@@ -318,34 +300,371 @@ is added, a corresponding `set_numa_ids` function would likely follow. The choic
...
+
+### Open questions
+
+1. Need to decide on the parameters and the return type of selectors
+2. What happens if all scores are negative? Options: return an error (exception), switch the parameter
+   to `automatic`, or leave the arena uninitialized.
+   - In SYCL, class constructors with a device selector must throw an exception if no device is selected.
+3. If selectors take a vector, what happens if the returned vector is smaller or greater than the input one?
+4. For a hypothetical platform having both a few NUMA nodes and different core types, would we allow
+   to select both at the same time, and if yes - how would that work?
```
dnmokhov, line 447: Maybe add another template:

```cpp
template <typename Selector1, typename Selector2>
task_arena(constraints a_constraints, Selector1 a_selector1, Selector2 a_selector2,
           unsigned reserved_slots = 1, priority a_priority = priority::normal);
```

with the following usage logic:

| `a_constraints.numa_id` value | `a_constraints.core_type` value | Selector Usage |
|----------|------|------|
| `selectable` | other | One `a_selector` for `numa_id` |
| other | `selectable` | One `a_selector` for `core_type ` |
| `selectable` | `selectable` | `a_selector1` for `numa_id` and `a_selector2` for `core_type` |

(The first two rows are using the same template.)

rfcs/proposed/core_types/README.md
```diff
@@ -318,34 +300,371 @@ is added, a corresponding `set_numa_ids` function would likely follow. The choic
...
+ // avoid the lest performant core type
+for (int i = core_types.size() - 2; i > 0; --i) {
+    avoid_LPE_cores.push_back(tbb::task_arena::constraints{.core_type = core_types[i]});
+}
+
+tbb::task_arena arena(avoid_LPE_cores);
+```
+
+Instead of the c
+
```
dnmokhov: ?

rfcs/proposed/core_types/README.md
akukanov: Yes, that's a possible option. I just think that maybe we can keep this question open and address at some later time.

Updated: let's keep the conversation unresolved for the details to be visible.

rfcs/proposed/core_types/README.md
```diff
@@ -318,34 +300,371 @@ is added, a corresponding `set_numa_ids` function would likely follow. The choic
...
+1. the whole vector of core types or NUMA nodes, as returned by `tbb::info` API calls, or
+2. a single `core_type_id` or `noma_node_id` value, packed into a tuple/struct with additional useful information,
+   specifically the number of entities to choose from (e.g., `tbb::info::core_types().size()`) and the position
+   (index) of the given ID value.
+
+In principle, it could be possible for the implementation to support both variations, distinguished at compile time.
+For each ID value, the selection function should return a signed integral number as the score for that ID. In case (1),
+the function should therefore return a vector of scores, one for each element of its input vector.
+
+A negative score would indicate that the corresponding core type or NUMA node should be excluded from use
```
aleksei-fedotov, line 381: Perhaps, for better readability the negative score could also have their own name such as `skip_resource`, for example.

rfcs/proposed/core_types/README.md
```diff
@@ -318,34 +300,371 @@ is added, a corresponding `set_numa_ids` function would likely follow. The choic
...
+The runtime version check can be hidden within the implementation, if we agree that only the first constraint
+applies in case multiple constraints cannot work.
+
+```cpp
+auto core_types = tbb::info::core_types();
+std::vector<tbb::task_arena::constraints> avoid_LPE_cores;
+
+// always use the most performant core type
+avoid_LPE_cores.push_back(tbb::task_arena::constraints{.core_type = core_types.back()});
+ // avoid the lest performant core type
```
aleksei-fedotov: A typo:
```suggestion
// avoid the least performant core type
```

rfcs/proposed/core_types/README.md
```diff
@@ -318,34 +300,371 @@ is added, a corresponding `set_numa_ids` function would likely follow. The choic
...
+In case selectors take the vector of core types, and without assuming there are exactly three types,
+the selector needs to deal with input of different sizes. It could be like this:
+
+```cpp
+tbb::task_arena arena(
+    tbb::task_arena::constraints{.core_type = tbb::task_arena::selectable},
+    [](auto /*std::vector*/ core_types) -> std::vector<int> {
+        std::size_t n_types = core_types.size();
+        std::vector<int> scores(n_types);
+        scores[0] = -1; // avoid the lest performant core type
```
aleksei-fedotov: A typo:
```suggestion
        scores[0] = -1; // avoid the least performant core type
```

rfcs/proposed/core_types/README.md
```diff
@@ -318,34 +300,371 @@ is added, a corresponding `set_numa_ids` function would likely follow. The choic
...
+by reusing the same set of arguments including the selector. It is also possible to have a proper copy
+constructor for `task_arena`, though we need to ensure that the internal state stores all the information needed
+to create another instance. But other APIs, such as the `tbb::info::default_concurrency` function that takes
+a `constraints` argument, currently do not have access to the same information. A possible way to address
+that is to add a parameter for either the selector or the vector(s) of scores it created; however that needs
+additional exploration.
+
+### Open questions
+
+1. Need to decide on the parameters and the return type of selectors
```
aleksei-fedotov, line 441: Arguments that I see in favor of returning a container of `constraints`:
1. Additional `selectable` constant is not needed as the constructor of `task_arena` would just accept `Selector` instead of constraints (not in addition to them).
2. User does not need to specify the rank for each core type, but provide only those constraints that are required.
3. Returning an empty vector of constraints naturally indicates that there is no constraints to the task arena to apply.
4. If needed, the position of constraints within this vector could indicate the preference of one constraints over the other. Although, the order within the vector is not that explicit as the ranking. Could a map of `{ rank : constraints }` instead of a vector help?
5. Container of constraints allows selecting various platform parts individually, and uniting them.
6. The `constraints` entity is an already established and explicit approach for selecting them. Whose generalization would allow not only selecting core types, but also NUMA nodes, and possibly other topology objects if those are added there later. (Is this the right time to change the layout of constraints and introduce versioning in this struct?)

Also, the more I think about the alternative with selectors the stronger I am inclined to thinking that the most (or the only?) differentiable thing of this alternative is the future potential to call that `selector` function multiple times during the `task_arena` lifetime, hence changing the constraints dynamically. However, we don't anticipate such a necessity ever happens.

rfcs/proposed/core_types/README.md
akukanov: Added as an open question.

rfcs/proposed/core_types/README.md
akukanov: Note that I did not propose to return a container of constraints, but a container of scores. Essentially, here you propose to sort of combine the two described alternatives, which is an option, but it requires more analysis. For now, I will add a short note to the description of the Alternative 1.

As for differentiating: I believe we are primarily discussing API convenience here and exposure of implementation details, not some unique additional capabilities.

rfcs/proposed/core_types/README.md
akukanov: An editing artifact, removed.

rfcs/proposed/core_types/README.md
akukanov: In principle, we can say that on platforms supporting both "hard" and "soft" thread affinity (like Windows), the types with positive scores define the hard affinity and the type with the highest score defines the soft affinity for arena threads.

I will add another open question for this.

vossmjp: Do you think we would make use of the "soft" affinity in the initial implementation?

akukanov: I think we could check if it provides any performance difference and so if it is worth doing.

dnmokhov (approved):
LGTM

PR #1928, opened 2025-12-12
Fix Task Dynamic Dependencies when RTTI is disabled

### LINKED ISSUE #1912: Cannot compile OneTBB 2022.3.0 without RTTI
Hi,

Due to this line:
https://github.com/uxlfoundation/oneTBB/blob/0cd32ab10a84eabf780bb699b17430deb028c0a4/include/oneapi/tbb/task_group.h#L672

OneTBB can no longer be compiled without RTTI enabled. This wasn't an issue in OneTBB 2022.2.0, but it is with 2022.3.0. Since `__TBB_PREVIEW_TASK_GROUP_EXTENSIONS` is enabled by `__TBB_BUILD`, there is no way to "opt-out" of this feature for now.

kboyarinov:
### Description 
Replace dynamic_cast with a static_cast to allow TBB build and Task Dynamic Dependencies usage when RTTI is disabled

Fixes #1912

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown


include/oneapi/tbb/detail/_task_handle.h
```diff
@@ -172,73 +216,41 @@ class task_handle_task : public d1::task {
...
-
-            if (m_dynamic_state.compare_exchange_strong(current_state, new_state)) {
-                current_state = new_state;
-            } else {
-                // CAS failed, current_state points to the dynamic state created by another thread
-                alloc.delete_object(new_state);
-            }
-        }
+inline task_dynamic_state* dynamic_state_task::get_dynamic_state() {
+#if __TBB_USE_OPTIONAL_RTTI && TBB_USE_DEBUG
```
vossmjp: So why the extra TBB_USE_DEBUG here? If not set, TBB_USE_ASSERT is set to TBB_USE_DEBUG. Why not just check __TBB_USE_OPTIONAL_RTTI? Is it to eliminate overheads?

include/oneapi/tbb/task_group.h
```diff
@@ -668,12 +674,12 @@ class task_group : public task_group_base {
...
-        // Not using __TBB_ASSERT(curr_th_task) to allow function_stack_task body to use this method
-        if (curr_th_task != nullptr) {
-            curr_th_task->transfer_completion_to(new_task);
-        }
+        __TBB_ASSERT(curr_task != nullptr, "transfer_this_task_completion_to was called outside of task body");
+#if __TBB_USE_OPTIONAL_RTTI && TBB_USE_DEBUG
+        __TBB_ASSERT(dynamic_cast<dynamic_state_task*>(curr_task) != nullptr,
+                     "transfer_this_task_completion_to was called from a task outside a task_group");
+#endif
+        static_cast<dynamic_state_task*>(curr_task)->transfer_completion_to(new_task);
```
vossmjp, line 682: So this not only removes dependence on RTTI, but also allows transfers from within a function_task_task, is that right? Before, that would have not have been possible.

include/oneapi/tbb/detail/_task_handle.h
```diff
@@ -134,9 +134,56 @@ class task_dynamic_state {
     std::atomic<std::size_t> m_num_references;
     d1::small_object_allocator m_allocator;
 };
-#endif // __TBB_PREVIEW_TASK_GROUP_EXTENSIONS
 
-class task_handle_task : public d1::task {
+class dynamic_state_task : public d1::task {
```
vossmjp, line 138: You'll need to update the implementation [RFC](https://github.com/uxlfoundation/oneTBB/blob/master/rfcs/proposed/task_group_dynamic_dependencies/implementation_details.md). It also needs to be taken out of proposed and put into experimental.

include/oneapi/tbb/detail/_task_handle.h
kboyarinov: It is not actually needed. Removed

include/oneapi/tbb/task_group.h
kboyarinov: No, transferring worked for `function_stack_task`s as well. Since it is not derived from `task_handle_task`, the old `dynamic_cast` returned `nullptr` and transferring was just no-op in this case. 

In the new implementation, `function_stack_task` is derived from `dynamic_state_task`, but it's dynamic state is guaranteed to be uninitialized. So transferring is no-op as well, but after one more `atomic` read.

include/oneapi/tbb/detail/_task_handle.h
kboyarinov: I have created an issue #1980 to do it separately

PR #1932, opened 2025-12-17
Build: do not include windows.h from header files

### LINKED ISSUE #573: Remove <Windows.h> include from headers

aras-p:
### Description 

Including `<windows.h>` is both 1) including a lot of code, and 2) leaving a lot of "damage" since it defines many things as preprocessor macros (e.g. `far`, `near`, `GetObject` etc. can't be used as identifiers).

In two cases that `<windows.h>` was used in TBB headers, the code needed only a tiny amount of functions, so just manually declare their prototypes:

- `_machine.h`: just needs `SwitchToThread()`,
- `enumerable_thread_specific.h`: just needs four `Tls` functions (this achieves same result as #576, but without any potential performance impacts due to extra function call).

The exact API shape of WinAPI functions (`__declspec(dllimport)`, `__stdcall`) and their argument types realistically can't ever change, or it would break all existing software. So it should be fine to just declare them manually like this.

Fixes #573

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/include/oneapi/tbb/detail/_machine.h
+++ b/include/oneapi/tbb/detail/_machine.h
@@ -1,6 +1,7 @@
 /*
     Copyright (c) 2005-2024 Intel Corporation
     Copyright (c) 2026 UXL Foundation Contributors
+
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
     You may obtain a copy of the License at


include/oneapi/tbb/enumerable_thread_specific.h
```diff
@@ -290,16 +294,16 @@ class ets_base<ets_key_per_instance>: public ets_base<ets_no_key> {
     using super = ets_base<ets_no_key>;
 #if _WIN32||_WIN64
 #if __TBB_WIN8UI_SUPPORT
-    using tls_key_t = DWORD;
+    using tls_key_t = unsigned long;
     void create_key() { my_key = FlsAlloc(nullptr); }
     void destroy_key() { FlsFree(my_key); }
     void set_tls(void * value) { FlsSetValue(my_key, (LPVOID)value); }
```
dnmokhov:
```suggestion
    void set_tls(void * value) { FlsSetValue(my_key, value); }
```

Alexandr-Konovalov:
Personally, I don't like the approach "let's add all we need manually and forget about windows.h". Can we explicitly write down other possible solutions and compare them?

For example, we can re-export required API from tbb.dll at the price of an additional function call. What else can be done?

aras-p:
> Personally, I don't like the approach "let's add all we need manually and forget about windows.h".

Yeah, but IMHO besides "uhh why we have to do this" aspect, this approach does not have many (any?) other downsides.

> Can we explicitly write down other possible solutions and compare them?
> For example, we can re-export required API from tbb.dll at the price of an additional function call. What else can be done?

Alternative could be to indeed have these functions not be called inline, but rather through a "wrapper"; and the `<windows.h>` include then could be only in some C++ source file. Downside of that approach: extra function call. So that would be very similar to #576 indeed then. I don't know how to quantify / measure the extra function call overhead.

Alexandr-Konovalov:
> Yeah, but IMHO besides "uhh why we have to do this" aspect, this approach does not have many (any?) other downsides.

I agree.

include/oneapi/tbb/detail/_machine.h
```diff
@@ -28,11 +28,8 @@
 #ifdef _WIN32
 #include <intrin.h>
 #ifdef __TBBMALLOC_BUILD
-#define WIN32_LEAN_AND_MEAN
-#ifndef NOMINMAX
-#define NOMINMAX
-#endif
-#include <windows.h> // SwitchToThread()
+// we only need this function; do not include whole of windows.h
+extern "C" __declspec(dllimport) int __stdcall SwitchToThread(void);
```
Alexandr-Konovalov, line 33: Can standard `std::this_thread::yield();` be used instead? Probably someone can remember that.

aleksei-fedotov: I found a comment in the same _machine.h file, indicating that `SwitchToThread` is necessary to avoid dependency on C++ runtime for tbbmalloc builds. I have not found/recalled other circumstances.

kboyarinov:
> As a reminder for when we merge this, this should be viewed as a source code breaking change. Users' code can break if they inadvertently have forgotten to include windows.h themselves and get it through TBB. So we need to note this as a potential source breaking change and ship it as part of an appropriate release.

Also, combination of `windows.h` and the `NOMINMAX` macro may break. Consider the following user-side code:
```cpp
#include <tbb/enumerable_thread_specific.h>
#include <windows.h>
```
With the current codebase, windows.h is included indirectly when including ets.h, with `NOMINMAX` defined. 
With the proposed change, `windows.h` will only be included explicitly after ets.h, and `NOMINMAX` will not be defined.
This could also be noted as a potential source break.

include/oneapi/tbb/detail/_machine.h
```diff
@@ -1,5 +1,5 @@
 /*
-    Copyright (c) 2005-2024 Intel Corporation
+    Copyright (c) 2005-2025 Intel Corporation
```
kboyarinov:
```suggestion
    Copyright (c) 2005-2024 Intel Corporation
    Copyright (c) 2026 UXL Foundation Contributors
```

include/oneapi/tbb/enumerable_thread_specific.h
```diff
@@ -1,5 +1,5 @@
 /*
-    Copyright (c) 2005-2024 Intel Corporation
+    Copyright (c) 2005-2025 Intel Corporation
```
kboyarinov:
```suggestion
    Copyright (c) 2005-2024 Intel Corporation
    Copyright (c) 2026 UXL Foundation Contributors
```

kboyarinov:
@aras-p, could you please apply my copyright suggestions above.
That is the only thing currently preventing this from being merged.
Thank you!

aras-p:
@kboyarinov ah right, forgot about this PR. Suggestions applied now!

Issue #1741, opened 2025-05-27
Does not build on clang on windows due to flag `-fPIC` not being suppored


solomoncyj:
# Summary
Does not build on clang on windows
# Version
commit 051fd44c11ce781de9b737c025367bf76c1f29bb
# Environment
Provide any environmental details that you consider significant for reproducing the issue.
The following information is important:
OS; win 11 24H2
clang version 20.1.2
Target: x86_64-pc-windows-msvc
Thread model: posix
InstalledDir: C:\Program Files\LLVM\bin

# Observed Behavior
build failed with:
```cmd
 cmake -S . -B build -G Ninja -DCMAKE_BUILD_TYPE=Release 
-- The CXX compiler identification is Clang 20.1.2 with GNU-like command-line
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: C:/Program Files/LLVM/bin/clang++.exe - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMake Warning at CMakeLists.txt:141 (message):
  Dependency signature verification during dynamic linking in run-time is
  disabled by default.  This may lead to security vulnerabilities.  See
  installation section in oneTBB documentation for more details.


-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Failed
-- Check if compiler accepts -pthread
-- Check if compiler accepts -pthread - no
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - not found
-- Found Threads: TRUE
-- Checking for one of the modules 'hwloc'
-- IPO enabled
-- The C compiler identification is Clang 20.1.2 with GNU-like command-line
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: C:/Program Files/LLVM/bin/clang.exe - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- HWLOC target HWLOC::hwloc_static doesn't exist. The tbbbind target cannot be created
-- HWLOC target HWLOC::hwloc_2 doesn't exist. The tbbbind_2_0 target cannot be created
-- HWLOC target HWLOC::hwloc_2_5 doesn't exist. The tbbbind_2_5 target cannot be created
-- Configuring done (7.5s)
[...]

PR #1936, opened 2025-12-25
Fix Windows Clang compiler flags in CMake configuration

justinzhuguangwen:
### Description 

This PR fixes Windows build compatibility issues when using Clang as the compiler on Windows (with the MSVC toolchain). The build system was previously attempting to use Linux-specific compiler and linker flags that are not supported on Windows, causing build failures or warnings. This change adds conditional logic to use platform-appropriate flags based on the target platform.

**Context:**
When Clang is used on Windows (not MSVC compiler, but Clang with MSVC toolchain), it has different capabilities and limitations compared to Clang on Linux. The previous code assumed a Linux-like environment, which caused issues on Windows.

**Key Changes:**

1. **DEF File Format Handling (lines 41-48):**
   - On Windows, use Windows `.def` file format via `CMAKE_LINK_DEF_FILE_FLAG` instead of Linux version-script format
   - Use `win${TBB_ARCH}` prefix for Windows DEF files instead of `lin${TBB_ARCH}`
   - Maintains backward compatibility for non-Windows platforms

2. **Compile Flags (lines 72-78):**
   - Excludes `-fPIC` on Windows (not supported by Clang with MSVC toolchain)
   - Excludes `-fstack-clash-protection` and `-fcf-protection` on Windows (not supported)
   - Adds `-D_CRT_SECURE_NO_WARNINGS` to suppress CRT security warnings for deprecated functions (e.g., `getenv`, `strncpy`) that TBB uses
   - Retains `-fstack-protector-strong` and format security checks on Windows

3. **Library Compile Flags (line 80):**
   - Explicitly excludes Windows from `-fstack-clash-protection` and `-fcf-protection=full` flags
   - These flags are only applied on non-Apple, non-Android, x86/x64 platforms that are not Windows

4. **Linker Flags (lines 84-88):**
   - Excludes Windows from `-z` linker options (relro, now, noexecstack)
   - These options are Linux-specific and not supported by `lld-link` on Windows
   - Updated comments to clarify that `-z` switch is not supported on both macOS and Windows

**Impact:**
- ✅ Windows builds with Clang (MSVC toolchain) now succeed without unsupported flag errors
- ✅ Proper Windows `.def` file format is used instead of Linux version-script
- ✅ CRT security warnings are suppressed for deprecated functions that TBB uses
- ✅ No impact on Linux, macOS, or other platforms (backward compatible)

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### Other information

**Testing:**
- ✅ Verified that the changes maintain existing behavior on Linux and macOS
- ✅ Tested that Windows builds with Clang (MSVC toolchain) now succeed without unsupported flag errors
- ✅ Confirmed that Windows-specific DEF file format is correctly used
- ✅ Verified that all conditional branches work correctly for their respective platforms

**Technical Details:**
- The changes use `WIN32` CMake variable to detect Windows platform (works for both native Windows and cross-compilation)
- Windows-specific flags are only applied when `WIN32` is true, ensuring other platforms are unaffected
- The `-z` linker options are Linux ELF-specific and not supported by `lld-link` on Windows
- `-fPIC` is not needed on Windows as DLLs use a different linking model

### DIFF
--- a/cmake/compilers/Clang.cmake
+++ b/cmake/compilers/Clang.cmake
@@ -38,8 +38,14 @@ elseif (MSVC)
     include(${CMAKE_CURRENT_LIST_DIR}/MSVC.cmake)
     return()
 else()
-    set(TBB_LINK_DEF_FILE_FLAG -Wl,--version-script=)
-    set(TBB_DEF_FILE_PREFIX lin${TBB_ARCH})
+    # On Windows, use Windows .def file format, not Linux version-script
+    if(WIN32)
+        set(TBB_LINK_DEF_FILE_FLAG ${CMAKE_LINK_DEF_FILE_FLAG})
+        set(TBB_DEF_FILE_PREFIX win${TBB_ARCH})
+    else()
+        set(TBB_LINK_DEF_FILE_FLAG -Wl,--version-script=)
+        set(TBB_DEF_FILE_PREFIX lin${TBB_ARCH})
+    endif()
     set(TBB_TEST_COMPILE_FLAGS ${TBB_TEST_COMPILE_FLAGS} $<$<NOT:$<VERSION_LESS:${CMAKE_CXX_COMPILER_VERSION},10.0>>:-ffp-model=precise>)
 endif()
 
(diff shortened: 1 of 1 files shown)


cmake/compilers/Clang.cmake
```diff
@@ -38,8 +38,14 @@ elseif (MSVC)
     include(${CMAKE_CURRENT_LIST_DIR}/MSVC.cmake)
     return()
 else()
-    set(TBB_LINK_DEF_FILE_FLAG -Wl,--version-script=)
-    set(TBB_DEF_FILE_PREFIX lin${TBB_ARCH})
+    # On Windows, use Windows .def file format, not Linux version-script
+    if(WIN32)
+        set(TBB_LINK_DEF_FILE_FLAG ${CMAKE_LINK_DEF_FILE_FLAG})
+        set(TBB_DEF_FILE_PREFIX win${TBB_ARCH})
```
dnmokhov, line 44: Wouldn't the `elseif (MSVC)` branch (lines 38-39) be taken when using Clang with MSVC toolchain?

justinzhuguangwen: Good question. This PR targets **native Clang** (clang++) on Windows, not clang-cl.

With clang-cl, `MSVC` is `true` and the existing MSVC branch handles it. But with native clang++ on Windows, `MSVC` is `false` while `WIN32` is `true` — that's the gap this PR fills.

A clarifying comment in the code might help avoid this confusion.

omalyshe:
Related Issue https://github.com/uxlfoundation/oneTBB/issues/1741

cmake/compilers/Clang.cmake
```diff
@@ -63,14 +69,21 @@ if (CMAKE_SYSTEM_PROCESSOR MATCHES "(AMD64|amd64|i.86|x86)" AND NOT EMSCRIPTEN)
...
 
-if (NOT APPLE AND NOT ANDROID_PLATFORM AND CMAKE_SYSTEM_PROCESSOR MATCHES "(AMD64|amd64|i.86|x86)")
+if (NOT APPLE AND NOT ANDROID_PLATFORM AND CMAKE_SYSTEM_PROCESSOR MATCHES "(AMD64|amd64|i.86|x86)" AND NOT WIN32)
     set(TBB_LIB_COMPILE_FLAGS ${TBB_LIB_COMPILE_FLAGS} -fstack-clash-protection $<$<NOT:$<BOOL:${EMSCRIPTEN}>>:-fcf-protection=full>)
 endif()
 
-# -z switch is not supported on MacOS
-if (NOT APPLE)
+# -z switch is not supported on MacOS and Windows
+# On Windows, lld-link doesn't support -z options (Linux-specific)
```
isaevil, line 85: Seems redundant since line just above already states that
```suggestion
```

cmake/compilers/Clang.cmake
```diff
@@ -63,14 +69,21 @@ if (CMAKE_SYSTEM_PROCESSOR MATCHES "(AMD64|amd64|i.86|x86)" AND NOT EMSCRIPTEN)
...
 
 # Clang flags to prevent compiler from optimizing out security checks
-set(TBB_COMMON_COMPILE_FLAGS ${TBB_COMMON_COMPILE_FLAGS} -Wformat -Wformat-security -Werror=format-security -fPIC $<$<NOT:$<BOOL:${EMSCRIPTEN}>>:-fstack-protector-strong>)
+# Don't use -fPIC, -fstack-clash-protection, -fcf-protection on Windows (not supported by Clang with MSVC toolchain)
+# Also disable CRT security warnings on Windows (getenv, strncpy, etc. are deprecated but TBB uses them)
+if(WIN32)
+    set(TBB_COMMON_COMPILE_FLAGS ${TBB_COMMON_COMPILE_FLAGS} -Wformat -Wformat-security -Werror=format-security -D_CRT_SECURE_NO_WARNINGS $<$<NOT:$<BOOL:${EMSCRIPTEN}>>:-fstack-protector-strong>)
+else()
+    set(TBB_COMMON_COMPILE_FLAGS ${TBB_COMMON_COMPILE_FLAGS} -Wformat -Wformat-security -Werror=format-security -fPIC $<$<NOT:$<BOOL:${EMSCRIPTEN}>>:-fstack-protector-strong>)
+endif()
```
isaevil, line 78: Consider using CMake's generator expressions for setting (or not setting) compiler flags just like it is done for EMSCRIPTEN

omalyshe:
@justinzhuguangwen Sorry, closed PR by mistake. Now as re-opened, could you please apply suggestions and address comments that we could proceed with the fix?

Issue #1907, opened 2025-11-07
2022.3.0: Malloc tests fail


Lastique:
# Summary

Running tests after building oneTBB 2022.3.0 fail:

```
136/140 Test #136: test_malloc_whitebox .....................   Passed    1.19 sec
test 137
        Start 137: test_malloc_atexit

137: Test command: /build/tbb/tbb/obj-x86_64-linux-gnu/gnu_13.3_cxx20_64_release/test_malloc_atexit "--force-colors=1"
137: Working Directory: /build/tbb/tbb/obj-x86_64-linux-gnu/gnu_13.3_cxx20_64_release
137: Test timeout computed to be: 10000000
137/140 Test #137: test_malloc_atexit .......................***Exception: SegFault  0.09 sec

test 138
        Start 138: test_malloc_overload

138: Test command: /build/tbb/tbb/obj-x86_64-linux-gnu/gnu_13.3_cxx20_64_release/test_malloc_overload "--force-colors=1"
138: Working Directory: /build/tbb/tbb/obj-x86_64-linux-gnu/gnu_13.3_cxx20_64_release
138: Test timeout computed to be: 10000000
138/140 Test #138: test_malloc_overload .....................***Exception: SegFault  0.07 sec

test 139
        Start 139: test_malloc_overload_disable

139: Test command: /build/tbb/tbb/obj-x86_64-linux-gnu/gnu_13.3_cxx20_64_release/test_malloc_overload_disable "--force-colors=1"
139: Working Directory: /build/tbb/tbb/obj-x86_64-linux-gnu/gnu_13.3_cxx20_64_release
139: Test timeout computed to be: 10000000
139/140 Test #139: test_malloc_overload_disable .............***Exception: SegFault  0.06 sec

test 140
        Start 140: test_malloc_new_handler

140: Test command: /build/tbb/tbb/obj-x86_64-linux-gnu/gnu_13.3_cxx20_64_release/test_malloc_new_handler "--force-colors=1"
140: Working Directory: /build/tbb/tbb/obj-x86_64-linux-gnu/gnu_13.3_cxx20_64_release
140: Test timeout computed to be: 10000000
140/140 Test #140: test_malloc_new_handler ..................***Exception: SegFault  0.07 sec


97% tests passed, 4 tests failed out of 140

Total Test time (real) = 138.23 sec

The following tests FAILED:
        137 - test_malloc_atexit (SEGFAULT)
[...]

PR #1956, opened 2026-02-02
Do not initialize pointers to original functions in (libc_)free

ldorau:
### Description 

Do not initialize pointers to original functions in `(libc_)free()`,
because `dlsym()` calls `(libc_)free()` internally, what can cause
the infinite recursion:
`(libc_)free -> InitOrigPointers -> dlsym -> (libc_)free` (see #1907).
This patch fixes this issue.

Fixes: #1907

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

@Lastique

### DIFF
--- a/src/tbbmalloc_proxy/proxy.cpp
+++ b/src/tbbmalloc_proxy/proxy.cpp
@@ -1,5 +1,6 @@
 /*
-    Copyright (c) 2005-2024 Intel Corporation
+    Copyright (c) 2005-2026 Intel Corporation
+    Copyright (c) 2026 UXL Foundation Contributors
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
@@ -196,7 +197,10 @@ TBBMALLOCPROXY_EXPORT void *PREFIX(calloc)(ZONE_ARG size_t num, size_t size) __T
 
 TBBMALLOCPROXY_EXPORT void PREFIX(free)(ZONE_ARG void *object) __THROW
 {
-    InitOrigPointers();
+    // InitOrigPointers() must NOT be called here,
+    // because it may cause the infinite recursion:
+    // free() -> InitOrigPointers() -> dlsym() -> free()
+    // (see issue #1907)
     __TBB_malloc_safer_free(object, (void (*)(void*))orig_free);
 }
 
@@ -293,7 +297,10 @@ TBBMALLOCPROXY_EXPORT void *__libc_valloc(size_t size) __TBB_ALIAS_ATTR_COPY(val
 // call original __libc_* to support naive replacement of free via __libc_free etc
 TBBMALLOCPROXY_EXPORT void __libc_free(void *ptr)
 {
-    InitOrigPointers();
+    // InitOrigPointers() must NOT be called here,
+    // because it may cause the infinite recursion:
+    // libc_free() -> InitOrigPointers() -> dlsym() -> libc_free()
+    // (see issue #1907)
     __TBB_malloc_safer_free(ptr, (void (*)(void*))orig_libc_free);
 }
 


ldorau:
@isaevil

src/tbbmalloc_proxy/proxy.cpp
```diff
@@ -180,10 +155,32 @@ extern "C" {
...
 #define PREFIX(name) impl_##name
 // not interested in original functions for zone overload
-inline void InitOrigPointers() {}
+static bool doInitOrigPointers = false;
 
 #endif // MALLOC_UNIXLIKE_OVERLOAD_ENABLED and MALLOC_ZONE_OVERLOAD_ENABLED
 
+struct initOrigPointers {
+    initOrigPointers() {
+        if (!doInitOrigPointers) {
```
Lastique: If I'm not mistaken, `doInitOrigPointers` is either always `true` or always `false`, depending on `MALLOC_UNIXLIKE_OVERLOAD_ENABLED`. Which means, `doInitOrigPointers` can be removed and this constructor should be empty unless `MALLOC_UNIXLIKE_OVERLOAD_ENABLED` is defined. If so, I would remove `doInitOrigPointers` and move the `initOrigPointers` class definition above the macro checks and define the constructor in the `MALLOC_UNIXLIKE_OVERLOAD_ENABLED` and `MALLOC_ZONE_OVERLOAD_ENABLED` sections.

ldorau: I removed `doInitOrigPointers` and used a preprocessor directive instead in order to keep the whole implementation in one place.
Done.

ldorau:
@Lastique Does this patch fix your issue?

Alexandr-Konovalov:
If I remember correctly, it was done via initialization on 1st use because some overloaded malloc were called before static ctor runs.  In the suggested code, are we sure that initOrigPointers::free etc contain 0 because they sit in bss?

Lastique:
> @Lastique Does this patch fix your issue?

I didn't test this yet. I will probably have a chance to test it later this week.

> If I remember correctly, it was done via initialization on 1st use because some overloaded malloc were called before static ctor runs. In the suggested code, are we sure that initOrigPointers::free etc contain 0 because they sit in bss?

The C++ standard [guarantees](https://en.cppreference.com/w/cpp/language/initialization.html#Non-local_variables) that namespace-scope static variables are zero-initialized before constructors are run.

`.bss` sections [are zero initialized](https://refspecs.linuxfoundation.org/LSB_1.1.0/gLSB/specialsections.html) on startup.

ldorau:
> > @Lastique Does this patch fix your issue?
> 
> I didn't test this yet. I will probably have a chance to test it later this week.
> 

OK. I have verified this fix using the [reproducer](https://github.com/user-attachments/files/24869062/tbb_malloc_malloc_test_fails.tar.gz) provided by @Lastique in https://github.com/uxlfoundation/oneTBB/issues/1907#issuecomment-3801975973 and it works:

```sh
root@dnp-02:/build/tbb/obj-x86_64-linux-gnu# make -j128 test ARGS\+=--verbose ARGS\+=-j128
[...]
100% tests passed, 0 tests failed out of 140

Total Test time (real) = 480.15 sec

root@dnp-02:/build/tbb/obj-x86_64-linux-gnu# ctest -I 137,140
Test project /build/tbb/obj-x86_64-linux-gnu
    Start 137: test_malloc_atexit
1/4 Test #137: test_malloc_atexit ...............   Passed    0.01 sec
    Start 138: test_malloc_overload
2/4 Test #138: test_malloc_overload .............   Passed    0.02 sec
    Start 139: test_malloc_overload_disable
3/4 Test #139: test_malloc_overload_disable .....   Passed    0.01 sec
    Start 140: test_malloc_new_handler
4/4 Test #140: test_malloc_new_handler ..........   Passed    0.01 sec

100% tests passed, 0 tests failed out of 4

Total Test time (real) =   0.05 sec
```

Alexandr-Konovalov:
> > If I remember correctly, it was done via initialization on 1st use because some overloaded malloc were called before static ctor runs. In the suggested code, are we sure that initOrigPointers::free etc contain 0 because they sit in bss?
> 
> The C++ standard [guarantees](https://en.cppreference.com/w/cpp/language/initialization.html#Non-local_variables) that namespace-scope static variables are zero-initialized before constructors are run.
> 
> `.bss` sections [are zero initialized](https://refspecs.linuxfoundation.org/LSB_1.1.0/gLSB/specialsections.html) on startup.

So, free() is no-op, that is acceptable, but `realloc` allocates nothing before the static constructor executed. Can this be addressed somehow?

ldorau:
> > > If I remember correctly, it was done via initialization on 1st use because some overloaded malloc were called before static ctor runs. In the suggested code, are we sure that initOrigPointers::free etc contain 0 because they sit in bss?
> > 
> > 
> > The C++ standard [guarantees](https://en.cppreference.com/w/cpp/language/initialization.html#Non-local_variables) that namespace-scope static variables are zero-initialized before constructors are run.
> > `.bss` sections [are zero initialized](https://refspecs.linuxfoundation.org/LSB_1.1.0/gLSB/specialsections.html) on startup.
> 
> So, free() is no-op, that is acceptable, but `realloc` allocates nothing before the static constructor executed. Can this be addressed somehow?

If no-op free() is acceptable then there is a much simpler fix that addresses the above mentioned issue.
@Alexandr-Konovalov @lplewa @KFilipek please re-review.

Lastique:
Note that the original code is not thread-safe, as accesses to `orig_free` etc. are not atomic (i.e. readers may observe a partially stored value of the function pointer). The previous version of the fix was also not quite thread-safe but it at least initialized the pointers during global initialization stage, where there are unlikely multiple threads running.

This doesn't mean this has to be fixed in this PR, just an observation.

Also, one other alternative to consider is the [`init_priority`](https://gcc.gnu.org/onlinedocs/gcc/C_002b_002b-Attributes.html) attribute that can be used to force early initialization of the function pointers.

ldorau:
> Note that the original code is not thread-safe, as accesses to `orig_free` etc. are not atomic (i.e. readers may observe a partially stored value of the function pointer). 
>

@Lastique  There is a comment: `// race is OK here, as different threads found same functions` in `InitOrigPointers()`.
I think the authors meant that the first `__TBB_malloc_safer_*()` function will be called when the first call to `InitOrigPointers()` is finished, so when all the pointers: `orig_free`, `orig_realloc`, `orig_msize`, `orig_libc_free`, `orig_libc_realloc` are set to the correct values. In meantime another thread can be storing those pointers again and one of those pointers can be "partially stored", but it does not matter, because a new value (partially stored) is equal to the old value, so a reader will read the correct value even if it is "partially stored", because the rest of the value (not-yet-stored) is already correct (stored by the thread that already returned from `InitOrigPointers()`).

ldorau:
@Alexandr-Konovalov @lplewa @KFilipek please re-review.

KFilipek (approved):

lplewa (approved):

ldorau:
@Alexandr-Konovalov @isaevil Can you merge it?

Lastique:
My local TBB package build with the current version of this PR applied passed. Thanks.

PR #1960, opened 2026-02-09
Mark functions and variables as inline to avoid "TU-local entity exposure" issue

isaevil:
### Description 
This PR fixes ODR violations when inline functions either invoke static functions or use static global variables (fixed by using inline variables since C++17). This also resolve the issue with exposure of TU-local entities when building C++20 modules and using TBB API in exported declarations. 

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [x] No - the only public global variable that was `static` is `tbb::flow::no_priority` which is constant predefined value. This change shouldn't affect TUs that used this constant in regular way (i.e. passing this value as parameter to nodes' constructors).
- [ ] Unknown

### DIFF
--- a/include/oneapi/tbb/detail/_flow_graph_impl.h
+++ b/include/oneapi/tbb/detail/_flow_graph_impl.h
@@ -36,7 +36,7 @@ namespace d2 {
 class graph_task;
 __TBB_GLOBAL_VAR graph_task* const SUCCESSFULLY_ENQUEUED = (graph_task*)-1;
 typedef unsigned int node_priority_t;
-__TBB_GLOBAL_VAR const node_priority_t no_priority = node_priority_t(0);
+__TBB_GLOBAL_VAR constexpr node_priority_t no_priority = node_priority_t(0);
 
 class graph;
 class graph_node;
--- a/include/oneapi/tbb/task_group.h
+++ b/include/oneapi/tbb/task_group.h
@@ -76,7 +76,7 @@ struct task_group_context_impl;
 namespace d2 {
 
 template<typename F>
-inline d1::task* task_ptr_or_nullptr(F&& f);
+d1::task* task_ptr_or_nullptr(F&& f);
 
 template<typename F>
 class function_task : public task_handle_task  {
@@ -130,7 +130,7 @@ public:
 
 #if __TBB_PREVIEW_TASK_GROUP_EXTENSIONS
     template<typename F>
-    inline d1::task* task_ptr_or_nullptr_impl(std::false_type, F&& f){
+    d1::task* task_ptr_or_nullptr_impl(std::false_type, F&& f){
         task_handle th = std::forward<F>(f)();
         task_handle_task* task_ptr = task_handle_accessor::release(th);
         // If task has unresolved dependencies, it can't be bypassed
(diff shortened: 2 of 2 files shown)


include/oneapi/tbb/detail/_flow_graph_impl.h
```diff
@@ -34,9 +34,9 @@ namespace detail {
 namespace d2 {
 
 class graph_task;
-static graph_task* const SUCCESSFULLY_ENQUEUED = (graph_task*)-1;
+__TBB_GLOBAL_VAR graph_task* const SUCCESSFULLY_ENQUEUED = (graph_task*)-1;
 typedef unsigned int node_priority_t;
-static const node_priority_t no_priority = node_priority_t(0);
+__TBB_GLOBAL_VAR const node_priority_t no_priority = node_priority_t(0);
```
kboyarinov:
```suggestion
__TBB_GLOBAL_VAR graph_task* const SUCCESSFULLY_ENQUEUED = (graph_task*)-1;
typedef unsigned int node_priority_t;
__TBB_GLOBAL_VAR constexpr node_priority_t no_priority = node_priority_t(0);
```

include/oneapi/tbb/task_group.h
```diff
@@ -75,10 +75,8 @@ struct task_group_context_impl;
 
 namespace d2 {
 
-namespace {
 template<typename F>
-d1::task* task_ptr_or_nullptr(F&& f);
-}
+inline d1::task* task_ptr_or_nullptr(F&& f);
```
kboyarinov: I don't think the explicit `inline` is needed for a function outside the unnamed namespace.
Same for all overloads and impl functions in this file.

isaevil: You're right. Removed

include/oneapi/tbb/detail/_flow_graph_trace_impl.h
```diff
@@ -42,14 +43,14 @@ template< typename T > class receiver;
         #define CODEPTR() nullptr
     #endif /* __TBB_FLOW_TRACE_CODEPTR */
 
-static inline void fgt_alias_port(void *node, void *p, bool visible) {
+inline void fgt_alias_port(void *node, void *p, bool visible) {
```
vossmjp, line 46: Someone (maybe me) should write an RFC about the direction for `_flow_graph_trace_impl.h`.   Flow Graph Analyzer, which is the only/primary user of `_flow_graph_trace_impl.h`, was deprecated as a tool within Advisor in [Intel(R) Advisor 2024.1](https://www.intel.com/content/www/us/en/docs/advisor/user-guide/2024-1/overview.html) and then removed in [Intel(R) Advisor 2025.0](https://www.intel.com/content/www/us/en/docs/advisor/user-guide/2025-0/overview.html). So we need to decide what to do with this support. For this release, we can fix the uses of `static` but ultimately we might remove much of this code.

isaevil:
> As far as I understand, this changes the linkage for all of the internal stuff. Previously, they had internal linkage, and now they start to have external and therefore visible to other TU so that the linker can deduplicate definitions and use only one of them.
> 
> What I don't understand is whether the external linkage makes these things visible outside the library, like the exported functions. Essentially, introducing new entry points. It seems it does, isn't it? And then, I guess, the .def files and/or default visibility settings conveniently cover this part by hiding them from the out-of-library world.
> 
> Is my understanding correct?

Yes, since we hide all symbols not listed in .def files by default, functions/constants will still remain local in library.

vossmjp (approved):
Looks good to me too.

PR #1964, opened 2026-02-10
Add resource_limited_node and resource_limiter to the Flow Graph

kboyarinov:
### Description 

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown


include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
```diff
@@ -0,0 +1,621 @@
...
+};
+
+template <typename ResourceHandle>
+class resource_provider_base {
+public:
+    using consumer_type = resource_consumer_base<ResourceHandle>;
+    using optional_type = resource_handle_optional<ResourceHandle>;
+
+    virtual void          request(consumer_type&, request_id) = 0;
+    virtual optional_type acquire(consumer_type&, request_id) = 0;
```
vossmjp, line 135:
```suggestion
    virtual optional_type acquire(consumer_type&, request_id) = 0;
    virtual void update_priority(consumer_type&, priority_type) = 0;
```

vossmjp: We need someway to dynamically update the node priority to enable more advanced provider types. Other names could be `report_pressure", etc.

kboyarinov: I guess you meant to add `update_priority` function without removing the `acquire`, right?

vossmjp: Yes, without removing acquire :)

vossmjp: And of course there needs to be a priority_type defined somewhere. Or it could just be node_priority_t. But if we want this to be independent of flow graph, it might be better to have a separate type alias.

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
```diff
@@ -0,0 +1,649 @@
...
+    using clock_type = std::chrono::high_resolution_clock;
+    using time_point = clock_type::time_point;
+
+    time_point    m_time_point;
+    std::uint64_t m_unique_integer;
+public:
+    // request_id()
+    //     : m_time_point(clock_type::now())
+    // {}
+
```
dnmokhov: Remove?

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
```diff
@@ -0,0 +1,649 @@
...
+    {
+        return consumers_tuple_type({std::get<Idx>(resource_providers), this}...);
+    }
+
+    void operator()(const Input& input, OutputPorts& ports) noexcept override {
+        auto& res = form_request(input, ports);
+        report_pressure(0); // TODO: report real pressure
+        request_resources(res.first);
+        release_self_ref(res.first, res.second);
+    }
```
dnmokhov: This is marked as `noexcept`, but can't `m_body` (potentially executed via `release_self_ref`) throw?

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
```diff
@@ -0,0 +1,649 @@
...
+
+template <typename Input, typename OutputPorts, typename Body, typename... ResourceProviders>
+class resource_limited_body_leaf
+    : public resource_limited_body<Input, OutputPorts>
+{
+    using handles_tuple_type = std::tuple<typename ResourceProviders::optional_type...>;
+    using consumers_tuple_type = std::tuple<resource_consumer<Input, OutputPorts, ResourceProviders>...>;
+    using request_data_type = request_data<Input, OutputPorts, handles_tuple_type>;
+    // TODO: should concurrent container be used instead?
+    using requests_map_type = std::unordered_map<request_id, request_data_type, request_id::hash, request_id::equal>;
```
dnmokhov, line 424: Maybe a concurrent container should indeed be used, because couldn't concurrent `operator()` calls result in rehashing and invalidating `res`?
```cpp
void operator()(const Input& input, OutputPorts& ports) noexcept override {
    auto& res = form_request(input, ports);
```

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
```diff
@@ -0,0 +1,649 @@
...
+    time_point    m_time_point;
+    std::uint64_t m_unique_integer;
+public:
+    // request_id()
+    //     : m_time_point(clock_type::now())
+    // {}
+
+    request_id(const std::uint64_t& unique_integer)
+        : m_time_point(clock_type::now())
+        , m_unique_integer(unique_integer)
```
dnmokhov, line 48: If both `m_time_point` and `m_unique_integer` are unique, why are both needed?

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
kboyarinov: It seems like you are right. Good catch!
The only option we have that fits all the requirements is `concurrent_hash_map`. I will see how to rewrite the code to correctly use the accessors.

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
kboyarinov: Initially, only the time point was used, but unfortunately, it is not unique. Imagine several threads requested the same resource at exactly the same time point. 
But we will need time point for future improvements of the provider for fairness.
The unique integer was added to improve the uniqueness of the `request_id`. Each consumer contains an `atomic<uint64_t>` that is increased for each request and never decreased. 
Only the time point is used for hashing, but the equality function compares both the time point and the integer to resolve collisions for requests happened at the same time point.

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
kboyarinov: I don't think it is really needed now - I just made some experiments with the lightweight policy and forgot to remove it.

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
kboyarinov: I have double-checked the C++ standard requirements for unordered_map iterators invalidation. It states that `emplace` (or any other insertion) invalidates iterators if the rehashing took place, but also guarantees that the reference to the corresponding element remains valid.
Since `res` is a reference in this context, there is no correctness issue here.

The only reason to consider using `concurrent_hash_map` here is the performance, but I guess we need to properly compare these maps on a real benchmark for resource limiting.

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
kboyarinov: Done

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
kboyarinov: Removed

vossmjp (commented):
What about feature-test macro?  TBB_HAS_FLOW_GRAPH_RESOURCE_LIMITED_NODE

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
```diff
@@ -0,0 +1,653 @@
...
+    time_point    m_time_point;
+    std::uint64_t m_unique_integer;
+public:
+    request_id(const std::uint64_t& unique_integer)
+        : m_time_point(clock_type::now())
+        , m_unique_integer(unique_integer)
+    {}
+
+    struct hash {
+        std::hash<time_point::duration::rep> m_hash;
```
vossmjp: Since the request_id has a unique_integer now, why not create a hash that uses it instead?

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
```diff
@@ -0,0 +1,653 @@
...
+#endif
+
+#include <chrono>
+#include <unordered_map>
+#include <forward_list>
+#include <functional>
+#include <utility>
+#include <atomic>
+#include <tuple>
+#include <initializer_list>
```
vossmjp:
```suggestion
```

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
```diff
@@ -0,0 +1,653 @@
...
+    bool has_value() const { return m_has_value; }
+
+    ResourceHandle& value() {
+        __TBB_ASSERT(has_value(), nullptr);
+        return m_resource_handle;
+    }
+};
+
+template <typename ResourceHandle>
+class resource_provider_base {
```
vossmjp, line 129: Since we've decided to rename as resource_limiter, probably all related names such as resource_provider_base should match (i.e. become resource_limiter_base) etc.

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
```diff
@@ -0,0 +1,653 @@
...
+
+    virtual void          request(consumer_type&, request_id) = 0;
+    virtual optional_type acquire(consumer_type&, request_id) = 0;
+    virtual void          report_pressure(consumer_type&, std::size_t) = 0;
+    virtual void          release(consumer_type&, request_id, optional_type&&) = 0;
+    virtual ~resource_provider_base() = default;
+};
+
+template <typename ResourceHandle>
+class resource_consumer_base {
```
vossmjp, line 142: Since `resource_limited_base` would be much too similar to `resource_limiter_base`, I think we should keep this old name of `resource_consumer_base`. So no change here.

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
```diff
@@ -0,0 +1,653 @@
...
+public:
+    using provider_type = resource_provider_base<ResourceHandle>;
```
vossmjp, line 144: But probably this should be `using limiter_type = resource_limiter_base<ResourceHandle>;

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
```diff
@@ -0,0 +1,653 @@
...
+    template <typename ConsumersTuple>
+    resource_limited_body_leaf(graph& g, ConsumersTuple&& consumers_tuple, const Body& body)
+        : resource_limited_body<Input, OutputPorts>(g)
+        , m_consumers(std::forward<ConsumersTuple>(consumers_tuple))
+        , m_body(body)
+        , m_counter(0)
+    {}
+
+public:
+    resource_limited_body_leaf(graph& g, std::tuple<ResourceProviders&...> resource_providers, const Body& body)
```
vossmjp, line 441: And `resource_providers` should probably be `resource_limiters`, etc.

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
```diff
@@ -0,0 +1,653 @@
...
+    using consumer_type = typename resource_provider_base<ResourceHandle>::consumer_type;
+    using optional_type = typename resource_provider_base<ResourceHandle>::optional_type;
+
+    template <typename Handle, typename... Handles>
+    resource_limiter(Handle&& handle, Handles&&... handles) {
+        emplace_handles(std::forward<Handle>(handle), std::forward<Handles>(handles)...);
+    }
+
+    void request(consumer_type& consumer, request_id id) override {
+        tbb::spin_mutex::scoped_lock lock(m_mutex);
```
vossmjp, line 164: Eventually, should we consider using an aggregator in the starvation-avoidance implementation of the resource_limiter.

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
```diff
@@ -0,0 +1,653 @@
...
+
+        m_resource_handles.emplace_front(std::move(handle.value()));
+        
+        auto consumers = std::move(m_consumers);
+        m_consumers.clear();
+
+        lock.release();
+        for (auto consumer_dt : consumers) {
+            consumer_dt.second->notify(*this, consumer_dt.first);
```
vossmjp, line 199: There's no way for a consumer to detach from a `resource_limiter` once it has a made a request. So the lifetime of any consumer that requests a resource must extend at least until it is notified. What happens if flow graph F1 is canceled but another flow graph, F2, is not. The canceled flow graph could be destroyed, along with all of its nodes. The non-canceled flow graph could make progress, with a node releasing a resource in F2, and then the `resource_limiter` would notify the deleted consumer in F1.

test/tbb/test_resource_limited_node.cpp
```diff
@@ -0,0 +1,416 @@
...
+// Test that ten resources are granted simultaneously to ten different nodes
+void test_several_resources() {
+    using namespace oneapi::tbb::flow;
+
+    resource_limiter<int> limiter{0, 1, 2, 3, 4, 5, 6, 7, 8, 9};
+
+    using node_type = resource_limited_node<int, std::tuple<int>>;
+    using ports_type = typename node_type::output_ports_type;
+
+    graph g;
```
vossmjp, line 93: Seems like, due to potential lifetime issue, we need some tests that use more than one flow graph concurrently. And also have flow graph cancellation and reset tests.

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
```diff
@@ -0,0 +1,653 @@
...
+            // Access to all resources is granted
+            call_body(req_data.input_message, req_data.output_ports, req_data.handles);
+            release_resources(id, req_data);
+            remove_request(id);
+        }
+    }
+
+    void release_resources(request_id id, request_data_type& req_data) {
+        release_resources_helper<sizeof...(ResourceProviders)>::run(m_consumers, id, req_data);
+        report_pressure(0); // TODO: report real pressure
```
vossmjp: I think pressure should be reported before the release so that notifies to this same consumer will be properly prioritized. But since this is a placeholder for now it doesn't matter.

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
kboyarinov: I thought these base classes are intended to represent the protocol that uses the provider-consumer terminology. In case of future extensions outside the Flow Graph, they would be moved as well.

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
kboyarinov: I think it definitely worth trying. I will add a TODO comment

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
kboyarinov: Fortunately, I have added an extra reference in the Flow Graph for each request made by the node. I was not sure if it's needed, but now I am 100% sure:)
It actually helps because even the canceled graph should be destroyed only after `wait_for_all()` returns. And any active request from the `resource_limited_node` would prevent it from completion until the node is notified. The only missed thing now is proper removal of the request in case of cancellation that can cause `wait_for_all()` to deadlock. I will fix that and will add some tests for cancellation.

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
kboyarinov: Sorry, I'm not sure I understood your proposal. Is it about calculating and reporting the real pressure anyway now even if it is not supported by the `resource_limiter`? Or is it more like a note to other reviewers about the decision we made in this direction?

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
kboyarinov: Done

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
kboyarinov: Added a comment

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
kboyarinov: Fixed task cancellation and added test for cancellation with active requests

test/tbb/test_resource_limited_node.cpp
kboyarinov: Fixed task cancellation and added test for cancellation with active requests

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
vossmjp: No, it is just a question about whether release should happen before report_pressure, or the other way around. Because release can cause a notify. And the resource_limiters will have out-of-date info about this consumer.

kboyarinov: Oops, I thought release is oneTBB 2023.0 release:)

vossmjp: That's really funny :) But, yes I meant the call to the `release_resources_helper::run`. And since nothing real is being reported, it doesn't need to be switched now anyway.

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
```diff
@@ -0,0 +1,669 @@
...
+namespace d2 {
+
+template <typename ResourceHandle>
+class resource_consumer_base;
+
+class request_id {
+    using clock_type = std::chrono::high_resolution_clock;
+    using time_point = clock_type::time_point;
+
+    time_point    m_time_point;
```
vossmjp, line 43: Is m_time_point needed anymore? Eventually, we may want to use time as part of starvation avoidance (or not) but its private here so seems unecessary.

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
```diff
@@ -0,0 +1,669 @@
...
+        __TBB_ASSERT(prev_value == 0, "Incorrect notify counter before acquisition");
+        tbb::detail::suppress_unused_warning(prev_value);
+        return acquire_resources_helper<0, sizeof...(ResourceProviders)>::run(this, m_consumers, id, req_data);
+    }
+
+    void try_acquire_resources_and_execute(request_id id, request_data_type& req_data) {
+        if (try_acquire_resources(id, req_data)) {
+            // Access to all resources is granted
+            call_body(req_data.input_message, req_data.output_ports, req_data.handles);
+            release_resources(id, req_data);
```
vossmjp, line 499: What if the body throws an exception? Since the resource_limiters should be usable across flow graphs, an exception in one graph should not affect another graph, but I think the resources will not be released.

vossmjp (dismissed):
There are some changes that should be made either now or after merge. First is a minor thing, which is the time point can just be completely removed from the request id now that its not used in the hash. And then more importantly, exception handling should be added since we expect that `resource_limiters` can be used across flow graphs. So we should be sure that resources are returned even if there's an exception in a body call.

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
kboyarinov: Will be addressed in a separate PR

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
kboyarinov: Will be addressed in a separate PR

isaevil (approved):

pranasge:
As someone that would use this functionality, I don't see a way to solve some problems.
For me resource_limiter would be really useful if it could work with async_node as even by your example with databaseHandle I would expect to do that in async_node. Other way to do that would be that resource_limiter could be passed to limiter_node so that multiple limiter_nodes could request same resource. Is there way with current implementation to limit multiple async_nodes with one resource_limiter?

kboyarinov:
> As someone that would use this functionality, I don't see a way to solve some problems. For me resource_limiter would be really useful if it could work with async_node as even by your example with databaseHandle I would expect to do that in async_node. Other way to do that would be that resource_limiter could be passed to limiter_node so that multiple limiter_nodes could request same resource. Is there way with current implementation to limit multiple async_nodes with one resource_limiter?

@pranasge thanks for reaching out with this question. At the moment, the current implementation limits `resource_limiter` support to `resource_limiter_node` only. In other words, only instances of this node can request one or more shared resources before executing something.

That said, extending `resource_limiter` and the resource-limiting protocol is currently under active development. One option we are considering is to make `resource_limiting` a policy that can be applied to multiple node types, possibly including `async_node`, or `limiter_node`. It would be very helpful to hear more about your specific use case, as this can help guide and motivate further improvements in this area.

pranasge:
My use case for this would be global rate limiting external service calls from multiple graphs. Currently I use kafka producer that blocks thread when it reaches message limit. If I could use resource_limiter on async_node onetbb would not keep threads idle in cases message limit is reached.

PR #1974, opened 2026-02-24
Add basic C++20 modules testing, extend test_tbb_header

isaevil:
### Description 
Add test that verifies that a module, that includes TBB headers as part of global module fragment, compiles without errors and all TBB definitions are present. This patch also adds some missing APIs definition testing to `test_tbb_header`

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown


isaevil:
@kboyarinov @aleksei-fedotov @dnmokhov @Alexandr-Konovalov please take a look.

test/tbb/test_tbb_header.cpp
```diff
@@ -1,6 +1,6 @@
 /*
     Copyright (c) 2005-2025 Intel Corporation
-    Copyright (c) 2025 UXL Foundation Contributors
+    Copyright (c) 2025-2026 UXL Foundation Contributors
```
isaevil:
```suggestion
    Copyright (c) 2025 UXL Foundation Contributors
```

test/tbb/test_tbb_header.cpp
```diff
@@ -92,8 +94,13 @@
...
 #endif
 
-static volatile size_t g_sink;
+#endif /* !__TBB_TEST_MODULE_EXPORT */
 
+#if __TBB_TEST_MODULE_EXPORT
+inline volatile size_t g_sink;
+#else
+static volatile size_t g_sink;
+#endif /* __TBB_TEST_MODULE_EXPORT */
```
kboyarinov, line 107: Can we leave just the inline variable for both testing modes?

test/tbb/test_tbb_header.cpp
```diff
@@ -135,10 +145,15 @@ struct Body3a { // for lambda-friednly parallel_scan
...
 struct Msg {};
+struct SuspendBody {
+    void operator()(tbb::task::suspend_point) const {}
+};
 
 // Test if all the necessary symbols are exported for the exceptions thrown by TBB.
 // Missing exports result either in link error or in runtime assertion failure.
+#if !__TBB_TEST_MODULE_EXPORT
 #include <stdexcept>
+#endif
```
kboyarinov: May be move this include up under the macro in the beginning of the file?

test/tbb/test_tbb_header.cpp
```diff
@@ -163,7 +178,11 @@ void TestExceptionClassExports ( const E& exc, tbb::detail::exception_id eid ) {
 #endif /* TBB_USE_EXCEPTIONS */
 }
 
+#if __TBB_TEST_MODULE_EXPORT
+inline void TestExceptionClassesExports () {
```
kboyarinov, line 180: Can we keep only the inline function? Here and in other places.

test/tbb/test_tbb_header.cpp
```diff
@@ -180,18 +199,36 @@ static void TestExceptionClassesExports () {
...
     TestTypeDefinitionPresence2( concurrent_lru_cache<int, int> );
     TestTypeDefinitionPresence( isolated_task_group );
 #if TBB_PREVIEW_MEMORY_POOL
     TestTypeDefinitionPresence( memory_pool_allocator<int> );
     TestTypeDefinitionPresence( memory_pool<std::allocator<int>> );
     TestTypeDefinitionPresence( fixed_pool );
 #endif
+#if __TBB_PREVIEW_TASK_GROUP_EXTENSIONS
+    TestTypeDefinitionPresence( task_completion_handle );
+    TestFuncDefinitionPresence( is_inside_task, (), bool );
```
kboyarinov, line 214: This function is actually interesting. It is provided publically, but it is not documented and I am not aware if it is used by someone. I would think twice if we need to export it from the module.
May be even remove this function completely

test/tbb/test_tbb_header_module.cppm
```diff
@@ -0,0 +1,47 @@
...
+
+#ifndef NOMINMAX
+#define NOMINMAX
+#endif
+
+#define __TBB_NO_IMPLICIT_LINKAGE 1
+#define __TBB_TEST_MODULE_EXPORT 1
+
+#define CHECK(x) do { if (!(x)) { std::terminate(); } } while (false)
+#define CHECK_MESSAGE(x, y) CHECK(x);
```
kboyarinov: Why not just include `common/test.h`?

test/CMakeLists.txt
```diff
@@ -557,8 +557,25 @@ if (TARGET TBB::tbb)
         tbb_add_test(SUBDIR tbb NAME test_tbb_fork DEPENDENCIES TBB::tbb)
     endif()
 
-    tbb_add_test(SUBDIR tbb NAME test_tbb_header DEPENDENCIES TBB::tbb)
+    tbb_add_test(SUBDIR tbb NAME test_tbb_header DEPENDENCIES TBB::tbb TBB::tbbmalloc)
```
kboyarinov, line 560: Why the extra dependency is needed?

test/tbb/test_tbb_header.cpp
isaevil: Second testing mode may use C++11/14, so I am afraid we cannot do that.

test/CMakeLists.txt
isaevil: Because I've added testing of scalable_memory_resource :)
The test ODR-uses all these declarations, so it will fail due to undefined symbols.

test/tbb/test_tbb_header.cpp
isaevil: I think since it is preview, we can actually keep it for now and decide to remove it later.

test/tbb/test_tbb_header.cpp
kboyarinov: Agreed to keep as-is

test/tbb/test_tbb_header_module.cppm
isaevil: Done.

test/tbb/test_tbb_header.cpp
isaevil: Done

kboyarinov (approved):
LGTM

PR #2024, opened 2026-03-26
Not define __TBB_CPP17_HW_INTERFERENCE_SIZE_PRESENT for Clang

fujunwei:
We are trying to build OpenVINO with Chromiumt, but the  `__TBB_CPP17_HW_INTERFERENCE_SIZE_PRESENT` is only defined for MSVC (`_MSC_VER >= 1911`), the Clang (used in Chromium builds) doesn't support
`std::hardware_destructive_interference_size` that is not available in Clang's standard library, so  adds `&& !__clang__` to the `__TBB_CPP17_HW_INTERFERENCE_SIZE_PRESENT` macro.

### DIFF
--- a/include/oneapi/tbb/detail/_config.h
+++ b/include/oneapi/tbb/detail/_config.h
@@ -265,7 +265,7 @@
 
 #define __TBB_CPP17_MEMORY_RESOURCE_PRESENT             (_MSC_VER >= 1913 && (__TBB_LANG > 201402L) || \
                                                         __TBB_GLIBCXX_VERSION >= 90000 && __TBB_LANG >= 201703L)
-#define __TBB_CPP17_HW_INTERFERENCE_SIZE_PRESENT        (_MSC_VER >= 1911)
+#define __TBB_CPP17_HW_INTERFERENCE_SIZE_PRESENT        (_MSC_VER >= 1911 && !__clang__)
 #define __TBB_CPP17_LOGICAL_OPERATIONS_PRESENT          (__TBB_LANG >= 201703L)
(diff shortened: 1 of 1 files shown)


dnmokhov:
@fujunwei, could you please rebase your branch from master to ensure all the checks pass? Thanks in advance!

petterreinholdtsen:
I notice your change also fail on the copyright check.  After some struggle I figured out that this check expect a line like "Copyright (c) 2026 UXL Foundation Contributors" in the headers where it claim there is missing copyright information.

fujunwei:
Rebased the latest code, PTAL. Thanks.

fujunwei:
@dnmokhov @petterreinholdtsen could you please take a look again, thanks.

include/oneapi/tbb/detail/_config.h
```diff
@@ -265,7 +265,7 @@
 
 #define __TBB_CPP17_MEMORY_RESOURCE_PRESENT             (_MSC_VER >= 1913 && (__TBB_LANG > 201402L) || \
                                                         __TBB_GLIBCXX_VERSION >= 90000 && __TBB_LANG >= 201703L)
-#define __TBB_CPP17_HW_INTERFERENCE_SIZE_PRESENT        (_MSC_VER >= 1911)
+#define __TBB_CPP17_HW_INTERFERENCE_SIZE_PRESENT        (_MSC_VER >= 1911 && !__clang__)
```
dnmokhov, line 268: This has an undesirable effect of disabling `std::hardware_destructive_interference_size` for ICX (which is Clang-based). Have you considered using something like `__has_feature` instead?

NsPro04:
I just ran into this issue.
I'm compiling using clang + libcxx on Windows
But in my case, the problem is solved just by adding the missing line `#include <new>`

Issue #15, opened 2017-04-05
occur "gmake[1]: *** [x86_rtm_rw_mutex.o] Error 1" when I install it


shen123shen:
I installed the tbb from source code. Run gmake occurs below error.
CentOS release 6.5 (Final), gcc 4.9.2

/tmp/ccK2bNjL.s: Assembler messages:
/tmp/ccK2bNjL.s:615: Error: no such instruction: `xtest'
/tmp/ccK2bNjL.s:634: Error: no such instruction: `xend'
/tmp/ccK2bNjL.s:801: Error: no such instruction: `xbegin .L54'
/tmp/ccK2bNjL.s:817: Error: no such instruction: `xabort $255'
/tmp/ccK2bNjL.s:972: Error: no such instruction: `xbegin .L70'
/tmp/ccK2bNjL.s:990: Error: no such instruction: `xabort $255'
gmake[1]: *** [x86_rtm_rw_mutex.o] Error 1
gmake[1]: Leaving directory `/export/servers/gcc/tbb-2017_U5/build/linux_intel64_gcc_cc4.9.2_libc2.12_kernel2.6.32_debug'
gmake: *** [tbb] Error 2

ntfshard:
Hello

It looks like you have newer version of GCC package and old version of gnu assembler( binutils package)
Try to update your Binutils

--Maksim

PR #2025, opened 2026-03-27
Removed artificial PATH_MAX limits.

petterreinholdtsen:
Rewrote all code using PATH_MAX to instead dynamically allocate the required buffers.  As the actually max path length depend on the file system used, it is not a good idea to pretend a compiled in buffer limit will work everywhere.

This also fixes build problems detected on GNU Hurd.

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/src/tbb/cgroup_info.h
+++ b/src/tbb/cgroup_info.h
@@ -149,11 +149,12 @@ private:
     }
 
     static bool try_read_cgroup_v1_num_cpus_from(const char* dir, int& num_cpus) {
-        char path[PATH_MAX] = {0};
-        if (std::snprintf(path, PATH_MAX, "%s/cpu.cfs_quota_us", dir) < 0)
+        std::size_t pathlen = strlen(dir) + 30;
+        std::unique_ptr<char[]> path(new char[pathlen]);
+        if (std::snprintf(path.get(), pathlen, "%s/cpu.cfs_quota_us", dir) < 0)
             return false;       // Failed to create path
 
-        unique_file_t fd(std::fopen(path, "r"), &close_file);
+        unique_file_t fd(std::fopen(path.get(), "r"), &close_file);
         if (!fd)
             return false;
 
@@ -166,9 +167,10 @@ private:
             return true;
         }
 
-        if (std::snprintf(path, PATH_MAX, "%s/cpu.cfs_period_us", dir) < 0)
+        if (std::snprintf(path.get(), pathlen, "%s/cpu.cfs_period_us", dir) < 0)
             return false;       // Failed to create path;
-        fd.reset(std::fopen(path, "r"));
+
+        fd.reset(std::fopen(path.get(), "r"));
         if (!fd)
             return false;
 
@@ -181,11 +183,12 @@ private:
     }
 
     static bool try_read_cgroup_v2_num_cpus_from(const char* dir, int& num_cpus) {
-        char path[PATH_MAX] = {0};
-        if (std::snprintf(path, PATH_MAX, "%s/cpu.max", dir) < 0)
+        std::size_t pathlen = strlen(dir) + 30;
+        std::unique_ptr<char[]> path(new char[pathlen]);
+        if (std::snprintf(path.get(), pathlen, "%s/cpu.max", dir) < 0)
             return false;       // Failed to create path
 
-        unique_file_t fd(std::fopen(path, "r"), &close_file);
+        unique_file_t fd(std::fopen(path.get(), "r"), &close_file);
         if (!fd)
             return false;
 
@@ -222,9 +225,10 @@ private:
 
     static int parse_cgroup_entry(const char* mnt_dir, process_cgroup_data& pcd) {
         int num_cpus = error_value; // Initialize to an impossible value
-        char dir[PATH_MAX] = {0};
-        if (std::snprintf(dir, PATH_MAX, "%s/%s", mnt_dir, pcd.relative_path) >= 0) {
-            if (try_read_cgroup_num_cpus_from(dir, num_cpus, pcd.version)) {
+        std::size_t dirlen = strlen(mnt_dir) + strlen(pcd.relative_path) + 2;
+        std::unique_ptr<char[]> dir(new char[dirlen]);
+        if (std::snprintf(dir.get(), dirlen, "%s/%s", mnt_dir, pcd.relative_path) >= 0) {
+            if (try_read_cgroup_num_cpus_from(dir.get(), num_cpus, pcd.version)) {
                 return num_cpus;
             }
         }
@@ -276,16 +280,20 @@ private:
     static int try_common_cgroup_mount_path(const process_cgroup_data& pcd,
                                             const cgroup_settings& cg_cfg) {
         int num_cpus = error_value;
-        char dir[PATH_MAX] = {0};
+        std::size_t dirlen = (strlen(pcd.relative_path)
+                              + strlen(cg_cfg.sys_fs_cgroup_dir_path)
+                              + 25);
+        std::unique_ptr<char[]> dir(new char[dirlen]);
         __TBB_ASSERT(*pcd.relative_path, nullptr);
-        if (0 <= std::snprintf(dir, PATH_MAX, "%s/%s", cg_cfg.sys_fs_cgroup_dir_path,
+        if (0 <= std::snprintf(dir.get(), dirlen, "%s/%s",
+                               cg_cfg.sys_fs_cgroup_dir_path,
                                pcd.relative_path))
-            try_read_cgroup_num_cpus_from(dir, num_cpus, pcd.version);
+            try_read_cgroup_num_cpus_from(dir.get(), num_cpus, pcd.version);
 
         if (error_value == num_cpus && pcd.version == process_cgroup_data::cgroup_version::v2) {
-            if (0 <= std::snprintf(dir, PATH_MAX, "%s/%s", "/sys/fs/cgroup/unified",
+            if (0 <= std::snprintf(dir.get(), dirlen, "%s/%s", "/sys/fs/cgroup/unified",
                                    pcd.relative_path))
-                try_read_cgroup_v2_num_cpus_from(dir, num_cpus);
+                try_read_cgroup_v2_num_cpus_from(dir.get(), num_cpus);
         }
         return num_cpus;
     }
--- a/src/tbb/dynamic_link.cpp
+++ b/src/tbb/dynamic_link.cpp
@@ -45,9 +45,6 @@
     #include <wintrust.h>
     #pragma comment (lib, "wintrust")     // Link with the Wintrust.lib file.
 #endif
-#ifndef PATH_MAX
-    #define PATH_MAX                MAX_PATH
-#endif
 #else /* _WIN32 */
     #include <dlfcn.h>
     #include <unistd.h>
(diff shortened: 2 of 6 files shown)


Alexandr-Konovalov:
What is the reason for memory management be C-style vs using `unique_ptr` etc?

petterreinholdtsen:
[Alexandr-Konovalov]
> What is the reason for memory management be C-style vs using
> `unique_ptr` etc?

Mostly because I am mostly familir with C style, and could not get
unique_ptr with char* working with the surrounding code.  Can you get it
working?

-- 
Happy hacking
Petter Reinholdtsen

src/tbb/cgroup_info.h
```diff
@@ -149,26 +149,43 @@ class cgroup_info {
     }
 
     static bool try_read_cgroup_v1_num_cpus_from(const char* dir, int& num_cpus) {
-        char path[PATH_MAX] = {0};
-        if (std::snprintf(path, PATH_MAX, "%s/cpu.cfs_quota_us", dir) < 0)
+        std::size_t pathlen = strlen(dir) + 30;
```
aleksei-fedotov, line 152: Why +30? The longest addition is only 18 bytes, if I am not mistaken.

src/tbb/cgroup_info.h
```diff
@@ -149,26 +149,43 @@ class cgroup_info {
...
+        char *path = (char*)calloc(pathlen, 1);
```
aleksei-fedotov: Usually, allocating memory dynamically takes a way more time than work with static arrays. Running over arrays using `strlen` can also add some penalties. We need to look at the performance data to make a weighed decision.

src/tbb/cgroup_info.h
```diff
@@ -149,26 +149,43 @@ class cgroup_info {
...
+        if (std::snprintf(path, pathlen, "%s/cpu.cfs_quota_us", dir) < 0) {
+            free(path);
```
aleksei-fedotov: Please use RAII approach. It is safer and cleaner.

src/tbb/dynamic_link.cpp
```diff
@@ -710,42 +743,50 @@ namespace r1 {
         dynamic_link_handle library_handle = nullptr;
 #if __TBB_DYNAMIC_LOAD_ENABLED
         const char* path = library;
-        std::size_t const len = PATH_MAX + 1;
-        char absolute_path[ len ] = {0};
+        std::size_t len = abs_path( library, NULL, 1);
```
aleksei-fedotov:
```suggestion
        std::size_t len = abs_path(library, NULL, 1);
```

petterreinholdtsen:
[Petter Reinholdtsen]
> Mostly because I am mostly familir with C style, and could not get
> unique_ptr with char* working with the surrounding code.  Can you get it
> working?

It is worth noting that the surrounding code using char* and snprintf()
instead of std::string and associated operations, made me believe C
style was the intended and wanted code style.  If C++ is wanted, I
suspect the entire file should be rewritten to use std::string.  I
decided against a complete rewrite to keep the change to the minimum
required to drop the arbitary limit.  I was unable to drop such limit on
Windows, where the API call used seem to require the limit, not the
calling code.

-- 
Happy hacking
Petter Reinholdtsen

aleksei-fedotov (commented):
Given that the patch does not take into account hard limitation on the length of a relative path, I conclude that the limitations on the maximum path length is not a real motivator for the patch. Therefore, to avoid cumbersome process of measuring performance penalties potentially implied by dynamic memory allocations, I would suggest simply defining `PATH_MAX` constant on the systems where it is missing.

Below are the comments to start with if we still want to go that way.

petterreinholdtsen:
[Aleksei Fedotov]
> Given that the patch does not take into account hard limitation on the
> length of a relative path, I conclude that the limitations on the
> maximum path length is not a real motivator for the patch.

Your conclusion is wrong.  The reason I try porting software to GNU
Hurd, is that it identifies artifical limits in Linux software that have
no real purpose.  Given that the limit also is not accurate, as that the
limits differ for different file system types, to me using PATH_MAX seem
like a useless limitation to include in programs.

> Below are the comments to start with if we still want to go that way.

Thank you for the fedback.

> Why +30? The longest addition is only 18 bytes, if I am not mistaken.

I picked a number I was sure was larger than the smallest possible
number.  If it is important for the onetbb developers to stay at the
exact limit, I can do more detailed calculations.

> Usually, allocating memory dynamically takes a way more time than work
> with static arrays. Running over arrays using `strlen` can also add
> some penalties. We need to look at the performance data to make a
> weighed decision.

Yes.  But as far as I can tell, the operations affected are conducted
once or very few times per program invocation, so the performance hit
should be negletiable and the advantage (ie not using the actually file
system dependent limit) should outweigh the disadvantages.

> Please use RAII approach. It is safer and cleaner.

What do you mean?

> ```suggestion
>         std::size_t len = abs_path(library, NULL, 1);
> ```

No problem.  Noticed a bit inconsistent style in the code base.

-- 
Happy hacking
Petter Reinholdtsen

Alexandr-Konovalov:
Can we make one step further and look at `PATH_MAX` [semantics](https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/limits.h.html) under Linux? According to my understanding, it limits number of bytes to be written by "system" to a buffer of unspecified size. As war as I can tell, TBB don't use such API, so for TBB `PATH_MAX` is used only as size of a buffer on stack that has enough size. So, can we remove `PATH_MAX` and set it manually to OS-independent value? There is no need in dynamic memory allocation after that.

petterreinholdtsen:
[Alexandr-Konovalov]
> Can we make one step further and look at `PATH_MAX` semantics under
> Linux.  According to my understanding, it limits number of bytes to be
> written by "system" to a buffer of unspecified size. As war as I can
> tell, TBB don't use such API, so for TBB `PATH_MAX` is used only as
> size of a buffer on stack that has enough size. So, can we remove
> `PATH_MAX` and set it manually to OS-independent value? There is no
> need in dynamic memory allocation after that.

I assume you are talking about this:

  Pathname Variable Values

  The values in the following list may be constants within an
  implementation or may vary from one pathname to another. For example,
  file systems or directories may have different characteristics.

  A definition of one of the symbolic constants in the following list
  shall be omitted from the <limits.h> header on specific
  implementations where the corresponding value is equal to or greater
  than the stated minimum, but where the value can vary depending on the
  file to which it is applied. The actual value supported for a specific
  pathname shall be provided by the pathconf() function.

  [...]

  {PATH_MAX} - Maximum number of bytes the implementation stores as a
               pathname in a user-supplied buffer of unspecified size,
               including the terminating null character. Minimum number
               the implementation shall accept as the maximum number of
               bytes in a pathname.
               Minimum Acceptable Value: {_POSIX_PATH_MAX}
  [...]

  {PATH_MAX} - IEEE PASC Interpretation 1003.1 #15 addressed the
               inconsistency in the standard with the definition of
               pathname and the description of {PATH_MAX}, allowing
               application developers to allocate either {PATH_MAX} or
               {PATH_MAX}+1 bytes. The inconsistency has been removed by
               correction to the {PATH_MAX} definition to include the
               null character. With this change, applications that
               previously allocated {PATH_MAX} bytes will continue to
               succeed.

As far as I know, PATH_MAX usage is not limited to what the system
writes, but intended also to limit what the maximum path a file on disk
can have, but the only proper way to figure out the latter is to use the
pathconf() function when one know which file system the program will be
operating on.  Such procedure is rarely worth it when the alternative is
to simply not have a limit and handle whatever string size is passed
from the user or system dynamically, which is what my patch implements.

How do you imagine determining "enough size"?  What is the program is
ported to a system with deeper path depths in the future?  Do you want
to keep extending the limit indefinitely to avoid dynamically handle any
size?

-- 
Happy ahcking
Petter Reinholdtsen

src/tbb/dynamic_link.cpp
petterreinholdtsen: Done.

Alexandr-Konovalov:
There are actually 2 different problem here we are discussing. 1st is `PATH_MAX` dependence, and 2nd is artificial limitations on paths. Sure, if 2nd is resolved, the 1st is also disappeared. But the developers are not happy with possible impact of dynamic memory allocation on performance, so thinking about possibility to keep buffers on stack somehow.

For microbenchmark with only
`volatile int num_cpus = tbb::detail::r1::cgroup_info<>::parse_cpu_constraints(tbb::detail::r1::default_cgroup_settings{});
`
I see ~1% performance drop, and most time-consuming things are syscalls. Hope this is the relevant benchmark.

> How do you imagine determining "enough size"?  What is the program is
ported to a system with deeper path depths in the future?  Do you want
to keep extending the limit indefinitely to avoid dynamically handle any
size?

Sorry, I think only about `cgroup_info.h`. That is not the complete picture.

src/tbb/dynamic_link.cpp
```diff
@@ -340,12 +337,15 @@ namespace r1 {
     static std::once_flag init_dl_data_state;
 
     static struct ap_data_t {
-        char _path[PATH_MAX+1];
+        char *_path;
+        std::size_t _capacity;
```
Alexandr-Konovalov, line 341: Is `_capacity` really need to be saved? If I understand correctly, it only used for doubling the buffer, so probably can be lokal.

Alexandr-Konovalov:
> Mostly because I am mostly familir with C style, and could not get unique_ptr with char* working with the surrounding code. Can you get it working?

It can be something like
```
        std::unique_ptr<char[]> dir(new char[dirlen]);
        __TBB_ASSERT(*pcd.relative_path, nullptr);
        if (0 <= std::snprintf(dir.get(), dirlen, "%s/%s", cg_cfg.sys_fs_cgroup_dir_path,
                               pcd.relative_path))
            try_read_cgroup_num_cpus_from(dir.get(), num_cpus, pcd.version);

```
I think.

src/tbb/dynamic_link.cpp
petterreinholdtsen: I believe it is required on Windows, and for consistency I used it the same way on non-Windows platforms.  I was tempted to switch also this to use std::string, but decided against it as std::string is not much used in this library.

petterreinholdtsen:
I switched to unique_ptr everywhere except in ap_data.  I do not fully understand the purpose of the internal ap_data_t structure, and am unsure about the consequences for this code.

petterreinholdtsen:
[Alexandr-Konovalov]
> But the developers are not happy with possible impact of dynamic
> memory allocation on performance, so thinking about possibility to
> keep buffers on stack somehow.

So, how can this fix be moved forward?  It is unclear to me who are the
decision makers and what I can do to assist further.

-- 
Happy hacking
Petter Reinholdtsen

src/tbb/dynamic_link.cpp
Alexandr-Konovalov: > I believe it is required on Windows.

I'm not sure in that. It seems it only used to keep value of `MAX_PATH` and then zeroed in case of unsuccessful allocation. At which part of the code are you talking about?

petterreinholdtsen:
[Petter Reinholdtsen]
> I believe it is required on Windows.

[Alexandr-Konovalov]
> I'm not sure in that. It seems it only used to keep value of
> `MAX_PATH` and then zeroed in case of unsuccessful allocation. At
> which part of the code are you talking about?

I am not sure either.

I am not very familiar with the code nor Windows APIs.  I do not know
how the methods are called, in which sequence or when, so I might have
misunderstood.  In general, I failed to understand the value of the
ap_data_t type and the ap_data variable, and why it could not simply be
a char * or a std::string, but did my best to try to keep the strange
structure around.

Feel free to change my proposal any way you see fit if you have a better
idea.

-- 
Happy hacking
Petter Reinholdtsen

PR #2031, opened 2026-03-31
Only look for old Linux kernel versions on Linux.

petterreinholdtsen:
### Description

This do not work well on GNU Hurd, where the kernel version is 0.9.

This get test_malloc_compliance working on GNU Hurd.

This fix is related to #2027.

### Type of change

- [ x tests - _change in tests_

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/test/common/memory_usage.h
+++ b/test/common/memory_usage.h
@@ -74,7 +74,8 @@ namespace utils {
         peakUsage
     };
 
-#if __unix__
+#if __linux__
+
     inline unsigned LinuxKernelVersion()
     {
         unsigned digit1, digit2, digit3;
@@ -125,8 +126,11 @@ namespace utils {
                 break;
             }
         }
-        // VmPeak is available in kernels staring 2.6.15
-        if (stat != peakUsage || LinuxKernelVersion() >= 2006015)
+        // VmPeak is available in Linux kernels staring 2.6.15
+        if (stat != peakUsage
+#if __linux__
+            || LinuxKernelVersion() >= 2006015)
+#endif
             ASSERT(size, "Invalid /proc/self/status format, pattern not found.");
         fclose(fst);
         return size * 1024;


test/common/memory_usage.h
```diff
@@ -125,9 +126,11 @@ namespace utils {
                 break;
             }
         }
-        // VmPeak is available in kernels staring 2.6.15
+#if !defined(__gnu_hurd__)
+        // VmPeak is available in Linux kernels staring 2.6.15
         if (stat != peakUsage || LinuxKernelVersion() >= 2006015)
```
Alexandr-Konovalov: Logic in the legacy is "size must be non-zero, except for VmPeak and really old Linux". And what is your assumption? Why you think is it reasonable to drop this check in every case?

petterreinholdtsen:
[Alexandr-Konovalov]
> Logic in the legacy is "size must be non-zero, except for VmPeak and
> really old Linux". And what is your assumption? Why you think is it
> reasonable to drop this check in every case?

My assumtion was that this was a check that was intended for Linux only.
I read from your message that only the latter part was for Linux only,
and have changed the commit to better reflect what I suspect is the
change you recommend.

-- 
Happy hacking
Petter Reinholdtsen

Alexandr-Konovalov:
> [Alexandr-Konovalov]
> Logic in the legacy is "size must be non-zero, except for VmPeak and really old Linux". And what is your assumption? Why you think is it reasonable to drop this check in every case?
> My assumtion was that this was a check that was intended for Linux only. I read from your message that only the latter part was for Linux only, and have changed the commit to better reflect what I suspect is the change you recommend.

Your current version assumes that Hurd may not have VmPeak in /proc/*/status. Is it another API go get that info?

petterreinholdtsen:
[Alexandr-Konovalov]
> Your current version assumes that Hurd may not have VmPeak in
> /proc/*/status. Is it another API go get that info?


I do not understand.  The latest edition look like this:

        if (stat != peakUsage
  #if __linux__
            || LinuxKernelVersion() >= 2006015)
  #endif
             ASSERT(size, "Invalid /proc/self/status format, pattern not found.");

I understand it to mean that every platform will check peakUsage, but
only old linux plattforms are allowed to skip the test.  Are we looking
at the same code?

-- 
Happy hacking
Petter Reinholdtsen

Alexandr-Konovalov:
I apologize for the delay in responding.

> I understand it to mean that every platform will check peakUsage, but
only old linux plattforms are allowed to skip the test. 

Yes. And for system where `__linux__`  is not defined we have just
```
        if (stat != peakUsage)
            ASSERT(size, "Invalid /proc/self/status format, pattern not found.");
```
So, for non-Linux `size` is never checked when `stat == peakUsage`. Why is this correct? If there is no `VmPeak` there, can we get peak memory consumption in some different way?

petterreinholdtsen:
I do not know why the old code was correct, nor why the new code is
correct when the linux specific test is removed.  I assumed the original
code was correct on Linux, and would stay correct on non-linux systems
with the linux spcifici test removed.  You seem to believe otherwise,
and I assume you know the code better than me.  Feel free to modify the
code or my proposal in any way you see fit.

-- 
Happy hacking
Petter Reinholdtsen

PR #2032, opened 2026-03-31
Add sub-RFC for interleaving memory allocation

Alexandr-Konovalov:


rfcs/proposed/numa_support/interleaved-allocation.md
```diff
@@ -0,0 +1,66 @@
+# API to allocate memory interleaved between NUMA nodes
+
+*Note:* This document is a sub-RFC of the [umbrella RFC about improving NUMA
+support](README.md). 
+
+## Motivation
+
+There are two kinds of NUMA-related performance bottlenecks: latency increasing due to
+access to a remote node and bandwidth-limited simultaneous access from different CPUs to
+a single NUMA memory node. A well-known method to mitigate both is random distribution of
```
vossmjp:
```suggestion
a single NUMA memory node. A well-known method to mitigate both is controlled placement of
```

The examples you provide in the next sentence are not "random". So perhaps you mean that you need to do something active to ensure memory placement in done in some thoughtful way. Not sure.

rfcs/proposed/numa_support/interleaved-allocation.md
```diff
@@ -0,0 +1,66 @@
...
+API. Generally, the latter requires less overhead.
+
+## Requirements to public API
+
+To perform allocation of large blocks of special kind of memory there is no need to keep
+state, so malloc-like function fits well. There are two possible parameters: `interleaving
+step` and `list of NUMA nodes to perform allocations on`. The function serves as a
+provider of memory blocks with at least page granularity and doesn't employ internal
+caching. So, to support high-performance, smaller and repetitive allocations `std::pmr` or
+other solutions should be used.
```
vossmjp:
```suggestion
Free, stateless functions, similar to malloc, are sufficient for the allocation of large blocks of memory. 
To guide the spreading of blocks across NUMA nodes, two additional parameters are proposed: `interleaving
step` and `list of NUMA nodes to perform allocations on`. This single function then serves as a
provider of memory blocks with at least page granularity and will not employ internal
caching. If high-performance, smaller and repetitive allocations are needed, then `std::pmr` or
other solutions should be used.
```

rfcs/proposed/numa_support/interleaved-allocation.md
```diff
@@ -0,0 +1,66 @@
...
+
+`interleaving step` has page granularity. Currently there are no clear use cases for
+granularity more than page size.
```
vossmjp:
```suggestion
`interleaving step` is the size of the distributed blocks and has page granularity. Currently there are 
no clear use cases for granularity more than page size.
```

rfcs/proposed/numa_support/interleaved-allocation.md
```diff
@@ -0,0 +1,66 @@
...
+
+`list of nodes for allocation` is conceptually set of `tbb::numa_node_id`. However,
```
vossmjp:
```suggestion
`list of nodes for allocation` is conceptually a set of `tbb::numa_node_id`. However,
```

rfcs/proposed/numa_support/interleaved-allocation.md
```diff
@@ -0,0 +1,66 @@
...
+because `tbb::numa_nodes()` returns `std::vector` and creating a `std::set` from it
+requires allocation, `vector` can be used. Because semantics of `tbb::numa_node_id` is
+not defined, we can't use it to construct e.g., a bit mask. Allocation that is unbalanced
+between NUMA nodes doesn't seem to have useful applications, so repeated elements in `list
+of nodes` is an error.
+
+One use case for `list of nodes` argument is desire to run parallel activity on subset of
```
vossmjp:
```suggestion
One use case for `list of nodes` argument is the desire to run parallel activity on subset of
```

rfcs/proposed/numa_support/interleaved-allocation.md
```diff
@@ -0,0 +1,66 @@
...
+nodes and so get memory only from those nodes.
+
+Most common usage of the allocation function is expected only with `size` parameter.
```
vossmjp, line 39: What happens in that case? Is the nodes automatically determined and is by default the full set of numa nodes. Or is the default to use a single node. Similarly with interleaving_step = 0, is the default to automatically select a size, or does this mean allocate on a single numa node?

Alexandr-Konovalov: Sure, I added the clarification.

rfcs/proposed/numa_support/interleaved-allocation.md
Alexandr-Konovalov: I tried to add more details hier.

vossmjp (dismissed):
LGTM

rfcs/proposed/numa_support/interleaved-allocation.md
```diff
@@ -0,0 +1,70 @@
...
+`interleaving step` is the size of the contiguous memory block from a particular NUMA
+node, it has page granularity. Currently there are no clear use cases for granularity more
+than page size.
+
+`list of nodes for allocation` is conceptually a set of `tbb::numa_node_id`. However,
+because `tbb::numa_nodes()` returns `std::vector` and creating a `std::set` from it
+requires allocation, `vector` can be used. Because semantics of `tbb::numa_node_id` is
+not defined, we can't use it to construct e.g., a bit mask. Allocation that is unbalanced
+between NUMA nodes doesn't seem to have useful applications, so repeated elements in `list
+of nodes` is an error.
```
aleksei-fedotov: Since we are unsure regarding repeated numa nodes having useful applications, may be we shouldn't strictly prohibit having them in a vector? If its support will not complicate the implementation, then I would suggest not prohibiting such a use case.
I would also add an open question regarding the use cases for which prohibiting could be beneficial.

akukanov: On the one hand, starting with more restrictive policies allows to relax it later, while restricting a policy is usually a breaking change. But on the other hand, it seems that enforcing the restriction would require some special effort.

If the purpose of this clause is to avoid dealing with potential implementation issues due to repeating elements, we can say that the behavior is undefined in that case. It will allow the implementation to assume that there are no repeating elements, without specifically checking for that and reporting an error. We might still want to do the checks in the debug mode (especially if we are aware of the related issues), but that's an implementation quality question.

Alexandr-Konovalov: That's interesting. Both Microsoft and libnuma have bitmasks here, and so no problems with repeated nodes. Somehow `tbb::numa_nodes()` returns `vector`, and we should think about semantics of repeated elements.

rfcs/proposed/numa_support/interleaved-allocation.md
```diff
@@ -0,0 +1,70 @@
...
+
+Most common usage of the allocation function is expected only with `size` parameter.
+In this case, `interleaving_step` defaults to the page size and memory is allocated on all
+NUMA nodes.
+
+```c++
+void *tbb::numa::alloc_interleaved(size_t size, size_t interleaving_step = 0,
+                                   const std::vector<tbb::numa_node_id> *nodes = nullptr);
+void tbb::numa::free_interleaved(void *ptr, size_t size);
+```
```
akukanov, line 50: The API may need further elaboration, One question is below, but I have more of those. At the same time, I do not want to postpone acceptance of the general proposal.

Do we prefer to accept the RFC now and work on the API in subsequent RFC updates? In that case, maybe it is better to either remove the specific APIs or at least mention that these are conceptual, rather than final.

rfcs/proposed/numa_support/interleaved-allocation.md
Alexandr-Konovalov: > Since we are unsure regarding repeated numa nodes having useful applications, may be we shouldn't strictly prohibit having them in a vector? If its support will not complicate the implementation, then I would suggest not prohibiting such a use case. I would also add an open question regarding the use cases for which prohibiting could be beneficial.

Good point. I'll add the open question.

rfcs/proposed/numa_support/interleaved-allocation.md
```diff
@@ -0,0 +1,70 @@
...
+least page granularity and will not employ internal caching. If high-performance, smaller
+and repetitive allocations are needed, then `std::pmr` or other solutions should be used.
+
+`interleaving step` is the size of the contiguous memory block from a particular NUMA
+node, it has page granularity. Currently there are no clear use cases for granularity more
+than page size.
+
+`list of nodes for allocation` is conceptually a set of `tbb::numa_node_id`. However,
+because `tbb::numa_nodes()` returns `std::vector` and creating a `std::set` from it
+requires allocation, `vector` can be used. Because semantics of `tbb::numa_node_id` is
```
akukanov: Ideally, I would use `std::span` but it is not available till C++20.
But taking a pointer to a const vector also looks strange in the API, while its only purpose is to support the default argument. Should we instead use a C-style "pointer and size" semantics? Or have two overloads, with and without the vector?

rfcs/proposed/numa_support/interleaved-allocation.md
akukanov: > Both Microsoft and libnuma have bitmasks here, and so no problems with repeated nodes.

I think that depends on the meaning of the masks and the use of the system APIs. For example, the list `{0, 1, 0, 2}` seems to indicate that every second chunk should go to the node 0, not just that the node 0 is in the set like the other two nodes.

Alexandr-Konovalov: Yes, exactly such semantics is expected. And [0, 1, 0, 2] is list, not set. 

I think bitmasks always mean compact representation of sets. In case of libnuma that's

> A struct bitmask controls a bit map of arbitrary  length containing a bit representation of nodes.

And from Microsoft

> A processor mask is a bit vector in which each bit represents a processor and whether it is in the node.

rfcs/proposed/numa_support/interleaved-allocation.md
Alexandr-Konovalov: Sadly, just-happens-to-be-standard `const std::vector<> &` for empty vector requires comparison with zero after return from `alloc_interleaved()`. And I found no arguments against version with pointer rather then `customer hates to write & in each non-default call.` 

Yet another overload looks like less evil wrt "pointer and size". But we have 2 parameters with default values. I;m not sure which one is "more default".

akukanov: Since this is only one API related question and there are more, I rather prefer a separate follow-up discussion of the API.

rfcs/proposed/numa_support/interleaved-allocation.md
akukanov: So, do you think it is feasible to implement such semantics with these system APIs. or would that be rather expensive in some way and so we should better stick to the "no repetition" rule at least for now?

Alexandr-Konovalov: It's not expected to be expensive and as it can be potentially used to load balancing, I think we may want to implement it.

rfcs/proposed/numa_support/interleaved-allocation.md
akukanov:
````suggestion
The following functions are provided to illustrate the conceptual API, not yet as the recommended new API.

```c++
void *alloc_interleaved(size_t size, size_t interleaving_step = 0,
                         const std::vector<tbb::numa_node_id> *nodes = nullptr);
void free_interleaved(void *ptr, size_t size);
```
````

Comment (Alexandr-Konovalov) [reply rfcs/proposed/numa_support/interleaved-allocation.md:50]: It seems discussing with names is simpler then with prose, so I'm for marking them conceptual.

File: rfcs/proposed/numa_support/interleaved-allocation.md [hunk 10]
```diff
@@ -0,0 +1,78 @@
...
+API. Generally, the latter requires less overhead.
+
+## Requirements to public API
+
+Free, stateless functions, similar to malloc, are sufficient for the allocation of large
+blocks of memory. To guide the spreading of blocks across NUMA nodes, two additional
+parameters are proposed: `interleaving step` and `list of NUMA nodes to perform
+allocations on`. This single function then serves as a provider of memory blocks with at
+least page granularity and will not employ internal caching. If high-performance, smaller
+and repetitive allocations are needed, then `std::pmr` or other solutions should be used.
```
akukanov: Trying to align the wording with my understanding of the requirements:
```suggestion
A free stateless function, similar to malloc, is sufficient for the allocation of large
blocks of memory, contiguous in the address space. To guide the mapping of memory
across NUMA nodes, two additional parameters are proposed: `interleaving step`
and `the list of NUMA nodes to get the memory from`. This function allocates whole
memory pages and does not employ internal caching. If smaller and repetitive allocations
are needed, then `std::pmr` or other solutions should be used.
```

Alexandr-Konovalov: > spreading of physical memory across NUMA nodes

It's virtual memory, not physical, right?

akukanov: It depends on the viewpoint, I guess - but after re-reading the suggestion, I agree it was not quite clear. Rephrased now.

rfcs/proposed/numa_support/interleaved-allocation.md
```diff
@@ -0,0 +1,78 @@
...
+Most common usage of the allocation function is expected only with `size` parameter.
+In this case, `interleaving_step` defaults to the page size and memory is allocated on all
+NUMA nodes.
+
+The following functions are provided to illustrate the conceptual API, not yet as the
+recommended new API.
+
+```c++
+void *alloc_interleaved(size_t size, size_t interleaving_step = 0,
+                        const std::vector<tbb::numa_node_id> *nodes = nullptr);
```
aleksei-fedotov, line 48: I saw the discussion regarding API and have seen that RFC now illustrates rather mandates the interface, but I still believe we can write:
```suggestion
void *alloc_interleaved(size_t size, size_t interleaving_step = 0,
                        const std::vector<tbb::numa_node_id>& nodes = tbb::info::numa_nodes());
```
As far as I remember, `tbb::info::numa_nodes()` cannot return an empty vector, but rather vector containing only `-1`. However, in either case it seems that the implementation may just fallback onto NUMA-unspecified one time allocation call.

Alexandr-Konovalov: This version of default argument leads to malloc/free on each `alloc_interleaved(16*1024)` call. That may be negligible, because of syscall internally. But could we eliminate excessive allocations?

akukanov: The default argument might be an empty vector; I believe (and we can check) that the major standard library implementations avoid allocation in that case.

But I think two overloads would be "cleaner" and perhaps a bit more efficient than any kind of default argument.

Alexandr-Konovalov: > The default argument might be an empty vector; I believe (and we can check) that the major standard library implementations avoid allocation in that case.

No allocation, but check against zero after returning from the function, because `const &` can became non-empty and we must call dtor.

So, set of overloads probably is less evil.

akukanov (approved):
The proposal requires further elaboration of details, but should be accepted as the idea/direction.

Alexandr-Konovalov:
Colleagues @aleksei-fedotov , @vossmjp , could you please look at this?

rfcs/proposed/numa_support/interleaved-allocation.md
aleksei-fedotov, line 48: I wonder what parameter with default argument is going to be specified more often. Can it be the case when users specify different NUMA nodes more often than changing the default interleaving step? In this case, it is reasonable to have NUMA list to be the second function parameter, i.e. the first parameter with default argument, and have `interleaving_step` to be the last one.

rfcs/proposed/numa_support/interleaved-allocation.md
```diff
@@ -0,0 +1,78 @@
...
+allocation, but those nodes are loaded equally. Having `vector` allows us to express an
+unbalanced load. Example: allocation over the list of nodes [3, 0, 3] uses 2/3 memory from
+node 3 and 1/3 from node 0.
+
+One use case for `list of nodes` argument is the desire to run parallel activity on subset
+of nodes and so get memory only from those nodes.
+
+Most common usage of the allocation function is expected only with `size` parameter.
+In this case, `interleaving_step` defaults to the page size and memory is allocated on all
+NUMA nodes.
```
aleksei-fedotov, line 41: I believe using a named constant instead of a zero as the default argument can improve code readability.
Something like 
```cpp
const size_t page_size = []() {
    // Find out what page size is and return it
    size_t found_page_size = ...;
    return found_page_size;
};
```
Then we can specify it as the default value for `interleaving_step` parameter:
```cpp
void *alloc_interleaved(size_t size, size_t interleaving_step = page_size,
                        const std::vector<tbb::numa_node_id> *nodes = nullptr);
```

aleksei-fedotov (approved):
A couple more comments from my side - just not to forget about them when the design is further elaborated.

rfcs/proposed/numa_support/interleaved-allocation.md
Alexandr-Konovalov: Can we see it as that? No named constants, and no need to define which argument is more often has non-default values.

```c++
TBB_EXPORT void *__TBB_EXPORTED_FUNC alloc_interleaved(size_t size, size_t interleaving_step,
                        const tbb::detail::d1::numa_node_id *nodes, size_t nodes_count);

inline void *alloc_interleaved(size_t size) {
    return alloc_interleaved(size, 0, nullptr, 0);
}

inline void *alloc_interleaved(size_t size, const std::vector<tbb::detail::d1::numa_node_id> &nodes) {
    return alloc_interleaved(size, 0, nodes.data(), nodes.size());
}

inline void *alloc_interleaved(size_t size, size_t interleaving_step) {
    return alloc_interleaved(size, interleaving_step, nullptr, 0);
}

inline void *alloc_interleaved(size_t size, size_t interleaving_step,
                               const std::vector<tbb::detail::d1::numa_node_id> &nodes) {
    return alloc_interleaved(size, interleaving_step, nodes.data(), nodes.size());
}

```

aleksei-fedotov: I think this is a good proposal! I like the part where we don't need to define which argument is more often used as all the variants are available.

When I look at this, only two disadvantages come to my mind, which might not be important to address at all:
1. It is not clear what `0` and `nullptr` mean. In this case, named constants that are not even required to be exposed to the user, can help as they will be used both in the interface and implementation part. They are only needed to a code reader to keep the intentions clear.
2. This approach might lead to exponentially growing number of functions. If we would ever need one more parameter to this function for some reason, we will end up with eight different functions. However, this is not of a big deal as addition of one more parameter probably never happens.

Alexandr-Konovalov: Thank you, good points.

rfcs/proposed/numa_support/interleaved-allocation.md
akukanov: With two overloads, either parameter can be omitted.

PR #2033, opened 2026-03-31
Adjusted test_lightweight to avoid zero threads in test.

petterreinholdtsen:
### Description 

When the hardware only support one core, hardware_concurrency() return 1, which when deviced by 2 using integer division end up as 0, a useless number of threads.  Instead, ensure it end up as 1, while the other return values still end up with a sensible number by changing (i/2) to ((i+1)/2).

This fix is related to #2027.

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown


kboyarinov (approved):
LGTM

PR #2040, opened 2026-04-01
Add RFC for using optional-like types with TBB

kboyarinov:
### Description 

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown

### DIFF
--- a/rfcs/proposed/tbb_optional/tbb_optional.md
+++ b/rfcs/proposed/tbb_optional/tbb_optional.md
@@ -1,4 +1,4 @@
-# ``tbb::optional`` and **Optional-Like** Named Requirement
+# Optional Semantics for oneTBB
 
 ## Table of Contents
 
@@ -7,19 +7,16 @@
   * 1.2 [Default Constructible Requirements on Flow Graph Node Inputs](#default-constructible-requirements-on-flow-graph-node-inputs)
   * 1.3 [Message Filtering for Flow Graph Nodes](#message-filtering-for-flow-graph-nodes)
 * 2 [Concrete Proposal](#concrete-proposal)
-  * 2.1 [`tbb::optional` API](#tbboptional-api)
-    * 2.1.1 [Functionality of C++17 `std::optional` Excluded from `tbb::optional`](#functionality-of-c17-stdoptional-excluded-from-tbboptional)
-    * 2.1.2 [C++17 Behavior](#c17-behavior)
-  * 2.2 [**Optional-Like** Named Requirement](#optional-like-named-requirement)
-  * 2.3 [`tbb::flow_control` deprecation](#tbbflow_control-deprecation)
+  * 2.1 [**Optional-Like** Named Requirement](#optional-like-named-requirement)
+  * 2.2 [Optional-Like API in oneTBB](#optional-like-api-in-onetbb)
+  * 2.3 [`tbb::flow_control` Deprecation](#tbbflow_control-deprecation)
 * 3 [Possible Additional Applications to Consider](#possible-additional-applications-to-consider)
-* 4 [Open Questions](#open-questions)
 
 ## Motivation
 
 ``std::optional`` is one of the most impactful Standard Library types introduced in modern C++. It represents a 
-fundamental concept of a "value that might not be present" and is widely used to replace ad-hoc conventions like
-sentinel values, using ``std::pair<T, bool>`` as a return type and others.
+fundamental concept of a "value that might not be present" and is widely used to replace ad-hoc conventions such as
+sentinel values, ``std::pair<T, bool>`` return types, and similar patterns.
 
 oneTBB supports C++11 as a minimum C++ Standard version, which means ``std::optional`` is not directly available
 across the entire support matrix. This forces the library to implement internal workarounds where optional-like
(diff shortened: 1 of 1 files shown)


rfcs/proposed/tbb_optional/tbb_optional.md
```diff
@@ -0,0 +1,392 @@
...
+
+This API requires the user to construct a throw-away value when signaling stop. TBB runtime guarantees
+to discard it if ``flow_control`` was stopped, but a valid output must still be provided, which is
+unintuitive and error-prone.
+
+In some cases, such an API implicitly requires the output type to be default constructible (although not
+formally required) since a default constructed object is the intuitive choice for signaling end-of-input when
+no sentinel is available. Non-default-constructible output types require contrived workarounds.
+
+With ``std::optional``, this same body becomes self-documenting:
```
isaevil, line 52: I think it is worth mentioning, how would the body of one-filter pipeline look like. Something like `optional<continue_msg>` could be used in this case:
```c++
std::optional<tbb::flow::continue_msg> operator()() {
    if (end-of-input) {
        return {};
    }
    return tbb::flow::continue_msg{};
}
```
Perhaps moving `continue_msg` out of `flow` should be considered.

vossmjp: I'm not sure I get the point about `continue_msg`? Are you suggesting to move `continue_msg` out of the flow namespace or that we can drop  `continue_msg` completely?

kboyarinov: As far as I understand, the point is that Optional-Like cannot be used for single pipeline filter (both input and output are `void`):

```cpp
void single_filter_body() {
     // Return type is void, how to signal end-of-pipeline?
}
```

The idea is to use `continue_msg` as a special flag class in this case and return non-empty `optional<continue_msg>` if the pipeline should be continued and an empty optional if the pipeline should be stopped. 

Since `continue_msg` is part of `flow` namespace now, the idea is to move it to `tbb` namespace. However, since we need to keep the old code that uses `tbb::flow::continue_msg`, I guess having it in both `tbb` and `tbb::flow` is a better option.

An alternative is to keep `flow_control` for single-filter (or even for entire `parallel_pipeline`) and to migrate to Optional-Like only for `input_node`.

vossmjp: Ok, I see now. For this use, the semantics of `continue_msg` is different. Usually, it means that the predecessor-successor dependence has been satisfied between two nodes. If used to signal the continue/stop condition for a single filter pipeline, it would be acting as-if there is an imaginary back edge from the single filter to itself. You call it a "special flag class". I wonder if any savings from eliminating an extra class is worth the possible sematic confusion?

rfcs/proposed/tbb_optional/tbb_optional.md
```diff
@@ -0,0 +1,392 @@
...
+Library internal usage of ``flow_control`` suppresses the warning.
+
+## Possible Additional Applications to Consider
+
+Some other TBB APIs can provide optional-aware alternatives. For example, ``try_pop(T&)`` in
+``concurrent_queue`` and ``concurrent_priority_queue`` could return ``tbb::optional``.
+Also basic Flow Graph ``sender`` class functions ``try_get(T&)`` and ``try_reserve(T&)`` could
+return ``tbb::optional`` (results in ABI break since the functions are virtual).
+
+## Open Questions
```
aleksei-fedotov, line 381: I wonder if `tbb::optional` is needed to be introduced at all. If it is introduced then it will be needed to be deprecated someday, similar to deprecation of `flow_control` proposed here, `src_node` in the past and even `tbb::atomic` if the current support model was used back that time. Also, introducing `tbb::optional` requires its own documentation, tests, check for necessary compatibility with `std::optional`. In short, lots of additional work and maintenance.

What if we go a bit different way - what if the use of `std::optional` will simply be allowed when it is available and is used in the proposed places? Is `tbb::optional` really necessary for someone in 2026?

kboyarinov: Yes, this is also an option we can consider. 
But the main concern is that `std::optional` is a C++17 feature, while we support C++11 as a minimal version. This means we cannot just introduce new named requirement that uses it and deprecate the `flow_control` class. We will have to support `flow_control` alternative for C++11/14 until we switch to C++17 as a minimal version. 

For new APIs (like FG message filters and other ideas that can benefit from optional semantics), either C++17 should be the minimal standard version, or an alternative should be provided. BTW, `flow_control` does not sound like a natural fit for this purpose.

Also, I feel that we will need to keep Optional-Like semantics in named requirements to allow user or 3-party types to be used instead, like `boost::optional`.

rfcs/proposed/tbb_optional/tbb_optional.md
kboyarinov: I think we can simply use a boolean flag instead - return true if the pipeline should be continued and false otherwise.

I have also rewritten a proposal in a step-by-step manner. First of all, I propose to allow Optional-Like for first-filter body and input_node. After that, provide our own Optional-Like object if we want to. 
And only after that, consider deprecating the flow-control. Until it's deprecation, I believe no replacement is required for a single-filter body.

PR #2043, opened 2026-04-02
Add RFC about C++20 modules support

isaevil:
### Description 

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown


rfcs/proposed/cxx20_modules/README.md
```diff
@@ -0,0 +1,199 @@
...
+    using tbb::parallel_for_each;
+
+   // other TBB API
+}
+```
+
+The proposed approach has several advantages over other alternatives:
+
+- It does not require the modification of existing headers, as the module includes them as
+  part of global module fragment and then selectively export the necessary `using`
```
vossmjp:
```suggestion
  part of a global module fragment and then selectively exports the necessary `using`
```

rfcs/proposed/cxx20_modules/README.md
```diff
@@ -0,0 +1,199 @@
...
+  declarations. Thus, the development can continue in headers.
```
vossmjp:
```suggestion
  declarations. Thus, development can continue in headers.
```

rfcs/proposed/cxx20_modules/README.md
```diff
@@ -0,0 +1,199 @@
...
+- Since the API is declared as part of global module fragment, the name of the module will
```
vossmjp:
```suggestion
- Since the API is declared as part of a global module fragment, the name of the module will
```

rfcs/proposed/cxx20_modules/README.md
```diff
@@ -0,0 +1,199 @@
...
+  not participate in the name mangling, making translation units compiled with modules,
```
vossmjp:
```suggestion
  not participate in name mangling, making translation units compiled with modules,
```

rfcs/proposed/cxx20_modules/README.md
```diff
@@ -0,0 +1,199 @@
...
+  compatible with the units compiled with regular headers.
+
+The disadvantage of this approach is an obligation to update the module interface unit each time new API is added.
```
vossmjp:
```suggestion
The disadvantage of this approach is an obligation to update the module interface unit each time a new API is added.
```

rfcs/proposed/cxx20_modules/README.md
```diff
@@ -0,0 +1,199 @@
...
+
+### CMake integration
+
+The library must remain at C++11 as the minimum standard, hence the module source cannot be
```
vossmjp: "must remain at C++11" is because of a project policy, not a technical reason. We may for example later, change to C++17 as the minimum standard as a policy change. As written, its unclear if you are stating a design constraint or a technical constraint.

rfcs/proposed/cxx20_modules/README.md
```diff
@@ -0,0 +1,199 @@
...
+compiled as part of the oneTBB build. Additionally, currently the built module constrains
+the consumer to use the specific version of the C++ standard: the one that was used during
```
vossmjp:
```suggestion
the consumer to use a specific version of the C++ standard: the one that was used during
```

rfcs/proposed/cxx20_modules/README.md
isaevil: Added more details.

rfcs/proposed/cxx20_modules/README.md
```diff
@@ -0,0 +1,201 @@
...
+  and preprocessing of header files.
+- Modules provide better encapsulation — macros, internal declarations, and implementation
+  details do not leak into the consumer's translation unit.
+
+As the C++ ecosystem gradually adopts modules, oneTBB should provide a module interface to
+allow users to use the library using `import tbb;` instead of `#include <oneapi/tbb.h>`.
+
+## Proposal
+
+### ABI non-breaking style with using-declaration
```
kboyarinov, line 18: I would also add a link to LLVM manual somewhere in this paragraph

rfcs/proposed/cxx20_modules/README.md
```diff
@@ -0,0 +1,201 @@
...
+   Partitions are internal to the module and not importable by consumers, but could
+   improve maintainability of the module interface. Alternatively, should multiple
+   fine-grained submodules (e.g., `tbb.flow_graph`, `tbb.containers`) be provided so
+   that consumers can import only what they need?
+
+5. How should module-based consumption be tested? Options include a dedicated test
+   that uses `import tbb;` instead of `#include`, or running the existing test suite
+   with module imports. Should whitebox tests be covered in the latter scenario?
+
+6. How to provide preview functionality with modules? Should it be a separate `tbb.preview` module
```
kboyarinov: Also, the important details about the preview features and feature-test macros are missed from the document. 
For the preview feature, the only way to enable them is to compile the source with -DTBB_PREVIEW..., defining before `import` does not work.

For feature-test macros, they are simply not available from the module. Initially, they can be available by `#include <tbb/version.h>` header. Core functionality is available using `import tbb`. 
We can consider providing exported functions replacing these macros in the future.

isaevil: > For the preview feature, the only way to enable them is to compile the source with -DTBB_PREVIEW...

That is what I meant by "should the `tbb` module be compiled with defined `TBB_PREVIEW_*` macros as needed?" in line below.

> For feature-test macros, they are simply not available from the module. Initially, they can be available by #include <tbb/version.h> header. Core functionality is available using import tbb.
We can consider providing exported functions replacing these macros in the future.

I think I mostly cover this in 7th question. But I'll add a little more details as well as for question 6.

rfcs/proposed/cxx20_modules/README.md
isaevil: Done.

rfcs/proposed/cxx20_modules/README.md
isaevil: Done.

rfcs/proposed/cxx20_modules/README.md
```diff
@@ -0,0 +1,229 @@
...
+   and a macro defined by the consumer before `import tbb;` cannot influence the module's
+   already-compiled interface. Should preview functionality be offered as a dedicated `tbb.preview`
+   module, or should the consumer compile the `.cppm` with the desired `TBB_PREVIEW_*` macros defined?
+
+7. How to support public macros such as `TBB_VERSION` or feature-test macros?
+   Modules do not export preprocessor definitions, so these macros are not visible
+   after `import tbb;`. Some options:
+   - Replace them with inline variables where possible or provide exported functions.
+   - Require the consumer to `#include <tbb/version.h>` alongside the import.
+
```
mropert, line 221: I would add a question about plans to handle transitive inclusion of the STL.
Since modules tend to break down when some project use the STL through `#include` directives and other through `import std` this should be considered.
I would recommend either toggling all `#include` into `import` when using TBB as a module, or at least provide a define to toggle between the two options.

isaevil: Do you mean that, since declarations in `std` module are attached to "unspecified" module, ABI compatibility may be broken when `std` module and standard headers are used in the same application?

mropert: Most (all?) compilers will support something like this:
```cpp
#include <vector>
import std;
```
But not this:
```cpp
import std;
#include <vector>
```
This is also true if the includes/imports happens transitively by import/including files that themselves do import/includes.
So if people do `import tbb`, then need to be able to control how transitive inclusion of the STL by TBB will be done. The most common I have seen is that when in module mode all `#include` should be turned into `import` directives, but if you're worried this is too drastic you can provide a define that allow people to pick a strategy.

vossmjp (dismissed):
LGTM

rfcs/proposed/cxx20_modules/README.md
isaevil: It seems more like a compiler implementation limitation, not limitation of TBB named module itself, so I wouldn't say it is an open question for supporting modules in TBB. But we'll keep that issue in mind for sure, thanks.

rfcs/proposed/cxx20_modules/README.md
```diff
@@ -0,0 +1,231 @@
...
+- Same cons as for ABI non-breaking style described above.
+
+## Open Questions
+
+1. Should the module be split into partitions (e.g., `tbb:algorithms`, `tbb:containers`,
+   `tbb:flow_graph`) to organize the `using`-declarations?
+   Partitions are internal to the module and not importable by consumers, but could
+   improve maintainability of the module interface. Alternatively, should multiple
+   fine-grained submodules (e.g., `tbb.flow_graph`, `tbb.containers`) be provided so
+   that consumers can import only what they need?
```
akukanov: I would separate the alternative (multiple modules) as a first-order question - I think it is more important than internal organization of declarations within the file.

With regard to that alternative - is it possible to combine both a single global module and submodules, and is it possible to add submodules later without breaking user code?

rfcs/proposed/cxx20_modules/README.md
```diff
@@ -0,0 +1,231 @@
...
+
+2. How should module-based consumption be tested? Options include a dedicated test
+   that uses `import tbb;` instead of `#include`, or running the existing test suite
+   with module imports. Should whitebox tests be covered in the latter scenario?
+    * Currently extending `test_tbb_header` is suggested to check public API availability via modules.
```
akukanov: Are there usage scenarios that are not covered by the suggested approach with `test_tbb_header`?
If the only practical difference is for using `import` instead of `#include`, then the API availability checks seem sufficient.

Also, is there any practical way to check that only the desired public APIs are exported, and nothing else?

rfcs/proposed/cxx20_modules/README.md
```diff
@@ -0,0 +1,231 @@
...
+
+3. How to provide preview functionality with modules? TBB preview features that extend existing
+   classes would create distinct types in a `tbb.preview` module, incompatible with the stable
+   `tbb` module, meaning that they cannot be used together.
```
akukanov: I do not quite understand what the sentence following the question mark represents. Is it a requirement, a future-looking proposal, a design limitation, or a combination of those?

rfcs/proposed/cxx20_modules/README.md
```diff
@@ -0,0 +1,231 @@
...
+2. How should module-based consumption be tested? Options include a dedicated test
+   that uses `import tbb;` instead of `#include`, or running the existing test suite
+   with module imports. Should whitebox tests be covered in the latter scenario?
+    * Currently extending `test_tbb_header` is suggested to check public API availability via modules.
+
+3. How to provide preview functionality with modules? TBB preview features that extend existing
+   classes would create distinct types in a `tbb.preview` module, incompatible with the stable
+   `tbb` module, meaning that they cannot be used together. 
+    * Current workaround is to require the consumer to compile `tbb.cppm` with desired
+    `TBB_PREVIEW_*` macros defined.
```
akukanov, line 215: Certainly good enough to begin with.

akukanov (dismissed):
The RFC looks good for initial acceptance and implementation. Open questions can be additionally clarified though.

rfcs/proposed/cxx20_modules/README.md
isaevil: > With regard to that alternative - is it possible to combine both a single global module and submodules, and is it possible to add submodules later without breaking user code?

I believe both are possible as long as exported declarations are attached to global module fragment (all declarations of the same entity must be attached to the same module) and ODR stays intact.

rfcs/proposed/cxx20_modules/README.md
isaevil: > Are there usage scenarios that are not covered by the suggested approach with test_tbb_header?
If the only practical difference is for using import instead of #include, then the API availability checks seem sufficient.

I would say perhaps ABI compatibility check (combining TUs compiled with modules and headers) is not implemented currently, but my understanding is that can be done the similar way as with `test_tbb_header_secondary.cpp`.

> Also, is there any practical way to check that only the desired public APIs are exported, and nothing else?

For now I can only think of marking somehow in headers what we believe is a public API (we usually have `inline`-namespace at the bottom of header with `using`-declarations) and auto-generating a module from that and comparing it with `tbb.cppm`.

rfcs/proposed/cxx20_modules/README.md
isaevil: I think I skipped an extra sentence about "what if we had a separate `tbb.preview` module", but essentially I am talking about a design limitation here: we have preview features that instead of adding new API actually extend the existing classes and it is unclear how `tbb` and `tbb.preview` modules can co-exist in such scenarios. If they can't, then there is no actual value of providing a separate module since `tbb.cppm` can be compiled with preview features enabled anyway.

isaevil: Seems like I accidentally removed some details that were here before. I've returned them back

rfcs/proposed/cxx20_modules/README.md
isaevil: > I would separate the alternative (multiple modules) as a first-order question - I think it is more important than internal organization of declarations within the file.

Separated these questions.

rfcs/proposed/cxx20_modules/README.md
akukanov: > I would say perhaps ABI compatibility check (combining TUs compiled with modules and headers) is not implemented currently, but my understanding is that can be done the similar way as with test_tbb_header_secondary.cpp.

Should we then add this to the testing plan and close the question, or are there more nuances?

Running the whole test suite with module imports instead of includes would check if the functionality is not just accessible but also properly usable. However, based on what we know of the way modules work and the internals of our implementation, would that add sufficient value? I feel like it would not.

rfcs/proposed/cxx20_modules/README.md
mropert: I'm pretty sure it's a limitation of how modules work in practice (and maybe even on spec, that part isn't entirely clear).

Either way, it's defintely a question you'll have to solve because else you'll get constant complaints from module users.

rfcs/proposed/cxx20_modules/README.md
isaevil: > Running the whole test suite with module imports instead of includes would check if the functionality is not just accessible but also properly usable. However, based on what we know of the way modules work and the internals of our implementation, would that add sufficient value? I feel like it would not.

Yeah, we reuse headers, where actual implementation lives, so running the whole testing suite but with `import tbb;` won't tell us much besides "test still runs". @kboyarinov any thoughts?

kboyarinov: In general, I agree. If we know for sure that the availability of the exported API is tested by test_tbb_header, each public API is present in both the module and the test, and we know how the module is implemented, then the test_tbb_header approach is sufficient.

The only value I can imagine in having full testing is catching scenarios where something is missed, for example when a new public API is not added to the module file and to the test. In this case, the regular test for this API in "import mode" will detect the error. 

However, providing pure "import mode" for our tests can be challenging. We will need to make sure import is the only way TBB is provided to the test. If we have both import and include (even from standard headers), the guarantees and the value I described above can be undermined. So I am not sure it pays off.

isaevil: I've removed the question about testing and updated "Testing" section.

akukanov (approved):

kboyarinov (approved):
LGTM

PR #2049, opened 2026-04-08
Document feature-test macros

kboyarinov:
### Description 
Add a table illustrating feature-test macros.
WARNING: The PR should be merged only once the PRs with features documentation are merged to avoid broken links

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown

### DIFF
--- a/doc/main/reference/examples/feature_test_macros.cpp
+++ b/doc/main/reference/examples/feature_test_macros.cpp
@@ -29,6 +29,8 @@ using parallel_loop2_body = parallel_loop_body;
 
 /*begin_feature_test_macros_example*/
 
+#define TBB_PREVIEW_PARALLEL_PHASE 1
+
 #include <oneapi/tbb/version.h>
 #include <oneapi/tbb/parallel_for.h>
 
--- a/doc/main/reference/examples/feature_test_macros.cpp
+++ b/doc/main/reference/examples/feature_test_macros.cpp
@@ -1,0 +1,55 @@
+/*
+    Copyright (c) 2026 UXL Foundation Contributors
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+
+int parallel_loop1_begin = 0;
+int parallel_loop1_end = 100;
+
+int parallel_loop2_begin = 100;
+int parallel_loop2_end = 1000;
+
+struct parallel_loop_body {
+    void operator()(int) const {}
+};
+
+using parallel_loop1_body = parallel_loop_body;
+using parallel_loop2_body = parallel_loop_body;
+
+/*begin_feature_test_macros_example*/
+
+#include <oneapi/tbb/version.h>
+#include <oneapi/tbb/parallel_for.h>
+
+#if TBB_HAS_PARALLEL_PHASE
+#include <oneapi/tbb/task_arena.h>
+#endif
+
+int main() {
+#if TBB_HAS_PARALLEL_PHASE
+    tbb::this_task_arena::start_parallel_phase();
+#endif
+
+    tbb::parallel_for(parallel_loop1_begin, parallel_loop1_end,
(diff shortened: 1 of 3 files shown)


doc/main/reference/examples/feature_test_macros.cpp
```diff
@@ -0,0 +1,55 @@
...
+
+using parallel_loop1_body = parallel_loop_body;
+using parallel_loop2_body = parallel_loop_body;
+
+/*begin_feature_test_macros_example*/
+
+#include <oneapi/tbb/version.h>
+#include <oneapi/tbb/parallel_for.h>
+
+#if TBB_HAS_PARALLEL_PHASE
```
vossmjp, line 37: Since parallel_phase is currently a preview feature, it will not be available unless the preview macro is also set, which it isn't in this example. Is that intentional?  Unfortunately, there are no feature test macros that currently guard non-preview features; otherwise, I'd suggest demonstrating with one of those.

doc/main/reference/feature_test_macros.rst
```diff
@@ -0,0 +1,65 @@
...
+
+Description
+***********
+
+|short_name| defines a set of preprocessor macros corresponding to the features
+provided by the library. They are intended for detecting the presence of these features.
+
+Each of these macros is defined in the header ``<oneapi/tbb/version.h>`` and in the feature headers
+specified in the table below.
+
```
vossmjp, line 18: Maybe add a paragraph explaining the relationship between preview features and feature test macros. For example, something like, "For preview features, the feature test macro will only be defined if the preview feature macro has been defined. You cannot use a feature test macro to guard the setting of a feature's preview macro."

And then maybe give a counter example like:

```
#include <oneapi/tbb/version.h>
#if TBB_HAS_FEATURE_X
#define TBB_PREVIEW_FEATURE_X
#include <feature_header.h>
#endif
```

doc/main/reference/examples/feature_test_macros.cpp
isaevil: Yes, it is missing by accident. I added missing preview macro.

doc/main/reference/feature_test_macros.rst
isaevil: Added suggested paragraph and also an example of incorrect/correct usage.

vossmjp (dismissed):
LGTM

doc/main/reference/examples/feature_test_macros.cpp
```diff
@@ -0,0 +1,57 @@
...
+#endif
+
+    tbb::parallel_for(parallel_loop1_begin, parallel_loop1_end,
+                      parallel_loop1_body{});
+
+    tbb::parallel_for(parallel_loop2_begin, parallel_loop2_end,
+                      parallel_loop2_body{});
+
+#if TBB_HAS_PARALLEL_PHASE
+    tbb::this_task_arena::end_parallel_phase(/*with_fast_leave=*/true);
```
isaevil, line 53: It turns out that `tbb::this_task_arena::end_parallel_phase(bool with_fast_leave)` misses the default value for `with_fast_leave` parameter, which is actually mentioned in [reference](https://github.com/uxlfoundation/oneTBB/blob/master/doc/main/reference/parallel_phase_for_task_arena.rst?plain=1#L154). I'll fix that in a separate PR and it shouldn't be a blocker for PR also.

doc/main/reference/feature_test_macros.rst
```diff
@@ -0,0 +1,84 @@
...
+For preview features, the feature test macro will only be defined if the preview feature macro has
+been defined. You cannot use a feature test macro to guard the setting of a feature's preview macro.
```
akukanov:
```suggestion
For preview features, the feature test macro are only defined if the feature is enabled by defining its
preview macro. You cannot use a feature test macro to guard setting of the feature preview macro.
```

doc/main/reference/feature_test_macros.rst
```diff
@@ -0,0 +1,84 @@
...
+Description
+***********
+
+|short_name| defines a set of preprocessor macros corresponding to the features
+provided by the library. They are intended for detecting the presence of these features.
+
+Each of these macros is defined in the header ``<oneapi/tbb/version.h>`` and in the feature headers
+specified in the table below.
+
+For preview features, the feature test macro are only defined if the feature is enabled by defining its
```
aleksei-fedotov:
```suggestion
For preview features, the feature test macro is only defined if the feature is enabled by defining its
```

akukanov (dismissed):

vossmjp (dismissed):
LGTM again

isaevil (approved):

PR #2052, opened 2026-04-10
Support IPO build for modern Intel compilers with old cmake under Linux

Alexandr-Konovalov:
### Description 
Previously, CMAKE_CXX_COMPILER_AR was not set for them and IPO detection failed because of that.

### Type of change

### Tests

- [X] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [X] No
- [ ] Unknown

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -249,28 +249,6 @@ foreach(FILE_WITH_EXTRA_TARGETS ${FILES_WITH_EXTRA_TARGETS})
     include(${FILE_WITH_EXTRA_TARGETS})
 endforeach()
 
-# - Enabling LTO on Android causes the NDK bug.
-#   NDK throws the warning: "argument unused during compilation: '-Wa,--noexecstack'"
-# - For some reason GCC does not instrument code with Thread Sanitizer when lto is enabled and C linker is used.
-if (TBB_ENABLE_IPO AND BUILD_SHARED_LIBS AND NOT ANDROID_PLATFORM AND NOT TBB_SANITIZE MATCHES "thread")
-    # Under Linux, ipo detection for 2024+ versions of Intel compilers is supported starting cmake 3.30.1.
-    # In this case and for older cmake versions without CheckIPOSupported, try to enable IPO unconditionally.
-    if (CMAKE_VERSION VERSION_LESS 3.9 OR (LINUX AND CMAKE_VERSION VERSION_LESS 3.30.1
-        AND CMAKE_CXX_COMPILER_ID STREQUAL "IntelLLVM" AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 2023))
-        set(TBB_IPO_FLAGS TRUE)
-    else()
-        cmake_policy(SET CMP0069 NEW)
-        include(CheckIPOSupported)
-        check_ipo_supported(RESULT TBB_IPO_PROPERTY)
-        if (NOT TBB_IPO_PROPERTY)
-            message(WARNING "IPO disabled: not supported by this compiler.")
-        endif()
-    endif()
-    if (TBB_IPO_PROPERTY OR TBB_IPO_FLAGS)
-        message(STATUS "IPO enabled")
-    endif()
-endif()
-
 if (TBB_FILE_TRIM)
     file(RELATIVE_PATH TBB_RELATIVE_BIN_PATH ${CMAKE_CURRENT_BINARY_DIR} ${CMAKE_SOURCE_DIR})
     file(TO_NATIVE_PATH ${CMAKE_SOURCE_DIR} NATIVE_TBB_PROJECT_ROOT_DIR)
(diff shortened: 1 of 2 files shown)


Alexandr-Konovalov:
I'm not sure, is it better to set `CMAKE_CXX_COMPILER_AR` manually for the affected configuration before `check_ipo_supported()` or simply enable it, as in the current patch.

Fix on cmake side is [IntelLLVM: Update toolchain layout for 2024+ versions](https://github.com/Kitware/CMake/commit/ceb58f0ccf6f5bbee531566fbaaa955f1d9de429).

CMakeLists.txt
```diff
@@ -253,15 +253,18 @@ endforeach()
 #   NDK throws the warning: "argument unused during compilation: '-Wa,--noexecstack'"
 # - For some reason GCC does not instrument code with Thread Sanitizer when lto is enabled and C linker is used.
 if (TBB_ENABLE_IPO AND BUILD_SHARED_LIBS AND NOT ANDROID_PLATFORM AND NOT TBB_SANITIZE MATCHES "thread")
-    if (NOT CMAKE_VERSION VERSION_LESS 3.9)
+    # Under Linux, ipo detection for 2024+ versions of Intel compilers is supported starting cmake 3.30.1.
+    # In this case and for older cmake versions without CheckIPOSupported, try to enable IPO unconditionally.
+    if (CMAKE_VERSION VERSION_LESS 3.9 OR (LINUX AND CMAKE_VERSION VERSION_LESS 3.30.1
+        AND CMAKE_CXX_COMPILER_ID STREQUAL "IntelLLVM" AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER 2023))
+        set(TBB_IPO_FLAGS TRUE)
```
aleksei-fedotov: On one hand, there are compiler-specific CMake configurations in `cmake/compilers` subdirectory where enabling of certain compiler features is taking place. On the other hand, the resolution about whether enable or disable IPO is made here. This looks a bit hackish to me because the compiler-specific check is mixed with the CMake-own way of checking existence of compiler-feature.

The way I see it is either CMake's approach or our understanding that compiler should support certain feature  - both constitute criteria for the feature enabling. Therefore, I think a more readable implementation should look like the following high-level pseudocode:
```cmake
    # First, include compiler-specific options where a particular-feature determination
    # is taking place, possibly, disregarding the version of CMake (if needed).
    # Among other flags, the following include sets TBB_COMPILER_IPO_FEATURE_EXISTS to TRUE
    # if that feature is available in a compiler
    include(${TBB_COMPILER_SETTINGS_FILE})

    # Then check with the CMake-own facilities.
    # These are skipped if compiler-specific check above determined the feature
    # existence, which makes CMake configurations work faster, and provides
    # a fallback approach in the same time.
    if (NOT TBB_COMPILER_IPO_FEATURE_EXISTS AND NOT CMAKE_VERSION VERSION_LESS 3.9)
        cmake_policy(SET CMP0069 NEW)
        include(CheckIPOSupported)
        # Check with CMake facilities, guarded by CMake version
        check_ipo_supported(RESULT TBB_CMAKE_IPO_FEATURE_EXISTS)
    endif()
    
    # Then at the place, where a final determination of a feature takes place, write
    if (TBB_COMPILER_IPO_FEATURE_EXISTS OR TBB_CMAKE_IPO_FEATURE_EXISTS)
        # Note the logical OR operation ^^ above, which makes using IPO
        # if it is either determined by manual or CMake check.
        set(TBB_IPO_FLAGS TRUE)
        message(STATUS "IPO enabled")
    else()
        message(WARNING "IPO disabled: not supported by this compiler.")
    endif()
```
I believe such approach ideally should be followed for each compiler feature we are interested in. 

@isaevil @Alexandr-Konovalov What do you think about that?

Alexandr-Konovalov: Probably an important detail here is that we don't check for a compiler feature, but rather working around a bug in not-so-fresh-cmake. Talking about the compiler feature, I hope all Intel compilers support ipo.

It's interesting to ask why we are using `check_ipo_supported` at all. Probably it's just user-friendly to report "ipo is not supported" at configure step, rather then during library building.

> This looks a bit hackish to me because the compiler-specific check is mixed with the CMake-own way of checking existence of compiler-feature.

Why is it "hackish"? It's easy to check for something and go further instead of creating and maintaining static tables with compiler features.

Moving the check "your cmake detection of ipo is broken for current compiler" to compiler settings is probably the right thing. Hope we can move `include(${TBB_COMPILER_SETTINGS_FILE})` up a little.

aleksei-fedotov: > I hope all Intel compilers support ipo.

As far as I know, this can be checked using CMake by compiling dummy program with necessary flags and see if compiler say anything about unrecognized option. I believe CMake's `check_ip_supported()` does something similar internally.

> Why is it "hackish"? It's easy to check for something and go further instead of creating and maintaining static tables with compiler features.

Because a pile of checks for different compilers, their various versions and whatnot in a single place makes things barely readable. If it can be done even more readable then I proposed, then let's do that, as the readability of a code is the most important thing I believe.

Alexandr-Konovalov: > As far as I know, this can be checked using CMake by compiling dummy program with necessary flags and see if compiler say anything about unrecognized option. I believe CMake's check_ip_supported() does something similar internally.

If we want to go this way, it's better to set `CMAKE_CXX_COMPILER_AR` manually and then run `check_ipo_supported()` as usual, as I describe in [1st comment](https://github.com/uxlfoundation/oneTBB/pull/2052#issuecomment-4222931345). I didn't do that and only describe it, because we are sure that ipo is working for Intel compiler, so why spend time checking for it.

CMakeLists.txt
```diff
@@ -289,6 +270,27 @@ else()
     message(WARNING "TBB compiler settings not found ${TBB_COMPILER_SETTINGS_FILE}")
 endif()
 
+# - Enabling LTO on Android causes the NDK bug.
+#   NDK throws the warning: "argument unused during compilation: '-Wa,--noexecstack'"
+# - For some reason GCC does not instrument code with Thread Sanitizer when lto is enabled and C linker is used.
+if (TBB_ENABLE_IPO AND BUILD_SHARED_LIBS AND NOT ANDROID_PLATFORM AND NOT TBB_SANITIZE MATCHES "thread")
+    # For older cmake versions without CheckIPOSupported and when ipo detection is broken,
+    # try to enable IPO unconditionally.
+    if (CMAKE_VERSION VERSION_LESS 3.9 OR DEFINED TBB_IPO_DETECTION_BROKEN)
```
aleksei-fedotov: According to [this documentation section of CMake](https://cmake.org/cmake/help/latest/command/if.html#variable) the following change suits better, because now `TBB_IPO_DETECTION_BROKEN` can also be set to `FALSE` or even undefined with expected consequences.
```suggestion
    if (CMAKE_VERSION VERSION_LESS 3.9 OR TBB_IPO_DETECTION_BROKEN)
```

aleksei-fedotov (approved):

PR #2057, opened 2026-04-20
[tbbmalloc] Avoid disabling 4267 and 4244 warnings

aleksei-fedotov:
4267 - 'var' : conversion from 'size_t' to 'type', possible loss of data.
4244 - 'argument' : conversion from 'type1' to 'type2', possible loss of data.

### DIFF
--- a/src/tbbmalloc/backend.h
+++ b/src/tbbmalloc/backend.h
@@ -1,5 +1,6 @@
 /*
     Copyright (c) 2005-2023 Intel Corporation
+    Copyright (c) 2026 UXL Foundation Contributors
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
--- a/src/tbbmalloc/backend.h
+++ b/src/tbbmalloc/backend.h
@@ -326,7 +326,8 @@ private:
         else if (size < minBinnedSize)
             return NO_BIN;
 
-        int bin = (size - minBinnedSize)/freeBinsStep;
+        size_t val = (size - minBinnedSize)/freeBinsStep;
+        int bin = (int)val;
 
         MALLOC_ASSERT(bin < HUGE_BIN, "Invalid size.");
         return bin;
--- a/src/tbbmalloc/backref.cpp
+++ b/src/tbbmalloc/backref.cpp
@@ -1,5 +1,6 @@
 /*
     Copyright (c) 2005-2023 Intel Corporation
+    Copyright (c) 2026 UXL Foundation Contributors
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
--- a/src/tbbmalloc/backref.cpp
+++ b/src/tbbmalloc/backref.cpp
@@ -40,8 +40,9 @@ struct BackRefBlock : public BlockI {
 
     BackRefBlock(const BackRefBlock *blockToUse, intptr_t num) :
         nextForUse(nullptr), bumpPtr((FreeObject*)((uintptr_t)blockToUse + slabSize - sizeof(void*))),
-        freeList(nullptr), nextRawMemBlock(nullptr), allocatedCount(0), myNum(num),
-        addedToForUse(false) {
+        freeList(nullptr), nextRawMemBlock(nullptr), allocatedCount(0),
+        myNum((BackRefIdx::main_t)num), addedToForUse(false)
+    {
         memset(static_cast<void*>(&blockMutex), 0, sizeof(MallocMutex));
 
         MALLOC_ASSERT(!(num >> CHAR_BIT*sizeof(BackRefIdx::main_t)),
@@ -181,14 +182,15 @@ bool BackRefMain::requestNewSpace()
 
     MallocMutex::scoped_lock lock(mainMutex); // ... and share under lock
 
-    const size_t numOfUnusedIdxs = BackRefMain::dataSz - lastUsed - 1;
+    const intptr_t numOfUnusedIdxs = BackRefMain::dataSz - lastUsed - 1;
     if (numOfUnusedIdxs <= 0) { // no space in main under lock, roll back
         backend->putBackRefSpace(newBl, blockSpaceSize, isRawMemUsed);
         return false;
     }
     // It's possible that only part of newBl is used, due to lack of indices in main.
     // This is OK as such underutilization is possible only once for backreferneces table.
-    int blocksToUse = min(numOfUnusedIdxs, blockSpaceSize / BackRefBlock::bytes);
+    unsigned blocksToUse =
+        (unsigned)(min((uintptr_t)numOfUnusedIdxs, blockSpaceSize / BackRefBlock::bytes));
 
     // use the first block in the batch to maintain the list of "raw" memory
     // to be released at shutdown
@@ -196,7 +198,9 @@ bool BackRefMain::requestNewSpace()
         newBl->nextRawMemBlock = backRefMain.load(std::memory_order_relaxed)->allRawMemBlocks;
         backRefMain.load(std::memory_order_relaxed)->allRawMemBlocks = newBl;
     }
-    for (BackRefBlock *bl = newBl; blocksToUse>0; bl = (BackRefBlock*)((uintptr_t)bl + BackRefBlock::bytes), blocksToUse--) {
+    for (BackRefBlock *bl = newBl; blocksToUse != 0;
+         bl = (BackRefBlock*)((uintptr_t)bl + BackRefBlock::bytes), blocksToUse--)
+    {
         initEmptyBackRefBlock(bl);
         if (active.load(std::memory_order_relaxed)->allocatedCount.load(std::memory_order_relaxed) == BR_MAX_CNT) {
             active.store(bl, std::memory_order_release); // active leaf is not needed in listForUse
--- a/src/tbbmalloc/backref.cpp
+++ b/src/tbbmalloc/backref.cpp
@@ -189,6 +189,12 @@ bool BackRefMain::requestNewSpace()
     }
     // It's possible that only part of newBl is used, due to lack of indices in main.
     // This is OK as such underutilization is possible only once for backreferneces table.
+
+    // Since numOfUnusedIdxs remains greater than zero after subtraction above, its value should be
+    // less than INT_MAX.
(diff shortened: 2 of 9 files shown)


src/tbbmalloc/frontend.cpp
```diff
@@ -999,13 +999,13 @@ TLSData* MemoryPool::getTLS(bool create)
...
  */
-inline Bin* TLSData::getAllocationBin(size_t size)
+inline Bin* TLSData::getAllocationBin(unsigned size)
 {
     return bin + getIndex(size);
 }
 
 /* Return an empty uninitialized block in a non-blocking fashion. */
-Block *MemoryPool::getEmptyBlock(size_t size)
+Block *MemoryPool::getEmptyBlock(unsigned size)
```
Alexandr-Konovalov, line 1008: I see in both call sites we use `getEmptyBlock()` with cast. To decrease number of casts, can we move the case to `getEmptyBlock()` body?

src/tbbmalloc/frontend.cpp
Alexandr-Konovalov, line 2572: After this point, we think it's always save to cast `size` to unsigned.

1. Can it be done one with some new variable? Or do you think that more explicit code is more readable?

2. This is correct, because `minLargeObjectSize` is fit in unsigned. Is it deserve an assert?

src/tbbmalloc/frontend.cpp
```diff
@@ -1583,7 +1583,7 @@ void Block::initEmptyBlock(TLSData *tls, size_t size)
 
     cleanBlockHeader();
     MALLOC_ASSERT(objSz <= USHRT_MAX, "objSz must not be less 2^16-1");
-    objectSize = objSz;
+    objectSize = (uint16_t)objSz;
```
Alexandr-Konovalov, line 1586: Can we return `uint16_t` directly from `getObjectSize()`? The assert also can be moved there.

src/tbbmalloc/frontend.cpp
aleksei-fedotov: Having a function which casts down its argument as the first step looks strange to me. I had been thinking about introducing another variable on the stack with appropriate type instead.

src/tbbmalloc/frontend.cpp
aleksei-fedotov: 1. The variable name is stumbling block for me here. Can you suggest a good one?
2. Do you mean `size <= UINT_MAX` as an assert? I don't think so, since the code flow checks this implicitly through comparison with `minLargeObjectSize`. Or, maybe you are suggesting different assertion?

src/tbbmalloc/frontend.cpp
aleksei-fedotov: I saw the TODO about that in the code just before the `getIndexOrObjectSize` function. Maybe, it is better to try addressing that with a separate patch?

Alexandr-Konovalov: Yes, sure. Haven't see that TODO.

src/tbbmalloc/frontend.cpp
Alexandr-Konovalov: 1. I'm not good in names. Probably, just `uSize`?

2. Not exactly. Probably, `static_assert()` with `minLargeObjectSize` and `UINT_MAX`. We use that implicitly, when write cast `(unsigned)size`.

src/tbbmalloc/large_objects.h
```diff
@@ -54,10 +54,11 @@ struct LargeBinStructureProps {
         return alignUp(size, CacheStep);
     }
 
-    static int sizeToIdx(size_t size) {
+    static unsigned sizeToIdx(size_t size) {
+        MALLOC_ASSERT(MaxSize <= UINT_MAX, ASSERT_TEXT);
```
Alexandr-Konovalov: Can it be `static_assert`?

src/tbbmalloc/backref.cpp
```diff
@@ -181,22 +182,25 @@ bool BackRefMain::requestNewSpace()
...
-    const size_t numOfUnusedIdxs = BackRefMain::dataSz - lastUsed - 1;
+    const intptr_t numOfUnusedIdxs = BackRefMain::dataSz - lastUsed - 1;
     if (numOfUnusedIdxs <= 0) { // no space in main under lock, roll back
         backend->putBackRefSpace(newBl, blockSpaceSize, isRawMemUsed);
         return false;
     }
     // It's possible that only part of newBl is used, due to lack of indices in main.
     // This is OK as such underutilization is possible only once for backreferneces table.
-    int blocksToUse = min(numOfUnusedIdxs, blockSpaceSize / BackRefBlock::bytes);
+    unsigned blocksToUse =
```
Alexandr-Konovalov, line 198: Why do we need make `unsigned` from `uintptr_t` and thinking about "is it fit"?

aleksei-fedotov: Added the comment in the code explaining the idea along with an assertion.

src/tbbmalloc/frontend.cpp
aleksei-fedotov: I think we addressed that in another comment.

src/tbbmalloc/frontend.cpp
aleksei-fedotov: 1. Renamed.
2. Added.

src/tbbmalloc/backref.cpp
Alexandr-Konovalov: Thank you for the clarification.

Alexandr-Konovalov:
LGTM.

omalyshe (commented):
@aleksei-fedotov Don't forget to fix the copyright

aleksei-fedotov:
> @aleksei-fedotov Don't forget to fix the copyright

Thanks for reminding. Updated.

Alexandr-Konovalov (dismissed):

src/tbbmalloc/large_objects.h
```diff
@@ -143,7 +146,7 @@ class LargeObjectCacheImpl {
 
 public:
     // The number of bins to cache large/huge objects.
-    static const uint32_t numBins = Props::NumBins;
+    static const unsigned numBins = Props::NumBins;
```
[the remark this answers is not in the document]
aleksei-fedotov: Loops were replaced with unsigned variant of assignment.

src/tbbmalloc/backend.h
```diff
@@ -326,7 +327,8 @@ class Backend {
         else if (size < minBinnedSize)
             return NO_BIN;
 
-        int bin = (size - minBinnedSize)/freeBinsStep;
+        size_t val = (size - minBinnedSize)/freeBinsStep;
+        int bin = (int)val;
```
Alexandr-Konovalov, line 331: Is `val` usable?

src/tbbmalloc/frontend.cpp
```diff
@@ -2469,7 +2470,9 @@ inline bool Block::isProperlyPlaced(const void *object) const
 FreeObject *Block::findAllocatedObject(const void *address) const
 {
     // calculate offset from the end of the block space
-    uint16_t offset = (uintptr_t)this + slabSize - (uintptr_t)address;
+    const uintptr_t tmp_offset = (uintptr_t)this + slabSize - (uintptr_t)address;
+    MALLOC_ASSERT(tmp_offset <= USHRT_MAX, ASSERT_TEXT);
```
Alexandr-Konovalov: Can we say in the assert text, that that is true, because of block size is limited to 16K and address points inside that block?

src/tbbmalloc/large_objects.h
```diff
@@ -48,16 +49,19 @@ struct LargeBinStructureProps {
 public:
     static const size_t   MinSize = MIN_SIZE, MaxSize = MAX_SIZE;
     static const size_t   CacheStep = 8 * 1024;
+    static_assert((MaxSize - MinSize) / CacheStep <= UINT_MAX,
+                  "The type for NumBins is small to represent the desired value.");
```
Alexandr-Konovalov: Can it be re-phrased to something like "the size of NNN is enough to ..." or something like that? And can we use sizeof(NumBins) in the assert instead of UINT_MAX because that's the size we want to fit?

src/tbbmalloc/backend.h
aleksei-fedotov: Do you mean if it is used anywhere else? No. Only in this function.

src/tbbmalloc/frontend.cpp
aleksei-fedotov: Sure. Added explanation to the assert text and used more strict condition in the assertion.

src/tbbmalloc/large_objects.h
aleksei-fedotov: It is not about size rather than the result value. Changed to a more scalable variant of the code.

src/tbbmalloc/frontend.cpp
```diff
@@ -2466,15 +2467,19 @@ inline bool Block::isProperlyPlaced(const void *object) const
...
+FreeObject *Block::findAllocatedObject(const void *address_ptr) const
 {
+    const uintptr_t address = (uintptr_t)address_ptr;
     // calculate offset from the end of the block space
-    uint16_t offset = (uintptr_t)this + slabSize - (uintptr_t)address;
-    MALLOC_ASSERT( offset<=slabSize-sizeof(Block), ASSERT_TEXT );
+    const uintptr_t tmp_offset = (uintptr_t)this + slabSize - address;
+    MALLOC_ASSERT(tmp_offset <= slabSize - sizeof(Block),
+                  "Address is within the small object allocation block");
+    MALLOC_ASSERT(slabSize <= USHRT_MAX, "Below cast is harmless");
```
Alexandr-Konovalov: Can it be `static_assert`?

src/tbbmalloc/large_objects.h
Alexandr-Konovalov: Definitely, `numeric_limits` is better.

src/tbbmalloc/frontend.cpp
aleksei-fedotov: Sure. Changed.

Alexandr-Konovalov (approved):
LGTM.

omalyshe (approved):
LGTM

ldorau (approved):

PR #2058, opened 2026-04-20
[tbbmalloc] Make indices unsigned

aleksei-fedotov:
Make `sizeToIdx` to return unsigned values, propagating the change up the stack.

### DIFF
--- a/src/tbbmalloc/large_objects.cpp
+++ b/src/tbbmalloc/large_objects.cpp
@@ -418,7 +418,8 @@ template<typename Props> void CacheBinFunctor<Props>::operator()(CacheBinOperati
 #if __TBB_MALLOC_WHITEBOX_TEST
             tbbmalloc_whitebox::locPutProcessed+=prep.putListNum;
 #endif
-            toRelease = bin->putList(prep.head, prep.tail, bitMask, idx, prep.putListNum, extMemPool->loc.hugeSizeThreshold);
+            toRelease = bin->putList(prep.head, prep.tail, bitMask, idx, prep.putListNum,
+                                     extMemPool->loc.hugeSizeThreshold);
         }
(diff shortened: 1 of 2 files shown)


src/tbbmalloc/large_objects.h
```diff
@@ -245,7 +245,7 @@ class LargeObjectCacheImpl {
 
     // Huge bins index for fast regular cleanup searching in case of
     // the "huge size threshold" setting defined
-    intptr_t     hugeSizeThresholdIdx;
+    uintptr_t     hugeSizeThresholdIdx;
```
Alexandr-Konovalov: We assign this variable from `uint32_t` and `unsigned` and have no arithmetic over it. Can it be narrower then `uintptr_t`?

src/tbbmalloc/large_objects.h
Alexandr-Konovalov: Is that a reason for one `updateUsedSize()` accept `unsigned` index, while another still accept `int`?

src/tbbmalloc/large_objects.cpp
Alexandr-Konovalov: This function accept `idx` as `int`.

src/tbbmalloc/large_objects.h
aleksei-fedotov: I think so. Narrowed the type.

src/tbbmalloc/large_objects.cpp
aleksei-fedotov: Thank you! That was my oversight.

src/tbbmalloc/large_objects.h
aleksei-fedotov: I think there is no reason. Updated to match the unsigned variant.

Alexandr-Konovalov (approved):
LGTM.

PR #2063, opened 2026-04-24
Propose the API for NUMA interleaved allocations

akukanov:
This is an update to #2032 proposing specific API for NUMA interleaved memory allocation.

### DIFF
--- a/rfcs/proposed/numa_support/interleaved-allocation.md
+++ b/rfcs/proposed/numa_support/interleaved-allocation.md
@@ -101,6 +101,10 @@ template <typename T>
 void free_numa_interleaved (T *ptr, size_t count);
 ```
 
+We may also want to add a function that queries the system memory page size, to make
+the use of custom interleaving steps simpler and less error-prone.
+Such a function should likely go into `info.h` and the `tbb::info` namespace.
+
 ### Error handling
 
 Two types of run-time errors might appear in these functions:
@@ -113,10 +117,10 @@ Since the API does not allow to return any error code directly, the options for 
 are:
 
 1. Do nothing except returning `nullptr`; not even argument checks. The requirements to
-   the arguments stem from those of the system API, where the arguments or their derivatives
+   the arguments stem from those of the system API, which the arguments or their derivatives
    are passed to. Therefore usage errors may effectively convert to system errors and result
    in a failed allocation or deallocation.
-   
+
    Together with declaring the behavior undefined if the requirements are not met, this is the simplest
    from the implementation viewpoint. On the downside, there is no way for users to known that
    `free_numa_interleaved` has failed (as it returns nothing), and no diagnostics for usage errors.
@@ -149,8 +153,31 @@ best approach, so we might need to use different ones, depending on a specific e
 
 ## ABI entry points
 
-TODO
+Since there is no memory management involved, the functionality belongs to the main TBB library,
+not the `tbbmalloc` library.
 
+The new ABI entry points should be stable and production-ready when added, so that we do not
+need to change these later. These functions should operate with raw bytes. Overloads are
+discouraged unless absolutely necessary. It is better to avoid the use of standard library types
+in the signatures, as that could cause dependencies on specific standard library implementations
+and/or versions and potentially require multiple sets of binaries for the same platform.
+
+Given that, the following conceptual function signatures are recommended:
+
+```c++
+void *allocate_interleaved (size_t bytes, tbb::numa_node_id *nodes, size_t node_count,
+                            size_t interleaving_step);
+void deallocate_interleaved (void *ptr, size_t bytes);
+```
+
+The names are subject to discussion. Internal namespaces will be used as appropriate.
+
+Note that the list of nodes is passed as a *{pointer, count}* pair of parameters instead of
+a single pointer or a reference. This both avoids using `std::vector` directly and allows
+changing/extending the public API with other contiguous storage types, e.g. `std::span`.
+
+We might consider changing the ABI signatures to also return an error code in some way,
+if we prefer error handling to be done strictly by the public API in the headers.
 
 ## Implementation details
 
@@ -169,11 +196,17 @@ There is no NUMA memory support under macOS, so the implementation can only fall
 
 ## Open Questions
 
+Are there reasons to release the API for preview first?
+
+In what header file should the API be defined?
+
+We need to decide on the way(s) to handle errors.
+
 When non-default `interleaving step` can be used?
 
-`size` argument for `free_interleaved()` appeared because what we have is wrappers over
-`mmap`/`munmap` and there is no place to put the size after memory is allocated. We can
-put it in, say, an internal cumap. Is it look useful?
+`bytes` argument for `free_numa_interleaved()` appeared because what we have is wrappers over
+`mmap`/`munmap` and there is no place in the allocated memory to store the size.
+We can put it in, say, an internal unordered map. Would it be better?
 
 Semantics of even distribution of data between NUMA nodes is straightforward: to equally
 balance work between the nodes. Why might someone want to distribute data unequally? Can
--- a/rfcs/proposed/numa_support/interleaved-allocation.md
+++ b/rfcs/proposed/numa_support/interleaved-allocation.md
@@ -9,10 +9,13 @@ There are two kinds of NUMA-related performance bottlenecks: latency increasing 
 access to a remote node and bandwidth-limited simultaneous access from different CPUs to
 a single NUMA memory node. A well-known method to mitigate both is a distribution of
 memory objects that are accessed from different CPUs to different NUMA nodes in such a way
-that matches an access pattern. If the access pattern is complex enough, a simple
-round-robin distribution can be good enough. The distribution can be achieved either by
-employing a first-touch policy of NUMA memory allocation or via special platform-dependent
-API. Generally, the latter requires less overhead.
+that matches the access pattern. If the access pattern is complex enough, a simple
+round-robin distribution can be good enough to address at least the second issue.
+To address the first issue best, depending on the access pattern it might be necessary
+to choose the set of NUMA nodes and/or the amount of memory mapped to a node.
+
+The distribution can be achieved either by employing a first-touch policy of NUMA memory
+allocation or via special platform-dependent API. Generally, the latter requires less overhead.
 
 ## The overall approach
 
@@ -26,9 +29,9 @@ To guide the mapping of memory across NUMA nodes, two additional parameters are 
 `interleaving step` is the size of the contiguous memory block from a particular NUMA node,
 it has page granularity. Currently there are no clear use cases for granularity more than page size.
 
-`list of nodes for allocation` is `std::vector<tbb::numa_node_id>` to be compatible with a
-value returned from `tbb::numa_nodes()`. `libnuma` supports a subset of NUMA nodes for
-allocation, but those nodes are loaded equally. Having `vector` allows us to express an
+`list of nodes for allocation` is `std::vector<tbb::numa_node_id>` to be compatible with
+the values returned from `tbb::info::numa_nodes()`. `libnuma` supports a subset of NUMA nodes
+for allocation, but those nodes are loaded equally. Using `vector` allows us to express an
 unbalanced load. Example: allocation over the list of nodes [3, 0, 3] uses 2/3 memory from
 node 3 and 1/3 from node 0.
 
@@ -83,13 +86,13 @@ The default value of `0` for the interleaving step (`bytes_per_chunk`) indicates
 the implementation should choose one automatically. If specified, the step should be
 a multiple of the memory page size.
 
-The functions that frees the memory take an allocation address and the memory size,
+The function that frees the memory takes an allocation address and the memory size,
 which should match the allocated size at the given address. This follows the standard practice
 for C++ memory allocators and memory resources, and it is needed for the implementation
 to properly call system routines. Alternative variants to store the size somehow between
 the calls (an internal map, an extra memory page, a shared pointer with custom deleter)
 were considered and rejected, as those come with some kind of overhead or/and deviate from
-the common API paradigms to allocate/free the memory.
+the common C++ API paradigms to allocate/free memory.
 
 We also propose function templates to allocate and free memory for objects of a certain type:
 
@@ -105,12 +108,12 @@ void deallocate_numa_interleaved (T *ptr, size_t count);
 ```
 
 Instead of raw size in bytes, these functions take a number of objects of the given type
-to indicate the memory size and the interleaving step. The latter (`count_per_chunk`) should be
+to indicate the memory size and the interleaving step. The latter, `count_per_chunk`, should be
 such that `count_per_chunk * sizeof(T)` is a multiple of the memory page size.
 
-The template argument for ``allocate_numa_interleaved`` cannot be deduced and should always be provided
-explicitly. The template argument for ``deallocate_numa_interleaved`` is deducible from the type of
-the first argument (pointer). As a downside, that requires an explicit pointer cast to ``void*``
+The template argument for `allocate_numa_interleaved` cannot be deduced and should always be provided
+explicitly. The template argument for `deallocate_numa_interleaved` is deducible from the type of
+the first argument (pointer). As a downside, an explicit pointer cast to `void*` is required
 in order to call the non-typed deallocation function.
 
 The allocated memory is not initialized (no constructors are called). Reading from it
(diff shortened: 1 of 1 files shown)


rfcs/proposed/numa_support/interleaved-allocation.md
```diff
@@ -36,19 +35,123 @@ node 3 and 1/3 from node 0.
...
+```c++
+void *allocate_numa_interleaved (size_t bytes,
+                                 const std::vector<tbb::numa_node_id>& nodes,
+                                 size_t bytes_per_chunk = 0);
+void *allocate_numa_interleaved (size_t bytes, size_t bytes_per_chunk = 0);
+```
+
+The default value of `0` for the interleaving step (`bytes_per_chunk`) indicates that
+the implementation should choose one automatically. If specified, the step should be
+a multiple of the memory page size.
```
aleksei-fedotov, line 82: Since it should be a multiple of the memory page size, maybe replace the parameter with the number of times by which the page size to be multiplied? The default value would be equal to `1` in this case.

rfcs/proposed/numa_support/interleaved-allocation.md
```diff
@@ -36,19 +35,155 @@ node 3 and 1/3 from node 0.
...
+with `tbb_allocator.h` being probably better than others.
+
+If a new header is added, it could be aimed to contain either these APIs only (e.g.,
+`numa_allocation.h`) or a broader set of NUMA-related APIs, both existing and added later
+(`numa.h`). In the latter case the header should also include other public headers such as
+`task_arena.h` and `info.h`.
+
+The choice might depend on whether the API is initially delivered as a preview feature or not.
+For the preview feature it seems fine to add it to an existing header, while for production
+a new header seems better.
```
aleksei-fedotov: I vote for having it in the _production-ready_ place from the very beginning. The reason is to minimize source code changes once the feature becomes of a production quality.
I believe having a dedicated _numa.h_ header that introduces NUMA-related API such as `tbb::info` and this proposal API by including or declaring them is a way to go.
For _task_arena.h_ I think it should be the other way around: _task_arena.h_ to include new _numa.h_.

rfcs/proposed/numa_support/interleaved-allocation.md
```diff
@@ -36,19 +35,155 @@ node 3 and 1/3 from node 0.
...
+                                 const std::vector<tbb::numa_node_id>& nodes,
+                                 size_t bytes_per_chunk = 0);
+void *allocate_numa_interleaved (size_t bytes, size_t bytes_per_chunk = 0);
 ```
 
+The default value of `0` for the interleaving step (`bytes_per_chunk`) indicates that
+the implementation should choose one automatically. If specified, the step should be
+a multiple of the memory page size.
+
+In addition, function templates are proposed to allocate memory for objects of a certain type:
```
aleksei-fedotov: The use cases for this API are not clear to me. Also, not calling the corresponding constructors and destructors seems to diminish the added value.

rfcs/proposed/numa_support/interleaved-allocation.md
```diff
@@ -36,19 +35,155 @@ node 3 and 1/3 from node 0.
...
+   in a failed allocation or deallocation.
+
+   Together with declaring the behavior undefined if the requirements are not met, this is the simplest
+   from the implementation viewpoint. On the downside, there is no way for users to known that
+   `free_numa_interleaved` has failed (as it returns nothing), and no diagnostics for usage errors.
+
+2. Throw an exception: `bad_alloc` or/and more specific types like `system_error` and
+   `invalid_argument`. Note that `bad_alloc`, while typically used as a signal that memory
+   could not be allocated, unfortunately does not allow to set a custom message, and also
+   does not fit well semantically to deallocation.
```
aleksei-fedotov, line 151: Another approach could be introducing own set of allocation/deallocation exceptions to distinguish runtime situations.

rfcs/proposed/numa_support/interleaved-allocation.md
```diff
@@ -36,19 +35,155 @@ node 3 and 1/3 from node 0.
...
+   https://uxlfoundation.github.io/oneAPI-spec/spec/elements/oneTBB/source/configuration/enabling_debugging_features.html#tbb-use-assert-macro)
+   and together with
+   [assertion handlers](https://uxlfoundation.github.io/oneTBB/main/reference/assertion_handler.html)
+   can be used for decent error handling and diagnostics.
+
+   However, there is no recovery from an assertion, so it does not fit well for handling system errors
+   during memory allocation.
+
+The choice can again be different for preview vs. production. Also it seems there is no universally
+best approach, so we might need to use different ones, depending on a specific error.
```
aleksei-fedotov: One more idea is to introduce a new data type that might include various things related to the status of invocation of this API. The minimum that comes to my mind is:
- The pointer to allocated memory, which could be `nullptr` if allocation fails.
- The status of allocation/deallocation.
- The size, which will be automatically used by corresponding `free`-functions.
This would allows us making the universal approach, disregarding whether exceptions are used or not. As always this comes with the price for users to learn the semantics of that new type as it also not quite "authentic" for C++.

rfcs/proposed/numa_support/interleaved-allocation.md
```diff
@@ -66,11 +201,17 @@ There is no NUMA memory support under macOS, so the implementation can only fall
...
+We need to decide on the way(s) to handle errors.
+
 When non-default `interleaving step` can be used?
 
-`size` argument for `free_interleaved()` appeared because what we have is wrappers over
-`mmap`/`munmap` and there is no place to put the size after memory is allocated. We can
-put it in, say, an internal cumap. Is it look useful?
+`bytes` argument for `free_numa_interleaved()` appeared because what we have is wrappers over
+`mmap`/`munmap` and there is no place in the allocated memory to store the size.
+We can put it in, say, an internal unordered map. Would it be better?
```
aleksei-fedotov: We could put it into objects of a new type I suggested earlier. This can be handled on a header side, not requiring to alter the ABI.

rfcs/proposed/numa_support/interleaved-allocation.md
Alexandr-Konovalov: > The status of allocation/deallocation.

Why may we want keep the status? It's propose it it be reported right away the allocation failed. So it's either an exception or in no-exception mode it's a return value (perhaps, in form of reference to it in additional argument).

In which use cases is it useful to check error code not at the call site?

rfcs/proposed/numa_support/interleaved-allocation.md
```diff
@@ -36,19 +35,155 @@ node 3 and 1/3 from node 0.
...
+
+Last, the functions that free the previously allocated memory take an allocation address
+and the memory size. The latter should match the allocated size at the given address.
+The function template overload does not call destructors; the type information is only
+necessary for the `count` argument to match the corresponding value at the allocation.
+
+```c++
+void free_numa_interleaved (void *ptr, size_t bytes);
+template <typename T>
+void free_numa_interleaved (T *ptr, size_t count);
```
Alexandr-Konovalov: What is the difference between typed and untyped `free_numa_interleaved`? It seems sycl has only version with `void *`.

rfcs/proposed/numa_support/interleaved-allocation.md
aleksei-fedotov: It is not about checking the result not at the call site. It is more about determining the reason of a failure, for which `nullptr` is not enough as it was highlighted earlier.

However, I can also relatively easy imagine the application that has a pipeline with possibly parallel branches where the earlier stages acquire resources while one of the later stages decide on processing based on the statuses from these earlier stages, one of which can be an allocation.

rfcs/proposed/numa_support/interleaved-allocation.md
akukanov: If `numa.h` is aimed to be the only header required to use any NUMA-related API, then "the other way around" for `task_arena.h` does not work. Remember that `create_numa_task_arenas` is already in `task_arena.h`, according to the specification.

rfcs/proposed/numa_support/interleaved-allocation.md
akukanov: Having the standard `uninitialized_*` family of functions, it is better to allow developers to choose how they initialize their data.

As for the use case - it's just syntax sugar over the non-typed allocation functions. Compare:
```
    int* pa = static_cast<int*>(tbb::allocate_numa_interleaved(num_elements * sizeof(int));
    int* pb = tbb::allocate_numa_interleaved<int>(num_elements);
```

rfcs/proposed/numa_support/interleaved-allocation.md
akukanov: I think custom exceptions would make sense if:

- TBB always used custom exceptions, perhaps derived from the same base type. AFAIK this is not the case.
- The available standard exceptions were semantically not appropriate for the types of errors.

I do not think any of these is true, therefore I do not see this approach being worth the effort.

rfcs/proposed/numa_support/interleaved-allocation.md
akukanov: > As always this comes with the price for users to learn the semantics of that new type as it also not quite "authentic" for C++.

For me these are significant downsides - essentially, the API would "play against" the common usage practice for allocation functions.

Also, building on Alexandr's comment, would it really be a good idea to combine in the same type both the status, which is only needed immediately after the call, and the size, which is only needed at the end-of-life point much later?

rfcs/proposed/numa_support/interleaved-allocation.md
akukanov: The difference is in the second parameter. The deallocation function template allows to pass the same value that was passed to the allocation function template, accounting for the difference between bytes and "elements". SYCL does not require to pass that value to `free`. See also https://en.cppreference.com/cpp/memory/allocator/deallocate

rfcs/proposed/numa_support/interleaved-allocation.md
akukanov: That means, essentially, expressing the step as the number of pages. Do we know if there are precedents of this approach?

That is not a bad idea, but I think it works not so well when you need to deal with more than one array of different types. The scenario I keep in mind is, essentially, a structure of arrays: think of several arrays of different types, which you need to process in a loop using the same index to access elements. In this case it is good, I believe, that the NUMA chunks in each array are of the same "logical" size (the number of elements), and not the same physical size.

akukanov: To illustrate, this is the code to allocate a "structure of arrays" with the typed API:
```c++
std::size_t items_per_chunk = page_size() / sizeof(int);
int * ints = tbb::allocate_numa_interleaved<int>(num_items, items_per_chunk);
double* doubles = tbb::allocate_numa_interleaved<double>(num_items, items_per_chunk);
```

The same with typeless API:
```c++
std::size_t items_per_chunk = page_size() / sizeof(int);
int * ints = (int*)tbb::allocate_numa_interleaved(num_items * sizeof(int), page_size());
double* doubles = (double*)tbb::allocate_numa_interleaved(num_items * sizeof(double),
                                                          items_per_chunk * sizeof(double));
```

With typeless API and the number of pages as the step:
```c++
std::size_t chunk_step = 1;
int * ints = (int*)tbb::allocate_numa_interleaved(num_items * sizeof(int), chunk_step);
double* doubles = (double*)tbb::allocate_numa_interleaved(num_items * sizeof(double),
                                                          chunk_step * sizeof(double) / sizeof(int));
```

With typed API and the number of pages as the step:
```c++
std::size_t chunk_step = 1;
int * ints = tbb::allocate_numa_interleaved<int>(num_items, chunk_step);
double* doubles = tbb::allocate_numa_interleaved<double>(num_items,
                                                         chunk_step * sizeof(double) / sizeof(int));
```

Maybe I miss some way to make the code more expressive, but so far using the number of pages for the step does not seem that convenient in this practical case.

rfcs/proposed/numa_support/interleaved-allocation.md
Alexandr-Konovalov: Thanks, missed that.

rfcs/proposed/numa_support/interleaved-allocation.md
Alexandr-Konovalov: Can it be just `shared_ptr` with custom deleter? An allocation function that return shared pointer is probably quite natural. There is no place for `allocation failed because an argument is not power-of-2`, however.

rfcs/proposed/numa_support/interleaved-allocation.md
vossmjp: I'm not convinced that NUMA is a first-order concern. I think that generally people will think of the base features first, such as task arenas, memory allocators, etc.  So I have a preference for keeping the numa extensions spread across other headers as appropriate, and don't think a numa.h is necessary. But maybe I can be convinced otherwise.

rfcs/proposed/numa_support/interleaved-allocation.md
vossmjp: I like the idea of a smart pointer type. Placing the burden on users to store/remember the size seems error prone.

rfcs/proposed/numa_support/interleaved-allocation.md
akukanov: Resolved as proposed in a team discussion.

rfcs/proposed/numa_support/interleaved-allocation.md
akukanov: With regard to the size/count, we decided to keep it consistent with the relevant C++ routines such as the `deallocate` method of allocators/memory resources. The error handling question is still unresolved.

rfcs/proposed/numa_support/interleaved-allocation.md
akukanov: See https://github.com/uxlfoundation/oneTBB/pull/2063#discussion_r3190871142

rfcs/proposed/numa_support/interleaved-allocation.md
vossmjp: Having looked at corresponding std C++ functions, I'm ok with having the size argument. I was thinking of this API as a high-level helper function, but if viewed as a set of low-level functions instead, then requiring the size is fine and can be the callers responsibility.

rfcs/proposed/numa_support/interleaved-allocation.md
```diff
@@ -66,11 +234,12 @@ There is no NUMA memory support under macOS, so the implementation can only fall
 
 ## Open Questions
 
-When non-default `interleaving step` can be used?
+The only major question left is that of error handling but it may impact but API and ABI.
```
Alexandr-Konovalov: `impact but API and ABI`
Typo? Both API and ABI?

rfcs/proposed/numa_support/interleaved-allocation.md
```diff
@@ -9,46 +9,214 @@ There are two kinds of NUMA-related performance bottlenecks: latency increasing
...
 ```
 
+Instead of raw size in bytes, these functions take a number of objects of the given type
+to indicate the memory size and the interleaving step. The latter, `count_per_chunk`, should be
+such that `count_per_chunk * sizeof(T)` is a multiple of the memory page size.
+
+The template argument for `allocate_numa_interleaved` cannot be deduced and should always be provided
+explicitly. The template argument for `deallocate_numa_interleaved` is deducible from the type of
+the first argument (pointer). As a downside, an explicit pointer cast to `void*` is required
+in order to call the non-typed deallocation function.
```
kboyarinov: We can consider changing the API of the templated deallocate_numa_interleaved overload to place T in a non-deduced context. For example, this could be done with std::decay_t:

```cpp
template <typename T>
void deallocate_numa_interleaved(std::decay_t<T>* ptr, size_t count);
```

With this API, T cannot be deduced, so any call to deallocate_numa_interleaved without an explicit template argument will resolve to the non-template overload, similarly to the allocation function.

rfcs/proposed/numa_support/interleaved-allocation.md
akukanov: Yes, thanks. But now this wording is removed.

rfcs/proposed/numa_support/interleaved-allocation.md
akukanov: Thanks, added this option to the text.

rfcs/proposed/numa_support/interleaved-allocation.md
```diff
@@ -9,46 +9,209 @@ There are two kinds of NUMA-related performance bottlenecks: latency increasing
...
+only the allocation size. In this case, `interleaving step` defaults to the page size (in bytes)
+and the memory is spread across all NUMA nodes.
+
+## API proposal
+
+### Header file
+
+None of the existing public TBB headers is really a good fit for the NUMA allocation API,
+with `tbb/tbb_allocator.h` being probably better than others. We propose therefore to add
+a new header, `tbb/numa_allocation.h`.
```
vossmjp, line 51: I think this is a sensible first step during preview feature. We can revisit an umbrella header later if we think that makes sense.

vossmjp (approved):
LGTM

kboyarinov (approved):
LGTM

PR #2074, opened 2026-05-08
Support interleaved NUMA allocations

Alexandr-Konovalov:
### Description 
Following RFC API to allocate memory interleaved between NUMA nodes

### Type of change

### Tests

- [ ] not needed

### Documentation

- [X] needs to be updated
- [ ] not needed

- [ ] Yes
- [X] No
- [ ] Unknown


test/common/utils.h
```diff
@@ -487,6 +487,27 @@ class UninitializedStorage {
...
+                if (((char*)(buf+i))[b])
+                    return sizeof(intptr_t)*i + b + 1;
+        }
+    for (size_t i=0; i<tailSz; i++)
+        if (bufTail[i]) {
+            return words*sizeof(intptr_t)+i+1;
+        }
+    return 0;
+}
+
```
aleksei-fedotov, line 510: Can you please add a comment why C++ standard algorithms such as `std::find_if_not` cannot be used here instead.

test/common/utils.h
```diff
@@ -487,6 +487,27 @@ class UninitializedStorage {
     T* operator&() { return &m_value; }
 };
 
+// if non-zero byte found, returns bad value address plus 1
```
aleksei-fedotov: Note that in the usage scenarios from the "test_malloc_compliance.cpp" below the one unit is subtracted from the returned value. Thus, consider returning the address without adding one to it.

include/tbb/numa_allocation.h
```diff
@@ -0,0 +1 @@
+#include "../oneapi/tbb/numa_allocation.h"
```
aleksei-fedotov: I think Copilot properly suggested to include the license preamble. All other headers follow this approach. Was the suggestion ignored by mistake?

src/tbb/governor.cpp
```diff
@@ -623,6 +623,12 @@ int __TBB_EXPORTED_FUNC constraints_threads_per_core(const d1::constraints&, int
     return system_topology::automatic;
 }
 
+// we need read-only access without making a copy
```
aleksei-fedotov: It is not clear to me what this comment means to signify. Is it about returning value or something else? Also, it does not say anything about why read-only access is needed. Consider elaborating it.

src/tbb/governor.cpp
```diff
@@ -623,6 +623,12 @@ int __TBB_EXPORTED_FUNC constraints_threads_per_core(const d1::constraints&, int
...
+const int *get_numa_nodes_indexes() {
```
aleksei-fedotov: Code style in this file attaches asterisks to a type name.
```suggestion
const int* get_numa_nodes_indexes() {
```

src/tbb/governor.h
```diff
@@ -151,6 +152,8 @@ class governor {
     }
 }; // class governor
 
+const int *get_numa_nodes_indexes();
```
aleksei-fedotov: Same here:
```suggestion
const int* get_numa_nodes_indexes();
```

include/tbb/numa_allocation.h
kboyarinov: The UXL Copyright and an empty line at the end are missed

test/tbb/test_numa_allocation.cpp
```diff
@@ -0,0 +1,185 @@
...
+*/
+
+#define TBB_PREVIEW_NUMA_ALLOCATION 1
+
+#include "common/test.h"
+#include "common/utils.h"
+#include "common/utils_dynamic_libs.h"
+
+#include "tbb/numa_allocation.h"
+
```
kboyarinov, line 24: Some briefs should be added to make brief scan happy after merge


```suggestion

//! \file test_numa_allocation.cpp
//! \brief Test for [preview] functionality

```

test/tbb/test_numa_allocation.cpp
```diff
@@ -0,0 +1,185 @@
...
+}
+
+#if __linux__
+    #define NUMA_EQ CHECK_EQ
+#else
+    // we set only desired NUMA node, but Windows may put page on any node
+    #define NUMA_EQ WARN_EQ
+#endif
+
+TEST_CASE("invalid parameters") {
```
kboyarinov, line 86:
```suggestion
//! \brief \ref requirement
TEST_CASE("invalid parameters") {
```

test/tbb/test_numa_allocation.cpp
```diff
@@ -0,0 +1,185 @@
...
+    REQUIRE(ptr != nullptr);
+    REQUIRE_EQ(utils::NonZero(ptr, bytes), 0);
+    if (use_find_node)
+        for (size_t i = 0; i < bytes; i += bytes_per_chunk) {
+            NUMA_EQ(find_numa_node(ptr + i), nodes[i / bytes_per_chunk % nodes.size()]);
+        }
+    tbb::deallocate_numa_interleaved(ptr, bytes);
+}
+
+TEST_CASE("test basics") {
```
kboyarinov, line 115:
```suggestion
//! \brief \ref interface \ref requirement
TEST_CASE("test basics") {
```

src/tbb/numa_allocation.cpp
```diff
@@ -0,0 +1,241 @@
...
+
+bool verify_args(size_t bytes, const tbb::detail::d1::numa_node_id *nodes_ids, size_t nodes_count,
+                 size_t bytes_per_chunk) {
+    if (bytes == 0) // to be consistent with mmap
+        return false;
+    if (bytes_per_chunk % governor::default_page_size() != 0)
+        return false;
+    if ((nodes_ids == nullptr && nodes_count != 0) || (nodes_ids != nullptr && nodes_count == 0))
+        return false;
+    return true;
```
kboyarinov:
```suggestion
    if (nodes_ids == nullptr) {
        return nodes_count == 0;
    }
    return nodes_count != 0;
```

src/tbb/numa_allocation.cpp
```diff
@@ -0,0 +1,241 @@
...
+                        const tbb::detail::d1::numa_node_id *nodes_ids, size_t nodes_count,
+                        size_t bytes_per_chunk) {
+    atomic_do_once(interleaved_initialization_impl, interleaved_initialization_state);
+
+    if (!verify_args(bytes, nodes_ids, nodes_count, bytes_per_chunk))
+        return nullptr;
+
+    if (!bytes_per_chunk)
+        bytes_per_chunk = governor::default_page_size();
+    const int *nodes = nodes_count? nodes_ids : get_numa_nodes_indexes();
```
kboyarinov:
```suggestion
    const int *nodes = nodes_count ? nodes_ids : get_numa_nodes_indexes();
```

include/oneapi/tbb/numa_allocation.h
```diff
@@ -0,0 +1,65 @@
...
+
+} // namespace r1
+
+namespace d1 {
+
+inline void *allocate_numa_interleaved(size_t bytes,
+                                       const std::vector<tbb::numa_node_id>& nodes,
+                                       size_t bytes_per_chunk = 0) {
+    if (nodes.empty())
+        return nullptr;
```
aleksei-fedotov, line 43: I know that the question about error handling is still open, but the discussion there suggests that we likely follow a combined approach. Therefore, I suggest using TBB asserts for the logic errors and, thus, replace this `if` with an assert. Besides following the TBB-like approach for error handling, it also helps users to figure out incorrect absence of HWLOC on early stages of development - [the problem with tbbbind](https://github.com/uxlfoundation/oneTBB/blob/master/rfcs/proposed/numa_support/tbbbind-link-static-hwloc.org#introduction:~:text=This%20lack%20of%20valid%20HW%20topology%2C%20caused%20by%20the%20absence%20of%20a%20third%2Dparty%20library%2C%20is%20the%20major%20problem%20with%20the%20current%20oneTBB%20behavior.%20The%20problem%20lies%20in%20the%20lack%20of%20diagnostics%20making%20it%20difficult%20for%20developers%20to%20detect.%20As%20a%20result%2C%20the%20code%20continues%20to%20run%20but%20fails%20to%20use%20NUMA%20as%20intended.) we specifically addressed.
With a `nullptr` being returned it is not clear to a client what the problem actually is. From the very beginning, let's expect decent values in arguments and don't hide user's issues in implementation.

src/tbb/numa_allocation.cpp
```diff
@@ -0,0 +1,241 @@
...
+
+    char *end_ptr = base_addr + bytes;
+    // move_pages() has no length parameter, so moving must be done per page
+    for (char *ptr = base_addr; ptr < end_ptr; ptr += governor::default_page_size()) {
+        size_t page_idx = (ptr - base_addr) / governor::default_page_size();
+        size_t stride_idx = (ptr - base_addr) / bytes_per_chunk;
+        pages[page_idx] = ptr;
+        nodes_per_page[page_idx] = nodes[stride_idx % nodes_count];
+    }
+    long ret = move_pages_ptr(0, count_pages, pages.get(), nodes_per_page.get(), status.get(), 0);
```
kboyarinov:
```suggestion
    long ret = move_pages_ptr(/*pid = */0, count_pages, pages.get(), nodes_per_page.get(), status.get(), /*flags = */0);
```

include/tbb/numa_allocation.h
Alexandr-Konovalov: Done.

test/tbb/test_numa_allocation.cpp
```diff
@@ -0,0 +1,185 @@
...
+#else // on Windows we use VirtualAllocEx and QueryWorkingSetEx, so always can check NUMA node
+    bool lib = true;
+#endif
+
+    for (size_t obj_size = 8; obj_size <= 1024 * 1024LLU; obj_size *= 2)
+    {
+        std::vector<tbb::numa_node_id> numa_nodes = tbb::info::numa_nodes();
+        // we treat no-NUMA as single-NUMA with node index 0, but numa_nodes() return -1 in this case
+        if (numa_nodes.size() == 1)
+            numa_nodes[0] = 0;
```
aleksei-fedotov, line 144: This looks strange from the user standpoint if it would always be needed. I think such checks needs to be made inside the implementation. I think we should support straightforward scenario, which is:

```cpp
std::vector<tbb::numa_node_id> numa_nodes = tbb::info::numa_nodes();
char *ptr = (char *)tbb::allocate_numa_interleaved(obj_size, numa_nodes);
```

test/tbb/test_numa_allocation.cpp
```diff
@@ -0,0 +1,185 @@
...
+
+#if _WIN32 || _WIN64
+#include <psapi.h>
+#else
+#include <unistd.h> // for sysconf(_SC_PAGESIZE)
+#endif
+
+#if __linux__
+static long (*move_pages_ptr)(int pid, unsigned long count,
+                void **pages, const int *nodes, int *status, int flags) = nullptr;
```
aleksei-fedotov: Align a bit more?
```suggestion
static long (*move_pages_ptr)(int pid, unsigned long count,
                              void **pages, const int *nodes,
                              int *status, int flags) = nullptr;
```

test/tbb/test_numa_allocation.cpp
```diff
@@ -0,0 +1,185 @@
...
+                    "allocation of 0 bytes must return nullptr");
+    REQUIRE_MESSAGE(tbb::detail::r1::allocate_interleaved(4096, nullptr, 1, 0) == nullptr,
+                    "nodes_count must be 0 if nodes_ids is nullptr");
+    int node = 0;
+    REQUIRE_MESSAGE(tbb::detail::r1::allocate_interleaved(4096, &node, 0, 0) == nullptr,
+                    "nodes_count must be greater than 0 if nodes_ids is not nullptr");
+}
+
+void VerifySizeAndNodes(bool use_find_node, size_t bytes, const std::vector<tbb::numa_node_id>& nodes,
+     size_t bytes_per_chunk) {
```
aleksei-fedotov: Usually, for functions with long list of parameters I use the following alignment:
```suggestion
void VerifySizeAndNodes(bool use_find_node, size_t bytes, const std::vector<tbb::numa_node_id>& nodes,
                        size_t bytes_per_chunk)
{
```

src/tbb/numa_allocation.cpp
Alexandr-Konovalov: Thanks, probably single comparison of `nodes_ids`  is better,

test/tbb/test_numa_allocation.cpp
```diff
@@ -0,0 +1,185 @@
...
+}
+
+int find_numa_node(void* addr) {
+#if __linux__
+    int numa_node = -1;
+    int status[1];
+    void* pages[1] = { addr };
+
+    // Query which node owns this page
+    if (move_pages_ptr(0, 1, pages, NULL, status, 0) == 0)
```
aleksei-fedotov, line 57: Consider making the code a bit more readable:

```suggestion
    if (move_pages_ptr(/*pid*/0, /*count*/1, pages, /*nodes*/NULL, status, /*flags*/0) == 0)
```

test/tbb/test_numa_allocation.cpp
```diff
@@ -0,0 +1,185 @@
...
+        numa_node = status[0];
+
+    return numa_node;
+#elif _WIN32 || _WIN64
+    PSAPI_WORKING_SET_EX_INFORMATION pv = { nullptr, {0} };
+    pv.VirtualAddress = addr;
+
+    // Query the working set extended information
+    if (!QueryWorkingSetEx(GetCurrentProcess(), &pv, sizeof(pv))) {
+        std::cerr << "QueryWorkingSetEx failed." << std::endl;
```
aleksei-fedotov: Shall we write using Doctest faciliities instead? Something like using `WARN`? I suppose it is going to be easier noticed on the console rather than a single line.

test/tbb/test_numa_allocation.cpp
```diff
@@ -0,0 +1,185 @@
...
+        // Function failed (e.g., invalid handle or pointer)
+        return -1;
+    }
+
+    // If Valid == 0, the page is either swapped out or not yet faulted in.
+    if (!pv.VirtualAttributes.Valid) {
+        std::cerr << "VirtualAttributes invalid" << std::endl;
```
aleksei-fedotov: Same comment about using of Doctest facilities here.

src/tbb/numa_allocation.cpp
```diff
@@ -0,0 +1,241 @@
...
+#include "oneapi/tbb/numa_allocation.h"
+
+#include <atomic>
+#include <memory> // for std::unique_ptr
+
+#if __linux__
+
+#include <sys/mman.h>
+
+// must support build without numaif.h
```
aleksei-fedotov: Can you please explain in the code why?

src/tbb/numa_allocation.cpp
```diff
@@ -0,0 +1,241 @@
...
+extern "C" long move_pages(int pid, unsigned long count,
+                           void **pages, const int *nodes, int *status, int flags);
+
+#elif _WIN32 || _WIN64
+
+template<typename T>
+static inline T alignUp(T arg, uintptr_t alignment) {
+    return T(((uintptr_t)arg+(alignment-1)) & ~(alignment-1));
+}
```
aleksei-fedotov: Consider moving this helper to common template helpers.

src/tbb/numa_allocation.cpp
aleksei-fedotov: How about?
```suggestion
    return !((uintptr_t)nodes_ids ^ nodes_count); // either both specified or unspecified.
```

src/tbb/numa_allocation.cpp
```diff
@@ -0,0 +1,241 @@
...
+void interleaved_initialization_impl() {
+#if __linux__
+    dynamic_link("libnuma.so", LibnumaLinkTable,
+                 sizeof(LibnumaLinkTable) / sizeof(dynamic_link_descriptor), nullptr,
+                 DYNAMIC_LINK_GLOBAL | DYNAMIC_LINK_LOAD | DYNAMIC_LINK_WEAK);
+#elif _WIN32 || _WIN64
+    dynamic_link("kernelbase.dll", LibnumaLinkTable,
+                 sizeof(LibnumaLinkTable) / sizeof(dynamic_link_descriptor));
+#endif
+}
```
aleksei-fedotov: So, what will happen if `dynamic_link` fails to link the symbol?

src/tbb/numa_allocation.cpp
```diff
@@ -0,0 +1,241 @@
...
+    if (!nodes_count)
+        nodes_count = numa_node_count();
+
+#if __linux__
+    char *base_addr = reinterpret_cast<char*>(
+        mmap(nullptr, bytes, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0));
+    if (base_addr == MAP_FAILED)
+        return nullptr;
+
+    auto unmap = [bytes](void* ptr) {
```
aleksei-fedotov: This file code style adjusts the pointer on the right side.
```suggestion
    auto unmap = [bytes](void *ptr) {
```

src/tbb/numa_allocation.cpp
```diff
@@ -0,0 +1,241 @@
...
+    dynamic_link("libnuma.so", LibnumaLinkTable,
+                 sizeof(LibnumaLinkTable) / sizeof(dynamic_link_descriptor), nullptr,
+                 DYNAMIC_LINK_GLOBAL | DYNAMIC_LINK_LOAD | DYNAMIC_LINK_WEAK);
+#elif _WIN32 || _WIN64
+    dynamic_link("kernelbase.dll", LibnumaLinkTable,
+                 sizeof(LibnumaLinkTable) / sizeof(dynamic_link_descriptor));
+#endif
+}
+
+bool verify_args(size_t bytes, const tbb::detail::d1::numa_node_id *nodes_ids, size_t nodes_count,
```
aleksei-fedotov: Consider renaming to something that can be used in boolean contexts, something that returns either `true` or `false`. For example:

```suggestion
bool has_proper_args(size_t bytes, const tbb::detail::d1::numa_node_id *nodes_ids, size_t nodes_count,
```

So that it can fit better in the conditional blocks.

src/tbb/numa_allocation.cpp
```diff
@@ -0,0 +1,241 @@
...
+    };
+    std::unique_ptr<void, decltype(unmap)> data_holder(base_addr, unmap);
+
+    // touch each page, otherwise move_pages() will fail with EFAULT
+    for (size_t i = 0; i < bytes; i += governor::default_page_size())
+        base_addr[i] = 0;
+
+    // no NUMA nodes or move_pages() not available, just return the memory as is
+    if (numa_node_count() == 1 || !move_pages_ptr)
+        return data_holder.release();
```
aleksei-fedotov: I wonder if the first loop is necessary in case it exits with the second one.

src/tbb/numa_allocation.cpp
```diff
@@ -0,0 +1,241 @@
...
+
+    if (!bytes_per_chunk)
+        bytes_per_chunk = governor::default_page_size();
+    const int *nodes = nodes_count? nodes_ids : get_numa_nodes_indexes();
+    if (!nodes_count)
+        nodes_count = numa_node_count();
+
+#if __linux__
+    char *base_addr = reinterpret_cast<char*>(
+        mmap(nullptr, bytes, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0));
```
aleksei-fedotov: Consider writing parameter names.

src/tbb/numa_allocation.cpp
```diff
@@ -0,0 +1,241 @@
...
+    if (numa_node_count() == 1 || !VirtualAlloc2_ptr)
+        // do not use VirtualAlloc(), because it compiled incorrectly by MSVC 2017 with
+        // -DCMAKE_MSVC_RUNTIME_LIBRARY=MultiThreadedDebug -DTBB_WINDOWS_DRIVER=ON
+        return VirtualAllocEx(GetCurrentProcess(), nullptr, bytes, MEM_RESERVE | MEM_COMMIT, PAGE_READWRITE);
+
+    // for VirtualAlloc2 it must be a multiple of the page size
+    bytes = alignUp(bytes, governor::default_page_size());
+    char* base_addr =
+        static_cast<char*>(VirtualAlloc2_ptr(nullptr, nullptr, bytes, MEM_RESERVE | MEM_RESERVE_PLACEHOLDER,
+                                             PAGE_NOACCESS, nullptr, 0));
```
aleksei-fedotov: Make the cast identical to the Linux variant of the code?
```suggestion
    char* base_addr =
        reinterpret_cast<char*>(VirtualAlloc2_ptr(nullptr, nullptr, bytes,
                                                  MEM_RESERVE | MEM_RESERVE_PLACEHOLDER,
                                                  PAGE_NOACCESS, nullptr, 0));
```

aleksei-fedotov: Also, comments for parameters are needed for easier following the code.

test/common/utils.h
Alexandr-Konovalov: Interesting. Probably, that was pre-C++11 code.

include/tbb/numa_allocation.h
Alexandr-Konovalov: Yes, Konstantin also advised that, and it was added.

test/tbb/test_numa_allocation.cpp
Alexandr-Konovalov: Yes, the question is right, but to me that's a problem on `tbb::info::numa_nodes()` side. I see it manifested in #2042. Could we somehow prevent it from be created instead of adding workaround to each usage of `tbb::info::numa_nodes()`?

test/common/utils.h
Alexandr-Konovalov: Sadly, `find()` from libstdc++ unable to compare whole word at once. And this is performance-critical.

test/common/utils.h
Alexandr-Konovalov: That's done to distinguish between "difference is found at 0 offset" and "no differences found" cases.

src/tbb/governor.cpp
Alexandr-Konovalov: Good point. Let's drop it.

include/oneapi/tbb/numa_allocation.h
Alexandr-Konovalov: I've agreed that having no clear processing of invalid arguments in not user-friendly in particular and that error processing is important in general.

But I think that we are postponing error processing subject, aren't we?

test/tbb/test_numa_allocation.cpp
Alexandr-Konovalov: Thanks, done.

test/tbb/test_numa_allocation.cpp
Alexandr-Konovalov: Done.

src/tbb/numa_allocation.cpp
Alexandr-Konovalov: "Because TBB build must not depend of numactl presence", right?

src/tbb/numa_allocation.cpp
Alexandr-Konovalov: In misc.h right place for this? I don't want to put it in public headers, and sadly it seems no common internal header for tbbmalloc build and tbb.

src/tbb/numa_allocation.cpp
Alexandr-Konovalov: I hope `move_pages_ptr` and `VirtualAlloc2_ptr` would not changed, so it would contain 0 and this is correctly processed in the subsequent code.

src/tbb/numa_allocation.cpp
Alexandr-Konovalov: Done.

src/tbb/numa_allocation.cpp
Alexandr-Konovalov: Sure, let's move zerofilling to the user side.

src/tbb/numa_allocation.cpp
Alexandr-Konovalov: If you wish. But that's one click in IDE, meanwhile.

src/tbb/numa_allocation.cpp
Alexandr-Konovalov: Done,

test/tbb/test_numa_allocation.cpp
Alexandr-Konovalov: Done.

src/tbb/numa_allocation.cpp
Alexandr-Konovalov: Done.

include/oneapi/tbb/numa_allocation.h
```diff
@@ -0,0 +1,65 @@
...
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+
+#ifndef __TBB_numa_allocation_H
+#define __TBB_numa_allocation_H
+
+#if !TBB_PREVIEW_NUMA_ALLOCATION
```
vossmjp, line 20: In addition to preview macro, we also need a feature-test macro, probably TBB_HAS_NUMA_ALLOCATION or something similar. It should be defined at the end of `detail/_config.h`

include/oneapi/tbb/numa_allocation.h
```diff
@@ -0,0 +1,65 @@
...
+
+#include "info.h"
+
+namespace tbb {
+namespace detail {
+
+namespace r1 {
+
+TBB_EXPORT void *__TBB_EXPORTED_FUNC allocate_interleaved(size_t bytes,
+    const tbb::detail::d1::numa_node_id *nodes, size_t nodes_count, size_t bytes_per_chunk);
```
vossmjp, line 32: BTW, the RFC doesn't have `const` for `*nodes`, so probably should update the RFC to match.

include/oneapi/tbb/numa_allocation.h
vossmjp: Right, the RFC describes the preview MVP as, "Do nothing special for error handling, except for returning nullptr".  So, I agree that there should be something done before production, but its not necessary for preview as we agreed to earlier.

include/oneapi/tbb/numa_allocation.h
```diff
@@ -0,0 +1,65 @@
...
+
+inline void *allocate_numa_interleaved(size_t bytes,
+                                       const std::vector<tbb::numa_node_id>& nodes,
+                                       size_t bytes_per_chunk = 0) {
+    if (nodes.empty())
+        return nullptr;
+    return r1::allocate_interleaved(bytes, nodes.data(), nodes.size(), bytes_per_chunk);
+}
+
+inline void *allocate_numa_interleaved (size_t bytes, size_t bytes_per_chunk = 0) {
```
vossmjp:
```suggestion
inline void *allocate_numa_interleaved(size_t bytes, size_t bytes_per_chunk = 0) {
```

Extra space.

src/tbb/numa_allocation.cpp
Alexandr-Konovalov: But not for multiplatform case, so that's useful.

src/tbb/numa_allocation.cpp
Alexandr-Konovalov: Thanks! Can it be `(nodes_ids == nullptr) ^ (nodes_count != 0);`, i.e. without cast?

include/oneapi/tbb/numa_allocation.h
Alexandr-Konovalov: Thanks, missed that. Done.

include/oneapi/tbb/numa_allocation.h
Alexandr-Konovalov: Sure, opened #2075.

test/common/utils.h
aleksei-fedotov: Not sure I understand. `find_if_not` is a function template. So, the compiler generates the function implementation on each template argument specified. [Here](https://godbolt.org/z/dEhY81Mdo) is how I managed to make it work.

include/oneapi/tbb/numa_allocation.h
aleksei-fedotov: Okay, let's postpone this question.

src/tbb/numa_allocation.cpp
aleksei-fedotov: Better now, but I would like it to be a bit more explanatory, so that the user does not need to figure out the details themselves.

If I correctly figured out the technical details of this, then I suggest something like: _move_pages requires "numaif.h" header file. However, this makes the library to link with numactl. To avoid linkage with new dependency, we forward-declare the function instead._

test/common/utils.h
Alexandr-Konovalov: Not sure the result is better in term of readability, but let's see.

src/tbb/numa_allocation.cpp
Alexandr-Konovalov: I tried to improve the explanation.

src/tbb/numa_allocation.cpp
aleksei-fedotov: Shall we create one such common place? Perhaps, in src/common?

src/tbb/numa_allocation.cpp
aleksei-fedotov: Of course, it can. However, as we have seen generated code is identical. Thus, I guess it should be the variant that reads easier.

src/tbb/numa_allocation.cpp
Alexandr-Konovalov: I'm not sure one trivial function deserves this.

src/tbb/misc.h
```diff
@@ -121,6 +121,14 @@ T max ( const T& val1, const T& val2 ) {
     return val1 < val2 ? val2 : val1;
 }
 
+#if _WIN32 || _WIN64
+// currently used only for this OS
+template<typename T>
+static inline T align_up(T arg, uintptr_t alignment) {
+    return T(((uintptr_t)arg+(alignment-1)) & ~(alignment-1));
+}
+#endif
```
aleksei-fedotov: I think it is clear from the guarding macro. Better is to write why not have it for other platforms. It requires resolution of compilation issues or something else? Note that _include/oneapi/tbb/detail/_template_helpers.h_ does not have such guarding macros to avoid definitions on specific platforms.

I also think that `static` keyword is unnecessary, since there is `inline`.

Also, this file code style seems to add extra spaces around entities.

```suggestion
template<typename T>
inline T align_up( T arg, uintptr_t alignment ) {
    return T( ((uintptr_t)arg + (alignment - 1)) & ~(alignment - 1) );
}
```

src/tbb/numa_allocation.cpp
aleksei-fedotov: I think it is much better now. Thank you!

src/tbb/misc.h
Alexandr-Konovalov: No, the guarding macro is to stress that the function is used only there. I'm fine with dropping it.

test/common/utils.h
```diff
@@ -487,6 +487,30 @@ class UninitializedStorage {
...
+    intptr_t *word_it = std::find_if_not(buf, buf + words, [](intptr_t v) { return v == 0; });
+    if (word_it != buf + words) {
+        // find exact byte in non-zero word
+        char *bytes = (char*)word_it;
+        char *byte_it = std::find_if_not(bytes, bytes + sizeof(intptr_t),
+                                         [](unsigned char c) { return c == 0; });
+        return (byte_it - (char*)ptr) + 1;
+    }
+
+    char *tail_it = std::find_if_not(buf_tail, buf_tail + tail_sz, [](char c) { return c == 0; });
```
aleksei-fedotov: Since it might be confusing why one version compares with `unsigned char` while the other with signed `char`, I suggest extracting common code into small lambda function. Plus it improves readability a bit more as now it reads something like _find if not is char zero_ => _find non-zero char_
```suggestion
    auto is_char_zero = [](const char c) { return c == 0; };
    if (word_it != buf + words) {
        // find exact byte in non-zero word
        char *bytes = (char*)word_it;
        char *byte_it = std::find_if_not(bytes, bytes + sizeof(intptr_t),
                                         is_char_zero);
        return (byte_it - (char*)ptr) + 1;
    }

    char *tail_it = std::find_if_not(buf_tail, buf_tail + tail_sz, is_char_zero);
```

test/tbb/test_numa_allocation.cpp
Alexandr-Konovalov: Your example is partially artificial, because it exactly equivalent to
`char *ptr = (char *)tbb::allocate_numa_interleaved(obj_size);
` 
And if someone want to use just half of the nodes, she would work with `tbb::info::numa_nodes();` semantics on her side, so workaround inside `tbb::allocate_numa_interleaved(obj_size, numa_nodes);` doesn't help much.

But I agree, a naturally looking code must work as it looks like.

test/common/utils.h
Alexandr-Konovalov: Thanks, `unsigned` out of nowhere is my oversight.

But I'm not sure about common code that compare with 0. To me, one level of indirection adds complication. Could we ask someone?

test/tbb/test_numa_allocation.cpp
Alexandr-Konovalov: Just in case,
```
std::vector<tbb::numa_node_id> numa_nodes = tbb::info::numa_nodes();
char *ptr = (char *)tbb::allocate_numa_interleaved(obj_size, numa_nodes);
```
is working now, that's all misunderstanding from my side, caused by the unclear comment.
```
numa_nodes[0] = 0;
```
is needed only for ownership testing.

aleksei-fedotov: You are right - no need to specify `numa_nodes` for the function that is designed to allocate memory on all of them by default.

Since the question is about the semantics of `tbb::info::numa_nodes`,  I am resolving this comment.

test/tbb/test_numa_allocation.cpp
```diff
@@ -0,0 +1,201 @@
...
+        std::vector<tbb::numa_node_id> numa_nodes_1 = tbb::info::numa_nodes();
+        // we treat no-NUMA as single-NUMA with node index 0, but numa_nodes() return -1 in this case
+        if (numa_nodes_1.size() == 1)
+            numa_nodes_1[0] = 0;
+        numa_nodes.insert(numa_nodes.end(), numa_nodes_1.begin(), numa_nodes_1.end());
+        VerifySizeAndNodes(lib, obj_size, numa_nodes, 3 * page_size);
+    }
+    // explicitly check the basic allocation scenario
+    {
+        size_t obj_size = 8;
```
aleksei-fedotov: This is the overall size to allocate, right? Is not this too small for multi-NUMA allocation? Does it matter?

test/common/utils.h
aleksei-fedotov: I just wanted to address readability issue. I am fine with just fixing the `char` signedness.

test/tbb/test_numa_allocation.cpp
Alexandr-Konovalov: Agreed, 1MB is better.

test/common/utils.h
Alexandr-Konovalov: Thanks, done.

include/oneapi/tbb/detail/_config.h
```diff
@@ -578,4 +582,8 @@
 #define TBB_HAS_TASK_GROUP_WAIT_FOR_SINGLE_TASK 202603
 #endif
 
+#if __TBB_PREVIEW_NUMA_ALLOCATION
+#define TBB_HAS_NUMA_ALLOCATION 202605
+#endif
```
akukanov, line 587: Are these macros specific to the interleaved allocation, or are they considered to be potentially broader and apply to everything included into the "numa_allocation.h" header? That might influence the names.

src/tbb/numa_allocation.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+void interleaved_initialization_impl() {
+#if __linux__
+    dynamic_link("libnuma.so", LibnumaLinkTable,
+                 sizeof(LibnumaLinkTable) / sizeof(dynamic_link_descriptor), nullptr,
+                 DYNAMIC_LINK_GLOBAL | DYNAMIC_LINK_LOAD | DYNAMIC_LINK_WEAK);
+#elif _WIN32 || _WIN64
+    dynamic_link("kernelbase.dll", LibnumaLinkTable,
+                 sizeof(LibnumaLinkTable) / sizeof(dynamic_link_descriptor));
+#endif
```
aleksei-fedotov: Can we reduce clutter with something like:

```suggestion
#if __linux__
const char* numa_lib_name = "libnuma.so";
#elif _WIN32 || _WIN64
const char* numa_lib_name = "kernelbase.dll";
#endif
    dynamic_link(numa_lib_name, LibnumaLinkTable,
                 sizeof(LibnumaLinkTable) / sizeof(dynamic_link_descriptor), /*handle*/nullptr,
                 DYNAMIC_LINK_GLOBAL | DYNAMIC_LINK_LOAD | DYNAMIC_LINK_WEAK);
```

src/tbb/numa_allocation.cpp
```diff
@@ -0,0 +1,239 @@
...
+        nodes_per_page[page_idx] = nodes[stride_idx % nodes_count];
+    }
+    long ret = move_pages_ptr(/*pid=*/0, count_pages, pages.get(), nodes_per_page.get(), status.get(),
+                              /*flags=*/0);
+    if (ret < 0)
+        return nullptr;
+
+    for (size_t i = 0; i < count_pages; ++i)
+        if (status[i] < 0)
+            return nullptr;
```
aleksei-fedotov, line 161: I wonder how long this loop can be. Shall we also change to something more standard? Perhaps, use `find_if` instead?

src/tbb/governor.h
```diff
@@ -151,6 +152,8 @@ class governor {
     }
 }; // class governor
 
+const int* get_numa_nodes_indexes();
+
```
akukanov: Not sure why, but I see other functions defined in `governor.cpp` but not belonging to the corresponding class are declared in `misc.h` instead of `governor.h`. Should this function follow the same pattern?

src/tbb/numa_allocation.cpp
```diff
@@ -0,0 +1,239 @@
...
+         curr_size += bytes_per_chunk, ++node_index) {
+        // must release every but last page
+        if (curr_size + bytes_per_chunk < bytes) {
+            BOOL ok = VirtualFree(base_addr + curr_size, bytes_per_chunk,
+                                  MEM_RELEASE | MEM_PRESERVE_PLACEHOLDER);
+            if (!ok)
+                return nullptr;
+        }
+
+        MEM_EXTENDED_PARAMETER param = { {0, 0}, {0} };
```
aleksei-fedotov, line 206: I think zero initialization is enough and looks simpler. Microsoft uses same initialization in [its example](https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-virtualalloc2#:~:text=int%20numaNode\)%0A%7B-,MEM_EXTENDED_PARAMETER%20param%20%3D%20%7B0%7D%3B,-param.Type%20%3D%20MemExtendedParameterNumaNode).
```suggestion
        MEM_EXTENDED_PARAMETER param = { 0 };
```

test/common/utils.h
```diff
@@ -487,6 +487,30 @@ class UninitializedStorage {
...
 
+// if non-zero byte found, returns bad value offset plus 1
+inline size_t NonZero(void *ptr, size_t size)
+{
+    size_t words = size / sizeof(intptr_t);
+    size_t tail_sz = size % sizeof(intptr_t);
+    intptr_t *buf = (intptr_t*)ptr;
+    char *buf_tail = (char*)(buf+words);
+
+    intptr_t *word_it = std::find_if_not(buf, buf + words, [](intptr_t v) { return v == 0; });
```
akukanov: For readability, I would change it to `find_if` with `v != 0`

test/common/utils.h
```diff
@@ -487,6 +487,30 @@ class UninitializedStorage {
...
+        char *bytes = (char*)word_it;
+        char *byte_it = std::find_if_not(bytes, bytes + sizeof(intptr_t),
+                                         [](char c) { return c == 0; });
+        return (byte_it - (char*)ptr) + 1;
+    }
+
+    char *tail_it = std::find_if_not(buf_tail, buf_tail + tail_sz, [](char c) { return c == 0; });
+    if (tail_it != buf_tail + tail_sz) {
+        return (tail_it - (char*)ptr) + 1;
+    }
```
akukanov, line 507: Handling the tail and the exact position within a word is the same piece of code, differing just by the starting point and the length.

src/tbb/numa_allocation.cpp
```diff
@@ -0,0 +1,239 @@
...
+
+#endif
+
+namespace tbb {
+namespace detail {
+namespace r1 {
+
+#if __TBB_WEAK_SYMBOLS_PRESENT
+#pragma weak move_pages
+#endif /* __TBB_WEAK_SYMBOLS_PRESENT */
```
akukanov: Should that be `#if __linux__ && __TBB_WEAK_SYMBOLS_PRESENT`?

src/tbb/numa_allocation.cpp
```diff
@@ -0,0 +1,239 @@
...
+  PVOID                  BaseAddress,
+  SIZE_T                 Size,
+  ULONG                  AllocationType,
+  ULONG                  PageProtection,
+  MEM_EXTENDED_PARAMETER *ExtendedParameters,
+  ULONG                  ParameterCount
+);
+
+static const dynamic_link_descriptor LibnumaLinkTable[] = {
+    DLD(VirtualAlloc2, VirtualAlloc2_ptr)
```
aleksei-fedotov, line 68: Seems like the [`VirtualAllocExNuma`](https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-virtualallocexnuma) is more straightforward for our use case because it does not require the use of `VirtualFree` (I may be wrong here).
Why did not we use it instead of the `VirtualAlloc2`? Perhaps, we have already discussed that question and I merely have forgotten. But given that only fact, I think we should consider writing the explaining comment in the code.

src/tbb/numa_allocation.cpp
```diff
@@ -0,0 +1,239 @@
...
+        return false;
+    if (bytes_per_chunk % governor::default_page_size() != 0)
+        return false;
+    // nodes_ids and nodes_count must be consistent with each other,
+    // i.e. either nodes_ids is nullptr and nodes_count is 0 or vice versa
+    return (nodes_ids == nullptr) ^ (nodes_count != 0);
+}
+
+// interleaved memory allocation is only supported for those platforms
+#if __linux__ || _WIN32 || _WIN64
```
akukanov, line 84: There is almost nothing common between the implementations for Linux and for Windows. Instead of interspersing those, would not full separation result in cleaner code overall?

src/tbb/governor.h
Alexandr-Konovalov: Sure, for uniformity.

src/tbb/numa_allocation.cpp
```diff
@@ -0,0 +1,239 @@
...
+#include <memory> // for std::unique_ptr
+
+#if __linux__
+
+#include <sys/mman.h>
+
+// TBB build must be done without numaif.h, but we need signature of move_pages()
+// for dynamic loading, so declare it here.
+extern "C" long move_pages(int pid, unsigned long count,
+                           void **pages, const int *nodes, int *status, int flags);
```
aleksei-fedotov, line 33: And I guess we do not use `mbind(..., MPOL_INTERLEAVE)` because it does not allow to specify the chunk granularity. So, if we used it, we would end up with similar to Windows pattern where we need to call the _pages moving_ function several times - one per each chunk. Am I right?

src/tbb/governor.h
```diff
@@ -1,5 +1,6 @@
 /*
     Copyright (c) 2005-2024 Intel Corporation
+    Copyright (c) 2026 UXL Foundation Contributors
```
dnmokhov: revert

aleksei-fedotov (commented):
Mostly looks good to me. I have a few principal questions though.

test/common/utils.h
Alexandr-Konovalov: Done.

src/tbb/governor.h
Alexandr-Konovalov: Thanks, fixed.

src/tbb/numa_allocation.cpp
akukanov: Below is why `VirtualAllocExNuma` does not work for our case:
> [in] nndPreferred
The NUMA node where the physical memory should reside.
_**Used only when allocating a new VA region (either committed or reserved). Otherwise this parameter is ignored when the API is used to commit pages in a region that already exists**_

In other words, it only allows to map a new memory region to a single node.

src/tbb/numa_allocation.cpp
Alexandr-Konovalov: Yes, syscalls are not cheap, it's better to not have them O(number of pages).

But this optimization can be applied for case of one-page interleaving. I'm going to do it after the feature freeze.

src/tbb/numa_allocation.cpp
Alexandr-Konovalov: I tried and unable to implement interleaving with this function.

test/common/utils.h
Alexandr-Konovalov: Done.

src/tbb/numa_allocation.cpp
Alexandr-Konovalov: Done.

src/tbb/numa_allocation.cpp
Alexandr-Konovalov: I'm not sure, but let's try.

src/tbb/numa_allocation.cpp
Alexandr-Konovalov: I wrote it this way because Intel compiler is unhappy with the code you suggested.

```
C:\Users\akonoval\src\oneTBB\src\tbb\numa_allocation.cpp(206,44): error: missing field 'Reserved' initializer [-Werror,-Wmissing-field-initializers]
  206 |         MEM_EXTENDED_PARAMETER param = { 0 };
      |                                            ^
```

include/oneapi/tbb/detail/_config.h
vossmjp: Good question. We could probably go either way. More limited would be `TBB_HAS_ALLOCATE_NUMA_INTERLEAVED`.  I don't have a very strong preference, but I'm ok with it as it is with `TBB_HAS_NUMA_ALLOCATION`, so wouldn't ask for a change. If we add more, then the check for newer things has to check the value, which I think is ok. Does anyone else have a strong preference either way?

src/tbb/numa_allocation.cpp
Alexandr-Konovalov: Could we somehow discuss, are all loops deserve to be replaced to `std::for_each` and why? Is current version not readable? Or has lower performance?

src/tbb/numa_allocation.cpp
```diff
@@ -0,0 +1,239 @@
...
+    atomic_do_once(interleaved_initialization_impl, interleaved_initialization_state);
+
+    if (!is_args_valid(bytes, nodes_ids, nodes_count, bytes_per_chunk))
+        return nullptr;
+
+    if (!bytes_per_chunk)
+        bytes_per_chunk = governor::default_page_size();
+    const int *nodes = nodes_count ? nodes_ids : get_numa_nodes_indexes();
+    if (!nodes_count)
+        nodes_count = numa_node_count();
```
dnmokhov, line 110: This introduces a dependency on TBBBind. There’s nothing wrong with that, but I wanted to highlight two points:

1. The dependency isn’t strictly necessary. Since `libnuma.so` is already being used, could we rely on it alone, even for the NUMA node indices? This could be helpful for users of the NUMA allocation API who don’t need arena constraints and therefore don’t require TBBBind.
2. If TBBBind is expected for the NUMA allocation API to function, we should consider documenting that requirement.

These are just suggestions for the future. The current preview implementation looks good.

Alexandr-Konovalov: Thanks, good point. Yes, the functionality can be made TBBBind-independent. At least for use cases when a user not provide us with list of NUMA nodes to allocate from, I think, that is exactly the common case.

include/oneapi/tbb/detail/_config.h
akukanov: I have no strong preference either.

test/common/utils.h
```diff
@@ -487,6 +487,27 @@ class UninitializedStorage {
...
+    if (word_it != buf + words) {
+        // find exact byte in non-zero word
+        buf_tail = (char*)word_it;
+        tail_sz = sizeof(intptr_t);
+    }
+    char *tail_it = std::find_if(buf_tail, buf_tail + tail_sz, [](char c) { return c != 0; });
+    if (tail_it != buf_tail + tail_sz) {
+        return (tail_it - (char*)ptr) + 1;
+    }
+    return 0;
```
akukanov, line 508: I think it can be further simplified. First, use `find_if` over the words to find a non-zero one; if not found, the result will point exactly to the tail. And so the second step is to search for a single byte, starting from the result of the first search and till the end of data - it will not go beyond a single word anyway. And maybe this last part is better done as the regular loop with array indices, computing the position to return right away.

src/tbb/numa_allocation.cpp
```diff
@@ -0,0 +1,260 @@
...
+// fallback implementation with malloc/free
+void *__TBB_EXPORTED_FUNC allocate_interleaved(size_t bytes,
+                        const tbb::detail::d1::numa_node_id *nodes_ids, size_t nodes_count,
+                        size_t bytes_per_chunk) {
+    return is_args_valid(bytes, nodes_ids, nodes_count, bytes_per_chunk) ?
+        calloc(bytes, 1) : nullptr;
+}
+
+void __TBB_EXPORTED_FUNC deallocate_interleaved(void *ptr, size_t bytes) {
+    (void)bytes;
```
akukanov, line 252:
```suggestion
void __TBB_EXPORTED_FUNC deallocate_interleaved(void *ptr, size_t) {
```

src/tbb/numa_allocation.cpp
```diff
@@ -0,0 +1,260 @@
...
+#endif // _WIN32 || _WIN64
+
+void __TBB_EXPORTED_FUNC deallocate_interleaved(void *ptr, size_t bytes) {
+    atomic_do_once(interleaved_initialization_impl, interleaved_initialization_state);
+
+    // TODO: process return value of munmap()/VirtualFree()
+#if __linux__
+    munmap(ptr, bytes);
+#elif _WIN32 || _WIN64
+    (void)bytes;
```
akukanov, line 236:
```suggestion
#elif _WIN32 || _WIN64
    suppress_unused_warning(bytes);
```

vossmjp (approved):
I think the remaining suggested changes can happen after feature freeze and even later before production. Some are stylistic choices. Overall, I think it's good enough for preview. If you want to make additional changes now to address some of the ongoing discussions that's ok.

akukanov (commented):
More suggestions for follow-up changes.

src/tbb/numa_allocation.cpp
aleksei-fedotov: I once saw such a manual loop performing much more slower than the standard-based variant. I admit that I might be lucky to encounter that situation. On the other hand, I believe that standard algorithms are better tested and tuned to various execution circumstances.
Ideally, if you could compare different versions of code pieces to make more weighted decision regarding the code version to use. This again brings us to the question about microbenchmarking.

For this particular case, I have considered these loops over 100GB of data with 4K page size, which results in more than 26 million loop iterations. Could be something to consider.

Alexandr-Konovalov: libstdc++'s `find_if` is pre-unrolled to 4 iterations, and that is just for-loop. It's strange a compiler need help in such case. Interesting.

test/common/utils.h
Alexandr-Konovalov: I opened #2078 for that.

> And so the second step is to search for a single byte, starting from the result of the first search and till the end of data - it will not go beyond a single word anyway.

Not sure I able to use all optimization possibility here. Now it's just either full word or rest of the buffer.

>  And maybe this last part is better done as the regular loop with array indices, computing the position to return right away.

I don't know, but for me it's more uniform when both cases are written with `std::find_if()`. But having index right away in the implementation with loop is also attractive.

PR #2077, opened 2026-05-13
Factor out align_up* functions in tbb library.

Alexandr-Konovalov:
### Description 
There are 2 kind of alignment up in TBB sources, alignment up to nearest aligned and alignment to next aligned, if an argument is already aligned. Readability is improved, if we name them explicitly. 

### Type of change

### Tests

- [X] not needed

### Documentation

- [X] needs to be updated
- [ ] not needed

- [ ] Yes
- [X] No
- [ ] Unknown

### DIFF
--- a/include/oneapi/tbb/detail/_utils.h
+++ b/include/oneapi/tbb/detail/_utils.h
@@ -248,7 +248,7 @@ inline T align_up(T arg, std::uintptr_t alignment) {
     return T( ((std::uintptr_t)arg + (alignment - 1)) & ~(alignment - 1) );
 }
 
-// if arg is aleady aligned, returns the next aligned value after arg
+// if arg is already aligned, returns the next aligned value after arg
 template<typename T>
 inline T align_up_next(T arg, std::uintptr_t alignment) {
     return T( ((std::uintptr_t)arg + (alignment)) & ~(alignment - 1) );
--- a/include/oneapi/tbb/detail/_utils.h
+++ b/include/oneapi/tbb/detail/_utils.h
@@ -243,6 +243,17 @@ constexpr bool is_aligned(T* pointer, std::uintptr_t alignment) {
     return 0 == (reinterpret_cast<std::uintptr_t>(pointer) & (alignment - 1));
 }
 
+template<typename T>
+inline T align_up(T arg, std::uintptr_t alignment) {
+    return T( ((std::uintptr_t)arg + (alignment - 1)) & ~(alignment - 1) );
+}
+
+// if arg is aleady aligned, returns the next aligned value after arg
+template<typename T>
+inline T align_up_next(T arg, std::uintptr_t alignment) {
+    return T( ((std::uintptr_t)arg + (alignment)) & ~(alignment - 1) );
+}
+
 #if TBB_USE_ASSERT
 __TBB_GLOBAL_VAR void* const poisoned_ptr = reinterpret_cast<void*>(-1);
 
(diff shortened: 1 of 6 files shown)


include/oneapi/tbb/detail/_utils.h
```diff
@@ -243,6 +243,19 @@ constexpr bool is_aligned(T* pointer, std::uintptr_t alignment) {
...
+    return T( ((std::uintptr_t)arg + (alignment - 1)) & ~(alignment - 1) );
+}
+
+// if arg is already aligned, returns the next aligned value after arg
+template<typename T>
+inline T align_up_next(T arg, std::uintptr_t alignment) {
+    __TBB_ASSERT(alignment != 0 && is_power_of_two(alignment), "Alignment should be a non-zero power of two");
+    return T( ((std::uintptr_t)arg + (alignment)) & ~(alignment - 1) );
+}
+
```
akukanov: Would it make more sense to remove this function, and instead call `align_up(arg + 1, alignment)`?

Alexandr-Konovalov: As there is a important difference in the semantics between the functions, I think they deserve different names in the code, not only explicit comments.

akukanov: That's fair, but unfortunately, I find it hard to comprehend what the semantical difference is, based on just the proposed names. And the result is only different when the argument is already aligned - i.e., the names should have the same meaning yet reflect the difference for that "already aligned" scenario. Perhaps more verbose name(s) could achieve that.

The use case for this function is to reserve some "header" space after the allocation. I think the code might be clearer with explicit "reserve, then align up" approach: `align_up(base + sizeof(uintptr_t), alignment)`.
As an option, we can also consider "align down, then reserve": `align_down(base, alignment) + alignment`.

Alexandr-Konovalov: I agree that `_next` is too cryptic. Can it it `align_up_unconditionally`?

include/oneapi/tbb/detail/_utils.h
```diff
@@ -243,6 +243,19 @@ constexpr bool is_aligned(T* pointer, std::uintptr_t alignment) {
     return 0 == (reinterpret_cast<std::uintptr_t>(pointer) & (alignment - 1));
 }
 
+template<typename T>
+inline T align_up(T arg, std::uintptr_t alignment) {
+    __TBB_ASSERT(alignment != 0 && is_power_of_two(alignment), "Alignment should be a non-zero power of two");
+    return T( ((std::uintptr_t)arg + (alignment - 1)) & ~(alignment - 1) );
+}
+
+// if arg is already aligned, returns the next aligned value after arg
```
aleksei-fedotov: What if `arg` comes unaligned? Shall the function enforce its expectation regarding `arg` alignment? Via an assert, for example.

Alexandr-Konovalov: Thanks, the comment must be extended.

aleksei-fedotov: I think it is the `if` part that allows confusion. I would remove it and write simply _Finds the next aligned value after arg_

include/oneapi/tbb/detail/_utils.h
akukanov: Maybe `align_to_greater` and `align_to_greater_or_equal` could work.

include/oneapi/tbb/detail/_utils.h
Alexandr-Konovalov: To me, it's definitely correct, but probably a bit too concise. I don't know.

include/oneapi/tbb/detail/_utils.h
Alexandr-Konovalov: Done.

akukanov (approved):

aleksei-fedotov:
_src/tbbmalloc/shared_utils.h_ contains similar definitions of alignment helpers. Will it be useful to combine and share the implementation for their clients as well?

aleksei-fedotov (approved):
Looks good to me as well.

PR #2078, opened 2026-05-15
Improve utils::NonZero and add test for it

Alexandr-Konovalov:
### Description 

Improve readability of utils::NonZero() and add test for it.

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [X] not needed

- [ ] Yes
- [X] No
- [ ] Unknown

### DIFF
--- a/test/common/utils.h
+++ b/test/common/utils.h
@@ -491,7 +491,7 @@ public:
 inline size_t NonZero(const void *ptr, size_t size)
 {
     REQUIRE_MESSAGE((uintptr_t)(ptr) % sizeof(intptr_t) == 0,
-                    "Unaligned access in unefficient, but usage model for NonZero assumes"
+                    "Unaligned access is inefficient, but usage model for NonZero assumes"
                     " that pointer is aligned to word size");
 
     size_t words = size / sizeof(intptr_t);
--- a/test/common/utils.h
+++ b/test/common/utils.h
@@ -488,24 +488,22 @@ public:
 };
 
 // if non-zero byte found, returns bad value offset plus 1
-inline size_t NonZero(void *ptr, size_t size)
+inline size_t NonZero(const void *ptr, size_t size)
 {
-    size_t words = size / sizeof(intptr_t);
-    size_t tail_sz = size % sizeof(intptr_t);
-    intptr_t *buf = (intptr_t*)ptr;
-    char *buf_tail = (char*)(buf+words);
+    REQUIRE_MESSAGE((uintptr_t)(ptr) % sizeof(intptr_t) == 0,
+                    "Unaligned access in unefficient, but usage model for NonZero assumes"
+                    " that pointer is aligned to word size");
 
-    intptr_t *word_it = std::find_if(buf, buf + words, [](intptr_t v) { return v != 0; });
-    if (word_it != buf + words) {
-        // find exact byte in non-zero word
-        buf_tail = (char*)word_it;
-        tail_sz = sizeof(intptr_t);
-    }
-    char *tail_it = std::find_if(buf_tail, buf_tail + tail_sz, [](char c) { return c != 0; });
-    if (tail_it != buf_tail + tail_sz) {
-        return (tail_it - (char*)ptr) + 1;
-    }
-    return 0;
+    size_t words = size / sizeof(intptr_t);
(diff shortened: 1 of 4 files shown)


Alexandr-Konovalov:
I don't know where is the right place for tests for `utils::NonZero()`. It seems a separate exe has too large overhead.

test/common/utils.h
```diff
@@ -488,24 +488,22 @@ class UninitializedStorage {
...
-    return 0;
+    const intptr_t *buf = (const intptr_t*)ptr;
+
+    const intptr_t *word_it = std::find_if(buf, buf + words, [](intptr_t v) { return v != 0; });
+    const char *char_it = (const char *)word_it;
+    // If points to end, search in tail. Otherwise, find exact byte in the non-zero word.
+    size_t tail_sz = word_it == buf + words ? size % sizeof(intptr_t) : sizeof(intptr_t);
+    const char *tail_it = std::find_if(char_it, char_it + tail_sz,
+                                       [](char c) { return c != 0; });
+    return tail_it == char_it + tail_sz ? 0 : (tail_it - (const char*)ptr) + 1;
```
akukanov: It's not critical, yet I still think the following code would be clearer, as it has less manipulations with variables:
```c++
size_t offset = (word_it - buf) * sizeof(intptr_t);
// Looking for the first non-zero byte in the non-zero word or in the buffer tail
for (; offset < size; ++offset) {
    if (*((const char*)ptr + offset) != 0)
        return offset + 1;
}
return 0;
```
It is sufficient to use `size` as the only loop bound (instead of computing a "tail size"), because in the case of successful `find_if` we know for sure a non-zero byte will be found within the first word and the loop bound is never reached.
Of course this code can be converted to a `find_if` call as well, but it will not have the same clear separation between two possible outcomes like in the regular loop.

Alexandr-Konovalov: Do you think is it also important not use array notation vs pointer arithmetic?

aleksei-fedotov: I would also avoid casting inside the loop and use increment instead of addition to have something like the following C-style laconic loop:
```cpp
const char *char_it = (const char*)word_it;
const char *end_it = (const char*)ptr + size;
for (; char_it != end_it; ++char_it) {
    if (*char_it)
        return char_it - (const char*)ptr + 1;
}
return 0;
```

test/common/utils.h
```diff
@@ -488,24 +488,22 @@ class UninitializedStorage {
 };
 
 // if non-zero byte found, returns bad value offset plus 1
-inline size_t NonZero(void *ptr, size_t size)
+inline size_t NonZero(const void *ptr, size_t size)
```
aleksei-fedotov: Conventionally, functions represent actions and therefore contain a verb in its name. Consider naming it in similar manner. Perhaps, `find_non_zero` or `find_non_zero_byte_offset` could suffice.

test/common/utils.h
akukanov: > Do you think is it also important not use array notation vs pointer arithmetic?

No, I would not mind using array notation either.

Alexandr-Konovalov: > It is sufficient to use `size` as the only loop bound (instead of computing a "tail size"), because in the case of successful `find_if` we know for sure a non-zero byte will be found within the first word and the loop bound is never reached.

Well, I don't know. In terms of correctness you are right, but I think it's might be important for the compiler to know that the number of iterations is always limited by `sizeof(intptr_t)`. In this case, the compiler might easily unroll the loop. In your code, `size` is run-time parameter and the compiler doesn't know that a non-zero byte must be found within the first word. 

It's hard to judge whether it's significant or not, however.

akukanov: This is not a performance sensitive code, is it?
Anyway, up to you, it is absolutely not critical.

test/common/utils.h
Alexandr-Konovalov: I think we use PascalCase in the tests, so it can be `FindNonZero()`.

test/common/utils.h
Alexandr-Konovalov: Sorry, it was misunderstanding from my side, actually all 3 major compilers can't eliminate the loop. And it's better to delegate unrolling to a compiler.

PR #2079, opened 2026-05-15
Fix a race in task group dynamic dependencies implementation

kboyarinov:
### Description 
Fix for a tricky race in task group dynamic dependencies.

Consider the following situation:
- Thread A is adding a successor to task T.
- Thread B is executing T which transfers its successors to TT, which executes and completes immediately.

To reproduce the issue, thread A should first detect T as not-completed, not-transferred to bypass initial checks. In this case, it allocates the successor notify node that will be inserted into T's notify list.
After that, thread B transfers T's successors to TT and completes TT (T's notify list is in TRANSFERRED state, TT's - in COMPLETED state).

thread A is then trying to insert new notify node into T's notify list, detects that it is in TRANSFERRED state and redirects the node to TT. The bug is that the state of TT is not checked at this point - the node is just inserted into TT's notify list, which was already processed. Therefore, a successor task will never be executed.

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown

### DIFF
--- a/include/oneapi/tbb/detail/_task_handle.h
+++ b/include/oneapi/tbb/detail/_task_handle.h
@@ -1,6 +1,6 @@
 /*
     Copyright (c) 2020-2025 Intel Corporation
-    Copyright (c) 2025 UXL Foundation Contributors
+    Copyright (c) 2025-2026 UXL Foundation Contributors
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
--- a/include/oneapi/tbb/detail/_task_handle.h
+++ b/include/oneapi/tbb/detail/_task_handle.h
@@ -1,6 +1,6 @@
 /*
     Copyright (c) 2020-2025 Intel Corporation
-    Copyright (c) 2025-2026 UXL Foundation Contributors
+    Copyright (c) 2025 UXL Foundation Contributors
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
--- a/include/oneapi/tbb/detail/_task_handle.h
+++ b/include/oneapi/tbb/detail/_task_handle.h
@@ -147,6 +147,7 @@ public:
     task_group_status wait_for_completion(d1::task_group_context&);
     task_group_status run_self_and_wait_for_completion(d1::task_group_context&);
     void add_notify_node(notify_list_node* new_notify_node, notify_list_node* current_notify_list_head);
+    void resolve_completion_state_and_add_notify_node(notify_list_node* new_notify_node);
     void add_notify_list(notify_list_node* notify_list);
 
     using notify_list_state_flag = std::uintptr_t;
@@ -411,10 +412,30 @@ inline bool operator!=(std::nullptr_t, task_handle const& th) noexcept {
 }
 
 #if __TBB_PREVIEW_TASK_GROUP_EXTENSIONS
+inline void task_dynamic_state::resolve_completion_state_and_add_notify_node(notify_list_node* new_notify_node) {
+    notify_list_node* current_notify_list_head = m_notify_list_head.load(std::memory_order_acquire);
+
+    if (represents_completed_task(current_notify_list_head)) {
+        new_notify_node->notify_on_completion();
+    } else if (represents_canceled_task(current_notify_list_head)) {
+        new_notify_node->notify_on_cancellation();
+    } else if (represents_transferred_completion(current_notify_list_head)) {
+        // Redirect notify_node to the task received the completion
+        task_dynamic_state* new_completion_point = m_new_completion_point.load(std::memory_order_relaxed);
+        __TBB_ASSERT(new_completion_point, "notify list is marked as transferred, but new dynamic state is not set");
+        new_completion_point->resolve_completion_state_and_add_notify_node(new_notify_node);
+    } else {
+        add_notify_node(new_notify_node, current_notify_list_head);
+    }
+}
+
 inline void task_dynamic_state::add_notify_node(notify_list_node* new_notify_node,
                                                 notify_list_node* current_notify_list_head)
 {
     __TBB_ASSERT(new_notify_node != nullptr, nullptr);
+    __TBB_ASSERT(!represents_completed_task(current_notify_list_head) &&
+                 !represents_canceled_task(current_notify_list_head) &&
+                 !represents_transferred_completion(current_notify_list_head), nullptr);
 
     new_notify_node->next_node = current_notify_list_head;
 
@@ -433,7 +454,7 @@ inline void task_dynamic_state::add_notify_node(notify_list_node* new_notify_nod
             // Redirect notify_node to the task received the completion
             task_dynamic_state* new_completion_point = m_new_completion_point.load(std::memory_order_relaxed);
             __TBB_ASSERT(new_completion_point, "notify list is marked as transferred, but new dynamic state is not set");
-            new_completion_point->add_notify_node(new_notify_node, new_completion_point->m_notify_list_head.load(std::memory_order_acquire));
+            new_completion_point->resolve_completion_state_and_add_notify_node(new_notify_node);
             break;
         }
 
--- a/include/oneapi/tbb/detail/_task_handle.h
+++ b/include/oneapi/tbb/detail/_task_handle.h
@@ -419,14 +419,16 @@ inline void task_dynamic_state::add_notify_node(notify_list_node* new_notify_nod
                  !represents_canceled_task(current_notify_list_head) &&
                  !represents_transferred_completion(current_notify_list_head), nullptr);
 
+    new_notify_node->next_node = current_notify_list_head;
+
     // current_state walks through the transfer chain to find the correct list on which to
     // retry the insertion. Rechecking for sentinel values is required for each list in the transfer chain
     task_dynamic_state* current_state = this;
     bool recheck_current_state = false;
 
     while (recheck_current_state ||
-           current_state->m_notify_list_head.compare_exchange_strong(current_notify_list_head, new_notify_node,
-                                                                     std::memory_order_release, std::memory_order_relaxed))
+           !current_state->m_notify_list_head.compare_exchange_strong(current_notify_list_head, new_notify_node,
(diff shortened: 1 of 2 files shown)


dnmokhov (dismissed):
LGTM (fix the failing copyright check though)

include/oneapi/tbb/detail/_task_handle.h
```diff
@@ -1,6 +1,6 @@
 /*
     Copyright (c) 2020-2025 Intel Corporation
-    Copyright (c) 2025 UXL Foundation Contributors
+    Copyright (c) 2025-2026 UXL Foundation Contributors
```
isaevil:
```suggestion
    Copyright (c) 2025 UXL Foundation Contributors
```

include/oneapi/tbb/detail/_task_handle.h
```diff
@@ -411,10 +412,30 @@ inline bool operator!=(std::nullptr_t, task_handle const& th) noexcept {
...
+    } else if (represents_transferred_completion(current_notify_list_head)) {
+        // Redirect notify_node to the task received the completion
+        task_dynamic_state* new_completion_point = m_new_completion_point.load(std::memory_order_relaxed);
+        __TBB_ASSERT(new_completion_point, "notify list is marked as transferred, but new dynamic state is not set");
+        new_completion_point->resolve_completion_state_and_add_notify_node(new_notify_node);
+    } else {
+        add_notify_node(new_notify_node, current_notify_list_head);
+    }
+}
+
```
aleksei-fedotov: As I see it, this is almost a complete extraction of a single iteration from the `while` loop below. It introduces a potential stackoverflow issue, because it might continue calling the same function again and again if transfers keep going.
Instead of introducing a new function, I suggest extracting the common part where you re-read the `m_notify_list_head` with acquire semantics and update the `current_notify_list_head` inside the loop to avoid duplication.

kboyarinov: Rewritten to avoid code duplication

include/oneapi/tbb/detail/_task_handle.h
```diff
@@ -437,28 +419,50 @@ inline void task_dynamic_state::add_notify_node(notify_list_node* new_notify_nod
...
+            // The recipient may already be completed or canceled, or its completion may itself have been transferred
+            // re-checking will be performed by the next iteration of the while loop
+            continue;
         }
 
+        // current_notify_list_head is a regular list node. Try to insert the node into the list
         new_notify_node->next_node = current_notify_list_head;
+        if (current_state->m_notify_list_head.compare_exchange_strong(current_notify_list_head, new_notify_node)) {
+            break;
+        }
```
aleksei-fedotov: To my taste the `do .. while ( !CAS );` loop looks more laconic here.

Also, given that the subsequent consumer of this pointer `acquires` the memory, I believe `std::memory_order_release` on success and `relaxed` on failure is enough here.

include/oneapi/tbb/detail/_task_handle.h
```diff
@@ -437,28 +419,50 @@ inline void task_dynamic_state::add_notify_node(notify_list_node* new_notify_nod
                  !represents_canceled_task(current_notify_list_head) &&
                  !represents_transferred_completion(current_notify_list_head), nullptr);
 
+    // Try to insert new_notify_node into the notify list
+    // Calling functions have already checked current_notify_list_head for sentinel values
```
aleksei-fedotov: I am not sure the CAS below is important here. I think that the loop body can take care on its own and does not produce much overhead from the double checks performed one more time in the loop body just before the same CAS because the values are most likely in the registers nearby since you say the functions up the stack has already checked for sentinels.
So, to me this CAS here is more of a readability sacrifice rather than an optimization.

aleksei-fedotov (commented):
Looks much better to me.
A couple of additional comments regarding the writing style and memory semantics.

include/oneapi/tbb/detail/_task_handle.h
kboyarinov: The issue with `do .. while (!CAS);` is that on the transfer branch we call `continue` after resetting the `current_notify_list_head`. After that, the `do... while(!CAS)` will check the condition, i.e. execute the CAS without re-checking the sentinel values on the recipient side.

Regarding the memory semantics, I agree.

include/oneapi/tbb/detail/_task_handle.h
kboyarinov: Agree

include/oneapi/tbb/detail/_task_handle.h
kboyarinov: The memory semantics is fixed

include/oneapi/tbb/detail/_task_handle.h
kboyarinov: Fixed

include/oneapi/tbb/detail/_task_handle.h
aleksei-fedotov: Oh, I have completely forgotten about that. Thanks. In this case, my suggestion won't work because the loop is used for two purposes: 1. Try inserting the node into the list; 2. Change the list where to insert the node with checking sentinel values of that updated list first.

In this case the loop condition can be modified to reflect that two-purpose intention more explicitly. E.g.:

```cpp
bool recheck;
task_dynamic_state* current_state = this;
do {
  recheck = false;
  if (completed) { /* ... */ break; }
  else if (cancelled) { /* ... */ break; }
  else if (transferred) {
    /* new list head and current_state are reloaded here */
    recheck = true;
  } else
    new_notify_node->next_node = current_notify_list_head;
while (recheck || !CAS(current_state->m_notify_list_head);
```
Furthermore, reading through one of you previous comments again _calling functions already checked current_notify_list_head for sentinel values_, makes me thinking that the `do .. while` can be replaced with `while { }` without issues. In this case, the execution will avoid checking for sentinel values again on the first loop iteration as it is likely to insert the node successfully at the first attempt, which I think is going to be in ~90% of the cases. So, the loop will be something like:
```cpp
bool recheck = false;
task_dynamic_state* current_state = this;
while (recheck || !CAS(current_state->m_notify_list_head)) {
  recheck = false;
  if (completed) { /* ... */ break; }
  else if (cancelled) { /* ... */ break; }
  else if (transferred) {
    /* new list head and current_state are reloaded here */
    recheck = true;
  } else
    new_notify_node->next_node = current_notify_list_head;
}
```

aleksei-fedotov (dismissed):
The code is good enough.

You may disregard comment below as just nitpicking, but I'll appreciate if you consider integrating it and get an A+ on this PR.

include/oneapi/tbb/detail/_task_handle.h
kboyarinov: Good catch! I would definitely prefer the loop with the clear exit condition instead of an endless one. 
Applied

include/oneapi/tbb/detail/_task_handle.h
```diff
@@ -415,29 +415,46 @@ inline void task_dynamic_state::add_notify_node(notify_list_node* new_notify_nod
...
+    // current_state walks through the transfer chain to find the correct list on which to
+    // retry the insertion. Rechecking for sentinel values is required for each list in the transfer chain
+    task_dynamic_state* current_state = this;
+    bool recheck_current_state = false;
+
+    while (recheck_current_state ||
+           !current_state->m_notify_list_head.compare_exchange_strong(current_notify_list_head, new_notify_node,
+                                                                      std::memory_order_release, std::memory_order_relaxed))
+    {
+        // The recheck was requested because of switching to the next node to the transfer chain
```
aleksei-fedotov: A typo?
```suggestion
        // The recheck was requested because of switching to the next node in the transfer chain
```

include/oneapi/tbb/detail/_task_handle.h
```diff
@@ -415,29 +415,46 @@ inline void task_dynamic_state::add_notify_node(notify_list_node* new_notify_nod
...
-    while (!m_notify_list_head.compare_exchange_strong(current_notify_list_head, new_notify_node)) {
-        // Other thread updated the head of the list
+    // current_state walks through the transfer chain to find the correct list on which to
+    // retry the insertion. Rechecking for sentinel values is required for each list in the transfer chain
+    task_dynamic_state* current_state = this;
+    bool recheck_current_state = false;
+
+    while (recheck_current_state ||
+           !current_state->m_notify_list_head.compare_exchange_strong(current_notify_list_head, new_notify_node,
+                                                                      std::memory_order_release, std::memory_order_relaxed))
```
aleksei-fedotov: I would keep the lines within 100 characters to simplify reading.
```suggestion
           !current_state->m_notify_list_head.compare_exchange_strong(current_notify_list_head,
                                            new_notify_node,
                                            std::memory_order_release,
                                            std::memory_order_relaxed))
```

aleksei-fedotov (dismissed):
Noticed small flaws though.

include/oneapi/tbb/detail/_task_handle.h
```diff
@@ -415,29 +415,48 @@ inline void task_dynamic_state::add_notify_node(notify_list_node* new_notify_nod
...
+    // current_state walks through the transfer chain to find the correct list on which to
+    // retry the insertion. Rechecking for sentinel values is required for each list in the transfer chain
+    task_dynamic_state* current_state = this;
+    bool recheck_current_state = false;
+
+    while (recheck_current_state ||
+           !current_state->m_notify_list_head.compare_exchange_strong(current_notify_list_head,
+                                            new_notify_node,
+                                            std::memory_order_release,
+                                            std::memory_order_relaxed))
```
aleksei-fedotov: Sorry, I meant a bit different alignment. I will try applying it myself.
```suggestion
           !current_state->m_notify_list_head.compare_exchange_strong(current_notify_list_head,
                                                                      new_notify_node,
                                                                      std::memory_order_release,
                                                                      std::memory_order_relaxed))
```

aleksei-fedotov (approved):
Looks good to me! Thanks!

PR #2082, opened 2026-05-19
[tbbmalloc] Add assertions about computed bin indices and masks

aleksei-fedotov:


src/tbbmalloc/frontend.cpp
```diff
@@ -1579,11 +1579,12 @@ void Block::initEmptyBlock(TLSData *tls, unsigned size)
...
     // allows better compiler optimization as they basically share the code.
-    unsigned int index = getIndex(size);
-    unsigned int objSz = getObjectSize(size);
+    const unsigned int index = getIndex(size);
+    const unsigned int objSz = getObjectSize(size);
 
     cleanBlockHeader();
-    MALLOC_ASSERT(objSz <= USHRT_MAX, "objSz must not be less 2^16-1");
+    MALLOC_ASSERT(index < numBlockBinLimit, "index must be within the array bounds");
+    MALLOC_ASSERT(objSz <= USHRT_MAX, "objSz must be less than 2^16-1");
```
Alexandr-Konovalov: Can we not re-phrase conditions in plain English, but write why we think it's impossible? I.e., objSz is limited by fittingSize5.

src/tbbmalloc/large_objects.cpp
```diff
@@ -844,7 +844,8 @@ void LargeObjectCache::reset()
 template<typename Props>
 LargeMemoryBlock *LargeObjectCacheImpl<Props>::get(ExtMemoryPool *extMemoryPool, size_t size)
 {
-    unsigned idx = Props::sizeToIdx(size);
+    const unsigned idx = Props::sizeToIdx(size);
+    MALLOC_ASSERT(idx < numBins, "Array out of bounds access found.");
```
Alexandr-Konovalov: Probably we may want to move the assert in the `sizeToIdx()` body, i.e. it never returns anything above `numBins` rather then repeat the assertion at each callsite.

src/tbbmalloc/frontend.cpp
aleksei-fedotov: Note that only _not_ particle was removed as the comment would be confusing once the message appeared on the console.

`fittingSize5` is indeed the maximum size of the object. I will use it instead of `USHRT_MAX` constant. Thus, making the condition even more strict.

In this case, the assert message again becomes just rewriting of the condition in plain English. Actually, I am not sure what to write in the message to make it useful.

Alexandr-Konovalov: Thanks, good point.

src/tbbmalloc/large_objects.cpp
aleksei-fedotov: Moved.

Alexandr-Konovalov (dismissed):
LGTM.

src/tbbmalloc/large_objects.cpp
aleksei-fedotov: Unfortunately, this does not work for all the cases. There is one case, where the `sizeToIdx` is called with the `maxHugeSize` value to initialize the `hugeSizeThresholdIdx` with the maximum possible value, which equals to 136 in the implementation. Since `sizeToIdx` cannot differentiate the reasons it was called, I moved the assertions out of the function, just before the array access. In some sense this is even better since now a reader can clearly see that the result of `sizeToIdx` is used to access the array of `numBins` size.

Alexandr-Konovalov: Ahh, 
`        hugeCache.hugeSizeThresholdIdx = HugeCacheType::sizeToIdx(hugeSizeThreshold);
`

Alexandr-Konovalov (approved):
LGTM.

PR #2087, opened 2026-05-21
Allocate memory directly from NUMA nodes to support interleaving

Alexandr-Konovalov:
### Description 
When interleaved memory allocated with default interleaving step, it's possible to allocate it directly from required nodes, not migrate after 1st touch.

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [X] not needed

- [ ] Yes
- [X] No
- [ ] Unknown

### DIFF
--- a/src/tbb/numa_allocation.cpp
+++ b/src/tbb/numa_allocation.cpp
@@ -29,8 +29,16 @@
 
 // TBB build must be done without numaif.h, but we need signature of move_pages()
 // for dynamic loading, so declare it here.
-extern "C" long move_pages(int pid, unsigned long count,
-                           void **pages, const int *nodes, int *status, int flags);
+extern "C" {
+long move_pages(int pid, unsigned long count,
+                void **pages, const int *nodes, int *status, int flags);
+struct bitmask *numa_bitmask_alloc(unsigned int);
+void numa_bitmask_free(struct bitmask *);
+int numa_bitmask_isbitset(const struct bitmask *, unsigned int);
+struct bitmask *numa_bitmask_setbit(struct bitmask *, unsigned int);
+void numa_interleave_memory(void *mem, size_t size, struct bitmask *mask);
+
+} // extern "C"
 
 #elif !(_WIN32 || _WIN64)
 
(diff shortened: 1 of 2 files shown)


src/tbb/numa_allocation.cpp
```diff
@@ -112,6 +136,15 @@ static const int *common_init(size_t bytes, const tbb::detail::d1::numa_node_id
     return nodes;
 }
 
+static bool is_numa_funcs_found() {
+#if __linux__
+    return move_pages_ptr && numa_bitmask_alloc_ptr && numa_bitmask_free_ptr &&
+           numa_bitmask_isbitset_ptr && numa_bitmask_setbit_ptr && numa_interleave_memory_ptr;
```
aleksei-fedotov, line 142: Not sure how important and probable that can be, but I see possibility to make the code more robust in case either one set of functions was not found/loaded correctly. The algorithm can fallback onto another behavior.

aleksei-fedotov (commented):
Not a full reivew, just skimmed through.

src/tbb/numa_allocation.cpp
Alexandr-Konovalov: Copilot also suggested this, but those functionality are in libnuma for 6+ years, so chances are low, that we meet a library without them. But 1-line fix is possible, if not try to cover all possible situations.

aleksei-fedotov (dismissed):

test/tbb/test_numa_allocation.cpp
```diff
@@ -98,16 +113,60 @@ TEST_CASE("invalid parameters") {
...
+    if (bytes_per_chunk == page_size) {
+        std::vector<tbb::numa_node_id> sorted_nodes(nodes);
+        std::sort(sorted_nodes.begin(), sorted_nodes.end());
+        auto adj = std::adjacent_find(sorted_nodes.begin(), sorted_nodes.end());
+        // numa_interleave_memory() can be used only for non-repeated nodes case
+        if (adj == sorted_nodes.end()) {
+            // touch each page, otherwise move_pages() will fail with EFAULT
+            // and so detecting NUMA node will be impossible
+            for (size_t i = 0; i < bytes; i += page_size)
+                ptr[i] = 0;
```
aleksei-fedotov: This part is a duplicate with the lines 158-159 below for a single NUMA node case. Consider extracting into one function or something.

test/tbb/test_numa_allocation.cpp
```diff
@@ -98,16 +113,60 @@ TEST_CASE("invalid parameters") {
...
+                const std::string nodes_str = NodesToString(sorted_nodes);
+                REQUIRE_MESSAGE(false, "Unexpected NUMA node " << start_node
+                                << " for the first page, expected one of: " << nodes_str);
+                return;
+            }
+
+            for (size_t offset = 0, page_index = it - sorted_nodes.begin(); offset < bytes;
+                 offset += page_size, ++page_index) {
+                NUMA_EQ(find_numa_node(ptr + offset), sorted_nodes[page_index % sorted_nodes.size()]);
+            }
```
aleksei-fedotov: This loop is just a particular case of the loop on lines 161-162 below. In fact, substituting `page_size` with `bytes_per_chunk` gives us the loop below. Consider re-using to avoid duplication.

aleksei-fedotov (commented):
Forgot to post two comments that I started writing. So, posting them now.

test/tbb/test_numa_allocation.cpp
Alexandr-Konovalov: I'm not sure that replacing 2-line loop with a function with 6 parameters improves readability, 
because a reader must match names in one place with logic in another. But probably you have something more readable.
```
void VerifySizeAndNodesImpl(char *ptr, size_t bytes, const std::vector<tbb::numa_node_id>& nodes,
                            size_t bytes_per_chunk, size_t start_page_index = 0) {
    for (size_t offset = 0, page_index = start_page_index; offset < bytes;
         offset += bytes_per_chunk, ++page_index)
        NUMA_EQ(find_numa_node(ptr + offset), nodes[page_index % nodes.size()]);
}
```

test/tbb/test_numa_allocation.cpp
Alexandr-Konovalov: Done.

test/tbb/test_numa_allocation.cpp
aleksei-fedotov: Why cannot we just remove this loop along with the `return` statement, allowing the code flow to fallthrough right into the loop at the end of the function?

Alexandr-Konovalov: Thanks, that's better. Indirection decreases readability, however.

aleksei-fedotov (approved):
Looks good to me!

PR #2088, opened 2026-05-26
Add documentation for interleaved memory allocation

Alexandr-Konovalov:
### Description 
Add documentation for interleaved memory allocation

### Type of change

### Tests

- [X] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [X] No
- [ ] Unknown

### DIFF
--- a/doc/main/reference/examples/allocate_numa_interleaved_basic.cpp
+++ b/doc/main/reference/examples/allocate_numa_interleaved_basic.cpp
@@ -1,0 +1,35 @@
+/*
+    Copyright (c) 2026 UXL Foundation Contributors
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+
+/*begin_allocate_numa_interleaved_example*/
+#define TBB_PREVIEW_NUMA_ALLOCATION 1
+
+#include <oneapi/tbb/numa_allocation.h>
+#include <oneapi/tbb/parallel_for.h>
+
+int main() {
+    std::size_t size = 10LLU * 1024 * 1024;
+    double* ptr = (double*)tbb::allocate_numa_interleaved(size*sizeof(double));
+    if (!ptr)
+        return -1;
+    tbb::parallel_for(std::size_t(0), size, [&ptr](std::size_t i) {
+        ((char*)ptr)[i] = 0;
+    });
+
+    tbb::deallocate_numa_interleaved(ptr, size*sizeof(double));
+}
+
+/*end_allocate_numa_interleaved_example*/
--- a/doc/main/reference/examples/allocate_numa_interleaved_basic.cpp
+++ b/doc/main/reference/examples/allocate_numa_interleaved_basic.cpp
@@ -22,13 +22,13 @@
 
 int main() {
     std::size_t array_size = 10LLU * 1024 * 1024;
-    double* ptr = (double*)tbb::allocate_numa_interleaved(array_size * sizeof(double));
+    double* ptr = (double*)oneapi::tbb::allocate_numa_interleaved(array_size * sizeof(double));
     if (!ptr)
         return -1;
-    tbb::parallel_for(std::size_t(0), array_size, [=](std::size_t i) {
+    oneapi::tbb::parallel_for(std::size_t(0), array_size, [=](std::size_t i) {
         ptr[i] = i;
     });
 
-    tbb::deallocate_numa_interleaved(ptr, array_size * sizeof(double));
+    oneapi::tbb::deallocate_numa_interleaved(ptr, array_size * sizeof(double));
 }
 /*end_allocate_numa_interleaved_example*/
--- a/doc/main/reference/examples/allocate_numa_interleaved_basic.cpp
+++ b/doc/main/reference/examples/allocate_numa_interleaved_basic.cpp
@@ -26,7 +26,7 @@ int main() {
     if (!ptr)
         return -1;
     tbb::parallel_for(std::size_t(0), size, [&ptr](std::size_t i) {
-        ((char*)ptr)[i] = 0;
+        ptr[i] = i;
     });
 
     tbb::deallocate_numa_interleaved(ptr, size*sizeof(double));
--- a/doc/main/reference/examples/allocate_numa_interleaved_basic.cpp
+++ b/doc/main/reference/examples/allocate_numa_interleaved_basic.cpp
@@ -31,5 +31,4 @@ int main() {
 
     tbb::deallocate_numa_interleaved(ptr, size*sizeof(double));
 }
-
 /*end_allocate_numa_interleaved_example*/
--- a/doc/main/reference/examples/allocate_numa_interleaved_basic.cpp
+++ b/doc/main/reference/examples/allocate_numa_interleaved_basic.cpp
@@ -21,14 +21,14 @@
 #include <oneapi/tbb/parallel_for.h>
 
 int main() {
-    std::size_t size = 10LLU * 1024 * 1024;
-    double* ptr = (double*)tbb::allocate_numa_interleaved(size*sizeof(double));
+    std::size_t array_size = 10LLU * 1024 * 1024;
+    double* ptr = (double*)tbb::allocate_numa_interleaved(array_size * sizeof(double));
     if (!ptr)
         return -1;
-    tbb::parallel_for(std::size_t(0), size, [&ptr](std::size_t i) {
+    tbb::parallel_for(std::size_t(0), array_size, [=](std::size_t i) {
         ptr[i] = i;
     });
 
-    tbb::deallocate_numa_interleaved(ptr, size*sizeof(double));
+    tbb::deallocate_numa_interleaved(ptr, array_size * sizeof(double));
 }
 /*end_allocate_numa_interleaved_example*/
--- a/doc/main/reference/examples/allocate_numa_interleaved_basic.cpp
+++ b/doc/main/reference/examples/allocate_numa_interleaved_basic.cpp
@@ -22,7 +22,8 @@
 
 int main() {
     std::size_t array_size = 10LLU * 1024 * 1024;
-    double* ptr = (double*)oneapi::tbb::allocate_numa_interleaved(array_size * sizeof(double));
+    double* ptr =
+        (double*)oneapi::tbb::allocate_numa_interleaved(array_size * sizeof(double));
     if (!ptr)
         return -1;
     oneapi::tbb::parallel_for(std::size_t(0), array_size, [=](std::size_t i) {
--- a/doc/main/reference/examples/allocate_numa_interleaved_pool.cpp
+++ b/doc/main/reference/examples/allocate_numa_interleaved_pool.cpp
@@ -37,7 +37,7 @@ public:
         return
             oneapi::tbb::allocate_numa_interleaved(num_of_objects * sizeof(value_type));
     }
-    void deallocate(void *ptr, size_t num_of_objects) {
+    void deallocate(void* ptr, std::size_t num_of_objects) {
         oneapi::tbb::deallocate_numa_interleaved(ptr, num_of_objects * sizeof(value_type));
     }
 };
--- a/doc/main/reference/examples/allocate_numa_interleaved_pool.cpp
+++ b/doc/main/reference/examples/allocate_numa_interleaved_pool.cpp
@@ -1,0 +1,52 @@
+/*
+    Copyright (c) 2026 UXL Foundation Contributors
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+
+/*begin_allocate_numa_interleaved_pool_example*/
+#define TBB_PREVIEW_MEMORY_POOL 1
+#define TBB_PREVIEW_NUMA_ALLOCATION 1
+
+#include <oneapi/tbb/numa_allocation.h>
+#include <oneapi/tbb/memory_pool.h>
+
+#include <cassert>
+#include <array>
+
+class numa_interleaved_allocator {
+public:
+    // Guarantee that each allocation is a multiple of the system page size,
+    // so allocate_numa_interleaved() requirements are satisfied.
+    typedef std::array<char, 4*1024> value_type;
+    numa_interleaved_allocator() {}
+    void *allocate(size_t size) {
+        return tbb::allocate_numa_interleaved(size*sizeof(value_type));
+    }
+    void deallocate(void *ptr, size_t size) {
+        tbb::deallocate_numa_interleaved(ptr, size*sizeof(value_type));
+    }
+};
+
+int main() {
+    // Memory pool requests memory in big chunks, slices them internally and uses caching,
+    // so may improve performance for many small allocations and reuse scenarios.
+    tbb::memory_pool<numa_interleaved_allocator> pool;
+    for (int i = 0; i < 10*1000; ++i) {
+        void* ptr = pool.malloc(1024);
+        assert(ptr);
+        pool.free(ptr);
+    }
+}
+
+/*end_allocate_numa_interleaved_pool_example*/
--- a/doc/main/reference/examples/allocate_numa_interleaved_pool.cpp
+++ b/doc/main/reference/examples/allocate_numa_interleaved_pool.cpp
@@ -52,7 +52,7 @@ int main() {
         // Temporary arrays allocated from the pool will reside in different
         // NUMA domains for better overall memory throughput.
         // As the pool caches the memory, on average it is faster than
-        // allocate_numa_interleaved()/deallocate_numa_interleaved()
+        // allocate_numa_interleaved()/deallocate_numa_interleaved().
         double* ptr = (double*)pool.malloc(10*1000*sizeof(double));
         // ...
         pool.free(ptr);
--- a/doc/main/reference/examples/allocate_numa_interleaved_pool.cpp
+++ b/doc/main/reference/examples/allocate_numa_interleaved_pool.cpp
@@ -26,10 +26,11 @@
 #include <vector>
 
 class numa_interleaved_provider {
+    static constexpr std::size_t page_size = 4 * 1024;
 public:
     // Guarantee that each allocation is a multiple of the system page size,
     // so allocate_numa_interleaved() requirements are satisfied.
-    typedef std::array<char, 4*1024> value_type;
+    typedef std::array<char, page_size> value_type;
     numa_interleaved_provider() {}
     // Like std::allocator<T>::allocate, these functions expect the number of
     // objects of the same size as sizeof(value_type).
@@ -47,7 +48,7 @@ int main() {
     // scenarios with the objects reuse.
     oneapi::tbb::memory_pool<numa_interleaved_provider> pool;
 
-    oneapi::tbb::parallel_for(0, 1024*1024, [&pool](std::size_t i) {
+    oneapi::tbb::parallel_for(0, 1024*1024, [&pool](std::size_t) {
         // Temporary arrays allocated from the pool will reside in different
         // NUMA domains for better overall memory throughput.
         // As the pool caches the memory, on average it is faster than
--- a/doc/main/reference/examples/allocate_numa_interleaved_pool.cpp
+++ b/doc/main/reference/examples/allocate_numa_interleaved_pool.cpp
@@ -33,9 +33,8 @@ public:
     numa_interleaved_provider() {}
     // Like std::allocator<T>::allocate, these functions expect the number of
     // objects of the same size as sizeof(value_type).
-    void *allocate(size_t num_of_objects) {
-        return
-            oneapi::tbb::allocate_numa_interleaved(num_of_objects * sizeof(value_type));
+    void* allocate(std::size_t num_of_objects) {
+        return oneapi::tbb::allocate_numa_interleaved(num_of_objects * sizeof(value_type));
     }
     void deallocate(void* ptr, std::size_t num_of_objects) {
         oneapi::tbb::deallocate_numa_interleaved(ptr, num_of_objects * sizeof(value_type));
--- a/doc/main/reference/examples/allocate_numa_interleaved_pool.cpp
+++ b/doc/main/reference/examples/allocate_numa_interleaved_pool.cpp
@@ -32,7 +32,7 @@ public:
     typedef std::array<char, 4*1024> value_type;
     numa_interleaved_provider() {}
     // Like std::allocator<T>::allocate, these functions expect the number of
-    // objects, where each object has size sizeof(value_type).
+    // objects of the same size as sizeof(value_type).
     void *allocate(size_t num_of_objects) {
         return
             oneapi::tbb::allocate_numa_interleaved(num_of_objects * sizeof(value_type));
@@ -49,10 +49,10 @@ int main() {
     oneapi::tbb::memory_pool<numa_interleaved_provider> pool;
 
     oneapi::tbb::parallel_for(0, 1024*1024, [&pool](std::size_t i) {
-        // If needed, a temporary array can be allocated from the pool that uses
-        // interleaved NUMA memory. It's faster than
+        // Temporary arrays allocated from the pool will reside in different
+        // NUMA domains for better overall memory throughput.
+        // As the pool caches the memory, on average it is faster than
         // allocate_numa_interleaved()/deallocate_numa_interleaved()
-        // because of the caching in the pool.
         double* ptr = (double*)pool.malloc(10*1000*sizeof(double));
         // ...
         pool.free(ptr);
--- a/doc/main/reference/examples/allocate_numa_interleaved_pool.cpp
+++ b/doc/main/reference/examples/allocate_numa_interleaved_pool.cpp
@@ -31,8 +31,8 @@ public:
     // so allocate_numa_interleaved() requirements are satisfied.
     typedef std::array<char, 4*1024> value_type;
     numa_interleaved_provider() {}
-    // As std::allocator<T>::allocate, those functions are expeting number of objects
-    // with sizeof(value_type)
+    // Like std::allocator<T>::allocate, these functions expect the number of
+    // objects, where each object has size sizeof(value_type).
     void *allocate(size_t num_of_objects) {
         return
             oneapi::tbb::allocate_numa_interleaved(num_of_objects * sizeof(value_type));
--- a/doc/main/reference/examples/allocate_numa_interleaved_pool.cpp
+++ b/doc/main/reference/examples/allocate_numa_interleaved_pool.cpp
@@ -20,21 +20,22 @@
 
 #include <oneapi/tbb/numa_allocation.h>
 #include <oneapi/tbb/memory_pool.h>
+#include <oneapi/tbb/parallel_for.h>
 
-#include <cassert>
 #include <array>
+#include <vector>
 
-class numa_interleaved_allocator {
+class numa_interleaved_provider {
 public:
     // Guarantee that each allocation is a multiple of the system page size,
     // so allocate_numa_interleaved() requirements are satisfied.
     typedef std::array<char, 4*1024> value_type;
-    numa_interleaved_allocator() {}
+    numa_interleaved_provider() {}
     void *allocate(size_t size) {
-        return tbb::allocate_numa_interleaved(size*sizeof(value_type));
+        return oneapi::tbb::allocate_numa_interleaved(size*sizeof(value_type));
     }
     void deallocate(void *ptr, size_t size) {
-        tbb::deallocate_numa_interleaved(ptr, size*sizeof(value_type));
+        oneapi::tbb::deallocate_numa_interleaved(ptr, size*sizeof(value_type));
     }
 };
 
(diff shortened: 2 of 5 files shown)


doc/main/reference/examples/allocate_numa_interleaved_pool.cpp
```diff
@@ -0,0 +1,52 @@
...
+#include <oneapi/tbb/memory_pool.h>
+
+#include <cassert>
+#include <array>
+
+class numa_interleaved_allocator {
+public:
+    // Guarantee that each allocation is a multiple of the system page size,
+    // so allocate_numa_interleaved() requirements are satisfied.
+    typedef std::array<char, 4*1024> value_type;
```
Alexandr-Konovalov: Using hard-coded page size is not good, but compile-time constant is required. Alternatively, we can inherit from `tbb::detail::d1::pool_base`, but this is an internal class. Or direct calls of `rml::pool_*`, that is also not public.

aleksei-fedotov: Since it is an example from the documentation, making the code more self-explaining will help. For example, I would introduce a dedicated constant for a page size, e.g.:
```cpp
// Popular page size in bytes in most of the platforms.
// In production code, retrieve actual page size from the system.
constexpr size_t page_size = 4096;
```

When doing allocation and deallocation, compute the size to be not more than required minimum rather than simply multiplying the size by page size. E.g.:
```cpp
// Guarantee that each allocation is a multiple of the system page size,
// so allocate_numa_interleaved() requirements are satisfied.
const size_t allocation_size = (size + page_size - 1) / page_size * page_size;
```

doc/main/reference/numa_interleaved_allocation.rst
```diff
@@ -0,0 +1,90 @@
...
+
+Synopsis
+--------
+
+.. code:: cpp
+
+    namespace tbb {
+        inline void *allocate_numa_interleaved(size_t bytes,
+                                       const std::vector<tbb::numa_node_id>& nodes,
+                                       size_t bytes_per_chunk = 0);
```
aleksei-fedotov: Aligh function parameters one by another?

```suggestion
        inline void *allocate_numa_interleaved(size_t bytes,
                                               const std::vector<tbb::numa_node_id>& nodes,
                                               size_t bytes_per_chunk = 0);
```

doc/main/reference/examples/allocate_numa_interleaved_pool.cpp
aleksei-fedotov, line 1: To avoid directory clutter and keep related examples together, can we place both examples in a single source file and name it something like "numa_allocation_examples.cpp"?

Alexandr-Konovalov: To me, the examples share nothing, so separating them makes things clearer.

doc/main/reference/examples/allocate_numa_interleaved_pool.cpp
Alexandr-Konovalov: > // In production code, retrieve actual page size from the system.

Sadly, there is no API to retrieve one in compile time. And we don't want to say "add nice ifdef-s in the production code".

doc/main/reference/numa_interleaved_allocation.rst
```diff
@@ -0,0 +1,95 @@
+.. _numa_interleaved_allocation:
+
+API to allocate memory interleaved between NUMA nodes
+=====================================================
+
+.. note::
+    To enable this feature, set the ``TBB_PREVIEW_NUMA_ALLOCATION`` macro to 1. When available and enabled,
+    the feature-test macro ``TBB_HAS_NUMA_ALLOCATION`` is defined.
```
kboyarinov, line 8: Please also update the `feature_test_macros.rst` page with the exact macro value and a link to this page.

doc/main/reference/numa_interleaved_allocation.rst
```diff
@@ -0,0 +1,95 @@
...
+
+A well-known method to improve performance on NUMA systems is to interleave memory between several NUMA
+nodes. There are two parameters that control the interleaving: the set of NUMA nodes across which memory is
+allocated and the chunk size used for interleaving. The first parameter allows users to select a subset of
+NUMA nodes, which may be desirable if a parallel algorithm uses only part of the available NUMA nodes. The
+second parameter controls the granularity of interleaving, which may be desirable to optimize for specific
+access patterns. The allocation/deallocation functions call the OS directly. If some form of caching is
+desirable, it can be implemented on top of this API.
+
+Under Linux, the API uses the ``libnuma`` library, which must be available at runtime. If the library is not
```
kboyarinov:
```suggestion
Under Linux*, the API uses the ``libnuma`` library, which must be available at runtime. If the library is not
```

doc/main/reference/numa_interleaved_allocation.rst
```diff
@@ -0,0 +1,95 @@
...
+available, the allocation functions fall back to standard memory allocation. On Windows, the API uses
```
kboyarinov:
```suggestion
available, the allocation functions fall back to standard memory allocation. On Windows*, the API uses
```

doc/main/reference/numa_interleaved_allocation.rst
```diff
@@ -0,0 +1,95 @@
...
+functionality available starting from Windows 10 / Windows Server 2016; on older versions of Windows, the
```
kboyarinov:
```suggestion
functionality available starting from Microsoft* Windows* 10 / Microsoft* Windows* Server 2016; on older versions of Microsoft* Windows*, the
```

doc/main/reference/numa_interleaved_allocation.rst
kboyarinov: It may also make sense to make it a caution.

doc/main/reference/numa_interleaved_allocation.rst
```diff
@@ -0,0 +1,95 @@
...
+
+    #define TBB_PREVIEW_NUMA_ALLOCATION 1
+    #include <oneapi/tbb/numa_allocation.h>
+
+Synopsis
+--------
+
+.. code:: cpp
+
+    namespace tbb {
```
kboyarinov: I guess `oneapi::tbb` namespace should be used, similarly to other reference pages

doc/main/reference/numa_interleaved_allocation.rst
```diff
@@ -0,0 +1,95 @@
...
+
+        inline void *allocate_numa_interleaved(size_t bytes, size_t bytes_per_chunk = 0);
+
+        inline void deallocate_numa_interleaved(void *ptr, size_t bytes);
+    } // namespace tbb
+
+Functions
+---------
+
+.. cpp:function:: void *tbb::allocate_numa_interleaved(size_t bytes, const std::vector<tbb::numa_node_id>& nodes, \
```
kboyarinov: Other reference pages omit the namespace in the function description.
```suggestion
.. cpp:function:: void* allocate_numa_interleaved(size_t bytes, const std::vector<tbb::numa_node_id>& nodes, \
```

doc/main/reference/numa_interleaved_allocation.rst
```diff
@@ -0,0 +1,95 @@
...
+                  size_t bytes_per_chunk = 0)
+
+    **Returns:** Allocated memory interleaved between specified NUMA ``nodes`` with interleaved chunk size of
+    ``bytes_per_chunk``. ``nodes`` can contain duplicates, in this case memory load is proportional to the
```
kboyarinov:
```suggestion
    ``bytes_per_chunk``. 
    
    If ``nodes`` contains duplicates, the memory load is proportional to the
```

doc/main/reference/numa_interleaved_allocation.rst
```diff
@@ -0,0 +1,95 @@
...
+    number of occurrences of each node. ``nodes`` must not be empty.
+    ``bytes_per_chunk`` must be a multiple of the system page size. If ``bytes_per_chunk`` is zero, a system
```
kboyarinov: I would better use "undefined behavior" wording.

Something like a note: `The behavior is undefined if ``nodes`` is empty, or ``bytes_per_chunk`` is not a multiple of the system page size.`.

doc/main/reference/numa_interleaved_allocation.rst
```diff
@@ -0,0 +1,95 @@
...
+    page size is used. Allocated memory contains zeros and is aligned to the system page size.
+    In case of allocation failure or invalid arguments, returns ``nullptr``.
```
kboyarinov:
```suggestion
    **Returns**: if the allocation succeeds, a pointer to the beginning of the memory region. ``nullptr`` otherwise.
```

doc/main/reference/numa_interleaved_allocation.rst
kboyarinov: Should we document what happens if `bytes`` is zero?

doc/main/reference/numa_interleaved_allocation.rst
```diff
@@ -0,0 +1,95 @@
...
+
+.. cpp:function:: void *tbb::allocate_numa_interleaved(size_t bytes, size_t bytes_per_chunk = 0)
+
+    Same as the above but allocates memory from all available NUMA nodes.
+
+.. cpp:function:: void tbb::deallocate_numa_interleaved(void *ptr, size_t bytes)
+
+    **Effects:** Deallocates memory allocated by ``allocate_numa_interleaved``. ``bytes`` must be the same as
```
kboyarinov: I would either use "Effects:" prefix for all three functions, or remove it in all places.

doc/main/reference/numa_interleaved_allocation.rst
```diff
@@ -0,0 +1,95 @@
...
+    the one passed to the corresponding allocation function. The behavior is undefined if ``ptr`` was not
```
kboyarinov: We may say that if ``bytes`` is not the same as the number passed to the corresponding allocation function, the behavior is undefined as well

doc/main/reference/numa_interleaved_allocation.rst
```diff
@@ -0,0 +1,95 @@
...
+*******
+
+The code below provides a simple example with direct use of interleaved memory allocation as arrays.
+
+.. literalinclude:: ./examples/allocate_numa_interleaved_basic.cpp
+    :language: c++
+    :start-after: /*begin_allocate_numa_interleaved_example*/
+    :end-before: /*end_allocate_numa_interleaved_example*/
+
+In this example, interleaved memory is wrapped in ``tbb::memory_pool``. This allows to amortize allocation
```
kboyarinov:
```suggestion
In the following example, interleaved memory is wrapped in ``tbb::memory_pool``. This allows to amortize allocation
```

doc/main/reference/numa_interleaved_allocation.rst
Alexandr-Konovalov: `Effects:` add no value and only distract with bold font, so I'm for dropping it, if this is acceptable.

doc/main/reference/numa_interleaved_allocation.rst
Alexandr-Konovalov: Is it clear enough from your suggestion that we return `nullptr` also in case of invalid arguments? I.e., not doing something undefined, not calling `abort()` etc?

doc/main/reference/numa_interleaved_allocation.rst
Alexandr-Konovalov: Is that because we want to keep possibility in future to change behavior to e.g. throwing exceptions?

Currently the behavior is defined: we check and return `nullptr`. Contrary to `deallocate_numa_interleaved()` with invalid `bytes`, where the behavior is definitely undefined.

doc/main/reference/numa_interleaved_allocation.rst
Alexandr-Konovalov: Thanks, missed that.

doc/main/reference/numa_interleaved_allocation.rst
Alexandr-Konovalov: Done.

doc/main/reference/numa_interleaved_allocation.rst
Alexandr-Konovalov: Done.

doc/main/reference/numa_interleaved_allocation.rst
Alexandr-Konovalov: Done.

doc/main/reference/numa_interleaved_allocation.rst
```diff
@@ -0,0 +1,103 @@
+.. _numa_interleaved_allocation:
+
+API to allocate memory interleaved between NUMA nodes
+=====================================================
```
akukanov:
```suggestion
Allocate Memory Interleaved between NUMA Nodes
==============================================
```

doc/main/reference/feature_test_macros.rst
```diff
@@ -71,6 +71,10 @@ are extended. The table below contains only the most recent values.
           -    ``202603``
           -    | ``<oneapi/tbb/task_group.h>``
                | ``<oneapi/tbb/task_arena.h>``
+        * -    :ref:`API to allocate memory interleaved between NUMA nodes<numa_interleaved_allocation>`
```
akukanov:
```suggestion
        * -    :ref:`Allocate Memory Interleaved between NUMA Nodes<numa_interleaved_allocation>`
```

doc/main/reference/numa_interleaved_allocation.rst
```diff
@@ -0,0 +1,103 @@
...
+Description
+***********
+
+A well-known method to improve performance on NUMA systems is to interleave memory between several NUMA
+nodes. There are two parameters that control the interleaving: the set of NUMA nodes across which memory is
+allocated and the chunk size used for interleaving. The first parameter allows users to select a subset of
+NUMA nodes, which may be desirable if a parallel algorithm uses only part of the available NUMA nodes. The
+second parameter controls the granularity of interleaving, which may be desirable to optimize for specific
+access patterns. The allocation/deallocation functions call the OS directly. If some form of caching is
+desirable, it can be implemented on top of this API.
```
akukanov:
```suggestion
second parameter controls the granularity of interleaving, which may be desirable to optimize for specific
access patterns.

The allocated memory is filled with zeros and aligned to the system page size. It is not cached
and is returned back immediately upon deallocation.
```

doc/main/reference/numa_interleaved_allocation.rst
```diff
@@ -0,0 +1,103 @@
...
+    } // namespace oneapi
+
+Functions
+---------
+
+.. cpp:function:: void* allocate_numa_interleaved(size_t bytes, const std::vector<tbb::numa_node_id>& nodes, \
+                  size_t bytes_per_chunk = 0)
+
+    **Returns:** Allocated memory interleaved between specified NUMA ``nodes`` with interleaved chunk size of
+    ``bytes_per_chunk``. In case of allocation failure or invalid arguments, returns ``nullptr``.
```
akukanov:
```suggestion
    **Returns:** The address of a memory block interleaved between the specified NUMA ``nodes`` in chunks of 
    ``bytes_per_chunk``. In case of allocation failure or invalid arguments, returns ``nullptr``.
```

doc/main/reference/numa_interleaved_allocation.rst
```diff
@@ -0,0 +1,103 @@
...
+
+    If ``nodes`` contains duplicates, the memory load is proportional to the number of occurrences of each
+    node. ``nodes`` must not be empty. ``bytes_per_chunk`` must be a multiple of the system page size. If
+    ``bytes_per_chunk`` is zero, a system page size is used. `bytes` must be non-zero. Allocated memory
+    contains zeros and is aligned to the system page size.
```
akukanov:
```suggestion
    **Requirements:** ``bytes`` must be non-zero, ``nodes`` must not be empty, and ``bytes_per_chunk``
    must be a multiple of the system page size.
    
    If ``nodes`` contains duplicate NUMA node IDs, each of these IDs independently participates
    in the interleaving order. If ``bytes_per_chunk`` is zero, the system page size is used.
    The allocated memory contains zeros and is aligned to the system page size.
```

doc/main/reference/numa_interleaved_allocation.rst
```diff
@@ -0,0 +1,103 @@
...
+    
+
+.. cpp:function:: void* allocate_numa_interleaved(size_t bytes, size_t bytes_per_chunk = 0)
+
+    Same as the above but allocates memory from all available NUMA nodes.
```
akukanov:
```suggestion
    Same as the above, but allocates memory interleaved across all available NUMA nodes.
```

doc/main/reference/numa_interleaved_allocation.rst
```diff
@@ -0,0 +1,103 @@
...
+
+.. cpp:function:: void deallocate_numa_interleaved(void* ptr, size_t bytes)
+
+    Deallocates memory allocated by ``allocate_numa_interleaved``. The behavior is undefined if ``bytes`` is
+    not the same as the one passed to the corresponding allocation function, ``ptr`` was not
+    allocated by ``allocate_numa_interleaved`` or if the same pointer is deallocated more than once.
```
akukanov:
```suggestion
    Deallocates memory allocated by ``allocate_numa_interleaved``.
    
    **Requirements:** ``ptr`` must be previously allocated by ``allocate_numa_interleaved`` and not yet
    deallocated, and ``bytes`` must be the same as the corresponding value used to allocate the memory.
    Otherwise, the behavior is undefined.
```

doc/main/reference/numa_interleaved_allocation.rst
```diff
@@ -0,0 +1,103 @@
...
+
+Example
+*******
```
akukanov:
```suggestion
Examples
********
```

doc/main/reference/numa_interleaved_allocation.rst
```diff
@@ -0,0 +1,103 @@
...
+
+The code below provides a simple example with direct use of interleaved memory allocation as arrays.
```
akukanov:
```suggestion
The code below provides a simple example with direct use of the allocated memory as a NUMA-interleaved array.
```

doc/main/reference/numa_interleaved_allocation.rst
```diff
@@ -0,0 +1,103 @@
...
+
+.. literalinclude:: ./examples/allocate_numa_interleaved_basic.cpp
+    :language: c++
+    :start-after: /*begin_allocate_numa_interleaved_example*/
+    :end-before: /*end_allocate_numa_interleaved_example*/
+
+In the following example, interleaved memory is wrapped in ``tbb::memory_pool``. This allows to amortize
+allocation overhead.
```
akukanov:
```suggestion
In the following example, interleaved memory is wrapped in ``tbb::memory_pool`` to amortize
allocation overhead for small objects.
```

doc/main/reference/examples/allocate_numa_interleaved_basic.cpp
```diff
@@ -0,0 +1,34 @@
...
+int main() {
+    std::size_t size = 10LLU * 1024 * 1024;
+    double* ptr = (double*)tbb::allocate_numa_interleaved(size*sizeof(double));
+    if (!ptr)
+        return -1;
+    tbb::parallel_for(std::size_t(0), size, [&ptr](std::size_t i) {
+        ptr[i] = i;
+    });
+
+    tbb::deallocate_numa_interleaved(ptr, size*sizeof(double));
```
akukanov:
```suggestion
    std::size_t array_size = 10LLU * 1024 * 1024;
    double* ptr = (double*)tbb::allocate_numa_interleaved(array_size * sizeof(double));
    if (!ptr)
        return -1;
    tbb::parallel_for(std::size_t(0), array_size, [=](std::size_t i) {
        ptr[i] = i;
    });

    tbb::deallocate_numa_interleaved(ptr, array_size * sizeof(double));
```

doc/main/reference/examples/allocate_numa_interleaved_pool.cpp
```diff
@@ -0,0 +1,52 @@
...
+#define TBB_PREVIEW_MEMORY_POOL 1
+#define TBB_PREVIEW_NUMA_ALLOCATION 1
+
+#include <oneapi/tbb/numa_allocation.h>
+#include <oneapi/tbb/memory_pool.h>
+
+#include <cassert>
+#include <array>
+
+class numa_interleaved_allocator {
```
akukanov: I would replace 'allocator' in the name with something else, perhaps 'provider'?

doc/main/reference/examples/allocate_numa_interleaved_pool.cpp
```diff
@@ -0,0 +1,52 @@
...
+public:
+    // Guarantee that each allocation is a multiple of the system page size,
+    // so allocate_numa_interleaved() requirements are satisfied.
+    typedef std::array<char, 4*1024> value_type;
+    numa_interleaved_allocator() {}
+    void *allocate(size_t size) {
+        return tbb::allocate_numa_interleaved(size*sizeof(value_type));
```
akukanov: Multiplication of `size` by 4K is strange. In my understanding, the argument to `allocate` here is the desired memory size in bytes (correct me if I am wrong), so it should be rounded up to 4K, not multipled.

doc/main/reference/examples/allocate_numa_interleaved_pool.cpp
```diff
@@ -0,0 +1,52 @@
...
+int main() {
+    // Memory pool requests memory in big chunks, slices them internally and uses
+    // memory caching, so may improve performance for many small allocations and
+    // scenarios with the objects reuse.
+    tbb::memory_pool<numa_interleaved_allocator> pool;
+    for (int i = 0; i < 10*1000; ++i) {
+        void* ptr = pool.malloc(1024);
+        assert(ptr);
+        pool.free(ptr);
+    }
```
akukanov: Would it make sense to use a bit more elaborated example? Maybe similar to that at https://uxlfoundation.github.io/oneTBB/main/reference/scalable_memory_pools/memory_pool_allocator_cls.html, with a container sitting on top of a NUMA-interleaved memory pool?

Alexandr-Konovalov: Here I want to show that creation of something like high-performance malloc/free from interleaved memory is simple. So I'm for adding a container example to this one, not for replacing it completely.

doc/main/reference/examples/allocate_numa_interleaved_pool.cpp
Alexandr-Konovalov: The name is definitely bad. `size` is number of items each having `sizeof(value_type)` size. It's like `nmemb` argument in `calloc()`.

doc/main/reference/numa_interleaved_allocation.rst
Alexandr-Konovalov: Both C++ standard and Linux man for `malloc` use wording "return a pointer to". Can we also follow this practice? So, can it be like

**Returns:** A pointer to the allocated memory interleaved between the specified NUMA ``nodes`` in chunks of 
    ``bytes_per_chunk``. In case of allocation failure or invalid arguments, returns ``nullptr``.

doc/main/reference/numa_interleaved_allocation.rst
Alexandr-Konovalov: > The allocated memory is filled with zeros and aligned to the system page size.

This is mentioned in the description of allocation functions as well. It seems it's more natural when we describe the functions, not in "general overview".

> It is not cached and is returned back immediately upon deallocation.

Probably it's unclear where the memory is returned. And in your version we say nothing about memory acquiring.

doc/main/reference/numa_interleaved_allocation.rst
akukanov: Works for me.

doc/main/reference/numa_interleaved_allocation.rst
akukanov: I think you have said everything about how the memory is obtained in the subsequent paragraph that mentions libnuma, malloc, and WinAPI.

So the key change I propose above is to make the note about absence of caching clear and concise. The other note about zero-filling and alignment serves as a segue, as well as helps to make a useful new paragraph, separate from the description of parameters. I do not mind repeating it later.

I also would not mind another way to rephrase the part about absence of caching.

doc/main/reference/examples/allocate_numa_interleaved_pool.cpp
akukanov: I think my proposal does not contradict to your desire - the "high-performance malloc/free" just becomes "the high performance allocator" that can also be used with containers.

Of course any simple and short example will be sketchy, but this one is sketchy in a strange way, to my taste. Maybe that is because of the loop - I do not clearly see what it is supposed to mean here, with just a malloc/free pair in the body. Replacing that with some operations with a container would be more meaningful, I think.

doc/main/reference/examples/allocate_numa_interleaved_pool.cpp
akukanov: OK, let then add some comments to explain what is going on.

doc/main/reference/numa_interleaved_allocation.rst
akukanov: Or maybe the note about immediate return could be moved to the description of `deallocate_numa_interleaved`.

doc/main/reference/numa_interleaved_allocation.rst
Alexandr-Konovalov: Done.

doc/main/reference/examples/allocate_numa_interleaved_pool.cpp
Alexandr-Konovalov: Sure, `allocator` is too overloaded. But in the [memory_pool docs](https://uxlfoundation.github.io/oneTBB/main/reference/scalable_memory_pools/memory_pool_cls.html) this entity is called "underlying allocator specified by the template argument".

doc/main/reference/examples/allocate_numa_interleaved_pool.cpp
Alexandr-Konovalov: I have a model of "a C-style HPC developer", for which an example with explicit malloc-like construction is preferred to an example with an STL container, where small allocations are done deep under the hood. I don't know, probably it's obsolete in the world of LLM.

An important point that should be stressed is that direct replacement of `malloc` to `allocate_numa_interleaved` is reasonable, but only for large and rare allocations.

doc/main/reference/numa_interleaved_allocation.rst
Alexandr-Konovalov: Done

doc/main/reference/numa_interleaved_allocation.rst
Alexandr-Konovalov: Can it be

```
...second parameter controls the granularity of interleaving, which may be desirable to optimize for 
specific access patterns.

An allocated memory is not cached und not splitted. It's returned back to OS immediately upon
deallocation.
```

akukanov: I do not think it is accurate to say it is returned to OS. If libnuma or malloc is used as the intermediate API, how do we know that it is returned to OS?

Again, I think the important information is that the memory is not cached by TBB but returned immediately (to the layer it was obtained from).

doc/main/reference/examples/allocate_numa_interleaved_pool.cpp
```diff
@@ -0,0 +1,66 @@
...
+#include <vector>
+
+class numa_interleaved_provider {
+public:
+    // Guarantee that each allocation is a multiple of the system page size,
+    // so allocate_numa_interleaved() requirements are satisfied.
+    typedef std::array<char, 4*1024> value_type;
+    numa_interleaved_provider() {}
+    // Like std::allocator<T>::allocate, these functions expect the number of
+    // objects, where each object has size sizeof(value_type).
```
akukanov:
```suggestion
    // Like std::allocator<T>::allocate, these functions expect the number of
    // objects of the same size as sizeof(value_type).
```

doc/main/reference/examples/allocate_numa_interleaved_pool.cpp
```diff
@@ -0,0 +1,66 @@
...
+    // Memory pool requests memory in big chunks, slices them internally and uses
+    // memory caching, so may improve performance for many small allocations and
+    // scenarios with the objects reuse.
+    oneapi::tbb::memory_pool<numa_interleaved_provider> pool;
+
+    oneapi::tbb::parallel_for(0, 1024*1024, [&pool](std::size_t i) {
+        // If needed, a temporary array can be allocated from the pool that uses
+        // interleaved NUMA memory. It's faster than
+        // allocate_numa_interleaved()/deallocate_numa_interleaved()
+        // because of the caching in the pool.
```
akukanov:
```suggestion
        // Temporary arrays allocated from the pool will reside in different
        // NUMA domains for better overall memory throughput.
        // As the pool caches the memory, on average it is faster than
        // allocate_numa_interleaved()/deallocate_numa_interleaved()
```

doc/main/reference/numa_interleaved_allocation.rst
Alexandr-Konovalov: Thanks for the discussion, I agree. We definitely do not want to dive into the discussion about "but on MacOs.,." or prohibit possible future optimization "on no-NUMA machine we always use malloc()". So, it's turns out to

```
...second parameter controls the granularity of interleaving, which may be desirable to optimize for 
specific access patterns.

An allocated memory is not splitted und not cached. It's returned back immediately upon deallocation.
```

doc/main/reference/examples/allocate_numa_interleaved_pool.cpp
Alexandr-Konovalov: I added the named constant, that's definitely good thing.

doc/main/reference/numa_interleaved_allocation.rst
Alexandr-Konovalov: Added. 

Also added 
```
Repeated IDs allow flexible load balancing between nodes.
```
to `allocate_numa_interleaved()` description.

doc/main/reference/numa_interleaved_allocation.rst
```diff
@@ -0,0 +1,109 @@
...
+    **Returns:** A pointer to the allocated memory interleaved between the specified NUMA ``nodes`` in chunks
+    of ``bytes_per_chunk``. In case of allocation failure or invalid arguments, returns ``nullptr``.
+
+    **Requirements:** ``bytes`` must be non-zero, ``nodes`` must not be empty, and ``bytes_per_chunk``
+    must be a multiple of the system page size.
+    
+    If ``nodes`` contains duplicate NUMA node IDs, each of these IDs independently participates
+    in the interleaving order. Repeated IDs allow flexible load balancing between nodes. If ``bytes_per_chunk``
+    is zero, the system page size is used. The allocated memory contains zeros and is aligned to the system
+    page size.
```
akukanov:
```suggestion
    If ``nodes`` contains some NUMA node IDs more than once, each of these IDs independently
    participates in the interleaving order. That allows flexible load balancing between nodes.
    If ``bytes_per_chunk`` is zero, the system page size is used. The allocated memory contains zeros 
    and is aligned to the system page size.
```

akukanov (approved):

aleksei-fedotov (approved):

doc/main/reference/numa_interleaved_allocation.rst
kboyarinov: Sorry, missed that while reading for the first time.

kboyarinov (approved):
LGTM

PR #2090, opened 2026-05-27
Load libnuma.so.1 under Linux, as libnuma.so goes from *-dev package

Alexandr-Konovalov:
### Description 
Currently for NUMA support libtbb dynamically loads libnuma.so, but in major Linux distributions it's a symlink that added as part of a *dev package. More universal is to load libnuma.so.1, that that is part of basic libnuma package.

### Type of change

### Tests

- [X] not needed

### Documentation

- [ ] needs to be updated
- [X] not needed

- [ ] Yes
- [X] No
- [ ] Unknown


isaevil (approved):
LGTM!

PR #2100, opened 2026-06-01
Explicitly tests alignment utility functions

Alexandr-Konovalov:
### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [X] not needed

- [ ] Yes
- [X] No
- [ ] Unknown


Alexandr-Konovalov:
In #2078 there is also need to put tests for utility functions somewhere. So it's probably justify creation of a separate test for that. I'm not sure the name is a good one, though.

aleksei-fedotov (approved):

PR #2104, opened 2026-06-09
Update the spec to clarify which parameters are overridden by `task_arena::initialize`

dnmokhov:
### Type of change

### Tests
- [x] not needed

### Documentation
- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/doc/main/specification/source/task_scheduler/task_arena/task_arena_cls.rst
+++ b/doc/main/specification/source/task_scheduler/task_arena/task_arena_cls.rst
@@ -244,11 +244,11 @@ Member functions
 
 .. cpp:function:: void initialize(int max_concurrency, unsigned reserved_slots = 1, priority a_priority = priority::normal)
 
-    Same as above, but overrides previous ``max_concurrency``, ``reserved_slots``, and ``a_priority`` arena parameters.
+    Same as ``initialize()``, but overrides ``max_concurrency``, ``reserved_slots``, and ``a_priority`` arena parameters set at construction.
 
 .. cpp:function:: void initialize(constraints constraints_, unsigned reserved_slots = 1, priority a_priority = priority::normal)
 
-    Same as above, but overrides previous :cpp:struct:`constraints`, ``reserved_slots``, and ``a_priority`` arena parameters.
+    Same as ``initialize()``, but overrides :cpp:struct:`constraints`, ``reserved_slots``, and ``a_priority`` arena parameters set at construction.
 
 .. cpp:function:: void initialize(oneapi::tbb::attach)
 
--- a/doc/main/specification/source/task_scheduler/task_arena/task_arena_cls.rst
+++ b/doc/main/specification/source/task_scheduler/task_arena/task_arena_cls.rst
@@ -244,11 +244,12 @@ Member functions
 
 .. cpp:function:: void initialize(int max_concurrency, unsigned reserved_slots = 1, priority a_priority = priority::normal)
 
-    Same as above, but overrides previous arena parameters.
+    Same as above, but overrides previous ``max_concurrency``, ``reserved_slots``, and ``a_priority`` arena parameters.
 
 .. cpp:function:: void initialize(constraints constraints_, unsigned reserved_slots = 1, priority a_priority = priority::normal)
 
-    Same as above.
+    Same as above, but overrides previous ``constraints_`` (``numa_id``, ``max_concurrency``, ``core_type``, ``max_threads_per_core``),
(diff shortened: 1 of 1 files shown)


doc/main/specification/source/task_scheduler/task_arena/task_arena_cls.rst
```diff
@@ -244,11 +244,12 @@ Member functions
...
 .. cpp:function:: void initialize(int max_concurrency, unsigned reserved_slots = 1, priority a_priority = priority::normal)
 
-    Same as above, but overrides previous arena parameters.
+    Same as above, but overrides previous ``max_concurrency``, ``reserved_slots``, and ``a_priority`` arena parameters.
 
 .. cpp:function:: void initialize(constraints constraints_, unsigned reserved_slots = 1, priority a_priority = priority::normal)
 
-    Same as above.
+    Same as above, but overrides previous ``constraints_`` (``numa_id``, ``max_concurrency``, ``core_type``, ``max_threads_per_core``),
+    ``reserved_slots``, and ``a_priority`` arena parameters.
```
aleksei-fedotov: Perhaps, better would be to avoid enumeration of what's inside the `constraints` structure and let the dedicated page/paragraph to do that. Otherwise, it would be harder to find and adjust all such places if its content structure changes. Consider placing a link here instead.

doc/main/specification/source/task_scheduler/task_arena/task_arena_cls.rst
```diff
@@ -244,11 +244,12 @@ Member functions
 
 .. cpp:function:: void initialize(int max_concurrency, unsigned reserved_slots = 1, priority a_priority = priority::normal)
 
-    Same as above, but overrides previous arena parameters.
+    Same as above, but overrides previous ``max_concurrency``, ``reserved_slots``, and ``a_priority`` arena parameters.
```
vossmjp: Should "previous" be defined to be specific about how, when they could have been set? For example, not by another call to initialize. Or do we expect, "Same as above" to be enough to make that clear.

dnmokhov: I think that "Same as above" is clear enough, because this note from above applies to all the `initialize` methods:
> [!NOTE]
> After the call to ``initialize``, the arena parameters are fixed and cannot be changed.

doc/main/specification/source/task_scheduler/task_arena/task_arena_cls.rst
dnmokhov: replaced with a link

doc/main/specification/source/task_scheduler/task_arena/task_arena_cls.rst
aleksei-fedotov: Actually, it is not 100% clear that the note applies to all `initialize` methods. It is written and rendered as if it applies only to method `intitialize()` without parameters. Since I am not sure how to render it in a way that it groups and applies to the `initialize` method all together, I think that the phrase _Same as above_ is essential here. I would make it a bit more precise though by saying that it applies to `initialize()` method without parameters. So the start of the sentence becomes something like _Same as in `initialize()`, but overrides previously specified ... arena parameters._

dnmokhov: I agree, other than the _previously specified_ wording. They are overridden regardless of whether they were specified or default, so I would keep it as _overrides previous_. What do you think?

aleksei-fedotov: Okay, I don't object. Also, instead of _previously specified_, _previously used_ fits as well.

aleksei-fedotov (dismissed):

vossmjp (dismissed):
LGTM too

doc/main/specification/source/task_scheduler/task_arena/task_arena_cls.rst
vossmjp: How about instead of "Same as above, but overrides previous ``max_concurrency``, ``reserved_slots``, and ``a_priority`` arena parameters." something like, "Same as initialize(), but overrides ``max_concurrency``, ``reserved_slots``, and ``a_priority`` arena parameters set at construction."

dnmokhov: updated

akukanov (approved):

Issue #1772, opened 2025-07-05
CMake build fails when building a universal binary on MacOS


alex-tee:
# Summary
[This line](https://github.com/uxlfoundation/oneTBB/blob/v2022.2.0/cmake/compilers/AppleClang.cmake#L41) introduces `-mrtm` and `-mwaitpkg` to the compiler flags even when building a universal binary, and [this line](https://github.com/uxlfoundation/oneTBB/blob/v2022.2.0/include/oneapi/tbb/detail/_config.h#L340) tries to use the `waitpkg` functionality, resulting in build errors (see below).

# Version
2022.2.0

# Environment
* Hardware: Mac Mini Apple M2
* OS name and version: Sequoia 15.4.1
* Compiler version arm64-apple-darwin24.4.0
* CMake generator used: Xcode (Xcode version 16.3)

# Observed Behavior
My builds fail due to the introduction of `-mrtm` and `-mwaitpkg` (xcode build invocation log):
```
    cd /Users/alex/builds/KydKbRm4x/0/zrythm/zrythm
    write-file /Users/alex/builds/KydKbRm4x/0/zrythm/zrythm/build_mac_arm-Release/build/tbb.build/Release/Objects-normal/arm64/tbb.LinkFileList
WriteAuxiliaryFile /Users/alex/builds/KydKbRm4x/0/zrythm/zrythm/build_mac_arm-Release/build/tbb.build/Release/Objects-normal/arm64/82b82416624d2658e5098eb0a28c15c5-common-args.resp (in target 'tbb' from project 'zrythm')
    cd /Users/alex/builds/KydKbRm4x/0/zrythm/zrythm
    write-file /Users/alex/builds/KydKbRm4x/0/zrythm/zrythm/build_mac_arm-Release/build/tbb.build/Release/Objects-normal/arm64/82b82416624d2658e5098eb0a28c15c5-common-args.resp
[...]

PR #2106, opened 2026-06-11
FIX: x86_64 flags in universal binary build (Xcode)

philippremy:
### Description

Xcode compiler invocations are unlike the invocations of other CMake generators when building universal binaries:

- Xcode manually invokes the compiler two (2) times with different target flags
- All other generators do a single compiler invocation with both architectures specified in one invocation

This causes Xcode to currently pass x86_64 flags to an arm64-only compiler invocation, which causes a compiler error.

Xcode does allow setting flags specific to the architecture, which must be passed with a `CMAKE_XCODE_ATTRIBUTE_OTHER_CPLUSPLUSFLAGS for x86_64` manually.

Fixes issue #1772. See my comment in that thread for more info.

### Type of change

- [x] bug fix

### Tests

- [x] not needed

### Documentation

- [x] not needed

- [x] No

@isaevil that would be the follow-up PR we talked about :).

### DIFF
--- a/cmake/compilers/AppleClang.cmake
+++ b/cmake/compilers/AppleClang.cmake
@@ -38,7 +38,18 @@ else()
     set(_tbb_target_architectures "${CMAKE_SYSTEM_PROCESSOR}")
 endif()
 if ("${_tbb_target_architectures}" MATCHES "(x86_64|amd64|AMD64)") # OSX systems are 64-bit only
-    set(TBB_COMMON_COMPILE_FLAGS ${TBB_COMMON_COMPILE_FLAGS} -mrtm $<$<NOT:$<VERSION_LESS:${CMAKE_CXX_COMPILER_VERSION},12.0>>:-mwaitpkg>)
+    if(CMAKE_GENERATOR STREQUAL "Xcode")
+        # Note: Must be a string with whitespaces, because it gets passed to
+        # the .pbxproj file directly.
+        # These flags need to be set on a per-target basis (see the relevant
+        # CMakeLists.txt).
+        set(TBB_X86_XCODE_COMPILE_FLAGS "-mrtm $<$<NOT:$<VERSION_LESS:${CMAKE_CXX_COMPILER_VERSION},12.0>>:-mwaitpkg>")
+    else()
+        # Can be inserted directly, generators other than Xcode have a single
+        # compiler invocation and silently ignore flags that do not apply to
+        # the target architecture
+        set(TBB_COMMON_COMPILE_FLAGS ${TBB_COMMON_COMPILE_FLAGS} -mrtm $<$<NOT:$<VERSION_LESS:${CMAKE_CXX_COMPILER_VERSION},12.0>>:-mwaitpkg>)
+    endif()
 endif()
 unset(_tbb_target_architectures)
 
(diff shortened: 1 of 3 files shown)


src/tbb/CMakeLists.txt
```diff
@@ -165,6 +165,17 @@ if(TBB_BUILD_APPLE_FRAMEWORKS)
         MACOSX_FRAMEWORK_SHORT_VERSION_STRING ${TBB_BINARY_VERSION})
 endif()
 
+# On Xcode when building a universal binary, the x86_64-specific flags need to
+# be set as an Xcode attribute because of the way Xcode invokes the compiler
+# compared to all other generators.
+if(DEFINED TBB_X86_XCODE_COMPILE_FLAGS AND CMAKE_GENERATOR STREQUAL "Xcode")
+    # $(OTHER_CPLUSPLUSFLAGS) must be passed as a literal to also use all
+    # common compilation flags
+    set_target_properties(tbb PROPERTIES
```
isaevil, line 174: `TBB_COMMON_COMPILE_FLAGS` is passed to other library targets (tbbmalloc[_proxy], tbbbind). I believe this attribute should be set to them as well.

cmake/compilers/AppleClang.cmake
```diff
@@ -38,7 +38,18 @@ else()
     set(_tbb_target_architectures "${CMAKE_SYSTEM_PROCESSOR}")
 endif()
 if ("${_tbb_target_architectures}" MATCHES "(x86_64|amd64|AMD64)") # OSX systems are 64-bit only
-    set(TBB_COMMON_COMPILE_FLAGS ${TBB_COMMON_COMPILE_FLAGS} -mrtm $<$<NOT:$<VERSION_LESS:${CMAKE_CXX_COMPILER_VERSION},12.0>>:-mwaitpkg>)
+    if(CMAKE_GENERATOR STREQUAL "Xcode")
+        # Note: Must be a string with whitespaces, because it gets passed to
+        # the .pbxproj file directly.
+        # These flags need to be set on a per-target basis (see the relevant
+        # CMakeLists.txt).
+        set(TBB_X86_XCODE_COMPILE_FLAGS "-mrtm $<$<NOT:$<VERSION_LESS:${CMAKE_CXX_COMPILER_VERSION},12.0>>:-mwaitpkg>")
```
isaevil, line 46: I think we can set this variable unconditionally if target architecture is x86 and without quotes. Then set `TBB_COMMON_COMPILE_FLAGS` when appropriate.
Shouldn't it be already guarded by quotes in src/tbb/CMakeLists.txt:175?

src/tbb/CMakeLists.txt
philippremy: Fair enough. I basically just did the minimum to get it working, but that seems sensible if any other library should ever call these special intrinsics...

cmake/compilers/AppleClang.cmake
philippremy: Fine with me. Makes it a little more readble then maybe...

Regarding the quotes, I am not 100% sure. I will do some testing and adjust that if necessary, but CMake and me are sometimes not the best friends when it comes to lists and strings. The only thing to make sure is that CMake preserves the whitespaces, because Xcode obviously does not understand CMake's list syntax and CMake makes no attempt at transforming the Xcode attributes, they just get passed as-is. Will report back :).

philippremy:
I don't know what the convention is for this project, but should I force push? Or do we add new commits and squash and merge eventually?

isaevil:
> I don't know what the convention is for this project, but should I force push? Or do we add new commits and squash and merge eventually?

Commits will be eventually squashed and merged into main branch. But as a reviewer I would actually prefer separate commits within the PR to better track what has changed since my last review.

PR #2107, opened 2026-06-16
Support interleaved allocations of objects > 2MB with non-default bytes_per_chunk

Alexandr-Konovalov:
### Description 
There is an issue: tbb::allocate_numa_interleaved(2*1024*1024, 8*1024) is not working when /sys/kernel/mm/transparent_hugepage/enabled is set to "always".

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [X] not needed

- [ ] Yes
- [X] No
- [ ] Unknown

### DIFF
--- a/src/tbb/numa_allocation.cpp
+++ b/src/tbb/numa_allocation.cpp
@@ -196,7 +196,7 @@ void *__TBB_EXPORTED_FUNC allocate_interleaved(size_t bytes,
     // If /sys/kernel/mm/transparent_hugepage/enabled is set to "always", allocations of objects
     // > 2MB with non-default bytes_per_chunk (say, 8K) fail, because move_pages() can't fulfill
     // the request. MADV_NOHUGEPAGE prevents this. Cover smaller sizes as well, because several
-    // allocations with smaller sizes can be potentially joined by THP.
+    // allocations with smaller sizes can be potentially joined by Transparent Huge Pages (THP).
     int r = madvise(base_addr, align_to_greater_or_equal(bytes, governor::default_page_size()),
                     MADV_NOHUGEPAGE);
     if (r != 0)
--- a/src/tbb/numa_allocation.cpp
+++ b/src/tbb/numa_allocation.cpp
@@ -197,8 +197,8 @@ void *__TBB_EXPORTED_FUNC allocate_interleaved(size_t bytes,
     // > 2MB with non-default bytes_per_chunk (say, 8K) fail, because move_pages() can't fulfill
     // the request. MADV_NOHUGEPAGE prevents this. Cover smaller sizes as well, because several
     // allocations with smaller sizes can be potentially joined by Transparent Huge Pages (THP).
-    int r = madvise(base_addr, align_to_greater_or_equal(bytes, governor::default_page_size()),
-                    MADV_NOHUGEPAGE);
+    // Expecting that madvise() is aligning bytes to page size.
+    int r = madvise(base_addr, bytes, MADV_NOHUGEPAGE);
     if (r != 0)
         return nullptr;
 
--- a/src/tbb/numa_allocation.cpp
+++ b/src/tbb/numa_allocation.cpp
@@ -27,6 +27,7 @@
 
 #include <algorithm> // for std::any_of
 #include <sys/mman.h>
+#include <stdio.h>
 
 // TBB build must be done without numaif.h, but we need signatures of those functions
 // for dynamic loading, so declare it here.
@@ -193,6 +194,15 @@ void *__TBB_EXPORTED_FUNC allocate_interleaved(size_t bytes,
         }
     }
     // process non-standard chunk size or repeated nodes
+
+    // If /sys/kernel/mm/transparent_hugepage/enabled is set to "always", allocations of objects
+    // > 2MB with non-default bytes_per_chunk (say, 8K) fail, because move_pages() can't fulfill
+    // the request. MADV_NOHUGEPAGE prevents this. Cover smaller sizes as well, because several
+    // allocations with smaller sizes can be potentially joined by THP.
+    int r = madvise(base_addr, align_to_greater_or_equal(bytes, governor::default_page_size()),
+                    MADV_NOHUGEPAGE);
+    if (r != 0)
+        return nullptr;
 
     // touch each page, otherwise move_pages() will fail with EFAULT
     for (size_t i = 0; i < bytes; i += governor::default_page_size())
--- a/src/tbb/numa_allocation.cpp
+++ b/src/tbb/numa_allocation.cpp
@@ -27,8 +27,6 @@
 
 #include <algorithm> // for std::any_of
 #include <sys/mman.h>
-#include <stdio.h>
-
 // TBB build must be done without numaif.h, but we need signatures of those functions
 // for dynamic loading, so declare it here.
 extern "C" {
--- a/test/tbb/test_numa_allocation.cpp
+++ b/test/tbb/test_numa_allocation.cpp
@@ -122,7 +122,8 @@ TEST_CASE("invalid parameters") {
 
 void VerifySizeAndNodes(char *ptr, size_t bytes, const std::vector<tbb::numa_node_id>& nodes,
                         size_t bytes_per_chunk, bool check_ownership) {
-    REQUIRE_MESSAGE(ptr != nullptr, "Failed to allocate NUMA interleaved memory for size " << bytes);
+    REQUIRE_MESSAGE(ptr != nullptr, "Failed to allocate NUMA interleaved memory for size "
+                                    << bytes << " and bytes_per_chunk " << bytes_per_chunk);
     REQUIRE_EQ(utils::NonZero(ptr, bytes), 0);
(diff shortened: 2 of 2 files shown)


src/tbb/numa_allocation.cpp
```diff
@@ -194,6 +193,15 @@ void *__TBB_EXPORTED_FUNC allocate_interleaved(size_t bytes,
     }
     // process non-standard chunk size or repeated nodes
 
+    // If /sys/kernel/mm/transparent_hugepage/enabled is set to "always", allocations of objects
+    // > 2MB with non-default bytes_per_chunk (say, 8K) fail, because move_pages() can't fulfill
+    // the request. MADV_NOHUGEPAGE prevents this. Cover smaller sizes as well, because several
+    // allocations with smaller sizes can be potentially joined by THP.
```
aleksei-fedotov: I believe it is better to decipher the abbreviations. Otherwise, not all readers could comprehend it at once. For example, I could not understand it right away. It requires me googling the meaning. And googling might take much of the time and can lead to obviously incorrect results such as [this one](https://github.com/Araozu/thp-zig). 🙂

So, I suggest writing it as:
```suggestion
    // allocations with smaller sizes can be potentially joined by Transparent Huge Pages (THP).
```

src/tbb/numa_allocation.cpp
```diff
@@ -194,6 +193,15 @@ void *__TBB_EXPORTED_FUNC allocate_interleaved(size_t bytes,
...
+    int r = madvise(base_addr, align_to_greater_or_equal(bytes, governor::default_page_size()),
+                    MADV_NOHUGEPAGE);
```
aleksei-fedotov: 1. Have you considered not aligning the `size` argument? The manual states it is done by the syscall itself. Perhaps, we can save a few cycles on that.
2. Is the situation with transparent pages being enabled by default a popular one? Otherwise, it might make sense to try `move_pages()` first and only if it fails, fallback to re-trying with `madvise(); move_pages()` sequence. What do you think?

src/tbb/numa_allocation.cpp
```diff
@@ -194,6 +193,15 @@ void *__TBB_EXPORTED_FUNC allocate_interleaved(size_t bytes,
...
+    if (r != 0)
+        return nullptr;
```
aleksei-fedotov, line 203: I would listen to the feedback from copilot here, and do best effort since `madvise` is an advisory call and some values simply may not be supported by the kernel.

Alexandr-Konovalov: To me, it's to ignore the error and to return `nullptr`. But this should be addressed when we add right error processing. Right now I want find all situations when `madvise` is unsuccessful. Ignoring doesn't help to do this.

src/tbb/numa_allocation.cpp
Alexandr-Konovalov: 1. In Ubuntu 24.04.4 I see

> madvise() only operates on whole pages, therefore addr must be page-aligned.  The value of length is rounded up to a multiple of page size.

Is that can be read as "The value of length is rounded up to a multiple of page size, because addr must be page-aligned"? To me, man is ambiguous, and I just do the safest thing.

2. There is a race condition between `mmap` and THP daemon's activity in case of setting `transparent_hugepage/enabled` to `always`. Before I understood that the root cause is THP, I tried to call `move_pages()` several times till success. And right after the total success of `move_pages()` we are asking where the pages are and finding them all on node 2, without any interleaving.

aleksei-fedotov: 1. I agree with you that the documentation in this regard sounds ambiguous. I usually use [the man7.org](https://www.man7.org) as an authoritative source of Linux APIs description. However, even there [the description of madvise function](https://www.man7.org/linux/man-pages/man2/madvise.2.html) is not clear enough. But, if we look a bit more careful we will find a couple of hints saying essentially that it is the function who aligns the argument. The hints I found are:
   - It says _The value of size is rounded up_ - I read it as the function gets some value of `size`, which it operates on. Note that it does not say that the value should be or must be page-aligned, as opposed to the statement about `addr`, which is made just in the previous sentence.
     > madvise() only operates on whole pages, therefore **addr must be page-aligned**.  The value of size _is rounded up_ to a multiple of page size.
   - Given [the relatively large list of possible error values](https://www.man7.org/linux/man-pages/man2/madvise.2.html#ERRORS) that can be returned from the function, I would expect it to have one for the case of unaligned `size`. However, the list mentions only about `EINVAL` error code being returned when _`addr` is not page-aligned or `size` is negative_, and nothing is said about unaligned `size`.

    So, to me it is likely that the function aligns the passed value of `size` itself, and, therefore, I would not do that manually.

2. That's what I am talking about. The question is how probable the situation that `transparent_hugepage/enabled` is set to `always`. Because, if that situation is unlikely, I would expect the `move_pages()` call to succeed, and, therefore, the code do not need to make `madvise` syscall at all. And, as I wrote previously, can use the fallback path (i.e., call `madvise()` followed by `move_pages()` in the unlikely case of enabled THP. In the opposite situation, however, when THP is likely to be enabled, it is reasonable to make the preventive `madvise` call, i.e. that the patch currently does.
Therefore, I wonder do we know anything about likelihood of THP being enabled? 
I guess it deserves an explaining commentary in the code as well.

src/tbb/numa_allocation.cpp
aleksei-fedotov: Hmmm... that's also an interesting question - should `allocate_interleaved` guarantee interleaving state of returned memory or just try its best, returning the memory it was able to allocate regardless of its interleaving state and report on the status. I guess if there is going to be lots of conditions that require additional calls to be made like this one with `madvise` and `move_pages` we might want to use the latter behavior.
Shall we add such a note in the RFC?

vossmjp: I agree with this. Try its best, make no guarantees and provide some way to know it is interleaved or not.

Alexandr-Konovalov: Actually, this is a good point, but not directly related to processing error from syscalls. I mean in MSDN they clearly wrote, that placing memory to nodes is only recommendation, it's not mandatory. So it must be in the RFC.

Wrt error. Checking return values for close() that opened for reading is useless, but once in decade it helps to find out a memory overrun. And in our case there is no error codes "marking the memory region with NOHUGEPAGE is impossible, but you can continue". All we get is "something strange occurred". Ignoring this important information seems counterproductive.

src/tbb/numa_allocation.cpp
Alexandr-Konovalov: (1) ok, let's drop alignment.

(2) Sadly, I can't estimate that probability. 

> And, as I wrote previously, can use the fallback path (i.e., call madvise() followed by move_pages()

Sadly, that is not work this way, because THP daemon, as far as I can see, only joins normal page to huge ones, not split huge pages once someone is asking for it (or it can do that in background, and I just should wait). So working solution is rollback to `mmap()` and then using `madvise()` before `move_pages()`. Even kluge solution is to read /sys/kernel/mm/transparent_hugepage/enabled on startup and fallback to `madvise()` when we got `-EBUSY` from `move_pages()`.

I see our contract as "we do our best to support user's request for data placement", not "make it somewhat faster, but how knows where it placed, because it's all recommendations after all".

src/tbb/numa_allocation.cpp
aleksei-fedotov: So I think we should start thinking about error reporting out from the `allocate_interleaved` function. I am thinking about using standard `errno` entity for that with our own set of `E<VALUE>`s that would indicate various runtime situations, including _marking the memory region with NOHUGEPAGE is impossible, but you can continue_.

Alexandr-Konovalov: Sure, error reporting is a huge topic.

src/tbb/numa_allocation.cpp
aleksei-fedotov: I also see it as _do our best to support user's request for data placement_, but with an addition that _we tell at the end if we were successful or not, explaining why, and let the user decide whether (s)he continues to use allocated memory that way or decide to free it and try allocate it differently using other techniques_

> Even kluge solution is to read /sys/kernel/mm/transparent_hugepage/enabled on startup and fallback to madvise() when we got -EBUSY from move_pages()

I agree that looking at settings at startup is a good idea. However, I thought we should definitely call `madvise` before the call to `move_pages` once we found THP enabled since  it is likely that `move_pages` failed. Am I wrong?

Alexandr-Konovalov: I agree, but let's postpone the error processing topic.

>  I thought we should definitely call madvise before the call to move_pages once we found THP enabled since it is likely that move_pages failed.

Partially yes, but problem is that `move_pages()` may not fail however interleaving is violated. And there is a race condition wrt changing  /sys/kernel/mm/transparent_hugepage/enabled, so the kluge solution may not work.

aleksei-fedotov (approved):
Looks good to me. The error processing topic is postponed.

PR #2110, opened 2026-06-23
[TCM] Add possibility to build TCM using oneTBB's CMake

aleksei-fedotov:
In accordance with the accepted RFC, this adds possibility to enable build of TCM while configuring TBB project. The patch also adds a few CI configurations to test TCM and its integration with TBB.


INSTALL.md
```diff
@@ -30,6 +30,8 @@ You may want to use some additional options for configuration:
...
 | `-DCMAKE_BUILD_TYPE=Debug`                  | Specify for Debug build                                                       | Not applicable for multi-configuration generators such as Visual Studio generator.                                                                                                                                           |
 | `-DTBB_VERIFY_DEPENDENCY_SIGNATURE=ON\|OFF` | Controls signature verification of dynamic dependencies loaded during runtime | If set to ON, only the signed dynamic dependencies are loaded. This is the recommended behavior. By default, the value is unspecified. Therefore, the warning is printed. To suppress the warning, set the value explicitly. |
+| `-DTCM_BUILD=ON\|OFF`                       | Controls build of Thread Composability Manager (TCM)                          | By default the value is unspecified, which builds TCM only if its dependencies are satisfied. When set explicitly to ON, stops configuration process if TCM cannot be built.                                                 |
```
isaevil: This option needs to be listed in `cmake/README.md`. Not sure if we need to list it here though.

thread_composability_manager/CMakeLists.txt
```diff
@@ -24,6 +32,29 @@ else()
     set(TCM_ARCH 64)
 endif()
 
+# Dealing with dependencies
+# --------------------------------------------------------------------------------------------------
+if (NOT DEFINED TCM_BUILD OR TCM_BUILD OR TCM_TEST)
```
isaevil: Why do we need to guard the build of TCM inside TCM CMakeLists? Unlike TBB, we don't have multiple library targets here. To me it seems like that it should be resolved (build or not build TCM) at TBB CMakeLists, which you already did.

CMakeLists.txt
```diff
@@ -410,5 +414,20 @@ if (TBB4PY_BUILD)
...
+  else()
+    if (TCM_BUILD)
+      message(FATAL_ERROR "Build of Thread Composability Manager (TCM) is disabled since "
+                          "${TCM_SUBDIR_NAME} subdirectory does not exist")
+    endif()
+  endif()
+  unset(TCM_SUBDIR_NAME)
+endif()
+# Thread Composability Manager (TCM) switches
+option(TCM_BUILD "Enable building of TCM" ON)
```
isaevil: Seems odd that we declare an option after its processing. Also, it would be preferable have this option declared next to other options.

INSTALL.md
aleksei-fedotov: Moved. Also, added a cross-reference link to a more detailed list of configuration options.

CMakeLists.txt
aleksei-fedotov: Agree, but can you suggest how to otherwise differentiate the following three situations?
As per RFC, the situation are:
1. TCM build is neither requested nor disabled a.k.a. the option is not specified => Try building TCM, but don't break configuration process if TCM dependencies are not specified, show a warning message instead and continue.
2. TCM build is explicitly requested => Configuration process breaks if TCM dependencies are not satisfied.
3.  TCM build is explicitly disabled => Do not attempt to configure and build TCM, without showing any warning messages.

This is similar behavior as we have for `TBB_VERIFY_DEPENDENCIES_SIGNATURE` configuration option.

thread_composability_manager/CMakeLists.txt
aleksei-fedotov: Elaborating on the previous comment about odd specification of the option after it is checked, it is responsibility of TCM configuration step to determine whether it can be actually built or not. Generally, parent project is not aware about TCM's configuration and build steps. Parent project simply does not attempt to configure TCM, if its build was explicitly disabled.

So, the whole story about that is to make TCM configuration and build unintrusive as much as possible unless explicitly requested.

CMakeLists.txt
isaevil: Let's at least keep options close to each other but leave adding the subdirectory here.

aleksei-fedotov: Addressed through introducing of an additional CMake variable.

thread_composability_manager/CMakeLists.txt
```diff
@@ -24,6 +32,29 @@ else()
...
+  # Find the HWLOC library
+  if (NOT TARGET HWLOC::hwloc)
+    list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake/modules)
+    if (IS_TCM_STANDALONE OR TCM_BUILD_OPTION_EXPLICITLY_SET)
+      find_package(HWLOC REQUIRED 2.5 VERSION)
+    else()
+      find_package(HWLOC 2.5 QUIET)
+    endif()
+    if (HWLOC_FOUND)
+      message(STATUS "Enabling build of Thread Composability Manager (TCM)")
```
aleksei-fedotov: Adding the sentence about steps to disable build of TCM.
```suggestion
      message(STATUS "Enabling build of Thread Composability Manager (TCM). Specify `-DTCM_BUILD=OFF` to disable.")
```

isaevil (approved):
LGTM!

PR #2119, opened 2026-07-01
Check that OS return non-zero system page size

Alexandr-Konovalov:
### Description 

System page size is acquired from OS, it's reasonable to check that it's valid.

### DIFF
--- a/src/tbb/misc.cpp
+++ b/src/tbb/misc.cpp
@@ -69,7 +69,7 @@ size_t DefaultSystemPageSize() {
 #else
     long page_size = sysconf(_SC_PAGESIZE);
     __TBB_ASSERT_RELEASE(page_size > 0, "Failed to get system page size");
-    return page_size;
+    return static_cast<size_t>(page_size);
 #endif
 }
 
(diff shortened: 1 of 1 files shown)


src/tbb/misc.cpp
```diff
@@ -64,9 +64,12 @@ size_t DefaultSystemPageSize() {
 #if _WIN32
     SYSTEM_INFO si;
     GetSystemInfo(&si);
+    __TBB_ASSERT_RELEASE(si.dwPageSize > 0, "Failed to get system page size");
     return si.dwPageSize;
 #else
-    return sysconf(_SC_PAGESIZE);
+    long page_size = sysconf(_SC_PAGESIZE);
+    __TBB_ASSERT_RELEASE(page_size > 0, "Failed to get system page size");
+    return static_cast<size_t>(page_size);
```
aleksei-fedotov, line 72: Let's move assert along with the `DefaultSystemPageSize` function to the interface function, that is [default_page_size](https://github.com/uxlfoundation/oneTBB/blob/d2addf4f7bc23db3e9c8c0e8b1464274134fd8e5/src/tbb/governor.h#L82). Use assert only in debug mode.

Alexandr-Konovalov: I'm not sure. I'd see debug-only asserts as a check against our own internal assumptions, and there we are validating external data.

PR #2120, opened 2026-07-02
Make task_arena constants constexpr

### LINKED ISSUE #529: Cannot link code using tbb::task_arena::automatic
Since `tbb::task_arena::automatic` is defined as **non-inline** static member, it cannot be odr-used:
https://github.com/oneapi-src/oneTBB/blob/v2021.3.0/include/oneapi/tbb/task_arena.h#L215
meaning that the symbol cannot be used in a library linked against TBB, because it results in a linker error.

Minimal reproducer: https://godbolt.org/z/nT3rzf76a

This would work if  `tbb::task_arena::automatic` was either `inline` or initialised in the implementation file instead of the header.

It is worth noting that any compiler optimisation inlines the definition (try adding -O1 in the godbolt example) so the problem is only observed in debug (unoptimised) builds of the software stack I work on.

kboyarinov:
### Description 
Make `task_arena::automatic`, `not_initialized` and `selectable` constexpr to allow their ODR use in scenarios like `std::make_unique<tbb::task_arena>(tbb::task_arena::automatic)` from C++17.

Fixes #529 

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown

### DIFF
--- a/include/oneapi/tbb/task_arena.h
+++ b/include/oneapi/tbb/task_arena.h
@@ -244,12 +244,12 @@ protected:
 
 public:
     //! Typedef for number of threads that is automatic.
-    static const int automatic = -1;
+    static constexpr int automatic = -1;
     //! Typedef for current thread index in an uninitialized arena.
-    static const int not_initialized = -2;
+    static constexpr int not_initialized = -2;
 #if __TBB_PREVIEW_TASK_ARENA_CORE_TYPE_SELECTOR
     //! Typedef for core type(s) to be specified by the provided selector.
-    static const int selectable = -2;
+    static constexpr int selectable = -2;
 #endif
 };
 
--- a/test/tbb/test_task_arena.cpp
+++ b/test/tbb/test_task_arena.cpp
@@ -2244,3 +2244,22 @@ TEST_CASE("Test enqueue guarantees when task_arena is combined with task_group")
 }
 
 #endif // TBB_USE_EXCEPTIONS
+
+#if __TBB_CPP17_PRESENT
+int force_constant_odr_use(const int& value) {
+    volatile const int* addr = &value;
+    return *addr;
+}
+
+//! \brief \ref regression
+TEST_CASE("ODR-use task_arena constants") {
+    CHECK(force_constant_odr_use(tbb::task_arena::automatic) == tbb::task_arena::automatic);
+    CHECK(force_constant_odr_use(tbb::task_arena::not_initialized) == tbb::task_arena::not_initialized);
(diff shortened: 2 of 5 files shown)


include/oneapi/tbb/task_arena.h
```diff
@@ -244,12 +244,12 @@ class task_arena_base {
...
     //! Typedef for number of threads that is automatic.
-    static const int automatic = -1;
+    static constexpr int automatic = -1;
     //! Typedef for current thread index in an uninitialized arena.
-    static const int not_initialized = -2;
+    static constexpr int not_initialized = -2;
 #if __TBB_PREVIEW_TASK_ARENA_CORE_TYPE_SELECTOR
     //! Typedef for core type(s) to be specified by the provided selector.
-    static const int selectable = -2;
+    static constexpr int selectable = -2;
```
dnmokhov, line 252: Should the documentation (`task_arena_cls.rst`, `core_type_selector.rst`) be updated accordingly?

test/tbb/test_task_arena.cpp
```diff
@@ -2244,3 +2245,22 @@ TEST_CASE("Test enqueue guarantees when task_arena is combined with task_group")
...
+    return *addr;
+}
+
+//! \brief \ref regression
+TEST_CASE("ODR-use task_arena constants") {
+    CHECK(force_constant_odr_use(tbb::task_arena::automatic) == tbb::task_arena::automatic);
+    CHECK(force_constant_odr_use(tbb::task_arena::not_initialized) == tbb::task_arena::not_initialized);
+#if __TBB_PREVIEW_TASK_ARENA_CORE_TYPE_SELECTOR
+    CHECK(force_constant_odr_use(tbb::task_arena::selectable) == tbb::task_arena::selectable);
+#endif
```
dnmokhov: The macro is currently not defined in this test. Instead of this check, `TBB_PREVIEW_TASK_ARENA_CORE_TYPE_SELECTOR` should be defined before all the includes.

include/oneapi/tbb/task_arena.h
kboyarinov: Changed `task_arena_cls.rst`. `core_type_selector.rst` already defines `selectable` as `constexpr`.

test/tbb/test_task_arena.cpp
kboyarinov: Moved this check to `test_arena_constraints`.

dnmokhov (approved):
LGTM

PR #2121, opened 2026-07-04
Fix slot index handling for constrained nested arenas

dnmokhov:
### Description 
Set thread slot index to 0 during constrained arena init and restore it after, preventing out-of-range access with TBBBind that could occur when a worker of a large outer arena constructed a smaller constrained arena (the worker's slot index can exceed the new arena size).

### Type of change

### Tests
- [ ] not needed

### Documentation
- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown


test/tbb/test_arena_constraints.cpp
```diff
@@ -136,6 +136,35 @@ TEST_CASE("Test constraints argument in create_numa_task_arenas") {
...
+        c.set_core_type(core_types().back());
+        c.set_max_threads_per_core(1);
+
+        // TBBBind changes thread affinity when entering the smaller constrained arena and restores it on exit.
+        // current_thread_index() exceeding the smaller arena size no longer causes out of range access in TBBBind.
+        tbb::task_arena{c}.execute([&outer_affinity] {
+            system_info::affinity_mask inner_affinity = system_info::allocate_current_affinity_mask();
+            REQUIRE_MESSAGE(hwloc_bitmap_weight(inner_affinity) < hwloc_bitmap_weight(outer_affinity),
+                "Nested constrained arena did not narrow the affinity mask.");
+        });
```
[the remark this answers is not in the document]
dnmokhov: Updated

src/tbb/arena.cpp
```diff
@@ -599,9 +600,12 @@ void task_arena_impl::initialize(d1::task_arena_base& ta) {
     observer = construct_binding_observer(
         static_cast<d1::task_arena*>(&ta), arena::num_arena_slots(ta.my_max_concurrency, ta.my_num_reserved_slots),
         ta.my_numa_id, ta.core_type(), ta.max_threads_per_core());
+    // Apply the constraints to this thread and make it appear as slot 0 during arena initialization.
+    auto current_slot = td->my_arena_index;
```
aleksei-fedotov: I propose to use explicit type as it makes the code more readable, i.e. reader does not need to do anything other than reading this line to figure out the type of the variables.

Also, since `current_slot` is meant to remember thread's slot index and isn't meant to change, let's express that explicitly by making the variable `const`ant.

test/tbb/test_arena_constraints.cpp
```diff
@@ -136,6 +136,34 @@ TEST_CASE("Test constraints argument in create_numa_task_arenas") {
...
+    using namespace tbb::info;
+    system_info::initialize();
+
+    tbb::task_arena::constraints c{numa_nodes().back()};
+    c.set_core_type(core_types().back());
+    c.set_max_threads_per_core(1);
+
+    if (default_concurrency() <= default_concurrency(c)) {
+        return; // The constraints do not reduce the arena size, so the test is not applicable.
+    }
```
isaevil: Can we just set `max_concurrency` in the constraints `c` to 1?

test/tbb/test_arena_constraints.cpp
```diff
@@ -136,6 +136,34 @@ TEST_CASE("Test constraints argument in create_numa_task_arenas") {
...
+
+    // parallel_for body runs in the implicit arena, which spans all available threads.
+    tbb::parallel_for(0, 1000, [c](int) {
```
isaevil: There is no guarantee that slots > size of nested arena will be occupied. To make the test reproducible, I'd suggest to saturate arena using `enqueue` and wait on the barrier.

src/tbb/arena.cpp
```diff
@@ -599,9 +600,12 @@ void task_arena_impl::initialize(d1::task_arena_base& ta) {
...
     if (observer) {
         // TODO: Consider lazy initialization for internal arena so
         // the direct calls to observer might be omitted until actual initialization.
+        td->my_arena_index = 0;
         observer->on_scheduler_entry(true);
```
aleksei-fedotov, line 612: Why is it needed to enter the scheduler at this place? Why cannot it wait until actual work execution starts? According to the `execute` method, the thread should first be attached to the arena, and only after that notify about that. If the attach => notify protocol had been followed, it would not cause such undefined behavior.

I guess implementation of a TODO item should resolve the issue automatically.

test/tbb/test_arena_constraints.cpp
dnmokhov: Core-type, NMA-node, or threads-per-core constraining is required for TBBBind to be used:
https://github.com/uxlfoundation/oneTBB/blob/master/src/tbb/arena.cpp#L64

isaevil: Sorry, I meant to additionally set `max_concurrency` to 1. With that we won't need this check as far as I understand.

src/tbb/arena.cpp
aleksei-fedotov: Discussed offline. Leaving the behavior as it is proposed in the patch seems to be the lesser of evils at this point. However, at least adding a commentary saying why entering observer matters here and why violation of observer's protocol cannot be avoided are necessary, I think. Finding a proper (micro)benchmark that would demonstrate the benefit of early entering the observer is another task rised from the discussion.

vossmjp: I also had offline discussion. Unless we have conclusive evidence that removing this first-touch-like behavior doesn't negatively impact workloads, we should keep it. But I think its open for discussion whether this pays off at all, and if not, then we may just remove it, instead of implement it more lazily.

src/tbb/arena.cpp
dnmokhov: Changed.

src/tbb/arena.cpp
dnmokhov: Comment added.

test/tbb/test_arena_constraints.cpp
dnmokhov: Changed.

test/tbb/test_arena_constraints.cpp
dnmokhov: Changed.

test/tbb/test_arena_constraints.cpp
```diff
@@ -136,6 +136,36 @@ TEST_CASE("Test constraints argument in create_numa_task_arenas") {
...
+    tbb::parallel_for(0, n, [&](int) {
+        // TBBBind changes thread affinity when entering the smaller constrained arena and restores it on exit.
+        // current_thread_index() exceeding the smaller arena size no longer causes out of range access in TBBBind.
+        tbb::task_arena{c, 0}.execute([&] {
+            system_info::affinity_mask inner_affinity = system_info::allocate_current_affinity_mask();
+            REQUIRE_MESSAGE(hwloc_bitmap_isincluded(inner_affinity, outer_affinity),
+                "Nested constrained arena affinity mask is not a subset of the outer affinity mask.");
+        });
+        barrier.wait();
+    }, tbb::static_partitioner{});
```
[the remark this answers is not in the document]
dnmokhov: The implicit arena's affinity mask stays unmodified.

This test is for a specific regression. The existing tests already verify worker affinity restoration.

vossmjp (dismissed):
Looks good to me. We can revisit later if we decide to completely remove the temporary binding, but until then, this is necessary bug fix and it looks good.

test/tbb/test_arena_constraints.cpp
```diff
@@ -136,6 +136,44 @@ TEST_CASE("Test constraints argument in create_numa_task_arenas") {
...
+                "Nested constrained arena affinity mask is not a subset of the outer affinity mask.");
+
+            // Indirectly verify that TBBBind applied the affinity for this arena.
+            if (tbb::detail::r1::constraints_default_concurrency(c) < n) { // affinity mask should be smaller
+                REQUIRE_MESSAGE(!hwloc_bitmap_isequal(inner_affinity, outer_affinity),
+                    "TBBBind did not apply affinity: the nested arena mask was not narrowed.");
+            }
+        });
+        barrier.wait();
+    }, tbb::static_partitioner{});
```
aleksei-fedotov: Perhaps, a bit better would be to use `tbb::simple_partitioner()` here to avoid any side effects related to potential future changes in splitting algorithm of the `tbb::static_partitioner`.
`tbb::simple_partitioner()` is much less likely to change, I guess. So, it will split to the smallest possible piece in any case.

aleksei-fedotov (dismissed):
Looks good, but consider a bit more future-proof version of the code.

test/tbb/test_arena_constraints.cpp
dnmokhov: Changed.

aleksei-fedotov (approved):

PR #2129, opened 2026-07-14
[TCM] Correct renegotiating for two requests w/o hyperthreading

aleksei-fedotov:
Fix and a regression test for over-subscription of a constrained CPU region during renegotiation.

Two active permits (A and B) share the same constrained region - the "one hardware thread per core" mask. That region has a capacity equal to the number of cores, while the process still owns the sibling hardware threads outside of it.

The bug: while computing how much can be granted to a permit that already owns some resources in
the shared region, try_satisfy() treated the region capacity minus the *competitors'* grants as
immediately available, and then added the permit's *own* current grant on top - double counting
the slots the permit already occupies. As a result a renegotiation triggered by freeing an
unrelated resource would grant one of the permits more than the region can hold, ending up with
(A_grant + B_grant) > region capacity. That later trips the "Incorrectly granted permit detected"
assertion in try_satisfy() when the over-subscribed region is scanned again.

The scenario in the test deterministically reconstructs that state:
  1. A requests the shared region and is granted the whole of it.
  2. B requests the same region; TCM negotiates A down so that A + B == capacity.
  3. C grabs the free sibling threads (the resources outside the region).
  4. Releasing C triggers renegotiate_permits(), whose active-permit saturation loop tries to
     grow A or B back towards its desired concurrency. With the bug this over-grants inside the
     already-full region.
The invariant checked is that the grants of the two permits sharing the region never exceed the
region capacity.

### DIFF
--- a/thread_composability_manager/test/test_constrained_requests.cpp
+++ b/thread_composability_manager/test/test_constrained_requests.cpp
@@ -451,8 +451,6 @@ TEST("Two constrained requests oversubscribing first core") {
 void test_two_requests_not_oversubscribe_after_renegotiation(tcm_cpu_constraints_t constraints,
                                                              const int32_t constraints_capacity)
 {
-  one_thread_per_core_mask region{};
-
   tcm_client_id_t clidA = connect_new_client();
 
   // A takes the whole shared region.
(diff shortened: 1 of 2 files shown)


thread_composability_manager/test/test_constrained_requests.cpp
```diff
@@ -426,6 +448,83 @@ TEST("Two constrained requests oversubscribing first core") {
   test_two_requests<first_core_mask, first_core_mask>{test_config}();
 }
 
+void test_two_requests_not_oversubscribe_after_renegotiation(tcm_cpu_constraints_t constraints,
+                                                             const int32_t constraints_capacity)
+{
+  one_thread_per_core_mask region{};
```
isaevil: Seems redundant.
```suggestion
```

aleksei-fedotov: Indeed. Thanks!

isaevil (approved):

PR #2130, opened 2026-07-17
Update leave_policy/parallel phase-related RFCs

isaevil:
Updated the RFCs with new API considerations and elaborating open questions in more details.
Note, that ideas proposed in new changes for `parallel_phase` RFC are not final - some of them suggested more confidently, while some should be discussed during the review.

### DIFF
--- a/rfcs/experimental/global_control_fast_leave/README.md
+++ b/rfcs/experimental/global_control_fast_leave/README.md
@@ -41,9 +41,13 @@ providing a global override mechanism.
 This proposal introduces a new `global_control` parameter called `leave_policy` that, when set to "fast",
 would override the default behavior for arenas that are initialized with `leave_policy::automatic`.
 Setting the parameter would not affect arenas that are already initialized or initialized with an
-explicit `leave_policy`. After initialization, the parallel phase API can independently modify the
-arena's leave behavior at runtime, allowing workers to be retained during active parallel phases
-regardless of the initial state set by the global control.
+explicit `leave_policy`. This follows the decision made in the
+[Completely disable new behavior](../parallel_phase_for_task_arena/README.md#completely-disable-new-behavior)
+section of the parallel phase RFC to keep the arena leave policy static for the lifetime of the arena,
+since there are no use cases for dynamic transition between the states yet. 
+After initialization, the parallel phase API can independently modify the arena's leave behavior
+at runtime, allowing workers to be retained during active parallel phases regardless of
+the initial state set by the global control.
 
 ### New Public API
 
@@ -96,7 +100,10 @@ The `leave_policy` parameter would control whether arenas are initialized with a
 | `task_arena::leave_policy::fast` | Workers leave immediately (fast leave enabled) |
 
 When multiple `global_control` objects exist for `leave_policy`, their logical disjunction would be
-used (consistent with the `terminate_on_exception` parameter). This means if any
+used (consistent with the `terminate_on_exception` parameter). For `leave_policy`, the request for
+`fast` is the restrictive one, since it asks the library to give up threads sooner, so it wins over
+the absence of such a request. As more leave policies appear, new strategy of choosing the next
+`leave_policy` can be implemented. At this point, it means that if any
 `global_control(leave_policy, task_arena::leave_policy::fast)` is active, fast leave would be enabled globally.
 
 ### Proposed Implementation Strategy
@@ -327,20 +334,6 @@ The default behavior could be changed to fast leave, making delayed leave opt-in
 
 **Cons:**
 - Fixes performance regression for some customers while causing it for others
-
-## Open Questions
-
-1. **Naming**: Should it convey "default override" semantics?
-   - Current proposal: parameter name matches `task_arena::leave_policy` for consistency
-   - Alternative: e.g., `override_default_leave_policy` to clarify no effect on already initialized arenas
-
-2. **Scope and Granularity**
-   - Current proposal: only affect arenas initialized after the `global_control` is set
-   - Alternative: allow retroactive effect on existing arenas
-   - Alternative: allow per worker control (e.g., via observers) to enable/disable fast leave on a per-thread basis
-
-3. **Composition**: Should the global control be a logical disjunction, first-registered wins, last-set wins, etc.?
-   - Current proposal: disjunction
 
 ## Exit Criteria
 
--- a/rfcs/experimental/global_control_fast_leave/README.md
+++ b/rfcs/experimental/global_control_fast_leave/README.md
@@ -99,12 +99,11 @@ The `leave_policy` parameter would control whether arenas are initialized with a
 | `task_arena::leave_policy::automatic` (default) | Workers follow the default system-specific policy (may spin before leaving) |
 | `task_arena::leave_policy::fast` | Workers leave immediately (fast leave enabled) |
 
-When multiple `global_control` objects exist for `leave_policy`, their logical disjunction is
-used (consistent with the `terminate_on_exception` parameter). For `leave_policy`, the request for
-`fast` is the restrictive one, since `automatic` leaves the choice up to the library, so it wins over
-the absence of such a request. If more leave policies appear, a new strategy of choosing the
-`leave_policy` can be implemented. At this point, it means that if any
-`global_control(leave_policy, task_arena::leave_policy::fast)` is active, fast leave is enabled globally.
+When multiple `global_control` objects request `leave_policy`, the first active
+request for `fast` determines the globally active policy, and it stays in effect until the
+corresponding `global_control` object is destroyed. The effect of any other `leave_policy` requests
+made while it is active is not guaranteed at this point. If more leave policies appear, a new
+strategy of choosing the `leave_policy` can be implemented.
 
 ### Proposed Implementation Strategy
 
@@ -171,7 +170,7 @@ While the feature is in preview state, the
 would need to be extended with documentation for the `global_control::leave_policy` parameter:
 - Add a new section "Global Control Integration" describing the `leave_policy` parameter
 - Update the Synopsis to include the `global_control` header and `leave_policy` parameter
-- Add a description of the parameter semantics and selection rule (logical disjunction)
+- Add a description of the parameter semantics and selection rule
 - Document the interaction between `global_control::leave_policy` and per-arena `leave_policy`
 - Add usage examples showing how to combine global fast leave with parallel phases
 - Include a note explaining that `global_control::leave_policy` provides application-wide control
@@ -236,7 +235,7 @@ int main() {
 }
 ```
 
-#### Global Control Scope, Initialization Order, and Disjunction
+#### Global Control Scope, Initialization Order, and Active Value Selection Rule
 
 ```cpp
 // No global_control is active yet; this call lazily initializes the implicit arena with automatic (the default).
@@ -255,7 +254,7 @@ tbb::parallel_for(0, 1000000, [](int i) { do_parallel_work(i); });
     {
         tbb::global_control gc2(tbb::global_control::leave_policy, tbb::task_arena::leave_policy::automatic);
         tbb::task_arena arena2;
-        // Both gc1 (fast) and gc2 (automatic) are active. Disjunction rule: any fast value present => fast wins.
+        // gc1 (fast) became active first, so it stays in effect; the effect of gc2 is not guaranteed.
         arena2.execute([&] { /* ... */ });
     }
 }
@@ -339,6 +338,4 @@ The default behavior could be changed to fast leave, making delayed leave opt-in
 
 The following conditions need to be met to move the feature from experimental to fully supported:
 - Open questions regarding the API should be resolved.
-- User feedback should confirm usability and performance improvements in mentioned scenarios and that no unforeseen
-  issues arise.
 - The feature must be added to the oneTBB specification and accepted.
--- a/rfcs/experimental/global_control_fast_leave/README.md
+++ b/rfcs/experimental/global_control_fast_leave/README.md
@@ -44,7 +44,7 @@ Setting the parameter would not affect arenas that are already initialized or in
 explicit `leave_policy`. This follows the decision made in the
 [Completely disable new behavior](../parallel_phase_for_task_arena/README.md#completely-disable-new-behavior)
 section of the parallel phase RFC to keep the arena leave policy static for the lifetime of the arena,
-since there are no use cases for dynamic transition between the states yet. 
+since we are not aware of practical use cases for dynamic transition between the states. 
 After initialization, the parallel phase API can independently modify the arena's leave behavior
 at runtime, allowing workers to be retained during active parallel phases regardless of
 the initial state set by the global control.
@@ -99,12 +99,12 @@ The `leave_policy` parameter would control whether arenas are initialized with a
 | `task_arena::leave_policy::automatic` (default) | Workers follow the default system-specific policy (may spin before leaving) |
 | `task_arena::leave_policy::fast` | Workers leave immediately (fast leave enabled) |
 
-When multiple `global_control` objects exist for `leave_policy`, their logical disjunction would be
+When multiple `global_control` objects exist for `leave_policy`, their logical disjunction is
 used (consistent with the `terminate_on_exception` parameter). For `leave_policy`, the request for
-`fast` is the restrictive one, since it asks the library to give up threads sooner, so it wins over
-the absence of such a request. As more leave policies appear, new strategy of choosing the next
+`fast` is the restrictive one, since `automatic` leaves the choice up to the library, so it wins over
+the absence of such a request. If more leave policies appear, a new strategy of choosing the
 `leave_policy` can be implemented. At this point, it means that if any
-`global_control(leave_policy, task_arena::leave_policy::fast)` is active, fast leave would be enabled globally.
+`global_control(leave_policy, task_arena::leave_policy::fast)` is active, fast leave is enabled globally.
 
 ### Proposed Implementation Strategy
 
--- a/rfcs/experimental/parallel_phase_for_task_arena/README.md
+++ b/rfcs/experimental/parallel_phase_for_task_arena/README.md
@@ -208,7 +208,8 @@ The same use case can be expressed using the `parallel_phase` object, assuming m
 
 ```cpp
 void handle_request(Request req) {
-    tbb::task_arena::parallel_phase phase{tbb::attach{}};
+    tbb::task_arena::parallel_phase phase{tbb::attach{},
+                                          tbb::task_arena::parallel_phase::end_flags::with_fast_leave};
     //
     // Some composition of parallel and serial computations
     //
--- a/rfcs/experimental/parallel_phase_for_task_arena/README.md
+++ b/rfcs/experimental/parallel_phase_for_task_arena/README.md
@@ -42,12 +42,10 @@ Let’s consider both “Delayed leave” and “Fast leave” as 2 different st
 
 <img src="completely_disable_new_behavior.png" width=800>
 
-There will be a question that we need to answer:
-* Do we see any value if arena potentially can transition from one to another state?
-
-To answer this question, the following scenarios should be considered:
-* What if different types of workloads are mixed in one application?
-* Different types of arenas can be used for different types of workloads.
+The question is whether we should allow the arena to dynamically transition from one state to
+another. At this point, there are no use cases for this, hence the proposal is to keep
+the state static for the lifetime of the arena. If we see a value in the future, the current
+implementation should be flexible enough to support this (see [technical details](#technical-details)).
 
 ### When threads should leave?
 
@@ -123,7 +121,7 @@ Summary of API changes:
 defaulted to "automatic".
 * Add functions to start and end the parallel phase to the `task_arena` class
 and the `this_task_arena` namespace.
-* Add RAII class to map a parallel phase to a code scope.
+* Add a `task_arena::parallel_phase` RAII class to map a parallel phase to a code scope.
 
 ```cpp
 class task_arena {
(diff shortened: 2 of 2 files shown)


rfcs/experimental/parallel_phase_for_task_arena/README.md
```diff
@@ -148,28 +146,128 @@ class task_arena {
...
 }
 ```
 The _parallel phase_ continues until each previous `start_parallel_phase` call
-to the same arena has a matching `end_parallel_phase` call.<br>
-Let's introduce RAII scoped object that will help to manage the contract.
+to the same arena has a matching `end_parallel_phase` call.
+Let's also introduce RAII object `parallel_phase` that will help to manage the contract.
+Rather than introducing a separate RAII type in the `this_task_arena` namespace,
+`task_arena::parallel_phase` can be constructed with the `tbb::attach` tag to map the phase to
+the implicit arena associated with the calling thread.
```
isaevil: This part relates to the open question about RAII parallel_phase object for "this_task_arena" ([see discussion](https://github.com/uxlfoundation/oneTBB/pull/1522#discussion_r1871955641)).

rfcs/experimental/parallel_phase_for_task_arena/README.md
```diff
@@ -148,28 +146,128 @@ class task_arena {
...
+is a high-level hint to the scheduler, it makes sense that other scheduling hints could be tied to
+it as well, so the API is designed to be configurable. A symmetric `start_flags` parameter
+could be added to `start_parallel_phase` in the future if needed.
 
 If the end of the parallel phase is not indicated by the user, it will be done automatically when
 the last public reference is removed from the arena (i.e., task_arena has been destroyed or,
 for an implicitly created arena, the thread that owns it has completed).
 This ensures correctness is preserved (threads will not be retained forever).
 
+Introduction of the explicit functions `start_parallel_phase` and `end_parallel_phase` opens
```
isaevil, line 229: This part with the examples tries to figure out the necessity of explicit "start/end_parallel_phase" functions vs. limiting the API only to RAII class ([see discussion](https://github.com/uxlfoundation/oneTBB/pull/1522#discussion_r1871505223)).

rfcs/experimental/parallel_phase_for_task_arena/README.md
```diff
@@ -148,28 +146,128 @@ class task_arena {
...
+    }
+
+    void on_idle_timeout() {
+        phase.reset();
+    }
+};
+```
+If the friction of dealing with the RAII object is considered too high, then it makes sense to provide the explicit functions as well.
+
+The start of _parallel phase_ can also be used as a warm-up hint for the workers to enter
```
isaevil: And this paragraph is about the expectations from parallel_phase API when a calling thread doesn't yet have an associated arena with it.

rfcs/experimental/parallel_phase_for_task_arena/README.md
```diff
@@ -148,28 +146,128 @@ class task_arena {
...
+        tbb::this_task_arena::end_parallel_phase(tbb::task_arena::parallel_phase::end_flags::with_fast_leave);
+    });
+}
+```
+
+The same use case can be expressed using the `parallel_phase` object, assuming move construction is supported:
+
+```cpp
+void handle_request(Request req) {
+    tbb::task_arena::parallel_phase phase{tbb::attach{}};
```
dnmokhov:
```suggestion
void handle_request(Request req) {
    tbb::task_arena::parallel_phase phase{tbb::attach{},
                                          tbb::task_arena::parallel_phase::end_flags::with_fast_leave};
```

rfcs/experimental/parallel_phase_for_task_arena/README.md
```diff
@@ -148,28 +146,129 @@ class task_arena {
...
     };
+    
+    void start_parallel_phase();
+    void end_parallel_phase(parallel_phase::end_flags e_flags = {});
 };
 
 namespace this_task_arena {
     void start_parallel_phase();
-    void end_parallel_phase(bool with_fast_leave = false);
+    void end_parallel_phase(task_arena::parallel_phase::end_flags e_flags = {});
```
aleksei-fedotov: This seems to change the default from `false` to `true` and makes an ending of a parallel phase using anything besides `with_fast_leave` value impossible. Is it what is intended?

isaevil: Not necessarily. The value of `with_fast_leave` is unspecified, so internally it can be set any non-zero value.

aleksei-fedotov: Can we use `task_arena::leave_policy` type instead?

Consider `void end_parallel_phase(task_arena::leave_policy end_phase_policy)`.

I think introducing separate `end_flags` type for essentially the same entity is not needed.

isaevil: [Again, parallel_phase is not about selecting `leave_policy` for task arena.](https://github.com/uxlfoundation/oneTBB/pull/1522#discussion_r1861130746)

isaevil:
@wangleis this is the RFC about task_arena `leave_policy` setting and `parallel_phase` API. Feel free to provide feedback on the API and maybe share some additional use-cases (if any) that would be useful to consider as well.

rfcs/experimental/parallel_phase_for_task_arena/README.md
isaevil: This question should be resolved by choosing the way to create RAII `parallel_phase` object (attach parameter vs using factory function).

isaevil:
@akukanov @vossmjp @aleksei-fedotov @kboyarinov @dnmokhov please take a look. I believe the remaining questions about `parallel_phase` API are about the necessity of explicit functions and the way to create RAII object.

rfcs/experimental/global_control_fast_leave/README.md
```diff
@@ -41,9 +41,13 @@ providing a global override mechanism.
 This proposal introduces a new `global_control` parameter called `leave_policy` that, when set to "fast",
 would override the default behavior for arenas that are initialized with `leave_policy::automatic`.
 Setting the parameter would not affect arenas that are already initialized or initialized with an
-explicit `leave_policy`. After initialization, the parallel phase API can independently modify the
-arena's leave behavior at runtime, allowing workers to be retained during active parallel phases
-regardless of the initial state set by the global control.
+explicit `leave_policy`. This follows the decision made in the
+[Completely disable new behavior](../parallel_phase_for_task_arena/README.md#completely-disable-new-behavior)
+section of the parallel phase RFC to keep the arena leave policy static for the lifetime of the arena,
+since there are no use cases for dynamic transition between the states yet.
```
vossmjp:
```suggestion
since we are not aware of practical use cases for dynamic transition between the states. 
```

vossmjp: There may be use cases already, we just don't know about them.

rfcs/experimental/global_control_fast_leave/README.md
```diff
@@ -96,7 +100,10 @@ The `leave_policy` parameter would control whether arenas are initialized with a
 | `task_arena::leave_policy::fast` | Workers leave immediately (fast leave enabled) |
 
 When multiple `global_control` objects exist for `leave_policy`, their logical disjunction would be
-used (consistent with the `terminate_on_exception` parameter). This means if any
+used (consistent with the `terminate_on_exception` parameter). For `leave_policy`, the request for
+`fast` is the restrictive one, since it asks the library to give up threads sooner, so it wins over
+the absence of such a request. As more leave policies appear, new strategy of choosing the next
```
vossmjp:
```suggestion
`fast` is the restrictive one, since `automatic` leaves the choice up to the library, so it wins over
the absence of such a request. If more leave policies appear, a new strategy of choosing the
```

rfcs/experimental/global_control_fast_leave/README.md
```diff
@@ -96,7 +100,10 @@ The `leave_policy` parameter would control whether arenas are initialized with a
 | `task_arena::leave_policy::fast` | Workers leave immediately (fast leave enabled) |
 
 When multiple `global_control` objects exist for `leave_policy`, their logical disjunction would be
```
vossmjp:
```suggestion
When multiple `global_control` objects exist for `leave_policy`, their logical disjunction is
```

rfcs/experimental/global_control_fast_leave/README.md
```diff
@@ -96,7 +100,10 @@ The `leave_policy` parameter would control whether arenas are initialized with a
...
-used (consistent with the `terminate_on_exception` parameter). This means if any
+used (consistent with the `terminate_on_exception` parameter). For `leave_policy`, the request for
+`fast` is the restrictive one, since it asks the library to give up threads sooner, so it wins over
+the absence of such a request. As more leave policies appear, new strategy of choosing the next
+`leave_policy` can be implemented. At this point, it means that if any
 `global_control(leave_policy, task_arena::leave_policy::fast)` is active, fast leave would be enabled globally.
```
vossmjp:
```suggestion
`global_control(leave_policy, task_arena::leave_policy::fast)` is active, fast leave is enabled globally.
```

rfcs/experimental/parallel_phase_for_task_arena/README.md
```diff
@@ -148,28 +146,155 @@ class task_arena {
...
-    void end_parallel_phase(bool with_fast_leave = false);
+    void start_parallel_phase(task_arena::parallel_phase::flags f = {});
+    void end_parallel_phase(task_arena::parallel_phase::flags f = {});
 }
 ```
 The _parallel phase_ continues until each previous `start_parallel_phase` call
-to the same arena has a matching `end_parallel_phase` call.<br>
-Let's introduce RAII scoped object that will help to manage the contract.
+to the same arena has a matching `end_parallel_phase` call.
+Let's also introduce RAII object `parallel_phase` that will help to manage the contract.
```
vossmjp:
```suggestion
Let's also introduce an RAII object `parallel_phase` that will help to manage the contract.
```

rfcs/experimental/parallel_phase_for_task_arena/README.md
```diff
@@ -148,28 +146,155 @@ class task_arena {
...
+```cpp
+class task_arena {
+    parallel_phase create_parallel_phase(parallel_phase::flags f = {});
+};
+
+namespace this_task_arena {
+    task_arena::parallel_phase create_parallel_phase(task_arena::parallel_phase::flags f = {});
+}
+```
+The approach is more consistent with the rest of task arena API.
```
vossmjp: Which one is "this" approach? You've just described two alternatives.

rfcs/experimental/parallel_phase_for_task_arena/README.md
```diff
@@ -148,28 +146,155 @@ class task_arena {
...
```
isaevil:
```suggestion
The factory function approach seems more consistent with the rest of task arena API.
```

rfcs/experimental/parallel_phase_for_task_arena/README.md
isaevil: Clarified, I meant the factory function approach.

rfcs/experimental/parallel_phase_for_task_arena/README.md
```diff
@@ -42,12 +42,10 @@ Let’s consider both “Delayed leave” and “Fast leave” as 2 different st
 
 <img src="completely_disable_new_behavior.png" width=800>
 
-There will be a question that we need to answer:
-* Do we see any value if arena potentially can transition from one to another state?
-
-To answer this question, the following scenarios should be considered:
-* What if different types of workloads are mixed in one application?
-* Different types of arenas can be used for different types of workloads.
+The question is whether we should allow the arena to dynamically transition from one state to
```
vossmjp: We've decided to not support dynamic transition at this point. So I think you should be more clear that this is a decision and no longer an open question. We can revisit if additional use cases arise (which you can state). But I'd suggest being more clear that for production, this is decided.

Essentially make the language stronger, removing things like "at this point" and maybe start with "One question was whether".

rfcs/experimental/parallel_phase_for_task_arena/README.md
```diff
@@ -148,28 +146,155 @@ class task_arena {
...
+    }
+};
+```
+If the friction of dealing with the RAII object is considered too high, then it makes sense to provide the explicit functions as well.
+
+The start of _parallel phase_ can also be used as a warm-up hint for the workers to enter
+the arena in advance, therefore the arena (including the implicit arena bound to the external thread)
+must be initialized during the first call to `start_parallel_phase`. It means that if a calling thread has
+no associated arena yet, the invocation of `this_task_arena::start_parallel_phase` will initialize
+an arena and bind it to the calling thread.
```
vossmjp: This equally would apply to `this_task_arena::create_parallel_phase` not just `start_parallel_phase`, right? So this is not a deciding factor in choosing RAII object over explicit start/stop functions.

rfcs/experimental/parallel_phase_for_task_arena/README.md
```diff
@@ -285,18 +414,8 @@ the `thread_leave_manager` during the execution of parallel phases. It shows how
...
 Some open questions that remain:
-* Are the suggested APIs sufficient?
-  * In the current version of proposed API, the `scoped_parallel_phase` object can be created
-    only for already existing `task_arena`. Should it be possible for `this_task_arena` as well?
-  * What should be expected from "Parallel Phase" API for `this_task_arena` when a calling thread
-    doesn't yet have any associated arena?
-  * Should parallel phase API be limited only to RAII-only style?
-    * Are there any scenarios where inconvenience of handling `scoped_parallel_phase` object is
-      not acceptable?
 * Are there additional use cases that should be considered that we missed in our analysis?
```
vossmjp: I'd remove this. As we move towards productization, we currently are not aware of any. If they arise, they become a new feature request later.

rfcs/experimental/parallel_phase_for_task_arena/README.md
vossmjp: This is more a process question. We are aware of improvements, but there's nothing documented in the RFC. Should there be?

rfcs/experimental/parallel_phase_for_task_arena/README.md
isaevil: Yes, it is actually a separate question elaborated here. Maybe it is better to separate them by subheaders.

rfcs/experimental/parallel_phase_for_task_arena/README.md
isaevil: Perhaps we could add at least some plan on measuring performance. I was thinking about updating the seismic example, since it showcases sequential composability scenario.

rfcs/experimental/parallel_phase_for_task_arena/README.md
isaevil: Updated.

rfcs/experimental/parallel_phase_for_task_arena/README.md
isaevil: Separated by subheaders.

rfcs/experimental/parallel_phase_for_task_arena/README.md
isaevil: Done.

rfcs/experimental/parallel_phase_for_task_arena/README.md
isaevil: Added "Performance testing" section with the suggestion to use seismic to measure performance.

rfcs/experimental/global_control_fast_leave/README.md
```diff
@@ -95,9 +99,12 @@ The `leave_policy` parameter would control whether arenas are initialized with a
...
 
-When multiple `global_control` objects exist for `leave_policy`, their logical disjunction would be
-used (consistent with the `terminate_on_exception` parameter). This means if any
-`global_control(leave_policy, task_arena::leave_policy::fast)` is active, fast leave would be enabled globally.
+When multiple `global_control` objects exist for `leave_policy`, their logical disjunction is
+used (consistent with the `terminate_on_exception` parameter). For `leave_policy`, the request for
+`fast` is the restrictive one, since `automatic` leaves the choice up to the library, so it wins over
+the absence of such a request. If more leave policies appear, a new strategy of choosing the
+`leave_policy` can be implemented. At this point, it means that if any
+`global_control(leave_policy, task_arena::leave_policy::fast)` is active, fast leave is enabled globally.
```
akukanov: I suggest not to call it "disjunction" in the documentation, especially the public one.

We should describe the behavior in the way that a) gives certain clear guarantees, and b) allows to extend/modify that in the future, ideally without breaking these guarantees. The current guarantee, I think is stated in the last sentence - so let's make it clear and upfront, and then clarify as necessary.

At the same time, I think this stated guarantee will be hard to keep if another policy is ever added - because of "any". Most likely in that case we will have to break the behavior.

rfcs/experimental/parallel_phase_for_task_arena/README.md
akukanov: I'd change wording to "implementation-specific", because in certain contexts (e.g., the C++ standard) "implementation-defined" means "chosen by implementors but clearly documented and guaranteed".

rfcs/experimental/parallel_phase_for_task_arena/README.md
```diff
@@ -42,12 +42,10 @@ Let’s consider both “Delayed leave” and “Fast leave” as 2 different st
...
-There will be a question that we need to answer:
-* Do we see any value if arena potentially can transition from one to another state?
-
-To answer this question, the following scenarios should be considered:
-* What if different types of workloads are mixed in one application?
-* Different types of arenas can be used for different types of workloads.
+One question was whether we should allow the arena to dynamically transition from one state to
+another. Due to the lack of use cases that would justify this, the decision is to keep the state
+static for the lifetime of the arena. The current implementation is flexible enough to support such
+extension (see [technical details](#technical-details)).
```
akukanov: It's not fully clear to me which extension do you mean here.

rfcs/experimental/parallel_phase_for_task_arena/README.md
```diff
@@ -148,28 +146,164 @@ class task_arena {
...
-    class scoped_parallel_phase {
-        scoped_parallel_phase(task_arena& ta, bool with_fast_leave = false);
+    class parallel_phase {
+        // Available only when each type in Flags is a parallel phase flag
+        class flags {
+            template <typename... Flags>
+            flags(Flags... f);
+        };
+        class end_with_fast_leave;
+        parallel_phase(attach, flags f = {});
```
akukanov, line 156: Do we really need the `attach` flag here? An alternative is to assume by default that the parallel phase acts in the current arena of a thread; would it have some downsides compared to the explicit flag?

rfcs/experimental/parallel_phase_for_task_arena/README.md
```diff
@@ -148,28 +146,164 @@ class task_arena {
...
+The start of _parallel phase_ can also be used as a warm-up hint for the workers to enter
+the arena in advance, therefore the arena (including the implicit arena bound to the external thread)
+must be initialized during the first call to `start_parallel_phase`. It means that if a calling thread has
+no associated arena yet, the invocation of `this_task_arena::start_parallel_phase` will initialize
+an arena and bind it to the calling thread.
 
 If the end of the parallel phase is not indicated by the user, it will be done automatically when
 the last public reference is removed from the arena (i.e., task_arena has been destroyed or,
 for an implicitly created arena, the thread that owns it has completed).
 This ensures correctness is preserved (threads will not be retained forever).
```
akukanov: What if the `end_parallel_phase` or `parallel_phase` destructor is called for an arena that is already destroyed?

rfcs/experimental/parallel_phase_for_task_arena/README.md
isaevil: I prefer having the tag because it is a bit more descriptive and in the context of task_arena `attach` tag already means that the instance is associated with the arena of a calling thread.

rfcs/experimental/parallel_phase_for_task_arena/README.md
isaevil: By "extension" I meant extending the functionality around `leave_policy` setting to allow changing it dynamically.

rfcs/experimental/parallel_phase_for_task_arena/README.md
isaevil: It seems like you won't be able to do so without creating a dangling reference to `task_arena` which seems more like "you already lost" situation. `enter/exit_parallel_phase` entry points accept `task_arena_base*` and if it is `nullptr`, arena bound to a calling thread is used.
If user explicitly calls `terminate` before ending the phase, while keeping the task arena object, we can add an assertion that during the phase end, the arena is still initialized.
Interesting thing is that, since we don't propagate `r1::arena` back, you can't start a parallel_phase for `this_task_arena` in one arena and end it in another.

rfcs/experimental/parallel_phase_for_task_arena/README.md
akukanov: Maybe you can adjust the wording to make that clear.

rfcs/experimental/parallel_phase_for_task_arena/README.md
akukanov: OK.

rfcs/experimental/parallel_phase_for_task_arena/README.md
akukanov: If it is a requirement that the lifetime of the arena object cannot end while a parallel phase is active, it should be stated as such. And then, "the end of the parallel phase is not indicated by the user" is not even a concern but simply undefined behavior.

rfcs/experimental/parallel_phase_for_task_arena/README.md
isaevil: Done.

rfcs/experimental/parallel_phase_for_task_arena/README.md
isaevil: Done.

rfcs/experimental/parallel_phase_for_task_arena/README.md
isaevil: Updated.

rfcs/experimental/global_control_fast_leave/README.md
isaevil: Updated.

vossmjp (approved):
Looks good to me.

kboyarinov (approved):

Issue #18, opened 2017-04-26
Reallocating memory using scalable_realloc when emulating Haswell crashes in tbbmalloc (32 bit only).


J-cztery:
`size_t size = 63;
void *ptr = scalable_malloc(size);
size += 4;
ptr = scalable_realloc(ptr, size);`

And then when running the following command: 
sde  -hsw -- TestTbbMalloc.exe
An access violation is received in tbbmalloc.dll!___intel_memcpy

The problem does not happen when compiling tbbmalloc from source using Visual Studio. I guess this is related to memcpy substituted by ICL. 
Reproduced on 2017 Update 5(Windows). 
The same code does not crash on real Haswell CPU without emulator.

alexey-katranov:
Hello! It turned out that I failed to reproduce the issue. Could you provide more details about your experiment, please? What compiler and options are used? Can you share the complete reproducer? What versions of the software are used (e.g. sde, compiler, OS). Any other relevant information that can be useful in your opinion.

Thank you in advance.

J-cztery:
Hi, 
Are you using tbbmalloc.dll as available in 2017 Update 5 binary?

I am sorry but i do not have access to that environment anymore. The compiler was Visual Studio 2015. 
It is a complete reproducer, a main method consisting only of the above code.

alexey-katranov:
It seems we did not move forward with the investigation. If the issue still affects you, feel free to reopen issue.

J-cztery:
? Why did you close the issue if it is not fixed?

alexey-katranov:
We failed to reproduce it. If it still persists, lets reopen and try something else to figure out the root cause.

J-cztery:
>We failed

Agreed.

Issue #14, opened 2017-04-02
Support for VS 2017


adityaatluri:
Hi,
Will there be support for visual studio 2017 community edition?

alexey-katranov:
According to the latest changes (Intel TBB 2017 Update 5) "Microsoft* Visual Studio* 2017" is supported. I suppose it does not depend on a specific edition. Do you observe some issues with the community edition?

Chandra03ram:
I don't think (Intel TBB 2017 Update 5) is supported for visual studio 2017. I'm using visual studio 2017 to build my project, where i need to use tbb, but there is no version of TBB that supports, visual studio 2017.
Generally, lib folder in the TBB should have vc15 folder to support vs2017. But, it is not provided in the TBB, thus there is no support of TBB for vs2017.

akukanov:
The [documentation](https://www.threadingbuildingblocks.org/docs/help/tbb_userguide/Windows_OS_ug.html) specifies that the same binaries (vc14) are used for both VS 2015 and VS 2017.

PR #2136, opened 2026-07-27
Fix lost wakeup in `notify_by_address` on AArch64: `mutex`/`rw_mutex` waiters sleep forever on an already-satisfied predicate

DarthSim:
This patch and the notes were generated by Claude based on a locally reproduced deadlock. This is not my area of expertise, but the evidence is clear, and the fix is straightforward. However, I'm not quite sure that `notify_by_address` is the right place to put the fence, so please review carefully.

## Summary

`tbb::detail::r1::notify_by_address*` call the `*_relaxed` notify variants of `concurrent_monitor` without issuing the full memory fence that the monitor's wakeup protocol requires on the notifier side. Their callers (`d1::mutex::unlock`, `d1::rw_mutex::unlock`, `d1::waitable_atomic`) only perform an atomic RMW on the waited address before calling them, and an atomic RMW is not a two-sided fence. On weakly-ordered architectures (AArch64) this permits the classic store-buffer outcome: the notifier's unfenced `my_waitset.empty()` early-exit check reads a stale "empty" waitset while the waiter's predicate re-check read the stale pre-unlock state. The notification is skipped, and the waiter sleeps forever on a mutex that is actually free.

We hit this in production on AWS Graviton 3 (random, unrecoverable process hangs after hours/days of load) and captured a live hung process whose memory state proves the mechanism end-to-end (full forensic evidence below). x86-64 is unaffected because TSO makes the RMW-then-load sequence behave as if fenced — consistent with this class of hang being reported on ARM only.

Affected: oneTBB 2021.13.x (including the `2021.13-ov` branch bundled with OpenVINO) through current mainline — `src/tbb/address_waiter.cpp`, `src/tbb/concurrent_monitor.h`, `include/oneapi/tbb/mutex.h`, and `include/oneapi/tbb/rw_mutex.h` are functionally unchanged from v2021.13.0 through v2023.1.0.

## Environment

- oneTBB 2021.13.0 (`libtbb.so.12.13`, prebuilt `oneapi-tbb-2021.13.0-lin-arm64` bundled with OpenVINO 2025.2)
- Linux aarch64: AWS Graviton 3 (64 cores, production) and a 10-core Apple M1 in Docker (reproduction)
- Workload: OpenVINO CPU inference (`ov::InferRequest::infer` → `CPUStreamsExecutor::execute`) called from a multithreaded host application; the deadlocked path itself is generic TBB scheduler code, not OpenVINO-specific

## Symptom

The process hangs permanently. All-thread stacks show three groups:

1. **One thread spinning** (100% CPU, `sched_yield` loop) inside the task dispatcher's stealing/wait loop, waiting for a nested parallel region to complete:

   ```
   #0 __sched_yield
   #1 libtbb.so.12.13   (task dispatcher wait/steal loop)
   #2 …openvino arm cpu plugin kernels…
   #14 libtbb.so.12.13  (task_arena wait/execute)
   #15 ov::threading::CPUStreamsExecutor::execute(std::function<void()>)
   #18 ov::IAsyncInferRequest::infer()
   ```

2. **One TBB worker blocked *inside* `spawn`** — this is the anomaly:

   ```
   #0 syscall (futex FUTEX_WAIT_PRIVATE, val=2)
   #1 tbb::detail::r1::wait_on_address(void*, delegate_base&, unsigned long) +0x390
   #2 libtbb +0x1e1d4   \
   #3 libtbb +0x1e220    > market::adjust_demand → d1::rw_mutex::lock (writer)
   #4 libtbb +0x1e998   /
   #5 tbb::detail::r1::spawn(task&, task_group_context&, unsigned short) +0x134
   #6 …caller in a parallel region…
   ```

   i.e. `spawn → arena::advertise_new_work` → `arena::request_workers` → `threading_control::adjust_demand` → `market::adjust_demand`, blocked acquiring `market::my_mutex` (a `d1::rw_mutex`) as a **writer**.

3. **All other TBB workers idle**, parked on their per-thread semaphores (`FUTEX_WAIT_PRIVATE`, val=2).

Because the blocked worker never finishes its in-flight task, the nested region's `wait_context` never completes, the spinning thread waits forever, and nothing ever wakes the idle workers (the code that would request/wake them is exactly the code that is blocked).

## Forensic evidence from the live hung process

All values read from `/proc/<pid>/mem` without perturbing the process:

**The rw_mutex the worker is waiting for** (`market::my_mutex`, identified by the containing object's vtable + member offset):

```
m_state = 0x2        // == WRITER_PENDING only; WRITER clear, no READERS
                     // => the mutex is FREE and the writer's wakeup condition
                     //    !(m_state & BUSY) is TRUE right now
```

**The waiter's `sleep_node<address_context>`** (on the blocked worker's stack; layout matches the source exactly):

```
next            = 0x…e4d8f0   // -> address_waiter_table bucket sentinel
prev            = 0x…e4d8f0   // -> same; node is linked, sole element
my_address      = &market::my_mutex
my_context      = 0           // WRITER_CONTEXT
my_initialized  = 1
my_is_in_list   = 1           // never cleared by any notify
my_epoch        = 0x2235
futex word      = 2           // committed, sleeping
```

**The `address_waiter_table` bucket** for that address:

```
sentinel.next = sentinel.prev = &node   // waiter visibly linked, both directions
my_epoch      = 0x2235                  // == node.my_epoch
```

The bucket epoch equals the node epoch. Every notify that enters the locked region increments the epoch (`notify`, `notify_one_relaxed`, `abort_all_relaxed` all do `my_epoch.store(my_epoch.load()+1)` under `my_mutex`), so **no notification ever entered the locked region after this waiter's `prepare_wait`** — despite the fact that `rw_mutex::unlock` is guaranteed to have called `notify_by_address(this, WRITER_CONTEXT)`: the current `m_state` still has `WRITER_PENDING` set, and unlock reads the state *after* its RMW (`curr_state = (m_state &= READERS | WRITER_PENDING)`), so it observed `WRITER_PENDING` and took the notify branch. The only way to reconcile this with the bucket state is the unfenced `my_waitset.empty()` early exit in `notify_relaxed` reading a stale (empty) list head.

## Root cause

The monitor protocol requires a full fence on **both** sides of the wait/notify handshake, and the code documents this:

- Waiter side — `concurrent_monitor.h`, end of `prepare_wait()`:

  ```cpp
  // Prepare wait guarantees Write Read memory barrier.
  // In C++ only full fence covers this type of barrier.
  atomic_fence_seq_cst();
  ```

- Notifier side — the non-relaxed variants do issue it:

  ```cpp
  void notify_one() {
      atomic_fence_seq_cst();
      notify_one_relaxed();
  }
  ```

- But `notify_by_address` / `notify_by_address_one` / `notify_by_address_all` call the `*_relaxed` variants with **no fence**, and their callers only perform an atomic RMW on the waited address. `mutex.h` even states the requirement:

  ```cpp
  // We need Write Read memory barrier before notify that reads the waiter list.
  // In C++ only full fence covers this type of barrier.
  my_flag.exchange(false);      // an RMW — not a fence
  my_flag.notify_one_relaxed();
  ```

An atomic RMW operation is not equivalent to `atomic_thread_fence(seq_cst)` for *independent* subsequent loads. On AArch64 with LL/SC lowering (`ldaxr`/`stlxr`), a later relaxed load of an unrelated address may be satisfied between the exclusive pair — i.e. before the RMW's store is globally visible; formally, the C++ memory model also does not forbid the outcome, as the two RMWs are on different objects and create no synchronizes-with edge with the relaxed loads. The resulting store-buffer interleaving:

```
waiter (rw_mutex::lock)                notifier (rw_mutex::unlock)
--------------------------------       -----------------------------------
insert node into waitset               m_state RMW (frees mutex, sees PENDING)
atomic_fence_seq_cst()                 my_waitset.empty() -> stale "empty"
re-check predicate -> stale BUSY       return, nobody woken
commit sleep (forever)
```

Both threads read the old values; the wakeup is lost. On x86-64 the locked RMW is a full barrier, the stale `empty()` read is impossible, and the bug cannot occur — matching field reports (ARM-only hangs).

Disassembly confirms the shipped aarch64 binaries contain **no `dmb` instruction** anywhere in the `notify_by_address*` entry paths.

## Fix

Issue `atomic_fence_seq_cst()` at the start of the three `notify_by_address*` entry points in `src/tbb/address_waiter.cpp`, restoring the documented fence-on-both-sides protocol for every `d1::mutex` / `d1::rw_mutex` / `waitable_atomic` user. The cost is confined to the notify slow path, which is reached only when a contended waiter is or may be present. The patch applies cleanly to v2021.13.0, v2021.13.3, and current mainline.

With the fence in place, GCC 12 emits `dmb ish` at the top of each function on
aarch64.

## Reproduction

No minimal reproducer yet — in our workload (OpenVINO inference on a 64-core Graviton 3 under sustained concurrent load) the hang appears after hours to days; the window is a few instructions wide and needs a contended `d1::rw_mutex` whose unlock races a committing waiter. The live-process forensics above (futex args from `/proc/<tid>/syscall`, mutex/node/bucket state from `/proc/<pid>/mem`) can be used to positively identify the condition in any hung process:

1. a thread parked in `wait_on_address` whose sleep node is still linked in its `address_waiter_table` bucket,
2. the waited address containing a value that satisfies the waiter's predicate (e.g. rw_mutex `m_state == WRITER_PENDING`), and
3. the bucket epoch equal to the node epoch.

## Second capture and intervention confirmation

We reproduced the hang a second time on the same setup. The second specimen had the identical shape but a different victim lock: a **`d1::mutex`** (`waitable_atomic<bool>`, taken in the same `spawn → advertise_new_work → request_workers` path) whose flag byte was `0` (**unlocked**) while a writer-context waiter slept on it — node linked as the sole element of its bucket, `my_is_in_list == 1`, and bucket epoch equal to the node epoch (`0x737ce`). So both notify entry points involved (`notify_by_address` from `rw_mutex::unlock`, `notify_by_address_one` from `mutex::unlock`) have been observed losing a wakeup in the wild.

We then confirmed the diagnosis by intervention. After a few hours stuck, we used a single-thread ptrace injection to make one idle thread in the hung process execute exactly the call the buggy path skipped:

```
tbb::detail::r1::notify_by_address_all(&stuck_mutex);
```

The sleeping waiter woke within a second, the bucket epoch advanced (`0x737ce → 0x737d1`) as organic notify traffic resumed, the thread that had been spinning in the dispatch loop completed its parallel region, and the process fully recovered — it served an inference request normally 120 ms later. No other state was modified: delivering the one lost notification was sufficient to restore the process, which isolates the failure to the skipped notification itself.

### DIFF
--- a/include/oneapi/tbb/mutex.h
+++ b/include/oneapi/tbb/mutex.h
@@ -1,5 +1,6 @@
 /*
     Copyright (c) 2021-2023 Intel Corporation
+    Copyright (c) 2026 UXL Foundation Contributors
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
@@ -74,6 +75,11 @@ public:
         // We need Write Read memory barrier before notify that reads the waiter list.
         // In C++ only full fence covers this type of barrier.
         my_flag.exchange(false);
+#if !(__TBB_x86_64 || __TBB_x86_32)
+        // RMW on x86 works as a full fence,
+        // but on other platforms we need to issue a seq_cst fence explicitly
+        atomic_fence_seq_cst();
+#endif
         my_flag.notify_one_relaxed();
     }
 
--- a/include/oneapi/tbb/rw_mutex.h
+++ b/include/oneapi/tbb/rw_mutex.h
@@ -1,5 +1,6 @@
 /*
     Copyright (c) 2021 Intel Corporation
+    Copyright (c) 2026 UXL Foundation Contributors
 
     Licensed under the Apache License, Version 2.0 (the "License");
     you may not use this file except in compliance with the License.
@@ -88,11 +89,11 @@ public:
         state_type curr_state = (m_state &= READERS | WRITER_PENDING); // Returns current state
 
         if (curr_state & WRITER_PENDING) {
-            r1::notify_by_address(this, WRITER_CONTEXT);
+            notify(WRITER_CONTEXT);
         } else {
             // It's possible that WRITER sleeps without WRITER_PENDING,
             // because other thread might clear this bit at upgrade()
-            r1::notify_by_address_all(this);
+            notify_all();
         }
     }
 
@@ -116,7 +117,7 @@ public:
         if (!(m_state.load(std::memory_order_relaxed) & has_writer)) {
             if (m_state.fetch_add(ONE_READER) & has_writer) {
                 m_state -= ONE_READER;
-                r1::notify_by_address(this, WRITER_CONTEXT);
+                notify(WRITER_CONTEXT);
             } else {
                 call_itt_notify(acquired, this);
                 return true; // successfully stored increased number of readers
@@ -133,11 +134,11 @@ public:
         state_type curr_state = (m_state -= ONE_READER); // Returns current state
 
         if (curr_state & (WRITER_PENDING)) {
-            r1::notify_by_address(this, WRITER_CONTEXT);
+            notify(WRITER_CONTEXT);
         } else {
             // It's possible that WRITER sleeps without WRITER_PENDING,
             // because other thread might clear this bit at upgrade()
-            r1::notify_by_address_all(this);
+            notify_all();
         }
     }
 
@@ -180,10 +181,32 @@ private:
         m_state += (ONE_READER - WRITER);
 
         if (!(m_state & WRITER_PENDING)) {
-            r1::notify_by_address(this, READER_CONTEXT);
+            notify(READER_CONTEXT);
         }
 
         __TBB_ASSERT(m_state.load(std::memory_order_relaxed) & READERS, "invalid state after downgrade: no readers");
+    }
+
+    using context_type = std::uintptr_t;
+
+    //! Notify waiting threads with the given context
+    void notify(context_type target_context) {
+#if !(__TBB_x86_64 || __TBB_x86_32)
+        // RMW on x86 works as a full fence,
+        // but on other platforms we need to issue a seq_cst fence explicitly
+        atomic_fence_seq_cst();
+#endif
+        r1::notify_by_address(this, target_context);
+    }
+
+    //! Notify all waiting threads
+    void notify_all() {
+#if !(__TBB_x86_64 || __TBB_x86_32)
+        // RMW on x86 works as a full fence,
+        // but on other platforms we need to issue a seq_cst fence explicitly
+        atomic_fence_seq_cst();
+#endif
+        r1::notify_by_address_all(this);
     }
 
     using state_type = std::intptr_t;
@@ -193,7 +216,6 @@ private:
     static constexpr state_type ONE_READER = 4;
     static constexpr state_type BUSY = WRITER | READERS;
 
-    using context_type = std::uintptr_t;
     static constexpr context_type WRITER_CONTEXT = 0;
     static constexpr context_type READER_CONTEXT = 1;
     friend scoped_lock;
--- a/src/tbb/address_waiter.cpp
+++ b/src/tbb/address_waiter.cpp
@@ -72,23 +72,6 @@ void wait_on_address(void* address, d1::delegate_base& predicate, std::uintptr_t
 }
 
 void notify_by_address(void* address, std::uintptr_t target_context) {
-    // The concurrent_monitor protocol requires a full fence between the store that satisfies
-    // a waiter's predicate and the (relaxed) read of the waitset inside notify_*_relaxed():
-    // the waiter side issues one at the end of prepare_wait() ("Prepare wait guarantees
-    // Write Read memory barrier"), and the notifier side must match it, as the non-relaxed
-    // concurrent_monitor::notify_one()/notify_all() do.
-    // The notify_by_address entry points are called by d1::mutex::unlock(),
-    // d1::rw_mutex::unlock(), and waitable_atomic, which only perform an atomic RMW on the
-    // waited address before calling here. An RMW operation is not a two-sided fence: on
-    // architectures with weak memory ordering (e.g. AArch64), the subsequent relaxed
-    // my_waitset.empty() check may read a stale (empty) value before a concurrent waiter's
-    // insertion becomes visible, while that waiter's predicate re-check read the pre-RMW
-    // state. Both threads then observe the old values (store-buffer pattern), the notifier
-    // takes the empty() early exit, and the wakeup is lost: the waiter sleeps forever on an
-    // already-satisfied predicate (e.g. a writer blocked on a free rw_mutex).
-    // Issue the fence here, on the notifier side, to restore the protocol.
-    atomic_fence_seq_cst();
-
     address_waiter& waiter = get_address_waiter(address);
 
     auto predicate = [address, target_context] (address_context ctx) {
@@ -99,9 +82,6 @@ void notify_by_address(void* address, std::uintptr_t target_context) {
 }
 
 void notify_by_address_one(void* address) {
-    // See the comment in notify_by_address about the memory ordering requirements.
-    atomic_fence_seq_cst();
-
     address_waiter& waiter = get_address_waiter(address);
 
     auto predicate = [address] (address_context ctx) {
@@ -112,9 +92,6 @@ void notify_by_address_one(void* address) {
 }
 
 void notify_by_address_all(void* address) {
-    // See the comment in notify_by_address about the memory ordering requirements.
-    atomic_fence_seq_cst();
-
     address_waiter& waiter = get_address_waiter(address);
 
     auto predicate = [address] (address_context ctx) {
--- a/src/tbb/address_waiter.cpp
+++ b/src/tbb/address_waiter.cpp
@@ -72,6 +72,23 @@ void wait_on_address(void* address, d1::delegate_base& predicate, std::uintptr_t
 }
 
 void notify_by_address(void* address, std::uintptr_t target_context) {
+    // The concurrent_monitor protocol requires a full fence between the store that satisfies
+    // a waiter's predicate and the (relaxed) read of the waitset inside notify_*_relaxed():
+    // the waiter side issues one at the end of prepare_wait() ("Prepare wait guarantees
+    // Write Read memory barrier"), and the notifier side must match it, as the non-relaxed
+    // concurrent_monitor::notify_one()/notify_all() do.
+    // The notify_by_address entry points are called by d1::mutex::unlock(),
+    // d1::rw_mutex::unlock(), and waitable_atomic, which only perform an atomic RMW on the
+    // waited address before calling here. An RMW operation is not a two-sided fence: on
+    // architectures with weak memory ordering (e.g. AArch64), the subsequent relaxed
+    // my_waitset.empty() check may read a stale (empty) value before a concurrent waiter's
+    // insertion becomes visible, while that waiter's predicate re-check read the pre-RMW
+    // state. Both threads then observe the old values (store-buffer pattern), the notifier
+    // takes the empty() early exit, and the wakeup is lost: the waiter sleeps forever on an
+    // already-satisfied predicate (e.g. a writer blocked on a free rw_mutex).
+    // Issue the fence here, on the notifier side, to restore the protocol.
+    atomic_fence_seq_cst();
+
     address_waiter& waiter = get_address_waiter(address);
 
     auto predicate = [address, target_context] (address_context ctx) {
@@ -82,6 +99,9 @@ void notify_by_address(void* address, std::uintptr_t target_context) {
 }
 
 void notify_by_address_one(void* address) {
+    // See the comment in notify_by_address about the memory ordering requirements.
+    atomic_fence_seq_cst();
+
     address_waiter& waiter = get_address_waiter(address);
 
     auto predicate = [address] (address_context ctx) {
@@ -92,6 +112,9 @@ void notify_by_address_one(void* address) {
 }
 
 void notify_by_address_all(void* address) {
+    // See the comment in notify_by_address about the memory ordering requirements.
+    atomic_fence_seq_cst();
+
     address_waiter& waiter = get_address_waiter(address);
 
     auto predicate = [address] (address_context ctx) {


DarthSim:
Could someone please take a look at this? This issue is affecting our users right now. @kboyarinov @aleksei-fedotov @isaevil

isaevil:
@DarthSim I am hesitant to put full fence into `notify_*` functions. The first reason is that at least based [on this code](https://github.com/uxlfoundation/oneTBB/blob/master/include/oneapi/tbb/detail/_waitable_atomic.h#L72), the intention for these functions was to have relaxed semantics, and it seems like they are simply not named properly. The second reason is that on x86 RMW atomic operations already feature full fence and additional fence will be an unnecessary overhead.
If I understand correctly, while `ldr` (relaxed load) can slip into `ldxar/stlxr` pair, acquire load `ldar` cannot do that due to sequential consistency guarantees. 
So I wonder if simply adding `acquire` memory order to currently relaxed reads will be enough.

DarthSim:
Hey @isaevil, thanks for checking this out!

As I understand it, the `wait` path doesn't have a problem, since `prepare_wait` already has a [full fence](https://github.com/uxlfoundation/oneTBB/blob/master/src/tbb/concurrent_monitor.h#L228). The problem is in the `notify` path, as it doesn't have a fence on ARM, and the `my_waitset.empty()` check slips between checking the predicate and adding a node to a waitlist. Thus, it seems to me that we can't fix the issue without adding a full fence before the `my_waitset.empty()` check.

> The first reason is that at least based on this code, the intention for these functions was to have relaxed semantics, and it seems like they are simply not named properly

Yeah, I noticed that too. However, this function is currently used only in mutexes, which is where the issue occurs. So _maybe_ this function should not be relaxed.

> The second reason is that on x86 RMW atomic operations already feature full fence and additional fence will be an unnecessary overhead

That's fair, but there are way more things that happen on the `notify` path, so the overhead looks negligible.

isaevil:
> The problem is in the `notify` path, as it doesn't have a fence on ARM, and the `my_waitset.empty()` check slips between checking the predicate and adding a node to a waitlist. Thus, it seems to me that we can't fix the issue without adding a full fence before the `my_waitset.empty()` check.

Yes, that is how I understand that as well. In Linux AArch64 build assembly of TBB mutex RMW operation in unlock for rw_mutex looks something like this:
```asm
.retry:
    ldaxr  x1, [x0] // load-acquire current value
    and  x1, x1, #0xfffffffffffffffe // m_state &= READERS(-4) | WRITER_PENDING(2)
    stlxr  w2, x1, [x0] // try to store new value back
    cbnz  w2,  .retry // retry if failed
```
And in [ARM documentation](https://support.arm.com/documentation/102336/0101/Load-Acquire-and-Store-Release-instructions?lang=en) in "Sequential consistent" section, it is said that relaxed load/stores can be reordered before `stlr`, but non-relaxed cannot (like acquire load `ldar`) and that basically guarantees sequential consistency:

> This means that, when a Load-Acquire appears in program order after a Store-Release, the memory access that is generated by the Store-Release instruction is observed before the memory access that is generated by the Load-Acquire instruction.

So, mapping this on mutex implementation, it seems like basically reading `my_waitset.empty()` with `std::memory_order_acquire` instead of relaxed should achieve the same result on ARM, while not adding extra fences on x86. Unfortunately I don't have Linux ARM system in use to check that and on MacOS system I have unlock uses `ldclral` instead of loop. 
You've said that you were able to reproduce the issue on 10-core Apple M1 as well. Do you have more details on that, the minimal reproducer would be the best?

DarthSim:
> So, mapping this on mutex implementation, it seems like basically reading my_waitset.empty() with std::memory_order_acquire instead of relaxed should achieve the same result on ARM, while not adding extra fences on x86

I'm not sure about that. I mean, this should fix reading a stale `empty()` value, but there are more things happening, and they won't be synchronized. Such as, the waiter can still read a stale predicate value. `mutex::unlock()` states:

> We need Write Read memory barrier before notify that reads the waiter list.
> In C++ only full fence covers this type of barrier.

Yet, there are no fences in notifiers. They are emulated with RMW on x86, but that doesn't work on ARM for independent read/writes. And I believe acquire-read in `empty()` won't work for them either.

> Do you have more details on that, the minimal reproducer would be the best?

That's the worst part. Reproduction is extremely random and complex. I managed to reproduce the issue in a project that uses OpenCV, which uses OpenVINO, which uses TBB. The project runs in a Docker image, so I basically have Linux on M1. Yet it takes hours to days to reproduce the deadlock; the window is very tight. I wrote a small program that simply locks/unlocks a mutex in multiple threads with some delays to skip `timed_spin_wait_until`, but no luck.

The TBB binaries are distributed along with OV, so I don't know how they are built. But the rest is built with GCC 12 on Ubuntu 22.04. I'm not sure whether the build flags matter here.

Considering the complexity of reproducing the issue, I'd play it safe here and restore the concurrent monitor protocol on ARM, adding a full fence. At least, for ARM only.

phprus:
Maybe we need to call `atomic_fence_seq_cst()` in all `*_relaxed` functions of https://github.com/uxlfoundation/oneTBB/blob/master/src/tbb/concurrent_monitor.h on platforms other than `__TBB_x86_64` and `__TBB_x86_32`?

isaevil:
> Maybe we need to call `atomic_fence_seq_cst()` in all `*_relaxed` functions of https://github.com/uxlfoundation/oneTBB/blob/master/src/tbb/concurrent_monitor.h on platforms other than `__TBB_x86_64` and `__TBB_x86_32`?

Doesn't seem necessary. `*_relaxed` functions already have `seq_cst` variants. And address_waiter.cpp `notify` functions seem to have `relaxed` semantics by design.
I would perhaps suggest adding full fence for non-x86 platforms before calling `notify_by_*` functions from address_waiter.cpp in corresponding mutex methods. I think it will give a better control over memory order.

> So, mapping this on mutex implementation, it seems like basically reading my_waitset.empty() with std::memory_order_acquire instead of relaxed should achieve the same result on ARM

Looks like it was incorrect assumption because since armv8.3-a, `ldapr` has been introduced, and it can be reordered with `stlr` and load with `std::memory_order_acquire` compiles into `ldapr` with `-march=armv8.3-a` (https://godbolt.org/z/xn31nM7z7). But overall I am pretty sure `stlr` followed by `ldar` in mutex unlock case will provide the same observability guarantee as with `dmb ish`.

DarthSim:
> Doesn't seem necessary. *_relaxed functions already have seq_cst variants. And address_waiter.cpp notify functions seem to have relaxed semantics by design.

I actually think this is a good bulletproof idea. It seems to me that `*_relaxed` functions are meant to be used in combination with RMW, which works as a full barrier on x86. But as we found out, this doesn't work for non-x86. So here's what I'd do:

1. Add `atomic_fence_seq_cst` to `*_relaxed` for non-x86 with a comment, explaining the situation.
2. Keep `atomic_fence_seq_cst` in  non-`*_relaxed` funcs only for x86 to not add redundant fence on non-x86

Given we already have two problematic cases, I won't be surprised if there are more. Fixing the issue at the concurrent monitor level rather than in every special case looks like a more stable solution to me.

isaevil:
> Add atomic_fence_seq_cst to *_relaxed for non-x86 with a comment, explaining the situation.
Keep atomic_fence_seq_cst in non-*_relaxed funcs only for x86 to not add redundant fence on non-x86

To be honest, this sounds confusing to me and maybe even error-prone in future. Now relaxed operations suddenly become sequentially consistent depending on architecture.

>  It seems to me that *_relaxed functions are meant to be used in combination with RMW.

I would say that memory order is intended to be controlled outside of those functions and `notify_all/notify_one/abort_all` do exactly that. I prefer having clear semantics for these functions. And for `address_waiter.cpp:notify_by_address` invocations in mutex methods, let's just wrap them into a function that actually controls ordering (relaxed on x86, full fence on others). What do you think?

DarthSim:
> To be honest, this sounds confusing to me and maybe even error-prone in future. Now relaxed operations suddenly become sequentially consistent depending on architecture.

IDK about this. The cost of an error with `atomic_fence_seq_cst` in `*_relaxed` is a redundant fence. The cost of an error without it is another lost wakeup or something like that.

Anyway, I moved fences from the address waiter to mutexes.

include/oneapi/tbb/rw_mutex.h
```diff
@@ -180,12 +180,32 @@ class rw_mutex {
...
         if (!(m_state & WRITER_PENDING)) {
-            r1::notify_by_address(this, READER_CONTEXT);
+            notify(READER_CONTEXT);
         }
 
         __TBB_ASSERT(m_state.load(std::memory_order_relaxed) & READERS, "invalid state after downgrade: no readers");
     }
 
+    //! Notify waiting threads with the given context
+    void notify(std::uintptr_t target_context) {
```
isaevil:
```suggestion
    void notify(state_type target_context) {
```

include/oneapi/tbb/mutex.h
isaevil, line 1: Please add UXL Foundation Contributors copyright after the Intel one.

include/oneapi/tbb/rw_mutex.h
isaevil, line 1: Same is here.

isaevil (commented):
Overall looks good. A few comments/suggestions though

include/oneapi/tbb/rw_mutex.h
DarthSim: I believe you mean `context_type`. Either way, this causes an "unknown type name" error. I'm not sure whether it's better to move the `context_type` definition up or to keep `std::uintptr_t` here.

include/oneapi/tbb/mutex.h
DarthSim: Done

include/oneapi/tbb/rw_mutex.h
DarthSim: Done

include/oneapi/tbb/rw_mutex.h
isaevil: You are right, sorry. I think you can move `context_type` definition before `notify` functions.

DarthSim: Done

include/oneapi/tbb/mutex.h
DarthSim: The copyright check still fails, not sure why

isaevil: It reports files that were already updated in `master`. So it is not a big deal

isaevil (approved):
LGTM, but let's see what others think

aleksei-fedotov:
> it seems like basically reading my_waitset.empty() with std::memory_order_acquire instead of relaxed should achieve the same result on ARM

Read the problem statement and analysis independently from @isaevil and came to same conclusion above with one addition that acquired load works only in pair with released store on same memory. If one is missing or different memory is used, the synchronization won't work. Therefore, loading `count` in `my_waitset::empty()` with acquire is only half of a deal. The other half I think is updating `count` with release during `my_waitset::add()` along with moving this write to the end of `add` function, that is after insertion of a node into the list. If the issue is only with slipped `my_waitset.empty()`, this should be enough for every platform that implements C++ standard.

> I'm not sure about that. I mean, this should fix reading a stale empty() value, but there are more things happening, and they won't be synchronized. Such as, the waiter can still read a stale predicate value.

If `waitset.empty()` reads `false` with acquired semantics (given that its write was with a release one), the notification path will use sequential consistent memory model on the same atomic inside semaphore, which should guarantee the predicate returns correct value since the flag is updated before notification path happens. So in theory, this should be enough to fix the issue.

> the minimal reproducer would be the best

We definitely need a reproducer for that. @DarthSim have you tried running one of the related tests from TBB's test suite such as test_mutex and test_concurrent_monitor on the platforms where the issue can be reproduced? @isaevil test_concurrent_monitor.cpp needs to be extended since it does not use relaxed notification operations.

DarthSim:
Sorry, I don't get it. Your code clearly states:

```
// We need Write Read memory barrier before notify that reads the waiter list.
// In C++ only full fence covers this type of barrier.
```

And that's absolutely correct. So why are you so opposed to putting a full fence where it belongs?

> If waitset.empty() reads false with acquired semantics (given that its write was with a release one), the notification path will use sequential consistent memory model on the same atomic inside semaphore, which should guarantee the predicate returns correct value since the flag is updated before notification path happens. So in theory, this should be enough to fix the issue.

As I understand it, this is not correct. I mean, this won't fix the issue. First, memory ordering in operations with atomics affects only those atomics. Acq/rel read/write on the list's `count` won't affect the predicate value. Second, this is a store-buffering race. Acq/rel read/write doesn't solve it; only seq_cst on both sides does. That's why we need a full fence on the notifier's side. A seq_cst fence only on the waiter side is basically meaningless.

> this should be enough for every platform that implements C++ standard

The standard states quite the opposite. Acq/rel only constrains what a thread sees after it reads the released value; it never _forces_ the read to see it. The lost wakeup is the store-buffering outcome – both sides reading stale – which [atomics.order](https://eel.is/c++draft/atomics.order) (4) forbids only via the seq_cst rules, and none of them apply to an acquire load with no fence before it (4.2 requires the load itself to be seq_cst, 4.3/4.4 require a fence on the notifier side). And (7) confirms that the bad outcome is allowed under the acquire scheme and is unreachable with the paired fences.

> We definitely need a reproducer for that. @DarthSim have you tried running one of the related tests from TBB's test suite such as test_mutex and test_concurrent_monitor on the platforms where the issue can be reproduced?

That's the catch. The error window is just a few instructions. The tests work fine (you also run them on ARM, as far as I can see). I even tried to write a small program that aims to hit that window; no luck. The only way I could reproduce this is with a real-world app that uses TBB, and I was lucky if I could do this in hours, 'cause it can take days. I could arrange access to that app for you, but I don't think that's what you're looking for. I guess you want something you can wrap into a test that reproduces the issue relatively quickly. But considering how tight the window is, I don't think this is really possible. Also, such tests are not reliable. Seeing them green may mean that everything is okay, but it can also mean that you're just (un)lucky today.

DarthSim:
I managed to build a reproduction: https://gist.github.com/DarthSim/292b2d2f739d5e5b8e4c6c8292d3c3ed. You need to build it with `-march=armv8-a -mno-outline-atomics` to make it work. It catches a deadlock in a second.

Acq/rel read/write of `count` seems to fix the issue. However:

1. It looks to me like it narrows the error window instead of eliminating the issue.
2. It looks like fixing a problem in the wrong place. `*_relaxed` methods become not-so-relaxed.
3. It introduces redundant ordering on x86, where RMW works as a full fence.

Adding seq_cst where the protocol requires it still seems like a more reliable solution to me.

aleksei-fedotov:
> So why are you so opposed to putting a full fence where it belongs?

I am not opposed to this, I wanted to find out if existing or even fewer number of operations affecting memory ordering can be used to enforce the desired behavior.

> First, memory ordering in operations with atomics affects only those atomics.

They affect surrounding memory accesses as well.

> Acq/rel read/write doesn't solve it; only seq_cst on both sides does.

That's why I was betting on the sequentially consisted exchanges happening around, hoping that they can help with the situation without introducing new fences.

> Also, such tests are not reliable. Seeing them green may mean that everything is okay, but it can also mean that you're just (un)lucky today.

That's true. However, it is also impossible that they stay green forever. Otherwise, we would not call this a problem.

Anyway, the topic is really a complex one and we can spend hours if not days figuring out on the details. I think the current state of a patch is good enough.

Thank you for the reproducer. I will consider it as a starting point in future analysis.

darknesswind:
may be [cppmem](http://svr-pes20-cppmem.cl.cam.ac.uk/cppmem/help.html) is helpful ?
buggy: 8 executions; 1 consistent, race free
```C
int main() {
  atomic_int flag=1; atomic_int inlist=0;
  {{{ { inlist.store(1,memory_order_relaxed);
        atomic_thread_fence(memory_order_seq_cst);
        r0=flag.load(memory_order_relaxed).readsvalue(1); }
  ||| { flag.store(0,memory_order_seq_cst);
        r1=inlist.load(memory_order_relaxed).readsvalue(0); }  }}}
  return 0;
}
```
fixed: 24 executions; no consistent
```C
int main() {
  atomic_int flag=1; atomic_int inlist=0;
  {{{ { inlist.store(1,memory_order_relaxed);
        atomic_thread_fence(memory_order_seq_cst);
        r0=flag.load(memory_order_relaxed).readsvalue(1); }
  ||| { flag.store(0,memory_order_seq_cst);
        atomic_thread_fence(memory_order_seq_cst); // add fence here
        r1=inlist.load(memory_order_relaxed).readsvalue(0); }  }}}
  return 0;
}
```

DarthSim:
Thanks for merging this! Could you also backport this to `onetbb_2021.13_ov`, please? I really hope OpenVINO will upgrade TBB in their distribution one day 😅

PR #2137, opened 2026-07-28
Introduce new constructors for flow::resource_limiter

### LINKED ISSUE #2131: `resource_limiter` constructor that accepts runtime-defined container of handles
# Summary

The recently added preview of `tbb::flow::resource_limiter` is excellent (!) and addresses a need we've had for some time.  To make it more user-friendly, we request a constructor of `resource_limiter` that can accept a runtime-defined container of handles.

# Problem Statement

It is not necessarily known at compile-time how many handles a given `resource_limiter` will be able to use (database instances may need to go offline between program executions, binary distributions installed on different platforms may need to support different resource limits, etc.).  The current `resource_limiter` constructor requires a compile-time-specified list of handles (e.g.):

```cpp
    template <typename Handle, typename... Handles>
    resource_limiter(Handle&& handle, Handles&&... handles) {
        emplace_handles(std::forward<Handle>(handle), std::forward<Handles>(handles)...);
    }
```

making a runtime-dependent construction of `resource_limiter` impractical.  Given that the underlying container of handles has a runtime-dependent size (currently, a `std::forward_list<ResourceHandle>`), we hope it will be possible to introduce an interface that allows users to specify something similar.

# Preferred Solution

We do not necessarily have a preferred solution.  A simple solution could be:

```cpp
    resource_limiter(std::forward_list<ResourceHandle> handles) { ... }
```

but that risks exposing the implementation to the user (i.e., using `std::forward_list<...>`), and a check would need to be made that the `handles` container is non-empty.  Another option could be:

```cpp
    template <typename Handle>
    resource_limiter(Handle handle, std::span<Handle> additional_handles) { ... }
```

where at least one handle is required, but additional handles are added via the C++20 construct `std::span`.       Instead of `std::span`, a pre-C++20 construct like `std::vector` could also be considered...or perhaps a dedicated oneTBB type that insulates the user from the underlying type.

kboyarinov:
### Description 
Change a set of constructors for `tbb::flow::resource_limiter` class:

```cpp
template <typename ResourceHandle>
class resource_limiter {
public:
    template <typename InputIterator>
    resource_limiter(InputIterator first, InputIterator last);
    
    template <typename ContainerBasedSequence>
    resource_limiter(ContainerBasedSequence&& sequence);
    
    resource_limiter(std::initializer_list<ResourceHandle> init);
    
    template <typename Tuple, typename... Tuples>
    resource_limiter(std::piecewise_construct_t, Tuple&& tuple, Tuples&&... tuples);
};
```

Updated the RFC with the motivation

Fixes #2131 

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown

### DIFF
--- a/rfcs/proposed/flow_graph_serializers/resource_consumer_node_api.md
+++ b/rfcs/proposed/flow_graph_serializers/resource_consumer_node_api.md
@@ -164,7 +164,7 @@ The concern with this constructor is that the number of resources should be know
 The current implementation of ``resource_limiter`` uses ``std::forward_list`` container which already provides the necessary runtime semantics, so the
 only concern is the ``resource_limiter`` constructor itself.
 
-The idea is to extend the set of constructors to accept arbitrary containers (or ranges) to define the number of resources and its values in runtime.
+The idea is to extend the set of constructors to accept arbitrary containers (or ranges) to define the number of resources and their values at runtime.
 
 ```cpp
 template <typename ResourceHandle>
--- a/rfcs/proposed/flow_graph_serializers/resource_consumer_node_api.md
+++ b/rfcs/proposed/flow_graph_serializers/resource_consumer_node_api.md
@@ -8,10 +8,11 @@ Note: This document is a sub-RFC for the [Resource-limited Nodes RFC](./README.m
   * 1.1 [Terminology](#terminology)
   * 1.2 [Proposed Design](#proposed-design)
   * 1.3 [API Details](#api-details)
-  * 1.4 [API Specification](#api-specification)
-    * 1.4.1 [`oneapi::tbb::flow::resource_limiter` Class](#oneapitbbflowresource_limiter-class)
-    * 1.4.2 [`ResourceLimitedBody` Named Requirements](#resourcelimitedbody-named-requirements)
-    * 1.4.3 [`oneapi::tbb::flow::resource_limited_node` Class](#oneapitbbflowresource_limited_node-class)
+  * 1.4 [Extending the ``resource_limiter`` Constructors](#extending-the-resource_limiter-constructors)
+  * 1.5 [API Specification](#api-specification)
+    * 1.5.1 [`oneapi::tbb::flow::resource_limiter` Class](#oneapitbbflowresource_limiter-class)
+    * 1.5.2 [`ResourceLimitedBody` Named Requirements](#resourcelimitedbody-named-requirements)
+    * 1.5.3 [`oneapi::tbb::flow::resource_limited_node` Class](#oneapitbbflowresource_limited_node-class)
 * 2 [API to Support Polymorphic Providers](#api-to-support-polymorphic-providers)
 * 3 [Exit Criteria and Open Questions](#exit-criteria-and-open-questions)
 * 4 [Usage Examples](#usage-examples)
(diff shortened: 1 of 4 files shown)


rfcs/proposed/flow_graph_serializers/resource_consumer_node_api.md
```diff
@@ -140,6 +141,200 @@ If access to one or several resources cannot be granted immediately, the node an
...
+    {}
+};
+
+resource_limiter<int> limiter(1, 2, 3, 4, 5);
+```
+
+In the code above, ``limiter`` manages 5 resources of type ``int``. Instances of resources are constructed in-place while constructing the limiter.
+The concern with this constructor is that the number of resources should be known at compile-time, while some workflows only knows it in runtime.
+
+Current underneath implementation of ``resource_limiter`` uses ``std::forward_list`` container which already provides a necessary runtime semantics, so the
```
vossmjp:
```suggestion
The current implementation of ``resource_limiter`` uses ``std::forward_list`` container which already provides the necessary runtime semantics, so the
```

rfcs/proposed/flow_graph_serializers/resource_consumer_node_api.md
```diff
@@ -140,6 +141,200 @@ If access to one or several resources cannot be granted immediately, the node an
...
+only concern is the ``resource_limiter`` constructor itself.
+
+The idea is to extend the set of constructors to accept arbitrary containers (or ranges) to define the number of resources and its values in runtime.
```
vossmjp:
```suggestion
The idea is to extend the set of constructors to accept arbitrary containers (or ranges) to define the number of resources and their values at runtime.
```

vossmjp (commented):
You need to fix test_flow_graph_whitebox.cpp:648, which I think you can do by just changing it to curly braces, "limiter{1};"

rfcs/proposed/flow_graph_serializers/resource_consumer_node_api.md
```diff
@@ -140,6 +141,200 @@ If access to one or several resources cannot be granted immediately, the node an
...
+    template <typename Handle, typename... Handles>
+    resource_limiter(Handle&& handle, Handles&&... handles) // Constructor (1) - existing
+    {}
+
+    template <typename InputIterator>
+    resource_limiter(InputIterator first, InputIterator last) // Constructor (2) - extension
+    {}
+
+    template <typename ContainerBasedSequence>
+    resource_limiter(ContainerBasedSequence&& sequence) // Constructor (3) - extension
```
vossmjp, line 182: Should we force failure on construction with an empty container / sequence? Since the resource_limited_node uses reserve_wait / release_wait pairs (in the current implementation), an empty container would lead to a call to graph::wait_for_all() being stuck. We can either document that such a thing is dangerous, assert or something else.

kboyarinov: Yes, but the only available way for the current API is to force it in runtime by adding an assert, exception or something like that.

Another option is to have 1 resource handle explicitly and others as a sequence, but I would prefer the existing API with an assert in debug/UB in release.

I have added the paragraph into the RFC about that.

test/tbb/test_resource_limited_node.cpp
```diff
@@ -379,6 +382,74 @@ void test_cancellation_with_active_requests(bool same_graph, bool exception) {
...
+    auto node_body = [&](int, ports_type&, resource_type resource) {
+        auto is_equal = [=](const resource_type& value) { return value == resource; };
+        CHECK_MESSAGE(std::any_of(resource_values.begin(), resource_values.end(), is_equal),
+                      "Unexpected resource");
+                      
+        ++counter;
+        for (std::size_t i = 0; i < 1000; ++i) {
+            CHECK_MESSAGE(counter <= resource_values.size(), "Detected more resources than expected");
+        }
+        --counter;
```
vossmjp, line 407: Can this test pass if the node always received the same resource and there was no concurrency exploited at all?  I think so. If I'm reading this right, this tests that the resource node_body gets one of the resources and the counter/concurrency never exceeds to number of resource values. But it doesn't test that there is any concurrency > 1 or that we see all or even more than one of the resource values. It does test that we do not exceed the concurrency limit set by the number of possible resources. Is that sufficient for this test?

test/tbb/test_resource_limited_node.cpp
```diff
@@ -176,7 +178,8 @@ void test_strict_resource_handle() {
     using namespace tbb::flow;
 
     int handle_value = 42;
-    resource_limiter<strict_resource_handle> limiter{strict_resource_handle_provider::construct(handle_value)};
+    resource_limiter<strict_resource_handle> limiter(std::piecewise_construct,
```
vossmjp, line 181: Do you test anywhere construction with tuples that have more than 1 element?

test/tbb/test_resource_limited_node.cpp
```diff
@@ -379,6 +382,74 @@ void test_cancellation_with_active_requests(bool same_graph, bool exception) {
...
+    g.wait_for_all();
+    CHECK(counter == 0);
+}
+
+template <std::size_t... Idx, typename ArrayType>
+void test_resource_limiter_handles_constructor(ArrayType& resources,
+                                               tbb::detail::index_sequence<Idx...>) {
+    CHECK(sizeof...(Idx) == resources.size());
+    test_resource_limiter_constructor(/*resource_values = */resources,
+                                      /*args = */std::piecewise_construct, std::forward_as_tuple(resources[Idx])...);
```
vossmjp, line 424: Another case of just 1 element per tuple.

vossmjp (approved):
I've previously commented about improved testing, but these can be addressed post feature freeze if needed.  Also, at some point `TBB_HAS_FLOW_GRAPH_RESOURCE_LIMITING` need to be updated but that can happen with this change or the starvation change.

kboyarinov:
> I've previously commented about improved testing, but these can be addressed post feature freeze if needed. Also, at some point `TBB_HAS_FLOW_GRAPH_RESOURCE_LIMITING` need to be updated but that can happen with this change or the starvation change.

I will address the remaining comments for tests in a separate PR.

PR #2147, opened 2026-08-03
Explicitely prohibit resuming after calling of assertion_handler

Alexandr-Konovalov:
### Description 
Motivation for the change is that we can't guarantee anything after assertion triggering, so let's say this explicitly. It's also matches the expectation for global terminate handler function.

### Type of change

### Tests

- [X] not needed

- [ ] Yes
- [ ] No
- [X] Unknown


aleksei-fedotov (approved):
Looks good to me.

kboyarinov (approved):

PR #2148, opened 2026-08-03
Added RFC about error processing during interleaved allocation

Alexandr-Konovalov:
### Type of change

### Tests

- [X] not needed

### Documentation

- [ ] needs to be updated
- [X] not needed

- [ ] Yes
- [X] No
- [ ] Unknown

### DIFF
--- a/rfcs/proposed/numa_support/interleaved-allocation-errors.md
+++ b/rfcs/proposed/numa_support/interleaved-allocation-errors.md
@@ -11,9 +11,9 @@ mappings. Therefore, returning an error is a valid error-handling policy.
 
 For deallocation, there is no reasonable fallback, and at least passing an invalid address to
 `munmap()` is serious enough to justify terminating the application. This termination behavior
-matches `free()`. The termination must be processed in a way customizable via a custom assertion
-handler. Advantage over `std::terminate_handler` is the possibility to report parameters of a failed
-function.
+matches `free()`. Instead of just using `std::terminate_handler`, termination must be processed in a
+way customizable via a custom assertion handler. Customization allows the possibility to report
+the parameters of the failed function.
 
 ## Error returning policy should match existing practices
 
--- a/rfcs/proposed/numa_support/interleaved-allocation-errors.md
+++ b/rfcs/proposed/numa_support/interleaved-allocation-errors.md
@@ -1,0 +1,76 @@
+# Allocate memory interleaved between NUMA nodes: error handling
+ 
+*Note:* This document is a sub-RFC of the [umbrella RFC about improving NUMA
+support](README.md). 
+
+## Returning vs termination
+
+For the allocation function, there is a reasonable fallback path: a `malloc()` call, and there are
+reasonable causes of run-time failure, for example hitting a quota on the maximum number of
+mappings. Therefore, returning an error is a valid error-handling policy.
+
+For deallocation, there is no reasonable fallback, and at least passing an invalid address to
+`munmap()` is serious enough to justify terminating the application. This termination behavior
+matches `free()`. The termination must be processed in a way customizable via custom assertion
+handler. Advatage over `std::terminate_handler` is possibility to report parameters of a failed
+function.
+
+## Error returning policy should match existing practices
+
+The common error-handling practice for `operator new` is to throw exceptions or, when explicitly
+requested, return `nullptr`. Combining that with the high-level principle that requirements of
+exotic customers should not worsen quality of life for ordinary customers, we get a function family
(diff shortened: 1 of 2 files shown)


rfcs/proposed/numa_support/interleaved-allocation-errors.md
```diff
@@ -0,0 +1,76 @@
...
+## Which exception?
+
+`std::bad_alloc` is a natural choice for allocation function, but it can’t provide information about
+the exact reason for failure, which is useful for analysis. `std::invalid_argument` matches cases
+where invalid arguments are passed to allocation functions. `std::system_error` matches cases
+where an OS call returns an error. Both can carry a message that describes the failure, and the
+latter also has room for error-code information such as errno.
+
+It is unclear how error codes would be used beyond writing logs, so enum values have no clear
+advantage over strings for indicating what went wrong.
```
aleksei-fedotov, line 35: I can think of a system that can do something useful based on the returned code. For example, collect garbage before attempting to allocate memory again. Such behavior can be implemented based on the returned error code indicating that the allocation did not happen because there is not enough memory. For such handlers comparing error codes is much simpler rather than parsing and comparing returned strings.

aleksei-fedotov (commented):
I think for such low-, system-level functions the approach with standard `errno` is sort of more conventional. Unfortunately, the language does not seem to allow extending it with own values. Nevertheless, may be existing values are sufficient for our use cases.

Another line of thought is that we may do nothing special about the error in these allocation functions and just return `nullptr`, relying on the functions we use internally to set reasonable `errno`. I guess we should check their guarantees in this regard and document that.

For the throwing use cases (if we ever decide to add support for them), I think we need to establish our own exceptions or extend standard ones somehow to include at least the `errno` value, since this variable is thread-specific.

rfcs/proposed/numa_support/interleaved-allocation-errors.md
```diff
@@ -0,0 +1,76 @@
...
+
+For the allocation function, there is a reasonable fallback path: a `malloc()` call, and there are
+reasonable causes of run-time failure, for example hitting a quota on the maximum number of
+mappings. Therefore, returning an error is a valid error-handling policy.
+
+For deallocation, there is no reasonable fallback, and at least passing an invalid address to
+`munmap()` is serious enough to justify terminating the application. This termination behavior
+matches `free()`. The termination must be processed in a way customizable via a custom assertion
+handler. Advantage over `std::terminate_handler` is the possibility to report parameters of a failed
+function.
```
vossmjp:
```suggestion
matches `free()`. Instead of just using `std::terminate_handler`, termination must be processed in a
way customizable via a custom assertion handler. Customization allows the possibility to report
the parameters of the failed function.
```

rfcs/proposed/numa_support/interleaved-allocation-errors.md
Alexandr-Konovalov: Good point, clearing buffers/caches and retry is valid scenario. Technically, there are 2 kinds of errors: invalid arguments and errors from syscalls. And retrying for 1st one is pointless. So we get along with returning an enum with error type (matched type of the exception) and a string with description.

To me, that's a little like overengineering, because check for invalid arguments is fast.

Alexandr-Konovalov:
> I think for such low-, system-level functions the approach with standard `errno` is sort of more conventional. Unfortunately, the language does not seem to allow extending it with own values. Nevertheless, may be existing values are sufficient for our use cases.
> 
> Another line of thought is that we may do nothing special about the error in these allocation functions and just return `nullptr`, relying on the functions we use internally to set reasonable `errno`. I guess we should check their guarantees in this regard and document that.
> 

Sure, errno can be good fully functional error reporting logic with several advantages over other approaches (say, low overhead, as it’s hard to beat access to a TLS variable), just because it was invented by very smart people. Main problem with it is that user of modern C++ expect exceptions are the error reporting technique. And a technical problem is that we have several syscalls with overlapping errno values, so understanding what exactly failed may not be trivial.

> For the throwing use cases (if we ever decide to add support for them), I think we need to establish our own exceptions or extend standard ones somehow to include at least the `errno` value, since this variable is thread-specific.

I hope `std::error_code` used in `std::system_error` ctor can serve for this.

Alexandr-Konovalov:
Wrt
> throw generates a lot of code, so it should not be used in the functions inlined to user code without strong justification.

there are the tiny [code snippet](https://godbolt.org/z/crKff9vsE) and [the reference](https://github.com/uxlfoundation/oneTBB/blob/3cdc6f6558ba23ec9ceed92078b49dc664ed5bf3/src/tbb/exception.cpp#L101) in the TBB codes.

PR #2149, opened 2026-08-03
Fix no member named 'hardware_destructive_interference_size'

NsPro04:
### Description 
The definition of `std::hardware_destructive_interference_size` is [found](https://en.cppreference.com/cpp/thread/hardware_destructive_interference_size) in the \<**new**\> header.

In some implementations of the standard library, it is included in the \<**vector**\> header.
For example, in the case of the MSVC STL, the \<**new**\> header is included as follows:
"**detail/_utils.h**" -\> \<**vector**\> -\> \<**xmemory**\> -\> \<**new**\>

However, this does not happen in libcxx, which is why the following error occurs:
image

The solution is simple - explicitly add `#include <new>`
### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### Other information
This might also resolve issue #2024, but I'm not sure

### DIFF
--- a/include/oneapi/tbb/cache_aligned_allocator.h
+++ b/include/oneapi/tbb/cache_aligned_allocator.h
@@ -1,4 +1,4 @@
-/*
+﻿/*
     Copyright (c) 2005-2022 Intel Corporation
     Copyright (c) 2026 UXL Foundation Contributors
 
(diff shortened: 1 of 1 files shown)


isaevil (approved):
LGTM

PR #2150, opened 2026-08-03
Add starvation avoidance based on priorities

vossmjp:
### Description 
Adds a starvation avoidance mechanism to the resource limiting preview feature.

### Type of change

### Tests

- [ X] added - _required for new features and some bug fixes_
- [ ] not needed

### Documentation

- [X] updated in this PR
- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown


include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
kboyarinov: It seems like `std::equal_to<request_id>` can be used directly instead of this structure since `operator==` was introduced.

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
```diff
@@ -157,48 +196,160 @@ class resource_limiter : public resource_provider_base<ResourceHandle> {
...
+        std::size_t num_notified = m_notified.size();
+
+        const std::size_t& pressure = increment_ref_count_for_consumer(consumer);
+
+        if (num_handles == 0) {
+            m_pending.push_back(consumer_data(id, &consumer, pressure));
+            return;
+        }
+
+        consumer_data new_consumer_data(id, &consumer, pressure);
```
kboyarinov:
```suggestion
        consumer_data new_consumer_data(id, &consumer, pressure);
        if (num_handles == 0) {
            m_pending.push_back(new_consumer_data);
            return;
        }

```

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
```diff
@@ -207,9 +358,133 @@ class resource_limiter : public resource_provider_base<ResourceHandle> {
...
+    void add_to_notified_list(const consumer_data& consumer_dt) {
+        m_notified.push_back(consumer_dt);
+    }
+
+    using pressure_map_type = std::map<resource_consumer_base<ResourceHandle>*,
+                                       std::pair<std::size_t /*pressure*/,
+                                                 std::size_t /*reference count*/>>;
+
+    tbb::spin_mutex            m_mutex;
+    const std::size_t          m_no_pressure = 0;
```
kboyarinov:
```suggestion
    static constexpr std::size_t          no_pressure = 0;
```

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
```diff
@@ -207,9 +358,133 @@ class resource_limiter : public resource_provider_base<ResourceHandle> {
...
+    }
+
+    // Called under the lock
+    void add_to_notified_list(const consumer_data& consumer_dt) {
+        m_notified.push_back(consumer_dt);
+    }
+
+    using pressure_map_type = std::map<resource_consumer_base<ResourceHandle>*,
+                                       std::pair<std::size_t /*pressure*/,
+                                                 std::size_t /*reference count*/>>;
```
kboyarinov: Is there a specific need to sort the consumers by address in the pressure map?
As far as I understand, it can be easily replaced by `unordered_map` and provide cheaper lookup.

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
```diff
@@ -207,9 +358,133 @@ class resource_limiter : public resource_provider_base<ResourceHandle> {
...
+            // rather than partition, since the threshold moves as requests are notified.
+            std::sort(m_notified.begin(), m_notified.end(), higher_priority_first());
+
+            while (it != m_pending.end() && !(*it < m_notified[num_handles - 1])) {
+                to_notify.push_back(std::make_pair(it->consumer_ptr, it->id));
+                // Keep m_notified sorted so that the threshold stays valid for the next
+                // iteration, which the newly notified request may itself have displaced.
+                m_notified.insert(std::upper_bound(m_notified.begin(), m_notified.end(),
+                                                   *it, higher_priority_first()),
+                                  *it);
```
kboyarinov: IMHO, it is easier to read

```suggestion
                auto position = std::upper_bound(m_notified.begin(), m_notified.end(), *it, higher_priority_first());
                m_notified.insert(position, *it);
```

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
```diff
@@ -207,9 +358,133 @@ class resource_limiter : public resource_provider_base<ResourceHandle> {
...
+        for (auto& notification : to_notify) {
+            notification.first->notify(*this, notification.second);
+        }
+    }
+
+    // Called under the lock with 0 < num_handles <= m_notified.size(). Partitions m_notified
+    // just enough for m_notified[num_handles - 1] to hold the lowest priority request among
+    // the num_handles highest priority ones, i.e. the threshold a request has to meet to be
+    // served while the notified requests outnumber the handles.
+    void ensure_priorities_current(std::size_t num_handles) {
```
kboyarinov: Probably, this function may be renamed to better describe it's purpose.
I am thinking of something like `select_nth_priority_to_notify` (and return the nth element from the function).

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
```diff
@@ -157,48 +196,160 @@ class resource_limiter : public resource_provider_base<ResourceHandle> {
...
+        if (it == m_consumer_pressure.end()) {
+            // No outstanding requests from this consumer, so there is nothing whose
+            // priority the new pressure could affect.
+            return;
         }
+        it->second.first = pressure; // only update the pressure, keep the reference count
+
+        // Notify on reported pressure so that a change in priority is reflected in the
+        // notified requests as soon as possible.
+        notify_pending(lock); // the lock may be released inside call, or it may be released after this call returns
```
kboyarinov: Is my understanding correct that the purpose of this notification is actually to return the notification back to the current caller as soon as possible since it's pressure may grow?

vossmjp: It could be the caller or another node. The pressure could go in either direction, up or down. If it goes down, this might mean that other nodes in the pending list could pass the caller and now be notifiable. Or it could mean that the caller might be bumped up and become notifiable as it becomes important enough.

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
vossmjp: Fixed

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
vossmjp: Addressed and simplified a bit further with the newly renamed `select_lowest_priority_to_serve`

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
vossmjp: Made it a global to match other special value variables in _flow_graph_impl.h

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
vossmjp: Nope. Made the change.

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
vossmjp: Done

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
vossmjp: Renamed

doc/main/reference/feature_test_macros.rst
```diff
@@ -50,7 +50,7 @@ are extended. The table below contains only the most recent values.
           -    Header(s)
         * -    :ref:`Resource Limiting in the Flow Graph<fg_resource_limiting>`
           -    ``TBB_HAS_FLOW_GRAPH_RESOURCE_LIMITING``
-          -    ``202603``
+          -    ``202608``
```
kboyarinov, line 53: I also thought about documenting the whole history of feature test macros and the corresponding changes.
This is more of a general question about documenting the feature-test macros and does not directly block this PR.

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
```diff
@@ -157,48 +174,124 @@ class resource_limiter : public resource_provider_base<ResourceHandle> {
...
-            m_resource_handles.pop_front();
-            return {typename optional_type::in_place_t{}, std::move(handle)};
         }
+
+        if (m_notified.size() <= num_handles) {
+            // There are at least as many resources available as notified requests, so every
+            // notified request can be served and there is no need to compare priorities.
+            if (remove_from_notified_list(acquisition_data)) {
+                return extract_handle();
+            }
```
kboyarinov: How can it happen on acquire that the consumer is not in the notification list? 
As I see, the request is only removed from the list while acquiring and while withdrawing the resource.
So, if the request is not in the list means either:
* acquire was already called for this request and the access was granted (request consumed, no further actions should be done), or denied (request() should be called again).
* The request was withdrawn on task cancellation. Why should the consumer call acquire after withdrawing the request?

Should not it be prohibited by the protocol?

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
```diff
@@ -157,48 +174,124 @@ class resource_limiter : public resource_provider_base<ResourceHandle> {
...
+        tbb::spin_mutex::scoped_lock lock(m_mutex);
+
+        consumer_data withdrawn(id, &consumer);
+        bool found = remove_from_notified_list(withdrawn)
+                     || remove_from_list(m_pending, withdrawn);
+
+        if (found) {
+            // A withdrawn notification frees up room under the notification limit, so a
+            // pending request may now be able to take its place.
+            notify_pending(lock); // the lock may be released
```
kboyarinov: Should we notify if we have removed from the pending list?

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
```diff
@@ -506,15 +741,16 @@ class resource_limited_body_leaf
...
+    }
 
     void release_concurrency_and_spawn_next(const Input& input_msg) {
-        __TBB_ASSERT(m_input_ptr, "m_input_ptr shouldn't be 0 when releasing concurrency slot");
-        // Call back to the input layer to release concurrency slot and get next task
-        // Only do this if concurrency is limited (not unlimited)
-        graph_task* next_task = m_input_ptr->release_concurrency_slot(input_msg);
-        if (next_task && next_task != SUCCESSFULLY_ENQUEUED) {
-            spawn_in_graph_arena(this->graph_reference(), *next_task);
+        if (m_input_ptr) {
```
kboyarinov: Why can this function now be called when m_input_ptr unset?

doc/main/reference/feature_test_macros.rst
vossmjp: There should also be some documentation in the API/ABI table if the increase adds features -- like the additional constructors.

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
vossmjp: It shouldn't happen. Its really just catching a misuse if someone tries to use outside of the resource_limiter. We could assert that is must be present in the list when acquire is called.

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
vossmjp: You're right, we only need to notify if it is removed from the notified list. I'll fix this.

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
vossmjp: It shouldn't be. Could be an assert.

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
vossmjp: Rewrote this function to assert at the top and simplify the conditions that follow.

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
vossmjp: Fixed

include/oneapi/tbb/detail/_flow_graph_resource_limiting.h
vossmjp: Changed to an assert

kboyarinov (dismissed):
LGTM

kboyarinov (approved):

PR #2157, opened 2026-08-11
Move leave_policy/parallel_phase functionality out of preview

isaevil:
### Description 
Update implementation of leave_policy/parallel_phase API with accordance to recent RFC changes.

### Type of change

### Tests

- [ ] not needed

### Documentation

- [x] needs to be updated
- [ ] not needed

- [ ] Yes
- [x] No
- [ ] Unknown


include/oneapi/tbb/detail/_config.h
```diff
@@ -573,9 +569,7 @@
 #define TBB_HAS_FLOW_GRAPH_RESOURCE_LIMITING 202603
 #endif
 
-#if __TBB_PREVIEW_PARALLEL_PHASE
 #define TBB_HAS_PARALLEL_PHASE 202603
```
kboyarinov: Don't forget to bump this value, as we discussed offline.

include/oneapi/tbb/detail/_parallel_phase.h
```diff
@@ -0,0 +1,74 @@
...
+
+template<>
+struct valid_flags<> {
+    static constexpr bool value = true;
+};
+
+template<typename Tag, typename... Tags>
+struct valid_flags<Tag, Tags...> {
+    static constexpr bool value = std::is_base_of<tag_base, Tag>::value && valid_flags<Tags...>::value;
+};
```
kboyarinov:
```suggestion
template <typename... Tags>
struct valid_flags : tbb::detail::conjunction<std::is_base_of<tag_base, Tag>...> {};
```

include/oneapi/tbb/task_arena.h
```diff
@@ -27,7 +27,9 @@
 #include "detail/_small_object_pool.h"
 #include "detail/_task.h"
 #include "detail/_task_handle.h"
+#include "detail/_parallel_phase.h"
 #include "info.h"
+#include "oneapi/tbb/detail/_utils.h"
```
kboyarinov:
```suggestion
#include "detail/_utils.h"
#include "info.h"
```

include/oneapi/tbb/task_arena.h
```diff
@@ -586,31 +526,73 @@ class task_arena : public task_arena_base {
...
-        scoped_parallel_phase(task_arena& ta, bool with_fast_leave = false)
-            : arena(ta), one_time_fast_leave(with_fast_leave)
+        class flags {
+            friend class parallel_phase;
+            friend class task_arena;
+            friend void start_parallel_phase(flags);
+            friend void end_parallel_phase(flags);
+
+            std::uint32_t my_start;
+            std::uint32_t my_end;
```
kboyarinov: I would probably better name them `my_start_flags` and `my_end_flags`, since I initially thought these values forms a "list" that starts from `my_start` and ends by `my_end`.

include/oneapi/tbb/task_arena.h
```diff
@@ -586,31 +526,73 @@ class task_arena : public task_arena_base {
...
+        parallel_phase(d1::attach, flags f = {}) : my_flags(f) {
+            r1::enter_parallel_phase(nullptr, /*reserved*/0);
+        }
+        parallel_phase(task_arena& ta, flags f = {})
+            : my_arena(&ta), my_flags(f)
         {
-            arena.start_parallel_phase();
+            suppress_unused_warning(reserved);
+            my_arena->initialize();
+            r1::enter_parallel_phase(my_arena, /*reserved*/0);
```
kboyarinov:
```suggestion
            my_arena.start_parallel_phase(my_flags);
```

include/oneapi/tbb/task_arena.h
```diff
@@ -586,31 +526,73 @@ class task_arena : public task_arena_base {
...
+        public:
+            flags() : my_start(0), my_end(0) {};
+            template <typename... Flags, typename = typename std::enable_if<phase::valid_flags<Flags...>::value>::type>
+            flags(Flags...) : my_start(phase::combine_tags<phase::start, Flags...>::value),
+                  my_end(phase::combine_tags<phase::end, Flags...>::value) {}
+        };
+        class end_with_fast_leave : public phase::tag<phase::end, phase::end_fast_leave> {};
+
+        parallel_phase(d1::attach, flags f = {}) : my_flags(f) {
+            r1::enter_parallel_phase(nullptr, /*reserved*/0);
```
kboyarinov, line 549:
```suggestion
            this_task_arena::start_parallel_phase(my_flags);
```

include/oneapi/tbb/task_arena.h
```diff
@@ -586,31 +526,73 @@ class task_arena : public task_arena_base {
...
+                r1::exit_parallel_phase(my_arena, static_cast<std::uintptr_t>(my_flags.my_end));
+            my_arena = other.my_arena;
+            my_flags = other.my_flags;
+            my_owns = other.my_owns;
+            other.my_owns = false;
+            return *this;
+        }
+        ~parallel_phase() {
+            if (my_owns)
+                r1::exit_parallel_phase(my_arena, static_cast<std::uintptr_t>(my_flags.my_end));
```
kboyarinov: my_arena.end_parallel_phase() or this_task_arena::end_parallel_phase() may also be called here, but will introduce additional check. It is up to you to apply it or keep as-is

include/oneapi/tbb/task_arena.h
```diff
@@ -685,16 +667,13 @@ inline void enqueue(F&& f, d2::task_group& tg) {
     d2::enqueue_impl(tg.defer(std::forward<F>(f)), nullptr);
 }
 
-#if __TBB_PREVIEW_PARALLEL_PHASE
 inline void start_parallel_phase() {
```
kboyarinov: Should it also accept flags?

include/oneapi/tbb/task_arena.h
```diff
@@ -586,31 +526,73 @@ class task_arena : public task_arena_base {
...
+
+            std::uint32_t my_start;
+            std::uint32_t my_end;
+        public:
+            flags() : my_start(0), my_end(0) {};
+            template <typename... Flags, typename = typename std::enable_if<phase::valid_flags<Flags...>::value>::type>
+            flags(Flags...) : my_start(phase::combine_tags<phase::start, Flags...>::value),
+                  my_end(phase::combine_tags<phase::end, Flags...>::value) {}
+        };
+        class end_with_fast_leave : public phase::tag<phase::end, phase::end_fast_leave> {};
```
kboyarinov: Does it makes sense to have a word "flag" explicitly in the name?
I was a bit confused by having both `phase.end()` function and `phase::end_with_fast_leave` that also sounds like a function

isaevil: I am not sure, the name of the flag is already pretty long

include/oneapi/tbb/detail/_parallel_phase.h
```diff
@@ -0,0 +1,64 @@
...
+
+template <typename Boundary, typename Tag, typename... Tags>
+struct combine_tags<Boundary, Tag, Tags...> {
+    static constexpr std::uint64_t value =
+       (std::is_same<typename Tag::boundary_type, Boundary>::value ? Tag::value : 0) | combine_tags<Boundary, Tags...>::value;
+};
+
+template <typename... Tags>
+struct valid_flags : tbb::detail::conjunction<std::is_base_of<tag_base, Tag>...> {};
```
isaevil:
```suggestion
struct valid_flags : tbb::detail::conjunction<std::is_base_of<tag_base, Tags>...> {};
```

include/oneapi/tbb/task_arena.h
isaevil: Seems unnecessary :)

include/oneapi/tbb/task_arena.h
isaevil: Added.

include/oneapi/tbb/task_arena.h

include/oneapi/tbb/detail/_config.h
isaevil: Updated.

include/oneapi/tbb/task_arena.h
```diff
@@ -586,31 +526,72 @@ class task_arena : public task_arena_base {
...
+        void end() {
+            if (my_owns) {
+                my_owns = false;
+                r1::exit_parallel_phase(my_arena, static_cast<std::uintptr_t>(my_flags.my_end_flags));
+            }
+        }
+    private:
+        task_arena* my_arena{nullptr};
+        flags my_flags;
+        bool my_owns{true};
```
kboyarinov: Ownership semantic is a bit confusing to me. Probably, the better name is something like `m_active` or `m_ended` with the default `false`

include/oneapi/tbb/task_arena.h
```diff
@@ -586,31 +526,72 @@ class task_arena : public task_arena_base {
...
+        std::uintptr_t reserved;
```
kboyarinov:
```suggestion
        std::uintptr_t m_reserved{};
```

src/tbb/arena.cpp
vossmjp, line 631: Should my_version_and_traits be copied here too? It seems like if a task_arena attaches to an existing explicit task_arena that has a fast policy (attaching from within a task executing in that explicit arena) the fast leave_policy could be lost.

include/oneapi/tbb/task_arena.h
```diff
@@ -586,31 +526,72 @@ class task_arena : public task_arena_base {
...
+        class flags {
+            friend class parallel_phase;
+            friend class task_arena;
+            friend void start_parallel_phase(flags);
+            friend void end_parallel_phase(flags);
+
+            std::uint32_t my_start_flags;
+            std::uint32_t my_end_flags;
+        public:
+            flags() : my_start_flags(0), my_end_flags(0) {};
```
vossmjp:
```suggestion
            flags() : my_start_flags(0), my_end_flags(0) {}
```

include/oneapi/tbb/task_arena.h
vossmjp: What about `phase::end_flag_fast_leave`? Its no longer, but maybe a bit awkward.

include/oneapi/tbb/task_arena.h
```diff
@@ -426,32 +383,22 @@ class task_arena : public task_arena_base {
...
-    )
+                    priority a_priority = priority::normal, leave_policy lp = leave_policy::automatic)
     {
         __TBB_ASSERT(!my_arena.load(std::memory_order_relaxed), "Impossible to modify settings of an already initialized task_arena");
         if( !is_active() ) {
             my_max_concurrency = max_concurrency_;
             my_num_reserved_slots = reserved_slots;
             my_priority = a_priority;
-#if __TBB_PREVIEW_PARALLEL_PHASE
             set_leave_policy(lp);
```
vossmjp, line 394: According to the spec, a `task_arena` can be terminated and then reused. So it could be reinitialized. Since `set_leave_policy` uses an OR in its implementation, a task_arena first initialized with fast leave would keep a fast leave policy even after a terminate followed by a re-initialize with automatic. If that's intended, then maybe it should be documented. But all of the other variables set here are changeable after terminate.

src/tbb/arena.cpp
isaevil: I think you are right. But It seems like `attach` constructor is already little bit broken at this point. For example, it does not save back other settings to `task_arena` (core_type_id, numa_id). If you create `task_arena` attached to constrained explicit arena, it won't know actual settings. It will work, because `r1::arena` is the same, but if you'll try to copy task arena, it won't be constrained anymore. For example:
```cpp
task_arena outer_ta{constraints.set_core_type(some_id)};
outer_ta.execute([&outer_ta] {
    task_arena inner_ta(attach); // will be constrained, but settings are not actually saved into "task_arena"
    task_arena inner_copy(inner_ta) // it won't be constrained!
});
```
I think that we probably should fix that as a part of another patch.

include/oneapi/tbb/task_arena.h
isaevil: I don't think that it should be kept between initializations. Since we don't provide an API to dynamically set leave_policy, I think it should be possible to do so at least using `terminate/initialize`.

include/oneapi/tbb/task_arena.h
isaevil: Renamed to `my_active`

include/oneapi/tbb/task_arena.h
isaevil: Renamed.

include/oneapi/tbb/task_arena.h
isaevil: Fixed.

kboyarinov (approved):
LGTM

vossmjp (approved):
LGTM

PR #2158, opened 2026-08-13
Use try-lock approach on stealing path

isaevil:
### Description 
This patch changes the behavior on stealing path: instead of unconditional spin-locking on victim's task_pool the worker only spins for a limited time (determined by `atomic_backoff` structure) and if failed (or task pool is empty), goes back to the stealing loop and tries to find another slot. This change improves performance by reducing contention arena slots' task pool atomic on high core count systems.

In contended scenarios, the retention of worker threads may decrease since pause count of `waiter` object will saturate faster, thusl `out_of_work` might be called earlier.


src/tbb/arena_slot.cpp
```diff
@@ -147,8 +147,18 @@ d1::task* arena_slot::get_task(execution_data_ext& ed, isolation_type isolation)
 }
 
 d1::task* arena_slot::steal_task(arena& a, isolation_type isolation, std::size_t slot_index) {
-    d1::task** victim_pool = lock_task_pool();
-    if (!victim_pool) {
+    d1::task** victim_pool = nullptr;
+    atomic_backoff backoff;
+    do {
+        victim_pool = try_lock_task_pool();
```
akukanov: Instead of a new try-lock function and an external loop, would it make sense to extend the existing function to support both bounded and unbounded spin-wait, depending on a template or function parameter?

isaevil: For the sake of better encapsulation, I think it makes sense. Perhaps something like:
```cpp
template<bool Timed>
task** lock_task_pool();
```

isaevil: Updated, I've settled on
```cpp
 template<typename WaitTag = pool_unbounded_wait>
 d1::task** lock_task_pool();
```
With available tags `pool_bounded_wait` and `pool_unbounded_wait`

src/tbb/arena_slot.h
```diff
@@ -116,6 +116,17 @@ class arena_slot : private arena_slot_shared_state, private arena_slot_private_s
         fill_with_canary_pattern( 0, my_task_pool_size );
     }
 
+    struct pool_bounded_wait{};
+    struct pool_unbounded_wait{};
+
+    bool wait_for_lock(atomic_backoff& backoff, pool_unbounded_wait) {
```
vossmjp, line 122: Is this reachable from anywhere now? Or is just the pool_bounded_wait called?

src/tbb/arena_slot.cpp
```diff
@@ -147,8 +147,9 @@ d1::task* arena_slot::get_task(execution_data_ext& ed, isolation_type isolation)
 }
 
 d1::task* arena_slot::steal_task(arena& a, isolation_type isolation, std::size_t slot_index) {
-    d1::task** victim_pool = lock_task_pool();
-    if (!victim_pool) {
+    d1::task** victim_pool = lock_task_pool<pool_bounded_wait>();
```
vossmjp, line 150: This explicitly uses `pool_bounded_wait`, so is the other variant just dead code?

isaevil: Yes, for now. It will be used by ["get_task refactoring" patch](https://github.com/uxlfoundation/oneTBB/tree/dev/restore-get-task-improvements) once we return it.

vossmjp (approved):
LGTM. But this needs to land with https://github.com/uxlfoundation/oneTBB/pull/1805, right?

dnmokhov (approved):
LGTM

PR #2160, opened 2026-08-19
Save settings from r1::arena to task_arena during attach

isaevil:
### Description
Currently, only some of the task_arena settings are saved during "attach" construction (max_concurrency, reserved_slots, priority), but not constraints and not leave policy. This leads to such scenario:
```cpp
task_arena outer_ta{constraints.set_core_type(some_id)};
outer_ta.execute([&outer_ta] {
    task_arena inner_ta(attach); // will be constrained, but settings are not actually saved into "task_arena"
    task_arena inner_copy(inner_ta) // it won't be constrained!
});
```
This patch stores that information in `r1::arena` and sets it during attach construction of `task_arena`

### Type of change

### Tests

- [ ] not needed

### Documentation

- [x] updated
- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/src/tbb/arena.h
+++ b/src/tbb/arena.h
@@ -330,6 +330,14 @@ struct arena_base : padded<intrusive_list_node> {
 
     threading_control_client my_tc_client;
 
+    tbb::task_arena::leave_policy my_leave_policy;
+
+    d1::numa_node_id my_numa_id;
+
+    d1::core_type_id my_core_type;
+
+    int my_max_threads_per_core;
+
 #if TBB_USE_ASSERT
     //! Used to trap accesses to the object after its destruction.
     std::uintptr_t my_guard;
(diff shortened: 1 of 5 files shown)


src/tbb/arena.h
```diff
@@ -330,6 +330,14 @@ struct arena_base : padded<intrusive_list_node> {
 
     threading_control_client my_tc_client;
 
+    tbb::task_arena::leave_policy my_leave_policy;
+
+    d1::numa_node_id my_numa_id;
+
+    d1::core_type_id my_core_type;
+
+    int my_max_threads_per_core;
```
kboyarinov, line 339: It seems like we can now remove these fields from `task_arena_base` and keep them in `r1::arena` only. 
Removal from `task_arena_base` is definitely a layout break and accessing the fields from `r1::arena` may require adding new entry points. 
So, not a stopper for a current fix, but probably it makes sense to open an issue not to forget to think about this in the future.

isaevil: Created #2162

kboyarinov (approved):

PR #2163, opened 2026-08-23
Fix FIFO task deadlock during nested waits

### LINKED ISSUE #2156: Potential deadlock: FIFO task enqueued during local_wait_for_all may not be consumed in single-slot configuration
# Summary

When local_wait_for_all sets m_properties.fifo_tasks_allowed = false (in src/tbb/task_dispatcher.h at commit fe9bfc5870679f87cfafe4f16d5735e718c1a4bc, line ~302), I believe a scenario can arise where a FIFO task enqueued after a thread goes into wait may not be consumed when there is only a single arena slot/worker, potentially resulting in a deadlock.
```C++
    oneapi::tbb::task_group tg;
    oneapi::tbb::task_arena ta(1, 0);

    ta.enqueue([&ta]()
    {
        oneapi::tbb::task_group inner;
        ta.enqueue([]() { /* do something */ }, inner);
        inner.wait();
    }, tg);
    ta.wait_for(tg);
```

# Version

v2022.3.0

# Environment

Reproduction is expected to be independent of OS/compiler because this is a scheduler logic concern visible in the source. If you need specific environment info I can provide details from my system.

# Observed Behavior

- In task_dispatcher::local_wait_for_all (src/tbb/task_dispatcher.h), the code saves the current properties into dl_guard.old_properties and immediately after sets:

    m_properties.outermost = false;
    m_properties.fifo_tasks_allowed = false;

- receive_or_steal_task (called later in the dispatch loop) receives fifo_allowed from dl_guard.old_properties.fifo_tasks_allowed, so the current dispatch iteration checks FIFO eligibility based on the *old* properties.

- My concern is a timing / single-slot scenario: if the runtime has only one arena slot (one worker) and that thread enters local_wait_for_all, then a task is enqueued into the FIFO stream while the thread is waiting. Depending on the ordering of idle/published flags and the m_properties change, that enqueued FIFO task might not become visible/consumable by the waiting thread (or by stealing logic), which could leave the system stuck waiting for work that is never executed.

- In short: with a single slot and FIFO tasks disallowed by properties at the relevant time, an enqueue that happens during the wait may not be consumed, potentially causing a deadlock.

# Expected Behavior

A FIFO task enqueued while a thread is in local_wait_for_all should be discovered and executed (or published for stealing) in a way that prevents the system from deadlocking, even in the single-slot case. Changing m_properties for the current dispatcher should not prevent FIFO tasks from being consumed if the outer/previous context allowed them or if the tasks were enqueued while the thread was idle.

# Steps To Reproduce
```C++
    oneapi::tbb::task_group tg;
    oneapi::tbb::task_arena ta(1, 0);

    std::mutex mtx;
    std::condition_variable cv;

    ta.enqueue([&ta, &mtx, &cv]()
    {
        oneapi::tbb::task_group inner;
        inner.run([] {
            printf("test1");
        });
        ta.enqueue([]() {
            printf("test2");
        }, inner);
        inner.wait();

        std::unique_lock<std::mutex> lock(mtx);
        cv.notify_all();
    }, tg);

    // do something else
    {
        std::unique_lock<std::mutex> lock(mtx);
        cv.wait(lock);
    }

    ta.wait_for(tg);
```

aryan-sapra:
### Description

When a single-slot `task_arena` executes a FIFO task that enqueues another FIFO task and waits for it, `local_wait_for_all` could disable FIFO consumption for nested waits. This caused the only worker to wait indefinitely.

This change preserves the inherited `fifo_tasks_allowed` state when entering `local_wait_for_all`, allowing nested waits to consume FIFO tasks when FIFO execution was already permitted.

Fixes #2156

### Type of change

- [x] bug fix
- [ ] new feature
- [x] tests
- [ ] infrastructure
- [ ] documentation

### Tests

- [x] added
- [ ] not needed

### Documentation

- [ ] updated
- [x] not needed

- [ ] Yes
- [x] No

### Other information

The regression test timed out on the original code and passes after the fix.

Related tests passed:
- `test_task_arena`
- `test_task_group`
- `conformance_task_arena`

### DIFF
--- a/src/tbb/task_dispatcher.h
+++ b/src/tbb/task_dispatcher.h
@@ -299,7 +299,7 @@ d1::task* task_dispatcher::local_wait_for_all(d1::task* t, Waiter& waiter ) {
     ed.wait_ctx = waiter.wait_ctx();
 
     m_properties.outermost = false;
-    m_properties.fifo_tasks_allowed = false;
+    m_properties.fifo_tasks_allowed = dl_guard.old_properties.fifo_tasks_allowed;
 
     if (!dl_guard.is_initially_registered) {
         m_thread_data->my_arena->my_tc_client.get_pm_client()->register_thread();
--- a/test/tbb/test_task_arena.cpp
+++ b/test/tbb/test_task_arena.cpp
@@ -2253,6 +2253,27 @@ TEST_CASE("Test enqueue guarantees when task_arena is combined with task_group")
 
 #endif // TBB_USE_EXCEPTIONS
 
+//! \brief \ref regression
+TEST_CASE("FIFO task enqueued during local_wait_for_all is consumed in single-slot arena") {
+    tbb::task_arena ta{1, 0};
+    tbb::task_group outer;
+    std::atomic<bool> inner_ran{false};
+
+    ta.enqueue([&] {
+        tbb::task_group inner;
+
+        ta.enqueue([&] {
+            inner_ran.store(true, std::memory_order_release);
+        }, inner);
+
+        inner.wait();
+    }, outer);
+
+    ta.wait_for(outer);
+
+    REQUIRE(inner_ran.load(std::memory_order_acquire));
+}
+
 #if __TBB_CPP17_PRESENT
 //! \brief \ref regression
 TEST_CASE("ODR-use task_arena constants") {


src/tbb/task_dispatcher.h
```diff
@@ -299,7 +299,7 @@ d1::task* task_dispatcher::local_wait_for_all(d1::task* t, Waiter& waiter ) {
     ed.wait_ctx = waiter.wait_ctx();
 
     m_properties.outermost = false;
-    m_properties.fifo_tasks_allowed = false;
+    m_properties.fifo_tasks_allowed = dl_guard.old_properties.fifo_tasks_allowed;
```
darknesswind, line 302: The `dl_guard` constructor directly copies `m_properties`, so this change is equivalent to simply deleting that line of code. Moreover, I believe there are deliberate design considerations behind prohibiting FIFO, and disabling it outright may be too drastic.

xhon-pelushi (commented):
Built this both ways and tried to break it. The diagnosis is right, the one-line change is exactly scoped, and there's a design-symmetry argument for it that the description doesn't make. One gap: the regression test covers a different shape than the one @isaevil confirmed in #2156.

## Why clearing that flag mattered at all

Worth spelling out, because the one-line diff looks like it should affect this dispatch level and doesn't. This level's own stealing already reads the *inherited* value:

```cpp
// task_dispatcher.h:379
t = receive_or_steal_task(
    *m_thread_data, ed, waiter, context_guard, isolation,
    dl_guard.old_properties.fifo_tasks_allowed, critical_allowed
);
```

So `m_properties.fifo_tasks_allowed = false` at line 302 never restricted the frame that set it. All it did was poison what a **nested** `local_wait_for_all` would later capture as *its* `dl_guard.old_properties`. That's why the failure is specific to nesting, and why touching only line 302 is the right scope.

## The change restores a symmetry that was already there for critical tasks

`critical_task_allowed` and `fifo_tasks_allowed` are set up as a pair — `nested_arena_context` enables both together when an external thread joins:

```cpp
// arena.cpp:697-699
m_orig_fifo_tasks_allowed = m_task_dispatcher->allow_fifo_task(true);
m_orig_critical_task_allowed = m_task_dispatcher->m_properties.critical_task_allowed;
m_task_dispatcher->m_properties.critical_task_allowed = true;
```

But at the top of `local_wait_for_all`, only `outermost` and `fifo_tasks_allowed` were reset — `m_properties.critical_task_allowed` is left alone and therefore inherited by nested waits, exactly the way this patch now makes `fifo_tasks_allowed` behave. So the old line was the odd one out, and `dl_guard.old_properties.fifo_tasks_allowed` is the value the rest of the function already treats as authoritative. That reads to me like the patch is restoring the intended invariant rather than papering over a symptom.

## Verified

Built `master` (`03bfb4e`) and this branch, `-DCMAKE_BUILD_TYPE=Release -DTBB_TEST=ON`, GCC 13.3.

**The regression test does fail without the fix.** Transplanted only `test/tbb/test_task_arena.cpp` from this PR onto unpatched `master`:

```
master:   test/tbb/test_task_arena.cpp:2257: FATAL ERROR: test case CRASHED: SIGTERM
          [doctest] test cases: 25 | 24 passed | 1 failed | 1 skipped
PR head:  [doctest] test cases: 25 | 25 passed | 0 failed | 1 skipped
```

**Suites on the PR head**, all clean:

| suite | result |
| --- | --- |
| `test_task_arena` | 25/25 passed, 1 skipped |
| `test_task_group` | 16/16 passed, 3 skipped |
| `conformance_task_arena` | 12/12 passed |
| `test_arena_priorities` | 1/1 passed |
| `test_resumable_tasks` | 5/5 passed |

I added the last two on the theory that a change to FIFO admission is most likely to show up in priority ordering or in resumable-task suspend/resume; neither moved.

**The risk I went looking for and didn't find.** Inheriting the flag means a nested wait can now consume arbitrary enqueued work, and each level holds a stack frame, so I expected a depth problem. Chained enqueue-then-wait in a single-slot arena:

```
depth 50   -> reached 51    ok
depth 500  -> reached 501   ok
depth 5000 -> reached 5001  ok
```

No stack exhaustion at 5000 levels on default thread stacks. Not unbounded-safe in principle, but it doesn't fall over anywhere near realistic depths.

## The gap: the test isn't @isaevil's reproducer

The added test enqueues from outside and then calls `ta.wait_for(outer)`, so the external thread joins the arena and becomes a participant. The reproducer @isaevil settled on in [#2156](https://github.com/oneapi-src/oneTBB/issues/2156) is deliberately different — the external thread *never* joins, it waits on an atomic instead, so the worker really is the only participant:

```cpp
tbb::task_arena ta(1);
std::atomic<bool> done{false};
ta.enqueue([&] {
    tbb::task_group inner;
    tbb::this_task_arena::enqueue([&] { done.store(true, std::memory_order_release); }, inner);
    inner.wait();
});
while (!done.load(std::memory_order_acquire)) { /* spin with a deadline */ }
```

Measured both shapes:

```
                              master            PR head
PR's shape (wait_for)         hang (30s kill)   inner_ran=1
#2156's shape (atomic flag)   hang (20s kill)   done=1
```

Good news: the fix covers both. But since the second shape is the one a maintainer confirmed and it exercises a genuinely different arena-participation path — `this_task_arena::enqueue` from a worker, with no external thread in the arena — I'd add it alongside the existing case rather than instead of it. It's cheap and it pins the scenario from the issue.

## What I did not test

I ran five suites, not the full `TBB_TEST` set, so I can't speak to the whole matrix. This changes scheduler admission for every nested wait inside an enqueued task at any arena size, and I only exercised single-slot arenas plus whatever the suites above cover — a broader run and a look on a non-x86 target seem worth having before merge, given the class of change. Nothing here was tested under TSan either.

PR #2165, opened 2026-08-25
Simplify constrained nested arena test

dnmokhov:
### Description 
The constrained nested arena test could fail on Windows systems with a NUMA node spanning across multiple processor groups due to its incorrect helper API usage: `allocate_current_affinity_mask()` was influenced by the current processor group (`get_arena_affinity()` could have been used instead). This patch makes the test independent of actual binding behavior and instead focuses on the regression that was fixed (#2121) by removing direct affinity mask comparisons and instead verifying that the nested arena has only 1 slot, and that `current_thread_index()` can exceed 0.

Also, updated the assertion messages in `common_arena_constraints.h` to clarify that the tests are checking for the union of NUMA node and core type masks being equal to process affinity, rather than intersection.

### Type of change

### Tests
- [ ] not needed

### Documentation
- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown

### DIFF
--- a/test/common/common_arena_constraints.h
+++ b/test/common/common_arena_constraints.h
@@ -313,7 +313,7 @@ private:
             hwloc_bitmap_or(buffer_cpu_set, buffer_cpu_set, numa_node_info.cpuset);
         }
         REQUIRE_MESSAGE( hwloc_bitmap_isequal(buffer_cpu_set, process_cpuset),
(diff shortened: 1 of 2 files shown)


vossmjp (approved):
LGTM

PR #2171, opened 2026-08-28
Add exported symbols check with baselines

isaevil:
### Description 
The patch implements GitHub Action that allows verifying exported symbols of a binary against baseline to detect possible compatibility breaks. It uses LIEF to support ELF, Mach-O and PE using the same Python script.

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown


.github/actions/abi-check/compare_symbols.py
```diff
@@ -0,0 +1,303 @@
...
+import argparse
+import pathlib
+import re
+import sys
+
+import lief
+
+# The job summary of a workflow run is capped, so only the beginning of a long
+# list of symbols is reported
+MAX_REPORTED_SYMBOLS = 51
```
omalyshe, line 24: Just curious - What is the reason for this limitation and how was this "magic" number found?

isaevil: GitHub Job summaries have a limit of [1MiB](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#step-isolation-and-limits). The intention is to have some cap to not exceed this limit when outputting multiple symbol reports (for each platform and each library). But to be honest, I did not do the calculation of precise value.

.github/actions/abi-check/action.yml
```diff
@@ -0,0 +1,80 @@
...
+      Directory holding the ABI baselines of the project.
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Install ABI tooling
+      shell: bash
+      run: |
+        set -e -o pipefail
```
aleksei-fedotov: Why not tracing what is being executed?
```suggestion
        set -e -x -o pipefail
```

.github/actions/abi-check/action.yml
```diff
@@ -0,0 +1,80 @@
...
+          "${venv_python}" -m pip install --upgrade pip
+          "${venv_python}" -m pip install -r "${GITHUB_ACTION_PATH}/requirements.txt"
+        fi
+
+        echo "ABI_PYTHON=${venv_python}" >> "${GITHUB_ENV}"
+
+    - name: Check ABI
+      shell: bash
+      run: |
+        set -e -o pipefail
```
aleksei-fedotov: Same about tracing here:
```suggestion
        set -e -x -o pipefail
```

.github/actions/abi-check/compare_symbols.py
```diff
@@ -0,0 +1,303 @@
...
+
+import argparse
+import pathlib
+import re
+import sys
+
+import lief
+
+# The job summary of a workflow run is capped, so only the beginning of a long
+# list of symbols is reported
```
aleksei-fedotov: Is the list always long? I would rephrase if not:
```suggestion
# The job summary of a workflow run is capped, so only the below specified number 
# of symbols that appeared first is reported
```

omalyshe (dismissed):
LGTM. Thank you.

thread_composability_manager/test/abi/linux-32/tcm.txt
aleksei-fedotov, line 1: Can we avoid duplicating lines for 32 and 64 bits, and, perhaps, even Windows and Linux? They are same since it is C-style interface.

.github/workflows/abi.yml
```diff
@@ -0,0 +1,355 @@
...
+          project: Thread Composability Manager
+          baseline-dir: thread_composability_manager/test/abi
+          binary: >-
+            libraries/*/lib/libtcm*
+            libraries/*/bin/tcm*.dll
+          output-dir: abi-artifacts/tcm
+      - name: Prepare the upgraded baseline as a patch
+        shell: bash
+        run: |
+          set -e -o pipefail
```
aleksei-fedotov: Same.

.github/workflows/abi.yml
```diff
@@ -0,0 +1,355 @@
...
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+        with:
+          name: baseline-patch
+          path: abi.patch
+          if-no-files-found: ignore
+          retention-days: 7
+      - name: Report the summary
+        shell: bash
+        run: |
+          set -e -o pipefail
```
aleksei-fedotov: And here as well.

.gitignore
```diff
@@ -59,4 +59,3 @@ CMakeSettings.json
 # -------- CTags --------
 .tags
 .ctags
-
```
aleksei-fedotov, line 62: Why removed?

.github/workflows/abi.yml
```diff
@@ -0,0 +1,355 @@
...
+            -DTBB_ENABLE_IPO=OFF \
+            ${{ matrix.cmake_extra_args }}
+          cmake --build build --target package -j$(nproc)
+
+      - name: Build debug libraries
+        shell: bash
+        run: |
+          set -e -x -o pipefail
+          cmake -S . -B build-debug -DCMAKE_BUILD_TYPE=debug -DTBB_TEST=OFF -DTCM_TEST=OFF \
+            -DTBB_ENABLE_IPO=OFF \
```
aleksei-fedotov: I think we should enable build of TCM explicitly, here and in other 3 places.

.gitignore
isaevil: I think I wanted to add something to .gitignore. But anyway, there is no need for extra blank line.

.github/actions/abi-check/action.yml
isaevil: Done.

.github/workflows/abi.yml
isaevil: Done.

thread_composability_manager/test/abi/linux-32/tcm.txt
isaevil: Good suggestion. Updated.

.github/actions/abi-check/requirements.txt
```diff
@@ -0,0 +1 @@
+lief==1.0.0
```
aleksei-fedotov, line 1: Shall this file include a copyright notice as well?

isaevil: I don't think that it is needed here. @omalyshe ?

omalyshe: > I don't think that it is needed here. @omalyshe ?

I agree.

aleksei-fedotov (approved):

PR #2172, opened 2026-08-28
Add RFC for resource_limited node policy

kboyarinov:
### Description 

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown

### DIFF
--- a/rfcs/proposed/flow_graph_serializers/resource_limited_policy.md
+++ b/rfcs/proposed/flow_graph_serializers/resource_limited_policy.md
@@ -1,0 +1,682 @@
+# Resource Limiting as a Functional Node Policy
+
+Note: this document is a sub-RFC for the [Resource-limited Nodes RFC](./README.md). 
+
+## Introduction
+
+The preview `flow::resource_limiter` and `flow::resource_limited_node` brought new functionality to manage
+access to the resources shared between multiple nodes in different parts of the Flow Graph.
+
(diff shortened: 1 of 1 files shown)


rfcs/proposed/flow_graph_serializers/resource_limited_policy.md
```diff
@@ -0,0 +1,682 @@
...
+acquire_concurrency --> |Yes| spawn[Spawn acquiring task]
+acquire_concurrency --> |No| withdraw_request
+spawn --> exit
+```
+
+The acquiring task is similar to the one shown above.
+
+Given that the implementation of a truly `rejecting` resource-limited node is complicated, requires tradeoffs where the resources would be held while the
+service tasks are scheduled, and significantly increases the time spent by the predecessor to put the item to the node, the current document proposes
+to have the `resource_limited` policy as a third message acceptance policy, not an alternative. The new policy would behave like the
```
vossmjp, line 208:
```suggestion
to have the `resource_limited` policy as a third message acceptance policy. In addition, we see `rejecting` policies used 
almost exclusively in patterns around token passing, where a token represents a resource. We therefore think a `resource_limited` policy would
not likely be combined with a `rejecting` policy even if possible, since these are a two distinct usage patterns. The new policy would behave like the
```

PR #2182, opened 2026-09-02
Test syscall failure during interleaved memory allocation

Alexandr-Konovalov:
### Description 
Initial implementation of failed syscalls testing.

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [X] not needed

- [ ] Yes
- [X] No
- [ ] Unknown

### DIFF
--- a/test/tbb/test_numa_allocation_whitebox.cpp
+++ b/test/tbb/test_numa_allocation_whitebox.cpp
@@ -1,0 +1,140 @@
+/*
+    Copyright (c) 2026 UXL Foundation Contributors
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+        http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+*/
+
+//! \file test_numa_allocation_whitebox.cpp
+//! \brief Test for [internal] functionality
+
+#define __TBB_NO_IMPLICIT_LINKAGE 1
+#define TBB_PREVIEW_NUMA_ALLOCATION 1
+
+#include "common/test.h"
+
+#if __linux__
+
+static bool overrided_madvise_called = false;
+static bool overrided_move_pages_called = false;
+static bool madvise_should_fail = true;
+
+#define madvise(addr, length, advice) failed_madvise(addr, length, advice)
+
+extern "C" {
+static int failed_madvise(void*, size_t, int) noexcept (true) {
+    overrided_madvise_called = true;
+    return madvise_should_fail ? -1 : 0;
+}
+}
+#elif !(_WIN32 || _WIN64)
+
+#endif
+
+#include "../../src/tbb/numa_allocation.cpp"
+
+#undef madvise
+
+namespace tbb {
+namespace detail {
+namespace r1 {
+
+#if __linux__
+
+static long dummy_move_pages(int, unsigned long, void**, const int*, int*, int) {
+    overrided_move_pages_called = true;
+    return -1;
+}
+
+static struct bitmask* dummy_numa_bitmask_alloc(unsigned int) {
+    return nullptr;
+}
+
+static void dummy_numa_bitmask_free(struct bitmask*) {}
+
+static int dummy_numa_bitmask_isbitset(const struct bitmask*, unsigned int) {
+    return 1;
+}
+
+static struct bitmask* dummy_numa_bitmask_setbit(struct bitmask* m, unsigned int) {
+    return m;
+}
+
+static void dummy_numa_interleave_memory(void*, size_t, struct bitmask*) {}
+
+#elif !(_WIN32 || _WIN64)
+
+#endif
+
+void assertion_failure(const char*, int, const char*, const char*) {
+    REQUIRE_MESSAGE(false, "Can't be called");
+}
+
+bool dynamic_link( const char* ,
+                   const tbb::detail::r1::dynamic_link_descriptor[],
+                   std::size_t ,
+                   tbb::detail::r1::dynamic_link_handle*,
+                   int) {
+#if __linux__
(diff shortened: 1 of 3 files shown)


test/tbb/test_numa_allocation_whitebox.cpp
```diff
@@ -0,0 +1,180 @@
...
+    size_t per_chunk = 2*4*1024;
+    tbb::detail::d1::numa_node_id nodes_ids_array[] = {0, 0};
+    tbb::detail::d1::numa_node_id *nodes_ids = nodes_ids_array;
+
+    // under Windows, must use < page size to call VirtualAlloc2_ptr in the committing loop
+    void *ptr = tbb::detail::r1::allocate_interleaved(tbb::detail::r1::DefaultSystemPageSize() / 2,
+                                                      nodes_ids, 2, per_chunk);
+    REQUIRE(ptr == nullptr);
+#if __linux__
+    REQUIRE_MESSAGE(overrided_madvise_called, "Failed madvise syscall was not called");
```
aleksei-fedotov:
```suggestion
    REQUIRE_MESSAGE(overridden_madvise_called, "Failed madvise syscall was not called");
```
Also, initialize it to `false` just before the call?

test/tbb/test_numa_allocation_whitebox.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+#define __TBB_NO_IMPLICIT_LINKAGE 1
+#define TBB_PREVIEW_NUMA_ALLOCATION 1
+
+#include "common/test.h"
+
+#if __linux__
+
+static bool overrided_madvise_called = false;
+static bool overrided_move_pages_called = false;
```
aleksei-fedotov: A typo:
```suggestion
static bool overridden_madvise_called = false;
static bool overridden_move_pages_called = false;
```

test/tbb/test_numa_allocation_whitebox.cpp
```diff
@@ -0,0 +1,186 @@
...
+static bool madvise_should_fail = true;
+
+#define madvise(addr, length, advice) failed_madvise(addr, length, advice)
+
+extern "C" {
+static int failed_madvise(void*, size_t, int) noexcept (true) {
```
aleksei-fedotov: `failed_madvise` is not a good name for this function. First, because it does not always fail. Second, because usually such entities are called mock or fake entities. Therefore, consider naming to something like `mock_madvise` or `fake_madvise`. `injected_madvise` will also look appropriate I think.

Also, consider applying this idea to the rest of injected functions, including those `dummy_*`.

test/tbb/test_numa_allocation_whitebox.cpp
```diff
@@ -0,0 +1,186 @@
...
+TEST_CASE("test failed syscall") {
+    // use non-default value to not call numa_interleave_memory under Linux
+    size_t per_chunk = 2*4*1024;
+    tbb::detail::d1::numa_node_id nodes_ids_array[] = {0, 0};
+    tbb::detail::d1::numa_node_id *nodes_ids = nodes_ids_array;
+
+#if __linux__
+    size_t size = 1024;
+    // make madvise failed
+    madvise_should_fail = true;
```
aleksei-fedotov: Similar to `move_pages_called` below, initialize just before the call?
```suggestion
    madvise_should_fail = true;
    overridden_madvise_called = false;
```

test/tbb/test_numa_allocation_whitebox.cpp
```diff
@@ -0,0 +1,186 @@
...
+#elif _WIN32 || _WIN64
+    // VirtualAlloc2_ptr is expected to fail, must use < page size to call VirtualAlloc2_ptr in the
+    // committing loop
+    void *ptr = tbb::detail::r1::allocate_interleaved(tbb::detail::r1::DefaultSystemPageSize() / 2,
+                                                      nodes_ids, 2, per_chunk);
+    REQUIRE(ptr == nullptr);
+    REQUIRE_MESSAGE(overrided_VirtualAlloc2_ptr_failed, "Failed VirtualAlloc2 syscall was not called");
+
+    // VirtualAlloc2_ptr is expected not to fail, VirtualFree in the committing loop is failing
+    overrided_VirtualFree_failed = false;
```
aleksei-fedotov: Initialization to `false` of this `overrided_VirtualFree_failed` variable as well as `overrided_VirtualAlloc2_ptr_failed` can be moved to the top of their corresponding mock implementations.

test/tbb/test_numa_allocation_whitebox.cpp
```diff
@@ -0,0 +1,186 @@
...
+}
+
+} // namespace r1
+} // namespace detail
+} // namespace tbb
+
+#if __linux__ || _WIN32 || _WIN64
+//! Testing correct behavior if syscall fails
+//! \brief \ref error_guessing
+TEST_CASE("test failed syscall") {
```
aleksei-fedotov: In my opinion, it would be more readable to have separate `TEST_CASE`s for separate tests.

aleksei-fedotov (commented):
Looks good! Although, there is always room for improvement.

Shall the `VirtualAllocEx`/`mmap` also be replaced to test that they is indeed invoked in certain cases and the tested function works as expected?

test/tbb/test_numa_allocation_whitebox.cpp
Alexandr-Konovalov: Done, use mock everywhere.

test/tbb/test_numa_allocation_whitebox.cpp
Alexandr-Konovalov: > initialize it to false just before the call?

Done.

test/tbb/test_numa_allocation_whitebox.cpp
Alexandr-Konovalov: I may not understand you suggestion. I want to have them locally.

test/tbb/test_numa_allocation_whitebox.cpp
Alexandr-Konovalov: I'm not sure. The test case does one thing, it tests `allocate_interleaved()` with failed syscalls, execution is fast enough to not want to skip part of it etc.

aleksei-fedotov (approved):
LGTM!

PR #2184, opened 2026-09-03
resource_limited_node: add test case for releasing and re-requesting the acquired resources

kboyarinov:
### Description 
release_and_rerequest_resources_helper is not covered by the existing tests since the `resource_limiter` do not notify more consumers then the resources available. Without the pressure reported, it is unlikely to get a denial once the consumer is notified - resource_provider orders the notifications and acquisitions statically by the time of the request.

The new test case adds a new provider class that always notifies immediately and denies first N acquisition requests. In a combination with two `resource_limiter` instances that are available during the request, it each input to release and re-acquire them N times.

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown

### DIFF
--- a/test/tbb/test_resource_limited_node.cpp
+++ b/test/tbb/test_resource_limited_node.cpp
@@ -1112,8 +1112,8 @@ TEST_CASE("Test re-requesting the resources when acquisition fails") {
     resource_limiter<int> second_provider{second_resource_value};
     denying_resource_provider third_provider;
 
-    static constexpr int input_value = 0;
-    static constexpr std::size_t num_inputs = 100;
+    constexpr int input_value = 0;
+    constexpr std::size_t num_inputs = 100;
     using node_type = resource_limited_node<int, std::tuple<>>;
 
     std::atomic<std::size_t> count{0};
--- a/test/tbb/test_resource_limited_node.cpp
+++ b/test/tbb/test_resource_limited_node.cpp
@@ -28,6 +28,7 @@
 #include <algorithm>
 #include <array>
 #include <mutex>
+#include <unordered_map>
 #include <vector>
 
 //! \file test_resource_limited_node.cpp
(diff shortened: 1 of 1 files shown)


test/tbb/test_resource_limited_node.cpp
```diff
@@ -1047,3 +1048,92 @@ TEST_CASE("resource_limiter constructors") {
...
+constexpr std::size_t denying_resource_provider::num_denials;
+
+// Test that if one of the required resource providers denied the access to the resource
+// The resource_limited_node correctly releases and re-requests the successfully acquired resources
+//! \brief \ref error_guessing
+TEST_CASE("Test re-requesting the resources when acquisition fails") {
+    using namespace tbb::flow;
+
+    static constexpr int first_resource_value = 500;
+    static constexpr int second_resource_value = 501;
```
isaevil:
```suggestion
    constexpr int first_resource_value = 500;
    constexpr int second_resource_value = 501;
```

test/tbb/test_resource_limited_node.cpp
```diff
@@ -1047,3 +1048,92 @@ TEST_CASE("resource_limiter constructors") {
...
+
+    resource_limiter<int> first_provider{first_resource_value};
+    resource_limiter<int> second_provider{second_resource_value};
+    denying_resource_provider third_provider;
+
+    static constexpr int input_value = 0;
+    static constexpr std::size_t num_inputs = 100;
```
isaevil:
```suggestion
    constexpr int input_value = 0;
    constexpr std::size_t num_inputs = 100;
```

test/tbb/test_resource_limited_node.cpp
```diff
@@ -1047,3 +1048,92 @@ TEST_CASE("resource_limiter constructors") {
...
+    }
+
+    void withdraw(consumer_type&, request_id) override {}
+};
+
+constexpr int denying_resource_provider::resource_value;
+constexpr std::size_t denying_resource_provider::num_denials;
+
+// Test that if one of the required resource providers denied the access to the resource
+// The resource_limited_node correctly releases and re-requests the successfully acquired resources
```
dnmokhov: For clarity,
```suggestion
// Test that if one of the required resource providers denied the access to the resource,
// the resource_limited_node correctly releases and re-requests the successfully acquired resources
```

test/tbb/test_resource_limited_node.cpp
```diff
@@ -1047,3 +1048,149 @@ TEST_CASE("resource_limiter constructors") {
...
+            // Using the barrier to increase the chance that 2 threads will have equal
+            // timestamps in requests
+            submit_barrier.wait();
+    
+            // Submit in different order from different threads
+            if (thread_index % 2 == 0) {
+                node1.try_put(input_value);
+            } else {
+                node2.try_put(input_value);
+            }
```
vossmjp, line 1181: There could also be a denial here even if the timestamps are not the same, but if the later timestamp reaches the limiter first. The consumer creates the timestamp (as part of the request id) and then requests. Since multiple threads are racing here, and id creation and request are not atomic, a request for a later timestamp can happen first. If so the limiter will notify it and any soon-to-be-seen earlier timestamp. The limiter would over notify as compared to the number of handles. At attempted acquisition time, the the later timestamp will be denied. This will exercise the rerequest path, but for a slightly different reason than what the comment implies. So the test is good for this possibility too. No change suggested to code, but maybe broaden the wording of the comment.

vossmjp (dismissed):
LGTM, though I would suggest broadening a comment.

isaevil (approved):

PR #2191, opened 2026-09-16
Fix setting GCC specific flag with TSAN

m-fila:
### Description 

Setting option `TBB_SANITIZE` to `thread` add compiler flags for enabling TSAN but also adds flag to ignore warnings `-Wno-tsan` which is supported with gcc but not by the other compilers.
I propose to add the flag only for gcc. In particular I'd like to be able to run tsan with clang (which now fails because `-Wno-tsan` isn't supported but otherwise works)

### Type of change

### Tests

- [x] not needed

### Documentation

- [ ] needs to be updated
- [x] not needed

- [ ] Yes
- [x] No
- [ ] Unknown


isaevil (approved):
LGTM

PR #2192, opened 2026-09-17
Throw exception from allocators when n*sizeof(T) causes overflow

kboyarinov:
### Description 
Current implementation of allocator classes (`scalable_allocator`, `cache_aligned_allocator`, `tbb_allocator` and `memory_pool_allocator`) unconditionally passes n*sizeof(T) into the underlying allocation function.

This may cause security issues in case n*sizeof(T) results in overflow:
* When n is large, overflowed n*sizeof(T) is small.
* allocate returns the memory allocated for the small overflowed amount without any error diagnostics
* The user may assume the returned memory can store n items which can result in out-of-bound access.

This patch adds throwing the `std::bad_array_new_length` exception from `allocate` in this case, which is aligned with the standard allocators behavior (`std::allocator` and `std::pmr::polymorphic_allocator`).

### Type of change

### Tests

- [ ] not needed

### Documentation

- [ ] needs to be updated
- [ ] not needed

- [ ] Yes
- [ ] No
- [ ] Unknown


include/oneapi/tbb/cache_aligned_allocator.h
```diff
@@ -53,7 +54,15 @@ class cache_aligned_allocator {
 
     //! Allocate space for n objects, starting on a cache/sector line.
     __TBB_nodiscard T* allocate(std::size_t n) {
-        return static_cast<T*>(r1::cache_aligned_allocate(n * sizeof(value_type)));
+        T* p = nullptr;
+
+        // Check overflow before multiplying
+        if (n > max_size()) {
```
kboyarinov: This line is an open question. With this change, each allocate will call `r1::cache_line_size` to determine `max_size`.

The alternative is to only throw `bad_array_new_length` in case of `n*sizeof(T)` overflow and let the `r1::cache_aligned_allocate` check for overflow when the cache line size is subtracted. In this case, for some `n>max_size()` the `std::bad_alloc` is thrown.

aleksei-fedotov: Note that the `cache_aligned_allocate` throws `bad_alloc` when it detects an overflow. [link](https://github.com/oneapi-src/oneTBB/blob/master/src/tbb/allocator.cpp#L195-L197). I guess we should be consistent and throw `bad_array_new_length` in that case as well. 
Fortunately, `bad_array_new_length` inherits `bad_alloc`, so any client code that catches `bad_alloc` will also catch `bad_array_new_length`. So, it won't be a breaking change.

include/oneapi/tbb/scalable_allocator.h
```diff
@@ -222,9 +223,17 @@ class scalable_allocator {
 
     //! Allocate space for n objects.
     __TBB_nodiscard T* allocate(std::size_t n) {
-        T* p = static_cast<T*>(scalable_malloc(n * sizeof(value_type)));
-        if (!p) {
-            throw_exception(std::bad_alloc());
+        T* p = nullptr;
+
+        // Check overflow before multiplying
+        if (n > ~(std::size_t(0)) / sizeof(value_type)) {
```
aleksei-fedotov: Parentheses around `std::size_t` do not change the order of computation here. It can be written as:
```suggestion
        if (n > ~std::size_t(0) / sizeof(value_type)) {
```

include/oneapi/tbb/tbb_allocator.h
```diff
@@ -57,7 +59,15 @@ class tbb_allocator {
 
     //! Allocate space for n objects.
     __TBB_nodiscard T* allocate(std::size_t n) {
-        return static_cast<T*>(r1::allocate_memory(n * sizeof(value_type)));
+        T* p = nullptr;
+
+        // Check overflow before multiplying
+        if (n > ~(std::size_t(0)) / sizeof(value_type)) {
```
aleksei-fedotov: Parentheses around `std::size_t` do not change the order of computation here. It can be written as:
```suggestion
        if (n > ~std::size_t(0) / sizeof(value_type)) {
```

include/oneapi/tbb/memory_pool.h
```diff
@@ -109,9 +109,17 @@ class memory_pool_allocator {
...
     //! Allocate space for n objects.
     pointer allocate( size_type n, const void* /*hint*/ = nullptr) {
-        pointer p = static_cast<pointer>( my_pool->malloc( n*sizeof(value_type) ) );
-        if (!p)
-            throw_exception(std::bad_alloc());
+        pointer p = nullptr;
+
+        // Check overflow before multiplying
+        if (n > max_size()) {
+            throw_exception(std::bad_array_new_length());
```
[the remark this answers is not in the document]
aleksei-fedotov: Was unable to find such tests as well. Why are memory pools not tested on exceptions? Should we add them?

include/oneapi/tbb/cache_aligned_allocator.h
kboyarinov: As far as I understand, my change made the code under the link unreachable because we detect possible overflow earlier in the first if.
I guess we can align the if condition here with other allocators and only check that `n*sizeof(T)` is not already an overflow. Then `cache_aligned_allocate` will check that `n*sizeof(T) + cache_line_size` for overflow and throw same exception type. 

It will address both accessing `r1::cache_line_size` from each allocation and throw `bad_array_new_length` on each in each overflow scenario. What do you think?

include/oneapi/tbb/memory_pool.h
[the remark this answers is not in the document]
kboyarinov: Fixed

include/oneapi/tbb/cache_aligned_allocator.h
aleksei-fedotov: Yeah, that is basically how I see it.

kboyarinov: Applied.

include/oneapi/tbb/cache_aligned_allocator.h
```diff
@@ -53,7 +54,17 @@ class cache_aligned_allocator {
...
     //! Allocate space for n objects, starting on a cache/sector line.
     __TBB_nodiscard T* allocate(std::size_t n) {
-        return static_cast<T*>(r1::cache_aligned_allocate(n * sizeof(value_type)));
+        T* p = nullptr;
+
+        // Check overflow before multiplying
+        if (n > ~std::size_t(0) / sizeof(value_type)) {
+            // r1::cache_aligned_allocate throws bad_array_new_length if
+            // n*sizeof(T) + cache_line_size causes overflow
+            throw_exception(exception_id::bad_array_new_length);
```
vossmjp, line 63: I don't suggest something different, but this only fixes the problem for recompiled code, not for drop in of new oneTBB binary library with previously compiled code.

aleksei-fedotov: Unfortunately, we cannot differentiate the _improper size_ situation in the library code even by using assembler injection. I think at best we can only recommend recompiling TUs where allocator is used.

aleksei-fedotov (approved):
Looks good to me!

